PR #5, opened 2024-10-09
WIP. Portfolio contract impl

### LINKED ISSUE #2: Portfolio contract
- Stores list of items (fka loans, assets), with
```solidity
struct Collateral {
  IERC6909 source;
  uint256 id;
}

struct Item {
    address owner;
    Collateral collateral;
    uint128 outstanding_quantity;
    uint128 outstanding_normalized_debt;
    IERC7726 valuation;
}

mapping (uint64 poolId => mapping (uint128 itemId => Item)) items;

```
- Write methods to
   - create(uint64 poolId, address collateralSource, uint256 collateralId, address owner)
     - locks ERC-6909 NFT
     - create item in storage 
   - increase_debt(uint64 poolId, itemId, amount)
     - only callable by owner 
   - decrease_debt(uint64 poolId, itemId, principal, interest, unscheduled)
     - only callable by owner
   - transfer_debt(uint64 poolId, ..)
     - only callable by owner

lemunozm:
# Description

Fixes #2

For the `Portfolio` contract

- we want:
  - Handling the accounting of items: tracking how much debt is increased/decreased
  - Be able to value items/portfolios.
  - Be able to update items.
- we don't want:
  - Handling permissions.
  - Distinguish between kinds of items.

### Lifecycle of an item from a portfolio.
- `create()` -> collateral attached to the new created item
  - Any calls to: `increase()`/`decrease()`/`updates()`/...
- `close()` -> collateral detached from the item

**Blocked by:**
- https://github.com/centrifuge/asset-pools/pull/19
- https://github.com/centrifuge/asset-pools/pull/22


src/Portfolio.sol
```diff
@@ -0,0 +1,158 @@
...
+    /// Contract where the collateral exists
+    IERC6909 source;
+    /// Identification of the collateral in that contract
+    uint256 id;
+}
+
+function collateralGlobalId(Collateral memory collateral) pure returns (address) {
+    // QUESTION: Could we have collisions in 20 bytes?
+    return address(uint160(uint256(keccak256(abi.encode(collateral.source, collateral.id)))));
+}
```
lemunozm: Here, I'm getting a unique address for our assets. I think with 20 bytes; we can not have collisions (otherwise, the own eth would not work)

src/Portfolio.sol
```diff
@@ -0,0 +1,158 @@
...
+    function hasRole(bytes32 role, address account) external view returns (bool);
+}
+
+interface IPoolRegistry {
+    function currencyOfPool(PoolId poolId) external view returns (address currency);
+}
+
+type PoolId is uint64;
+
+type ItemId is uint32;
```
lemunozm: I feel this is quite useful for IDs (which do not require methods to manipulate them), but not sure about our rules regarding this

src/Portfolio.sol
```diff
@@ -0,0 +1,158 @@
...
+    // TODO: extend with more properties
+    event Create(PoolId, ItemId);
+    event DebtIncreased(PoolId, ItemId);
+    event DebtDecreased(PoolId, ItemId);
+    event TransferDebt(PoolId, ItemId from, ItemId to);
+    event Closed(PoolId, ItemId);
+
+    mapping(PoolId => uint32 nonce) public itemNonces;
+    mapping(PoolId => mapping(ItemId => Item)) public items;
+
```
wischli: One result of the `InterestAccrual` spec call was that the `Portfolio` contract is the entity to identify the rate group of an item such that `LinearAccrual` contract only needs to map a `Group` to a `Rate` : https://www.notion.so/Interest-Accrual-11a2eac24e1780eca702c117cbc487d1

```solidity
    enum CompoundingPeriod {
        Secondly,
        Daily,
        Quarterly,
        Biannually,
        Annually
    }
    
    struct Group {
        uint128 ratePerPeriod;
        CompoundingPeriod period;
    }

    struct Rate {
        // accumlated rate index over time (prev chi)
        uint128 accumulatedRate;
        // last time the rate was accumulated
        uint64 lastUpdated;
    }

    mapping(Group => Rate) public rates;
```

src/Portfolio.sol
```diff
@@ -0,0 +1,158 @@
...
+    IAccessControl poolAdmin;
+    IPoolRegistry poolRegistry;
+
+    constructor(IAccessControl _poolAdmin, IPoolRegistry _poolRegistry) {
+        poolAdmin = _poolAdmin;
+        poolRegistry = _poolRegistry;
+    }
+
+    function create(PoolId poolId, ItemInfo calldata info) external {
+        bytes32 poolAdminRole = bytes32(uint256(PoolId.unwrap(poolId)));
```
wischli: I might be wrong but IIUC this is not collision resistant, ie it is possible to generate the same bytes for instance from an item id.

src/Portfolio.sol
```diff
@@ -0,0 +1,158 @@
...
+
+struct Item {
+    /// Base info of this item
+    ItemInfo info;
+    /// Total amount decreased by `decreaseDebt`. Measured in pool currency denomination
+    uint128 totalDecreasedDebt;
+    /// Total amount increased by `increaseDebt`. Measured in pool currency denomination
+    uint128 totalIncreasedDebt;
+    /// Total interest paid. Measured in pool currency denomination.
+    uint128 totalInterestPaid;
```
hieronx: I don't think these are needed

src/Portfolio.sol
```diff
@@ -0,0 +1,158 @@
...
+    Collateral collateral;
+    /// Fixed point rate
+    Decimal18 interestRate;
+    /// Fixed point number with the amount of asset hold by this item.
+    /// Usually for Price valued items it will be > 1. Other valuations will normally set this value from 0-1.
+    Decimal18 quantity;
+    /// Valuation method
+    IERC7726 valuationMethod;
+    /// Unix timestamp measured in secs
+    uint64 maturity;
```
hieronx: I dont think this is needed here

src/Portfolio.sol
```diff
@@ -0,0 +1,158 @@
...
+    address owner;
+    /// The RWA used for this item as a collateral
+    Collateral collateral;
+    /// Fixed point rate
+    Decimal18 interestRate;
+    /// Fixed point number with the amount of asset hold by this item.
+    /// Usually for Price valued items it will be > 1. Other valuations will normally set this value from 0-1.
+    Decimal18 quantity;
+    /// Valuation method
+    IERC7726 valuationMethod;
```
hieronx: I would just call it
```suggestion
    IERC7726 valuation;
```

src/Portfolio.sol
```diff
@@ -0,0 +1,158 @@
...
+    /// Identification of the collateral in that contract
+    uint256 id;
+}
+
+function collateralGlobalId(Collateral memory collateral) pure returns (address) {
+    return address(uint160(uint256(keccak256(abi.encode(collateral.source, collateral.id)))));
+}
+
+/// Struct used for user inputs and "static" item data
+struct ItemInfo {
```
peculiarity:
```suggestion
struct ItemMetadata {
```

src/Portfolio.sol
lemunozm: This comment is related to the usage of `RateId`? that we've discussed in slack, right?

src/Portfolio.sol
lemunozm: Why that? `PoolId` (uint64) is smaller than bytes32 (equivalent to uint256), so it always fits with no collision.

Anyways, this will be removed and use `Auth` instead.

src/Portfolio.sol
lemunozm: Maybe the `valuation()` method should be named differently then? To avoid confusion, because that method processes the result coming from this IERC7726 valuation, so both are differents

src/Portfolio.sol
lemunozm: Just curious, could you rationalize it?

Not sure if metadata is just what we store there... 🤔

src/Portfolio.sol
```diff
@@ -0,0 +1,158 @@
...
+            return computeDebt(poolId, itemId);
+        }
+    }
+
+    /// The valuation of this item
+    function valuation(PoolId poolId, ItemId itemId) external view returns (uint128 value) {
+        uint128 price = this.getPrice(poolId, itemId);
+        return items[poolId][itemId].outstandingQuantity.mulInt(price);
+    }
+
```
hieronx: Add two methods:
```solidity
function nav(PoolId poolId) external view returns (uint128 value) {
  // Iterate over all items, call getQuote on each, sum all values and return this
}

function indicativeNav(PoolId poolId) external view returns (uint128 value) {
  // Iterate over all items, call getIndicativeQuote on each, sum all values and return this
}
```

src/Portfolio.sol
```diff
@@ -0,0 +1,158 @@
...
+    function close(PoolId poolId, ItemId itemId) external {
+        require(items[poolId][itemId].outstandingQuantity.inner() == 0, "The item must not have outstanding quantity");
+
+        delete items[poolId][itemId];
+
+        emit Closed(poolId, itemId);
+    }
+
```
hieronx: Add `updateRate` and `updateValuation` methods

src/Portfolio.sol
```diff
@@ -0,0 +1,158 @@
...
+        item.outstandingQuantity = item.outstandingQuantity + quantity;
+        item.totalIncreasedDebt += quantity.mulInt(price);
+
+        emit DebtIncreased(poolId, itemId);
+    }
+
+    function decreaseDebt(PoolId poolId, ItemId itemId, Decimal18 principal, uint128 interest) external {
+        //TODO: opposite to increaseDebt
+    }
```
lemunozm: From the Frederik specs, this method is divided into two, one for the return of the principal and another one for the interest.

Does it make sense to replicate that? I think the only difference is regarding the gas. Two more accurate calls vs one generic call. Depends on the likelihood of each scenario

src/Portfolio.sol
lemunozm: I'm wondering. To avoid using `getIndicateQuote()` and be more tied to ERC7726, could make sense to encode the "indicative" knowledge as part of the `base` part in `getQuote()`?

Right know we do: 
```solidity
address base = address(uint160(uint256(keccak256(abi.encode(collateral.source, collateral.id)))));
uint128 normalPrice = getQuote(1, base, currency)
uint128 indicativePrice = getIndicativeQuote(1, base, currency)
```

Could we do:
```solidity
enum PriceMode {
    Default
    Indicative
}

address base = address(uint160(uint256(keccak256(abi.encode(collateral.source, collateral.id, Default)))));
uint128 normalPrice = getQuote(1, base, currency)

address base = address(uint160(uint256(keccak256(abi.encode(collateral.source, collateral.id, Indicative)))));
uint128 indivativePrice = getQuote(1, base, currency)

Comment (lemunozm) [reply src/Portfolio.sol]: Later we can pass the same `PriceMode` enum to `nav()`

Comment (hieronx) [reply src/Portfolio.sol]: That just makes it more implicit. The main mistake I think we made on Centrifuge Chain with this was being implicit, when you were getting which price/NAV. I would really like to make it as explicit as possible now, to avoid incorrect usage. I think introducing a new method on the oracle side is fine for that, and helps the case.

File: src/Portfolio.sol (hunk 11 above)
Comment (lemunozm) [inline src/Portfolio.sol]: @hieronx I think at this contract level, `transferDebt()` is just exactly calling `decreaseDebt(from) + increaseDebt(to)`, so I think there is no reason to create such call. The top layer can offer this.

I mean, `transferDebt()` do not add any new information

Comment (hieronx) [reply src/Portfolio.sol]: It might be worth the gas cost though. But agreed it's worth considering. Once we have a more complete integrated system, we can write gas snapshot tests, and see.

File: src/Portfolio.sol [hunk 12]
```diff
@@ -0,0 +1,178 @@
...
+    ILinearAccrual linearAccrual;
+
+    constructor(IPoolRegistry _poolRegistry, ILinearAccrual _linearAccrual) {
+        poolRegistry = _poolRegistry;
+        linearAccrual = _linearAccrual;
+    }
+
+    function create(PoolId poolId, ItemInfo calldata info) external {
+        bool ok = info.collateral.source.transfer(address(this), info.collateral.id, 1);
+        require(ok, CollateralCanNotBeTransfered());
```
lemunozm: I do not fully remember what was the latest decision regarding transferring the collateral in the Portfolio. Should be delegated to the higher level?

@hieronx @ilinzweilin

src/Portfolio.sol
lemunozm: Ok. By now, internally, I can not save any read/write operation by doing it in the same method

lemunozm: But it's true that I need to check both amounts matches, so it deserves doing here the operation and checks. Never mind, we keep it!

src/Portfolio.sol
hieronx: I think keep it in the portfolio for now, but we might move it out later.

lemunozm: Ok. If we keep it, we then require some logic in the `close()` method to return the collateral to the creator of the item and also store the creator's address.

lemunozm:
Ready for another review phase. 

Some points could require discussion:
- After some play around, I've decided to make the `increaseDebt()`/`decreaseDebt()` methods to accept an `amount` instead of a `quantity`. It makes life much easier later for transferring debt (which can be done directly). Also, the interface is more "friendly" I guess. The difference in achieving this is that the `getQuote()` method now receives the `base` and `quote` parameters interchanged. Still open to discussion and see how it evolves.

- Added a `creator` field to be able to return the collateral to the creator of the asset in the close() method.

- As we talked @Jeroen, I've added `getQuote` and `getIndicativeQuote()`. Nevertheless, later, for the usage in the `nav()`/`itemValuation()`, I'm using an enum to discriminate which underlying method to use and avoid duplicating the same code to just end up calling different a `getXXXQuote()` method. I hope you agree with it. 
  - If you still think we need to have an explicit method named as `indivativeNav()` to abstract from the enum indexes, then we could add a thin wrapper over `nav(PricingMode.Indicative)`.

- `owner` (understood as borrowed) removed from the `Item`. The next layer should handle this.

- Removed `PoolId` and `ItemId` types to not add extra noise now.

hieronx:
> Added a creator field to be able to return the collateral to the creator of the asset in the close() method.

I think you can just make this a parameter in the close method. Since this is anyways permissioned, a higher level contract can ensure it is the correct address.

This also allows building for example a liquidation layer on top, where the collateral is transferred to a different user than the initial creator.

src/Portfolio.sol
```diff
@@ -0,0 +1,245 @@
...
+
+enum PricingMode {
+    Real,
+    Indicative
+}
+
+contract Portfolio is Auth {
+    using MathLib for uint256;
+
+    error ItemNotFound();
```
hieronx: Let's move events and errors to the interface

src/Portfolio.sol
```diff
@@ -0,0 +1,245 @@
...
+    event TransferDebt(uint64 poolId, uint32 fromItemId, uint32 toItemId, Decimal18 quantity, uint128 interest);
+    event Closed(uint64 poolId, uint32 itemId);
+
+    /// Latest incremental item identifier per pool.
+    mapping(uint64 poolId => uint32 nonce) public itemNonces;
+
+    /// A list of items (a portfolio) per pool.
+    mapping(uint64 poolId => Item[]) public items;
+
+    IPoolRegistry poolRegistry;
```
hieronx: Both should have the `public` modifier

src/Portfolio.sol
```diff
@@ -0,0 +1,245 @@
...
+    ILinearAccrual linearAccrual;
+
+    constructor(address owner, IPoolRegistry _poolRegistry, ILinearAccrual _linearAccrual) Auth(owner) {
```
hieronx: Should have postfixed underscore, not prefixed. Why: our style guide is that we use prefixed underscore for internal variables, and postfixed underscore for avoiding variable shadowing (like we have here, where the names otherwise match)

src/Portfolio.sol
```diff
@@ -0,0 +1,245 @@
...
+
+        emit DebtDecreased(poolId, itemId, principal, interest);
+    }
+
+    /// Transfer debt `from` an item `to` another item.
+    function transferDebt(uint64 poolId, uint32 fromItemId, uint32 toItemId, uint128 principal, uint128 interest)
+        external
+        auth
+    {
+        this.decreaseDebt(poolId, fromItemId, principal, interest);
```
hieronx: Adding `this` makes it an external call, which makes it more expensive. Instead these 2 methods should be `public` rather than `external`, and then `this.` should be removed

src/Portfolio.sol
```diff
@@ -0,0 +1,245 @@
...
+struct Item {
+    /// Base info of this item
+    ItemInfo info;
+    /// A representation of the debt used by LinealAccrual to obtain the real debt
+    uint128 normalizedDebt;
+    /// Outstanding quantity
+    Decimal18 outstandingQuantity;
+}
+
+function exists(Item storage item) view returns (bool) {
```
hieronx: Imo this is very rusty solidity ;) I admit it is more stylistic, but I would have an internal method `_doesItemExist` (or something like that, or if it's one line like here, actually just put it inline.

src/Portfolio.sol
```diff
@@ -0,0 +1,245 @@
...
+        emit DebtIncreased(poolId, itemId, amount);
+    }
+
+    /// Decrease the debt of an item
+    function decreaseDebt(uint64 poolId, uint32 itemId, uint128 principal, uint128 interest) external auth {
+        Item storage item = items[poolId][itemId];
+        require(item.exists(), ItemNotFound());
+
+        Decimal18 quantity = _getQuantity(poolId, item, principal);
+        uint128 amount = principal + interest;
```
hieronx: Seems like this definition is not really required since it's only used once, can just be inline on L173

src/Portfolio.sol
```diff
@@ -0,0 +1,245 @@
...
+        this.increaseDebt(poolId, toItemId, principal + interest);
+    }
+
+    /// Close a non-outstanding item returning the collateral to the creator of the item
+    function close(uint64 poolId, uint32 itemId) external {
+        Item storage item = items[poolId][itemId];
+        require(item.exists(), ItemNotFound());
+        require(item.outstandingQuantity.inner() == 0, ItemCanNotBeClosed());
+
+        bool ok = item.info.collateral.source.transfer(item.info.creator, item.info.collateral.id, 1);
```
hieronx: I would move the `transfer` to after the `delete` call, to prevent any reentrancy issues. See Solidity re-entrancy and the "checks-effects-interactions (CEI) pattern"

src/Portfolio.sol
lemunozm: Ok. Curious why. Is it intended to be accessible from other contracts that access the `Portfolio`?

src/Portfolio.sol
lemunozm: Ok, that's a question I had. It is supposed that I need to create an `IPortfolio` then.

Because with solidity you can create any contract with any method that matches, not fully sure when an interface is needed. For example, the `Pools` contract could have on top of its file the following:

```solidity
contract Portfolio {
    function nav() extend returns (uint128);
}
```
And I do not need to create a portfolio interface to call `nav()`

src/Portfolio.sol
lemunozm: Ok, thanks for the declaration, I did not fully understand why, for some methods, I needed `this`. Now it makes sense!

src/Portfolio.sol
lemunozm: Ok, I wanted to encapsulate in a method because it's not a super simple line, and maybe in the future how we identify a "null" item change. So, let's go with `_doesItemExist()` 👍🏻

src/Portfolio.sol
lemunozm: Is there any cost for the EVM machine (plus optimizations) on declaring internal function variables?

src/Portfolio.sol
lemunozm: Thanks! I still do not have a sharp eye to realize these things

lemunozm:
> I think you can just make this a parameter in the close method. Since this is anyways permissioned, a higher level contract can ensure it is the correct address.

Makes sense! I guess the same can be done for `create()`. We can pass the creator address from where the asset belongs, right? As far as the `creator` has approved the transfer for the caller to `create()`, it will work.

src/Portfolio.sol
hieronx: A common style (and one we follow) is to create an interface file for every single contract, and put all the events, errors, and natspec comments per function here. See for example this: https://github.com/centrifuge/liquidity-pools/blob/main/src/interfaces/IInvestmentManager.sol

Benefits of this approach is:
- Keeps the core contract as simple as possible
- Allows having extensive natspec separately, that will be included in UIs etc
- Allows integrators to pull in the interface with all the documentation, for any version, while we hardcode the contract version

src/Portfolio.sol
hieronx: 1. There is not really any downside to making a variable public.
2. It is helpful to be able to inspect dependencies, for many reasons. One is simply for us to build fork tests that check the dependencies are wired correctly.

src/Portfolio.sol
hieronx: I think you should never need `this`. Although I might miss some case

src/Portfolio.sol
hieronx: Probably minimal, this is more of a subjective opinion from my side, but I'd argue a separate variable only makes sense if it's either:
- Used multiple times
- Not intuitive what the definition of the contents of the variable is.

In this case, imo defining it as `amount` doesn't actually make it clearer.

hieronx:
> We can pass the creator address from where the asset belongs, right? As far as the creator has approved the transfer for the caller to create(), it will work.

The approval should be to the portfolio contract actually, not to the creator.

src/Portfolio.sol
lemunozm: Thanks for the detailed explanation!

src/Portfolio.sol
lemunozm: I guess `this` also changes the caller's context, so `msg.sender` will be the same contract address (?)

src/Portfolio.sol
```diff
@@ -0,0 +1,175 @@
...
+    using MathLib for uint256;
+
+    struct Item {
+        /// Base info of this item
+        ItemInfo info;
+        /// A representation of the debt used by LinealAccrual to obtain the real debt
+        uint128 normalizedDebt;
+        /// Outstanding quantity
+        Decimal18 outstandingQuantity;
+    }
```
lemunozm: I am not fully sure about our style of putting structs/enums in/out from contracts/interfaces. 

Here, I think I'm forced because `ItemInfo` belongs to the interface, so it's the only way to access `ItemInfo` from here (if not, I do not know how).

src/Portfolio.sol
```diff
@@ -0,0 +1,175 @@
...
+
+    constructor(address owner, IPoolRegistry poolRegistry_, ILinearAccrual linearAccrual_) Auth(owner) {
+        poolRegistry = poolRegistry_;
+        linearAccrual = linearAccrual_;
+    }
+
+    /// Creates a new item based of a collateral.
+    /// The owner of the collateral will be this contract until close is called.
+    function create(uint64 poolId, ItemInfo calldata info, address creator) external auth {
+        bool ok = info.collateral.source.transferFrom(creator, address(this), info.collateral.id, 1);
```
lemunozm: > The approval should be to the portfolio contract actually, not to the creator.

Regarding this comment, I do not fully understand it. AFAIK, `transferFrom` can transfer  the tokens if `creator` has approved some amount of `collateral.id` (aka `1`) for `msg.sender`. Is that wrong?

hieronx: Some `sender` gives an approval for `receiver` to transfer `amount` tokens on their behalf. Setting `allowance[sender][receiver] += amount`.

Then `spender` can call `transferFrom` with `from=sender`, where it checks `allowance[from][msg.sender]`.

If user A gives approval for user B to spend tokens, and B calls the portfolio contract, the portfolio contract cannot actually call `transferFrom` with `from=A` since it will check the allowance of `allowance[from][PortfolioContract]`, which is 0.

So the user is approving the portfolio contract, not the caller of the portfolio contract, to transfer the collateral.

lemunozm: > since it will check the allowance of allowance[from][PortfolioContract]

Not sure why you set the `PortfolioContract` account in the allowance. Looking at the code, it seems to work as follows:

1. User A gives approval for user B:
  - `allowance[A][B] = 1`
2. User B calls `create()`using A as a `creator`:
  - `transferFrom(A, address(Portfolio))` and `B = msg.sender`, which under the hood it checks if `allowance[A][msg.sender] >= 1` that is `allowance[A][B] >= 1` which matches with the given approval above. No `PortfolioContract` address is used in the allowance mapping.
  
Am I blind? 🤔

hieronx: The portfolio contract calls the ERC6909 contract, so
```
which under the hood it checks if allowance[A][msg.sender] >= 1 that is allowance[A][B] >= 1 
```

is instead
```
which under the hood it checks if allowance[A][msg.sender] >= 1 that is allowance[A][address(Portfolio)] >= 1 
```

since the `msg.sender` from the perspective of the ERC6909 contract is the Portfolio contract.

lemunozm: Right! That's correct! I still do not internalize that each call done to a new contract changes the `sender`. This is quite different from Substrate.

Basically, that means that the `allowance` feature from `IERC6909` lost most of its potential because of this 😞. So, you use ERC6909 directly, or you can not delegate to other accounts through contracts

hieronx: Indeed. The allowance is really intended as the main smart contract pattern for token transfers, not for delegating token approval to other users.

src/interfaces/IPortfolio.sol
```diff
@@ -0,0 +1,109 @@
...
+    error CollateralCanNotBeTransfered();
+
+    /// @notice Dispatched after the creation of an item.
+    event Create(uint64 indexed poolId, uint32 itemId, Collateral collateral);
+
+    /// @notice Dispatched when the item valuation has been updated.
+    event ValuationUpdated(uint64 indexed poolId, uint32 itemId, IERC7726);
+
+    /// @notice Dispatched when the interest rate has been updated.
+    event InterestRateUpdated(uint64 indexed poolId, uint32 itemId, bytes32 rateId);
```
wischli, line 67: We are also dispatching a `RateAccumulated` event after updating the rate. Suppose both events can co-exist because `LinearAccrual` doesn't know about pool or item ids. Just wanted to point that out in case the `LinearAccrual` one can be removed due to being redundant.

src/Portfolio.sol
```diff
@@ -0,0 +1,174 @@
...
+        poolRegistry = poolRegistry_;
+        linearAccrual = linearAccrual_;
+    }
+
+    /// @inheritdoc IPortfolio
+    function create(uint64 poolId, ItemInfo calldata info, address collateralOwner) external auth {
+        bool ok = info.collateral.source.transferFrom(collateralOwner, address(this), info.collateral.id, 1);
+        require(ok, CollateralCanNotBeTransfered());
+
+        uint32 itemId = uint32(items[poolId].length);
```
wischli: Question towards Solidity experts: Curious if we can assume such casting to be safe.

src/Portfolio.sol
```diff
@@ -0,0 +1,174 @@
...
+        items[poolId].push(Item(info, 0, d18(0)));
+
+        emit Create(poolId, itemId, info.collateral);
```
wischli: Nit question: AFAIU, events should still have some prefix/suffix related to the contract (e.g. `PortfolioCreated` or `NewPortfolio`) to make them unique in contrast of how we did it in Substrate where that was redundant.

src/Portfolio.sol
```diff
@@ -0,0 +1,174 @@
...
+
+        emit InterestRateUpdated(poolId, itemId, rateId);
+    }
+
+    /// @inheritdoc IPortfolio
+    function updateValuation(uint64 poolId, uint32 itemId, IERC7726 valuation) external auth {
+        Item storage item = items[poolId][itemId];
+        require(_doItemExists(item), ItemNotFound());
+
+        item.info.valuation = valuation;
```
wischli: Question for solidity experts: Curious if it makes more sense to move this repetitive pattern to a modifier? Especially since some functions such as this one only mutate one field of the unterlying `Item` struct such that we can save some gas by not declaring the entire item as `storage` but only the field.

```diff
- Item storage item = items[poolId][itemId];
+ IERC7726 storage valuation = items[poolId][itemId].valuation;
```

src/Portfolio.sol
```diff
@@ -0,0 +1,174 @@
...
+        Item storage item = items[poolId][itemId];
+        require(_doItemExists(item), ItemNotFound());
+
+        Decimal18 quantity = _getQuantity(poolId, item, principal);
+
+        item.normalizedDebt =
+            linearAccrual.decreaseNormalizedDebt(item.info.interestRateId, item.normalizedDebt, principal + interest);
+        item.outstandingQuantity = item.outstandingQuantity - quantity;
+
+        emit DebtDecreased(poolId, itemId, principal, interest);
```
wischli: Curious why for `DebtIncreased` event we don't include the principal but for `DebtDecreased` we do.

src/Portfolio.sol
```diff
@@ -0,0 +1,174 @@
...
+        Item storage item = items[poolId][itemId];
+        require(_doItemExists(item), ItemNotFound());
+        require(item.outstandingQuantity.inner() == 0, ItemCanNotBeClosed());
+
+        Collateral memory collateral = item.info.collateral;
+
+        delete items[poolId][itemId];
+
+        bool ok = collateral.source.transfer(collateralOwner, collateral.id, 1);
+        require(ok, CollateralCanNotBeTransfered());
```
wischli: Nit: In case the transfer fails, we can save minor gas by checking before deleting
```suggestion
        bool ok = collateral.source.transfer(collateralOwner, collateral.id, 1);
        require(ok, CollateralCanNotBeTransfered());
        
        delete items[poolId][itemId];
```

src/Portfolio.sol
```diff
@@ -0,0 +1,174 @@
...
+        } else {
+            // mode == PricingMode.Indicative
+            return item.info.valuation.getIndicativeQuote(quantity.inner(), base, quote).toUint128();
+        }
+    }
+
+    /// @inheritdoc IValuation
+    function itemValuation(uint64 poolId, uint32 itemId, PricingMode mode) external view returns (uint128 value) {
+        Item storage item = items[poolId][itemId];
+        require(_doItemExists(item), ItemNotFound());
```
wischli: Nit related to my above question of replacing the repetitive check with a modifier: If `_doItemExists` did not require `item` to be `storage`, then this view function could declare it as `memory` instead as we don't perform mutation; and same in `function nav`. But maybe irrelevant because we don't care too much for `view` functions.

src/libraries/Decimal18.sol
```diff
@@ -0,0 +1,37 @@
+// SPDX-License-Identifier: UNLICENSED
+pragma solidity 0.8.28;
```
wischli: Nit
```suggestion
pragma solidity ^0.8.28;
```

test/unit/Portfolio.t.sol
```diff
@@ -0,0 +1,91 @@
+// SPDX-License-Identifier: MIT
+pragma solidity ^0.8.28;
+
+import "forge-std/Test.sol";
+import "src/Portfolio.sol";
+import "src/interfaces/IPortfolio.sol";
+
+contract MockERC6909 is IERC6909 {
```
wischli: Comment which can immediately be resolved: The Substrate mockbuilder rises from its ashes looking much **solid** 🚀

wischli (commented):
Loving the return of mocks already in your first code PR! A few questions, mostly to "Solidity experts" from my side and a few nits.

Great documentation in the `IPortfolio`. A minor nit there: A few function params are missing docs.

Looking forward to seeing your tests!

src/Portfolio.sol
lemunozm: Is it true even if the event belongs to an interface? The interface puts some kind of namespace. Here, I can use them without the namespace because I'm inheriting from that interface, but from the outside, I need to write `IPortfolio.Create`

src/interfaces/IPortfolio.sol
lemunozm: `InterestRateUpdated` is dispatched when `Portfolio` calls `renormalizedDebt()`, so I think in this path, there is no such collision.

src/Portfolio.sol
lemunozm: I also thought about this. 

If I move `_doItemExists()` as a modifier. I would need to access the item there and also later in the function body. So 2 reads from the item from the storage

Anyways, I'm interested to know exactly when to use a modifier and when to use a simple call to a private method in the body of the function. Seems like it's exactly identical in most cases (and the same amount of code factorized), and sometimes the code under modifiers seems more tricky to follow (?)

src/interfaces/IPortfolio.sol
wischli: Perfect, then no collision or redundancy as `ILinearAccrual.RateAccumulated` is dispatched during rate update after periods have passed.

src/Portfolio.sol
lemunozm: `increase` and `decrease` are not identical. 

When you increase, you just "borrow" money, if later that debt is added as an amount to the account. So basically, i.e. you ask for money because you want to buy your house, so now you have a debt.

Later, when you decrease, you can "pay" or "return" 2 kinds of amounts: the principal (the amount related to what you increased or borrowed) and the interest accrued over the time you had that money. Depending on what amount is returned for decreasing the debt, the accounting will behave differently.

So basically, there is no concept of principal when increasing because everything is principal. But note that `DebtIncreased` still register the amount used to increase.

src/Portfolio.sol
lemunozm: I had a previous comment from Jeroen asking me to do it in the other way due to re-entrance issues and follow the CEI pattern 😆 : https://github.com/centrifuge/asset-pools/pull/5#discussion_r1799874965

src/Portfolio.sol
lemunozm: Not sure if I follow your comment correctly. `storage` is faster than `memory` though. Basically AFAIK ref vs copy, so `storage` is faster.

src/libraries/Decimal18.sol
lemunozm: Good point

test/unit/Portfolio.t.sol
lemunozm: Yeah, that's cool! Although I still miss some kind of closure for giving different implementations for each test case.

src/Portfolio.sol
hieronx: We should probably add a `MAX_ITEMS` constant that limits the total number of items to `type(uint32).max` on `create()`, and then we can safely cast it here

src/Portfolio.sol
hieronx: Events are generally closely tied to the contract, so you don't need to add a prefix for the contract.

src/Portfolio.sol
hieronx: If you use a modifier, the code is repeated inline in every place, so the bytecode size grows by the number of usages of the modifier. An internal method instead is not repeated in every place.

hieronx: The pattern I've seen most codebases use is to limit modifiers to access/authorization related checks, and use internal methods to everything else.

src/libraries/Decimal18.sol
hieronx: Generally it's considered better practice to hardcode a specific Solidity version. That is also what we did in liquidity-pools.

src/Portfolio.sol
lemunozm: Makes sense. Also, good catch William. In substrate, lengths could not be longer than `uint32`, but here, we could exceed that.

src/Portfolio.sol
lemunozm: Thanks! Super accurate explanation in two sentences!

src/libraries/Decimal18.sol
lemunozm: But for libraries/interfaces is it also true?

src/Portfolio.sol
```diff
@@ -0,0 +1,174 @@
...
+
+    /// A list of items (a portfolio) per pool.
+    mapping(uint64 poolId => Item[]) public items;
+
+    IPoolRegistry public poolRegistry;
+    ILinearAccrual public linearAccrual;
+
+    constructor(address owner, IPoolRegistry poolRegistry_, ILinearAccrual linearAccrual_) Auth(owner) {
+        poolRegistry = poolRegistry_;
+        linearAccrual = linearAccrual_;
```
ilinzweilin, line 39: we need an auth `file` function where we can update the linearAccrual contract in case we swap it

src/Portfolio.sol
```diff
@@ -0,0 +1,174 @@
...
+    }
+
+    /// @inheritdoc IPortfolio
+    function create(uint64 poolId, ItemInfo calldata info, address collateralOwner) external auth {
+        bool ok = info.collateral.source.transferFrom(collateralOwner, address(this), info.collateral.id, 1);
```
ilinzweilin: I would not create local variables that are only used once...
Unfortunately in makes the code less readable, but safes gas

src/Portfolio.sol
ilinzweilin: I think , we also should always use `MathLib` or OpenZepplin `SafeCast` functions for casting though. 
Same as we are doing inside `liquidity-pools` https://github.com/centrifuge/liquidity-pools/blob/main/src/InvestmentManager.sol#L65

src/Portfolio.sol
```diff
@@ -0,0 +1,174 @@
...
+    IPoolRegistry public poolRegistry;
+    ILinearAccrual public linearAccrual;
+
+    constructor(address owner, IPoolRegistry poolRegistry_, ILinearAccrual linearAccrual_) Auth(owner) {
+        poolRegistry = poolRegistry_;
+        linearAccrual = linearAccrual_;
+    }
+
+    /// @inheritdoc IPortfolio
+    function create(uint64 poolId, ItemInfo calldata info, address collateralOwner) external auth {
```
ilinzweilin: I think there might be items, like cash / fee items that are not bind to a collateral.
That's why I would move the locking function outside of `create` into a separate `lock` function. 
Within create: if a collateral id is provided, I would check if the collateral is locked in order to create the item. If no collateral id is provided, I would create the item without collateral.
I also think that we should be locking the collateral nfts in a seperate escrow contract.

ilinzweilin (changes requested):

src/Portfolio.sol
hieronx: I think this is a good idea! 

For the last line, @lemunozm you can use this contract: https://github.com/centrifuge/liquidity-pools/blob/main/src/Escrow.sol

src/Portfolio.sol
wischli: Thanks for the thorough explanation, makes perfect sense!

src/Portfolio.sol
lemunozm: I thought compiling with some minimal optimization level solved those kinds of things:

https://docs.soliditylang.org/en/latest/internals/optimizer.html#commonsubexpressioneliminator

> All subexpressions that are identifiers themselves are replaced by their current value if the value is an identifier.

which I would say, it's true for `ok`

src/Portfolio.sol
lemunozm: > I think there might be items, like cash items that are not bind to a collateral.

Makes sense! 

> If no collateral id is provided, I would create the item without collateral.

Does it mean that I should model that as `collateral.source = address(0)`? That would mean that later we can not get the `quote`. Although I guess this does not affect because cash items are valued just with the debt (what Jeroen said in some meeting/thread)

> That's why I would move the locking function outside of `create` into a separate `lock` function.

Could you extend it? Not sure I'm following you with the `lock` part

src/Portfolio.sol
lemunozm: Need to dive into this 👍🏻

src/Portfolio.sol
hieronx: > Does it mean that I should model that as collateral.source = address(0)? Does would mean that later we can not get the quote. Although I guess this does not affect because cash items are valued just with the debt (what Jeroen said in some meeting/thread)

Yeah exactly.

> Could you extend it? Not sure I'm following you with the lock part

I think the idea would be something like
```solidity
function lock(address collateral, uint256 id) external auth {
  // Transfer collateral to escrow
}

function create(address collateral, uint256 id, ..) external auth {
  // If not 0, check collateral is locked in escrow
  // Create item
}
```

src/libraries/Decimal18.sol
hieronx: For interfaces, we intentionally specify a much more open version, since any integrators need to match the version range. So in liquidity pools we specify >=0.5.0: https://github.com/centrifuge/liquidity-pools/blob/main/src/interfaces/IInvestmentManager.sol#L2

For libraries, I think the same applies, that hardcoding a specific version is better for security.

src/Portfolio.sol
lemunozm: Rewarding the escrow contract, do we expect a custom implementation of such contract? The liquidity pools one only handles ERC20, but here, we want to use ERC6909

src/Portfolio.sol
```diff
@@ -0,0 +1,205 @@
...
+    }
+
+    /// @notice Updates a contract parameter
+    /// @param what Accepts a bytes32 representation of 'poolRegistry', 'linearAccrual'
+    function file(bytes32 what, address data) external auth {
+        if (what == "poolRegistry") poolRegistry = IPoolRegistry(data);
+        else if (what == "linearAccrual") linearAccrual = ILinearAccrual(data);
+        else revert("Portfolio/file-unrecognized-param");
+        emit File(what, data);
+    }
```
lemunozm: Do I need for this:
- Move it to `IPortfolio`?
- Create a custom error instead of revert?
  - And move it probably to `IPortfolio`
- Put the File event in `IPortfolio`?

If yes, what are the advantages?

src/libraries/Decimal18.sol
lemunozm: This is fixed in https://github.com/centrifuge/asset-pools/pull/19, which should be merged first

src/Portfolio.sol
lemunozm: Implemented as a separate PR: https://github.com/centrifuge/asset-pools/pull/22

Please avoid to review the current Escrow contract of this PR

src/interfaces/IPortfolio.sol
```diff
@@ -0,0 +1,119 @@
...
+import {IERC7726, IERC6909} from "src/interfaces/Common.sol";
+import {D18} from "src/types/D18.sol";
+
+/// @notice Defines a set of items that can be valued
+interface IValuation {
+    /// @notice How the price is used to value each item.
+    enum PricingMode {
+        /// @dev Underlaying call to `ERC6909.getQuote()`
+        Real,
+        /// @dev Underlaying call to `ERC6909.getIndicativeQuote()`
```
peculiarity: There are no such methods defined in the ERC6909 spec or impl. 
What exactly is this supposed to do ?

src/interfaces/IPortfolio.sol
```diff
@@ -0,0 +1,119 @@
...
+    }
+
+    /// @notice Dispatched when the item can not be found.
+    error ItemNotFound();
+
+    /// @notice Dispatched when the item can not be closed yet.
+    error ItemCanNotBeClosed();
+
+    /// @notice Dispatched after the creation of an item.
+    event Create(uint64 indexed poolId, uint32 itemId, IERC6909 source, uint256 tokenId);
```
peculiarity:
```suggestion
    event Created(uint64 indexed poolId, uint32 itemId, IERC6909 source, uint256 tokenId);
```

src/libraries/MathLib.sol
```diff
@@ -155,6 +155,23 @@ library MathLib {
         return uint8(value);
     }
 
+    /// @notice Safe type conversion from uint256 to uint32.
+    function toUint32(uint256 value) internal pure returns (uint32) {
```
peculiarity, line 159: Some tests are missing I see :)

src/interfaces/IPortfolio.sol
lemunozm: I wrote ERC6909 in the comment but I wanted to write ERC7726 xD

lemunozm: Fixed

src/interfaces/IPortfolio.sol
lemunozm: Fixed

src/libraries/MathLib.sol
lemunozm: Added

src/Portfolio.sol
peculiarity: Yes to all of them.
1. I think we talked about that :)
2. Having strings is pretty expensive because the string must be encoded into the bytecode and you pay shit ton of gas for non-zero byte. ( if it's like with the calldata, then for zero-byte you pay 4 gas and for non-zero 16 gas. ) When you have custom error, actually all you have is 4 bytes of the hashed signature and that's it .
3. We put events in interfaces.

src/interfaces/Common.sol
```diff
@@ -0,0 +1,33 @@
+// SPDX-License-Identifier: AGPL-3.0-only
+pragma solidity >=0.8.0;
+
+// NOTE: understand this file as a backlog for interfaces that do not have an specific file yet (but should).
+
+interface IERC6909 {
```
peculiarity, line 6: Probably all of these Interfaces  will be already defined. IERC7726 could be in a separate file. Doesn't see much benefit of hacking multiple interfaces defined in this file. What does common mean ? What defines an interfaces being common?

src/interfaces/IPortfolio.sol
```diff
@@ -0,0 +1,119 @@
...
+interface IPortfolio is IValuation {
+    /// @notice Struct used for user inputs and "static" item data
+    struct ItemInfo {
+        /// @notice Rate identification to compute the interest.
+        bytes32 interestRateId;
+        /// @notice Fixed point number with the amount of asset hold by this item.
+        /// Usually for Price valued items it will be > 1.
+        /// Other valuations will normally set this value from 0-1.
+        D18 quantity;
+        /// @notice Valuation used for this item.
```
peculiarity, line 32: Probably it's nice to write that this will be an address to an oracle providing valuation.

src/libraries/MathLib.sol
```diff
@@ -155,6 +155,23 @@ library MathLib {
...
+        if (value > type(uint32).max) {
+            revert("MathLib/uint32-overflow");
+        }
+        return uint32(value);
+    }
+
+    /// @notice Safe type conversion from uint256 to int128.
+    function toInt128(uint256 _value) internal pure returns (int128 value) {
+        if (_value > uint128(type(int128).max)) {
+            revert("MathLib/int128-overflow");
```
peculiarity, line 169: Let's move to custom errors ;)

src/Portfolio.sol
```diff
@@ -0,0 +1,197 @@
...
+
+    /// @inheritdoc IPortfolio
+    function create(uint64 poolId, ItemInfo calldata info, IERC6909 source, uint256 tokenId) external auth {
+        uint32 itemId = items[poolId].length.toUint32() + 1;
+
+        uint256 uniqueItemId = uint256(poolId) << 64 + itemId;
```
peculiarity: That's pretty clever but...
`uint256(poolId) << 64 + itemId` != `(uint256(poolId) << 64) + itemId`
and I'm pretty sure you want to have the second one :D

peculiarity: Even though it is more expensive by some 26 gas I would suggest using `abi.encodePacked(poolId, itemId)`
This will give you in total 96 bytes. 
From left to right, first 64 bytes will be for the `poolId` and the following 32 bytes will be for the `itemId`. Which can be casted to `uint96`.
`uint96(bytes12(abi.encodePacked(poolId, itemId))`

Even if you decide to go with your implementation, I'm not sure why you shift by `64`and not `32`. You want to make space for the `itemId`'s 32 bytes, right? The way you do it, you leave 32 bytes empty. 
So you final number of 256 will consist.
 - 128 empty bytes
 - 64 bytes pool id ( at most )
 - 32 empty bytes
 - 32 bytes item id ( at most )

src/Portfolio.sol
```diff
@@ -0,0 +1,197 @@
...
+pragma solidity 0.8.28;
+
+import {D18, d18} from "src/types/D18.sol";
+import {MathLib} from "src/libraries/MathLib.sol";
+import {Auth} from "src/Auth.sol";
+import {INftEscrow} from "src/interfaces/INftEscrow.sol";
+import {IPortfolio, IValuation} from "src/interfaces/IPortfolio.sol";
+import {IERC7726, IERC6909, IPoolRegistry, ILinearAccrual} from "src/interfaces/Common.sol";
+
+struct Item {
```
peculiarity, line 11: Move it to the `IPortfolio` interface ;)

src/Portfolio.sol
```diff
@@ -0,0 +1,197 @@
...
+    /// Base info of this item
+    IPortfolio.ItemInfo info;
+    /// A representation of the debt used by LinealAccrual to obtain the real debt
+    int128 normalizedDebt;
+    /// Outstanding quantity
+    D18 outstandingQuantity;
+    /// Identification of the asset used for this item
+    uint160 collateralId;
```
peculiarity, line 19: Is there a situation where you can have an item with `collateralId` == 0 ?

src/Portfolio.sol
```diff
@@ -0,0 +1,197 @@
...
+    /// @param what Accepts a bytes32 representation of 'poolRegistry', 'linearAccrual'
+    function file(bytes32 what, address data) external auth {
+        if (what == "poolRegistry") poolRegistry = IPoolRegistry(data);
+        else if (what == "linearAccrual") linearAccrual = ILinearAccrual(data);
+        else revert("Portfolio/file-unrecognized-param");
+        emit File(what, data);
+    }
+
+    /// @inheritdoc IPortfolio
+    function create(uint64 poolId, ItemInfo calldata info, IERC6909 source, uint256 tokenId) external auth {
```
peculiarity: I think the name is a little bit misleading.
You don't create anything (well probably the first time);
You put new item into the same `poolId`.
Suggestion: `addItem`, `add`

src/NftEscrow.sol
```diff
@@ -0,0 +1,67 @@
...
+        bool ok = source.transferFrom(from, address(this), tokenId, 10 ** source.decimals(tokenId));
+        require(ok, CanNotBeTransferred());
+
+        emit Locked(source, tokenId);
+
+        return nftId;
+    }
+
+    /// @inheritdoc INftEscrow
+    function unlock(IERC6909 source, uint256 tokenId, address to) external auth {
```
peculiarity, line 21: How do you know what value to use for `to` ?

src/Portfolio.sol
lemunozm: Yeah, we have talked and I finally understood the reasons. 

Nevertheless, the `file` method is a bit special because, AFAIK, it will be always called directly by an account. So, no contracts will call it, and no interface is needed theoretically. It's something we should add in most of our contract, but it does not nothing to do with the contract itself. 

In case we really need an interface, does it make sense to move it to its own `IFile`? I think so

src/interfaces/Common.sol
lemunozm: Have you checked the note (line 4)?

src/interfaces/IPortfolio.sol
lemunozm: Do you mean `IERC7726 valuation`? Why an address?

src/libraries/MathLib.sol
lemunozm: I was following the file style for MathLib. I agree we should use custom error. Nevertheless, I would say it's out of scope to do this in this PR. But we can do in a following one

src/Portfolio.sol
lemunozm: - I tested, and `uint256(poolId) << 64 + itemId` is enough without the parenthesis. The shift is first done over the `uint256(poolId)`.

- Agree, `encodePacked` seems more elegant. 26 gas is deprecable, I guess. Good suggestion.

- Right, I'm shifting 64 when only 32 are needed. Will go with `encodePacked` anyways

src/Portfolio.sol
lemunozm: Why? This type is not used in the interface, only by the implementation.

src/Portfolio.sol
lemunozm: Shouldn't

src/Portfolio.sol
lemunozm: You're creating an underlying `Item`, note that `ItemInfo` is just another parameter used to create the internal `Item` with extra information.

src/NftEscrow.sol
lemunozm: It can be whatever, the upper layer through `auth` will handle that. Usually it will be the same who called lock() or some allowed account by them.

src/libraries/MathLib.sol
wischli: Agreeing with @lemunozm. Happy to create an issue for the custom error migration of MathLib and potentially other reused libs.

src/Portfolio.sol
lemunozm: Added the interface, which I think it makes sense: https://github.com/centrifuge/asset-pools/pull/24

test/unit/Portfolio.t.sol
```diff
@@ -0,0 +1,302 @@
...
+    }
+
+    function testWithInterest() public {
+        _prepareItem();
+
+        _mockDebt(int128(INCREASE_AMOUNT), int128(INCREASE_AMOUNT + ITEM_INTEREST));
+        _mockQuoteForQuantities(0, d18(0));
+        _mockModifyNormalizedDebt(
+            -int128(INTEREST), int128(INCREASE_AMOUNT), int128(INCREASE_AMOUNT + ITEM_INTEREST - INTEREST)
+        );
```
lemunozm: This is why I think the interaction against `LinearAccrual` should be done only by one interface method called `getAccRate()` instead of `increase/decrease (aka modify) /renormalize...`. [Thread](https://kflabs.slack.com/archives/C07PG2EUR9C/p1728631897006459). It is very difficult to mock `LinearAccrual` and know what's happening in those lines. And this kind of lines appears in all test cases.

Note that although integration tests are something that helps here, we need unitary tests anyways to check easily all the combinatory produced by increase/decrease with principal,interests,price changes,...

Maybe something we can discuss in the offsite @wischli @ilinzweilin @hieronx

lemunozm: I left some a doc showing the approaches: https://www.notion.so/Minimize-LinearAccrual-interactions-12e2eac24e1780b79784f49004fd6f43

wischli: Thanks for the great doc @lemunozm! I hope we can find consensus before the offsite to get some PRs closed and have more capacity for other open questions. Personally, I am in favor of option B. Nevertheless, I believe this should be done in a separate PR after merging #9 when everyone is happy.

lemunozm: Fully agree, in case of doing it, to do in a new PR

src/Portfolio.sol
```diff
@@ -0,0 +1,212 @@
...
+
+        item.outstandingQuantity = newOutsandingQuantity;
+        item.normalizedDebt =
+            linearAccrual.modifyNormalizedDebt(item.info.interestRateId, item.normalizedDebt, amount.toInt128());
+
+        emit DebtIncreased(poolId, itemId, amount);
+    }
+
+    /// @inheritdoc IPortfolio
+    function decreasePrincipalDebt(uint64 poolId, uint32 itemId, uint128 principal) public auth {
```
lemunozm, line 117: Does it make sense for you to use the term "face value" instead of "principal" here? @hieronx @mustermeiszer 

I ask this because it seems that principal is used only for what we called loans/internal pricing, not for external pricing items. Maybe "FaceValue" is a more generic term for this.

src/interfaces/IPortfolio.sol
peculiarity: Discard this

src/Portfolio.sol
peculiarity: This is the approach we used in LPs but frankly I think that it shouldn't be there. @hieronx  can elaborate more.
In general it is a best practice to define structs in contracts because structs are used to organize data ( sometimes only for gas optimization only ) and interfaces just defines the behavior of the contract. 
The think is that we have to agree on a common approach and not have different ways of doing the same thing.

src/Portfolio.sol
peculiarity: I think the question was related to the `isValue` field. If `collateralId` is `0` can we assume that it just doesn't exist hence do we need the `isValid` flag at all? I might be missing something though.

src/Portfolio.sol
lemunozm: I think interfaces should have as little information as possible. More information means more friction with the consumers of those interfaces. So, if they are minimal, we minimize the coupling between entities. 

I think this aligns with good developing practices, but open to listen other arguments

lemunozm:
Closed because now the source of truth will be the Accounting contract. 

Probably most of the code can be rescued to build the Bonds contract

PR #6, opened 2024-10-10
Add liquidity-pools as dependency

lemunozm:
Linked to `release-v2.0` branch

~~Currently, `liquidity-pools` uses Solidity `0.8.26`, but in our repo, we use `0.8.28`. Will this generate any issues when linking `liquidity-pool` stuff? Should we upgrade `liquidity-pools`?~~ No

### DIFF
--- a/.gitmodules
+++ b/.gitmodules
@@ -4,3 +4,6 @@
 [submodule "lib/forge-gas-snapshot"]
 	path = lib/forge-gas-snapshot
 	url = https://github.com/marktoda/forge-gas-snapshot
+[submodule "lib/liquidity-pools"]
+	path = lib/liquidity-pools
+	url = https://github.com/centrifuge/liquidity-pools
--- a/.gitmodules
+++ b/.gitmodules
@@ -1,9 +1,6 @@
 [submodule "lib/forge-std"]
 	path = lib/forge-std
 	url = https://github.com/foundry-rs/forge-std
-[submodule "lib/forge-gas-snapshot"]
-	path = lib/forge-gas-snapshot
-	url = https://github.com/marktoda/forge-gas-snapshot
 [submodule "lib/liquidity-pools"]
 	path = lib/liquidity-pools
 	url = https://github.com/centrifuge/liquidity-pools
--- a/remappings.txt
+++ b/remappings.txt
@@ -1,3 +1,5 @@
 @chimera/=lib/liquidity-pools/lib/chimera/src/
+chimera/=lib/liquidity-pools/lib/chimera/src/
+ds-test/=lib/liquidity-pools/lib/forge-std/lib/ds-test/src/
 forge-std/=lib/forge-std/src/
-liquidity-pools/=lib/liquidity-pools/src/
+liquidity-pools/=lib/liquidity-pools/
--- a/remappings.txt
+++ b/remappings.txt
@@ -1,5 +1,3 @@
 @chimera/=lib/liquidity-pools/lib/chimera/src/
-chimera/=lib/liquidity-pools/lib/chimera/src/
-ds-test/=lib/liquidity-pools/lib/forge-std/lib/ds-test/src/
 forge-std/=lib/forge-std/src/
 liquidity-pools/=lib/liquidity-pools/
--- a/remappings.txt
+++ b/remappings.txt
@@ -1,2 +1,3 @@
+@chimera/=lib/liquidity-pools/lib/chimera/src/
 forge-std/=lib/forge-std/src/
-forge-gas-snapshot/=lib/forge-gas-snapshot
+liquidity-pools/=lib/liquidity-pools/src/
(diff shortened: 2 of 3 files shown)


peculiarity:
Why do you need the `liquidity-pools` here? Looking for a specific type? In general it's not recommended to import modules.
Also you can remove `solc-version` and foundry should default to using `auto_detect_solc` ( by default it should be true and if not you can explicitly set it to true )

lemunozm:
Probably @hieronx can rationale this better. 

But TLDR. We will use some contracts already maintained by `liquidity-pools` as `MathLib` or `Auth`.

Maybe a more correct solution is to factorize some common centrifuge libraries and use them from `liquidity-pools` and `asset-pools`. Because I'm not sure if explicitly, we need something of `liquidity-pools` itself from `asset-pools`

lemunozm:
@hieronx could you give more info about my above comment?

Specially curious about whether we have this dependency because we need the common `libraries` from `liquidity pools` or if we will end up needing to call the business contract logic from `liquidity-pools`

hieronx:
> Specially curious about whether we have this dependency because we need the common libraries from liquidity pools or if we will end up needing to call the business contract logic from liquidity-pools

It is both. For now, we just need the libraries (but also Auth.sol, which is not a library. And likely we'll also re-use Escrow.sol). Later though, we likely will need the interfaces from liquidity-pools as well, once we start building the interactions.

Given `liquidity-pools` is already audited at the current commit hash, I would use a dependency.

hieronx (approved):

.gitmodules
```diff
@@ -4,3 +4,6 @@
 [submodule "lib/forge-gas-snapshot"]
 	path = lib/forge-gas-snapshot
 	url = https://github.com/marktoda/forge-gas-snapshot
+[submodule "lib/liquidity-pools"]
```
peculiarity, line 4: Can you put it as an entry into the `remappings.txt` ? 
so that we use `liquidity-pools` ...
We can even make it `@liquidity-pools/=/lib/liquidity-pools/src/`

hieronx: @lemunozm you can just run `forge remappings > remappings.txt` to regenerate the file

lemunozm: Good point!

lemunozm: Curious about what the `@` does?

lemunozm: > @lemunozm you can just run `forge remappings > remappings.txt` to regenerate the file

This command not only not add the `liquidity-pools` but it also removes the `forge-gas-snapshot` 🤔

lemunozm: Ok, I needed to update submodules first

remappings.txt
```diff
@@ -1,2 +1,3 @@
+@chimera/=lib/liquidity-pools/lib/chimera/src/
 forge-std/=lib/forge-std/src/
-forge-gas-snapshot/=lib/forge-gas-snapshot
```
hieronx, line 2: I think this should still not be removed. Not sure why it would be. @peculiarity any ideas?

peculiarity: Yeah, this shouldn't be removed @lemunozm .
I tried it locally an for some reason the command `forge remappings` removes it but no idea why. Will investigate further.

lemunozm: Could it be because it's recently archived? https://github.com/marktoda/forge-gas-snapshot

lemunozm: At least `git submodule update --init --recursive` no longer download that repo under `lib/`

hieronx: Ah cool,  it was upstreamed https://github.com/foundry-rs/foundry/pull/8952

hieronx: Not sure if that is the reason why, but it does mean we can remove the dependency, and rather just update foundry and use that going forward!

In that case, can you also remove the submodule?

peculiarity: https://github.com/marktoda/forge-gas-snapshot?tab=readme-ov-file - this is deprecated anyways. So probably we can remove it anyways .

lemunozm: Done. I've also called this again `forge remappings > remappings.txt` but with all submodules downloaded recursively

lemunozm: If that's not the correct way of do this I can revert it

peculiarity: The problem comes when manually add the entry in `.gitmodules`

remappings.txt
```diff
@@ -1,2 +1,5 @@
+@chimera/=lib/liquidity-pools/lib/chimera/src/
+chimera/=lib/liquidity-pools/lib/chimera/src/
+ds-test/=lib/liquidity-pools/lib/forge-std/lib/ds-test/src/
```
peculiarity: I think this can be remove

remappings.txt
```diff
@@ -1,2 +1,5 @@
+@chimera/=lib/liquidity-pools/lib/chimera/src/
+chimera/=lib/liquidity-pools/lib/chimera/src/
```
peculiarity: This also can be removed. We have it as first line.

peculiarity (approved):

remappings.txt
lemunozm: Then I guess I do not need to update submodules recursively

Issue #1, opened 2024-10-04
Interest accrual contract


hieronx:
https://github.com/centrifuge/tinlake/blob/main/src/borrower/pile.sol
https://github.com/centrifuge/tinlake-math/blob/f1dd5aeb79bccc9ecba7d8ea0fb834853420f00e/src/interest.sol

Spec: https://www.notion.so/Interest-Accrual-11a2eac24e1780eca702c117cbc487d1

PR #9, opened 2024-10-11
feat: linear accrual

wischli:
Related issue #1

Adds 
* library `Compounding` which provides means of mapping different compounding intervals to their duration in seconds as well as determining the number of full compounding periods which have passed for a given timestamp and compounding schedule
* contract `LinearAccrual` which provides means of mapping a `Group` to a rate identifier as well determining normalized debt for a given `rateId` and current `normalizedDebt`
* Assumes rates to be "prefixed" by 1, i.e. 5% is assumed to be represented by `1.05` (or in solidity `d18(1e18 + 5e16)`)

Will be done in follow-up PR
* `LinearAccrual` into library `DebtAccrual` and contract `LinearAccrual` to simplify debt mutation for consuming contracts

### TODO

- [x] Discuss `TODO(@review)` sections in code
- [x] Fuzz tests
- [x] Custom errors
- [x] Test custom errors
- [x] Reduce compounding periods to Secondly and Daily
- [x] Split `getRateId`


src/LinearAccrual.sol
```diff
@@ -0,0 +1,129 @@
+// SPDX-License-Identifier: AGPL-3.0-only
+pragma solidity 0.8.28;
+
+import {ILinearAccrual} from "./ILinearAccrual.sol";
+import {Compounding, CompoundingPeriod} from "./Compounding.sol";
+
+/// @dev Represents the rate accumulator and the timestamp of the last rate update
+struct Rate {
+    /// @dev Accumulated rate index over time
+    uint128 accumulatedRate;
```
lemunozm: Where the `accumulatedRate` is accumulated?

lemunozm (commented):
Wow!! It seems to just click the `deploy` buttom 💯 

Just a question after a fast overlook

src/LinearAccrual.sol
wischli: Inside drip, which is not fully implemented yet because I need to copy some math stuff. Will continue after lunch, just wanted to get it out.

src/Compounding.sol
```diff
@@ -0,0 +1,85 @@
...
+
+library Compounding {
+    uint256 constant SECONDS_PER_DAY = 86400; // 60 * 60 * 24
+    uint256 constant SECONDS_PER_YEAR = 31104000; // 360 days for a commercial year (360 * 86400)
+    uint256 constant SECONDS_PER_MONTH = SECONDS_PER_YEAR / 12;
+
+    /// @notice Returns the amount of seconds for the given compounding period.
+    ///
+    /// @dev    In finance, one year is composed of 12 months of 30 days such that 360 days
+    ///         represent a commercial year.
```
peculiarity: I guess you can pass this link for ref. https://en.wikipedia.org/wiki/360-day_calendar#Financial_use

src/Compounding.sol
```diff
@@ -0,0 +1,74 @@
...
+    function getSeconds(CompoundingPeriod period) public pure returns (uint256) {
+        if (period == CompoundingPeriod.Secondly) return 1;
+        if (period == CompoundingPeriod.Daily) return SECONDS_PER_DAY;
+        if (period == CompoundingPeriod.Quarterly) return SECONDS_PER_YEAR / 4; // 3 months
+        if (period == CompoundingPeriod.Biannually) return SECONDS_PER_YEAR / 2; // 6 months
+        if (period == CompoundingPeriod.Annually) return SECONDS_PER_YEAR; // 12 months
+        revert("invalid-compounding-period");
```
lemunozm: If this is infallible, why not an `if/else` chain avoiding the `revert()` (and gas?). Is this some kind of pattern? (I agree it is more beautiful than if/else chain)

src/Compounding.sol
```diff
@@ -0,0 +1,74 @@
...
+    }
+
+    /// @notice Returns the number of full compounding periods that have passed since a given timestamp.
+    ///
+    /// @dev    In finance, one year is composed of 12 months of 30 days such that 360 days
+    ///         represent a commercial year. Uses 0am UTC as the cutoff time for daily, quarterly, biannually, and
+    /// annually.
+    function getPeriodsPassed(CompoundingPeriod period, uint256 startTimestamp) public view returns (uint256) {
+        // TODO: Discuss revert vs. return
+        if (startTimestamp >= block.timestamp) return 0;
```
lemunozm: I think return `0` is fine. I see it as a valid result (that later can be treated as wrong if required)

src/Compounding.sol
```diff
@@ -0,0 +1,74 @@
...
+        }
+        revert("invalid-compounding-period");
+    }
+
+    /// @dev Get the year and month from a timestamp (approximate)
+    ///      Based on a commercial year of 360 days, with 12 months of 30 days each.
+    function getYearAndMonth(uint256 timestamp) internal pure returns (uint256 year, uint256 month) {
+        year = (timestamp / SECONDS_PER_YEAR);
+        uint256 daysIntoYear = (timestamp % SECONDS_PER_YEAR) / SECONDS_PER_DAY;
+        month = daysIntoYear / 30;
```
lemunozm: I'm not sure if this is correct. `timestamp` contains actual years with 365 days from 1970.

So the above divisions will shift the date 5 days for each year passed from 1970. Ending up in different months.

Dealing with dates is very horrible... I'm not sure if there exists some Date library to use, at least to subtract the years passed from 1970 accurately (as 365/366 days) before getting the months from the last year

src/LinearAccrual.sol
```diff
@@ -0,0 +1,112 @@
...
+    /// @param      normalizedDebt Normalized debt from which we derive the debt
+    function debt(bytes32 rateId, uint128 normalizedDebt) public view onlyUpdatedRate(rateId) returns (uint256) {
+        // TODO(@review): Discuss desired precision (10 ** 27 taken from tinlake)
+        return MathLib.mulDiv(normalizedDebt, rates[rateId].accumulatedRate, MathLib.One27);
+    }
+
+    /// @notice     Updates the accumulated rate of the corresponding identifier based on the periods which have passed
+    /// since the last update
+    /// @param      rateId the id of the interest rate group
+    function drip(bytes32 rateId) public {
```
lemunozm: Curious, how do we incentivize calling this? The rate always needs to be up to date to not make fail increase/decrease methods.

If this method is used as a pre-call in most use cases to ensure the rate is up to date before doing some operation, then it's possible that most of the time, for the same block, the rate is already updated by another call. 

NIT. Maybe we can skip computing `getPeriodsPassed` or some read if we check ASAP if `lastUpdated == block.timestamp` and make this noop in that case.

src/LinearAccrual.sol
```diff
@@ -0,0 +1,112 @@
...
+        Group memory group = Group(rate, period);
+
+        // TODO(@review): Discuss how to be future-proof if Group is altered which would lead to new hash
+        rateId = keccak256(abi.encode(group));
+        if (groups[rateId].ratePerPeriod == 0) {
+            groups[rateId] = group;
+            rates[rateId] = Rate(rate, uint64(block.timestamp));
+            emit NewRateId(rateId, rate, period);
+        }
+    }
```
lemunozm: Seems weirds that a method named `getXXX()` actually performs a write. Why not `registerRate()` or similar?

src/LinearAccrual.sol
```diff
@@ -0,0 +1,112 @@
...
+    mapping(bytes32 rateId => Group group) public groups;
+
+    modifier onlyUpdatedRate(bytes32 rateId) {
+        require(rates[rateId].accumulatedRate != 0, "rate-not-initialized");
+        require(rates[rateId].lastUpdated == block.timestamp, "rate-not-updated");
+        _;
+    }
+
+    /// @inheritdoc ILinearAccrual
+    function getRateId(uint128 rate, CompoundingPeriod period) public returns (bytes32 rateId) {
```
lemunozm: How `rate` input is specified? If it's an interestPerYear, but the period is secondly. It needs to be transformed to ratePerSeconds before set it as `accumulatedRate`.

I think all rates are usually always specified by year, and later, depending on the compounding period transformed to the period unit internally for compounding it

src/interfaces/ILinearAccrual.sol
```diff
@@ -0,0 +1,54 @@
...
+    event NewRateId(bytes32 rateId, uint128 rate, CompoundingPeriod period);
+    event RateAccumulated(bytes32 rateId, uint128 rate, uint256 periodsPassed);
+
+    /// @notice     Returns the rate identifier for the given rate and compound period.
+    ///
+    /// @dev        Initializes storage if the rate identifier has not existed yet.
+    ///
+    /// @param      rate Rate
+    /// @param      period Compounding schedule
+    function getRateId(uint128 rate, CompoundingPeriod period) external returns (bytes32 rateId);
```
lemunozm: The consumer of `getRateId()` differs from the consumer of the below methods. Are we sure `getRateId()` should be in this interface? Maybe it is just directly an external method of the contract impl instead?

src/LinearAccrual.sol
```diff
@@ -0,0 +1,112 @@
...
+struct Group {
+    /// @dev Rate per compound period
+    uint128 ratePerPeriod;
+    /// @dev Duration of compound period
+    CompoundingPeriod period;
+}
+
+contract LinearAccrual is ILinearAccrual {
+    mapping(bytes32 rateId => Rate rate) public rates;
+    mapping(bytes32 rateId => Group group) public groups;
```
lemunozm, line 33: Could `Rate` and `Group` be merged in the same struct? Why are they split? It seems they have the same lifetime

lemunozm (commented):
Amazing amount of work here! 🚀 

Some comments regarding the internal functionality. Most of them comments/questions, but I think two are important: how months are computed and how the accumulated rate is measured.

src/interfaces/ILinearAccrual.sol
```diff
@@ -0,0 +1,54 @@
...
+
+    /// @notice     Returns the sum of the current normalized debt and the normalized increment.
+    ///
+    /// @dev        Initializes storage if the rate identifier has not existed yet.
```
lemunozm: NIT. I think this comment is outdated (other occurrences)

src/Compounding.sol
wischli: I agree solely using `if` is probably not desired but this function cannot be infallible since enums are internally just `uint` such that you could call `getSeconds(CompoundingPeriod(type(uint256).max))`. Moreover, pattern matching only works in assembly AFAIK which does not support reverting with an error message, just with an integer.

src/LinearAccrual.sol
wischli: > NIT. Maybe we can skip computing `getPeriodsPassed` or some read if we check ASAP if `lastUpdated == block.timestamp` and make this noop in that case.

Thanks for pointing this out, I agree. This will save a read for `group` and computing `getPeriodsPassed`.

src/Compounding.sol
wischli: Thanks for pointing this out, I didn't consider that. I will add more fuzz UTs for bounds of quarters, half-years as well as years and hope they just turn out correctly. Otherwise, not sure at this point how to properly deal with that issue. 

If anyone knows a helpful library, please let me know @peculiarity @ilinzweilin @hieronx .

src/Compounding.sol
lemunozm: 🥲, then I should do the same in my PR, I guess, and follow your pattern for extra security

src/LinearAccrual.sol
wischli: > Seems weirds that a method named `getXXX()` actually performs a write. 

Valid point.

> Why not `registerRate()` or similar?

On the contrary, it feels incorrect to call `registerRate` just to get a rate, right? How about `getOrRegisterRateId`?

src/LinearAccrual.sol
wischli: Thanks for pointing this out. Do you agree that adding a `pure` getter function which renormalizes a rate from one compounding schedule to another would solve this issue?

src/interfaces/ILinearAccrual.sol
wischli: > The consumer of `getRateId()` differs from the consumer of the below methods.

Can you elaborate on why you believe the consumer differs? I disagree but am lacking knowledge of the `Portfolio` contract which I believe to be the main consumer. IMO, it wants to deal with normalized debt (getting or updating) and in order to do that, first needs to get the corresponding `rateId`.

src/interfaces/ILinearAccrual.sol
wischli: Thanks, remnant from older impl.

src/LinearAccrual.sol
wischli: IMO, they cannot because this would require the `Portfolio` contract to have more knowledge than just the `rateId` for an item which is used to identify the corresponding rate `Group`. It's true, both have the same lifetime in `LinearAccrual` but it increases storage overload for `Portfolio`. Moreover, `Group` might change over time by adding new fields.

src/LinearAccrual.sol
lemunozm: 🤔, I think _register_ is not "wrong". I mean, the intention of calling that method is to register it, but if it is already found, it will return the already existent (so the method becomes idempotent, an idempotent registration).

If you're not convinced, maybe `useRate()` also fits with the behavior

src/interfaces/ILinearAccrual.sol
lemunozm: The portfolio already stores the `rateId`, but the portfolio doesn't know the value of that rate. It does not belong to the portfolio to know about that. So the user will configure the rate value directly in your contract, and later will use the id in the portfolio. The same applies when updating the rate in the portfolio

src/LinearAccrual.sol
lemunozm: I am not sure if I get your point, but I think we're not fully aligned, given the thread below. 

Basically, I miss these maths: https://github.com/centrifuge/centrifuge-chain/blob/main/pallets/interest-accrual/src/lib.rs#L559 to transform from "perYear" to "perPeriod". However, I do not need to use that from the portfolio. It's something internal to the `LinearAccrual`

src/LinearAccrual.sol
lemunozm: > it increases storage overload for `Portfolio`

That seems a valid contra-argument

src/LinearAccrual.sol
wischli: That's what I referred to, i.e. a conversion of rate from one `CompoundingPeriod` to another one. 

> However, I do not need to use that from the portfolio. It's something internal to the `LinearAccrual`.

Not sure if that is true because currently, the `rateId` is determined by the underlying `Group` and therefore the compounding schedule of that group. So when a consumer wants to get (or register) the rate id for some given rate, they need to know the compounding schedule and the corresponding rate per period.

lemunozm: > they need to know the compounding schedule and the corresponding rate per period.

Ok, I know what's happening. From the `Portfolio` I'm just receiving a `RateId`, but probably the top-level contract that uses the `Portfolio` will require to "register" that Id based on that rate value and compounding.

But anyway, no contract will still require to use that conversion method. I think (if I'm not wrong) all rates values will be given in "per year". So when a compounding is set, that "per year" must be transformed to be operated with internally in LinearAccrual. But from the outside it will be like:

Transform this 0.2% APR with annual compounding into this 0.3% APR with daily compounding.

But the user never treats the internal rate measured in the compounding units (that you need to have to accumulate correclty). Not sure If I got explaining correctly

wischli: @hieronx Curious about your thoughts:

When some top-level `X` contracts retrieves or registers a rate id, is it wrong to assume that `X` knows about the rate per period or is it more realistic that it knows about the annual rate such that `LinearAccrual` should calculate the rate per period internally based on the given compounding schedule?

src/interfaces/ILinearAccrual.sol
lemunozm: This is related to some point in the thread above, and now I think it could be in the same interface 👍🏻

src/LinearAccrual.sol
wischli: As discussed, the current implementation of `getRateId` needs to split into a `pure` getter (i.e. `getRateId`) and a callable `register` function (i.e. `registerRateId`). The assumed consumer of those functions is the user, not the `Portfolio` contract.

src/Compounding.sol
wischli: As discussed, will refactor to only support secondly and daily due to the above issue.

src/LinearAccrual.sol
wischli: As discussed, we need both storages in order to reduce the likelihood of updating the `Portfolio` contract as much as possible. This design allows `Portfolio` to only know about the `rateId`, any rate specific data can be queried via `LinearAccrual` storage or getters.

src/LinearAccrual.sol
```diff
@@ -0,0 +1,112 @@
...
+
+    /// @inheritdoc ILinearAccrual
+    function getRateId(uint128 rate, CompoundingPeriod period) public returns (bytes32 rateId) {
+        Group memory group = Group(rate, period);
+
+        // TODO(@review): Discuss how to be future-proof if Group is altered which would lead to new hash
+        rateId = keccak256(abi.encode(group));
+        if (groups[rateId].ratePerPeriod == 0) {
+            groups[rateId] = group;
+            rates[rateId] = Rate(rate, uint64(block.timestamp));
```
lemunozm: Do we need to support `rates == 0` at this level (i.e. interestRate of 0%)?

If so, we need some kind of protection to deal with an `accumulatedRate == 0` after the creation of this Rate. Because that value acts as a guard in `onlyUpdateRate()`.

wischli: I don't think we need to support that because any zero rate is constant independent of the underlying compounding period.

lemunozm: But methods will panic when they reach `onlyUpdateRate()` if 0, right? I mean, from the portfolio, I do not have any way of setting a zero value, I just have a `RateId`. Not sure how to represent a zero-rate from my side 🤔.

Also keep in mind that rates can be mutated, so maybe some computation results in a zero rate...

src/Compounding.sol
wischli: I was wrong, when I try to cast an integer with value `x` to an enum which has less than `x` elements, then Solidity reverts with `panic: failed to convert value into enum type (0x21)`. So it seems impossible to reach the `revert` line here. I will remove it and just use the `Compounding.Secondly` as default case.

src/LinearAccrual.sol
wischli: > But methods will panic when they reach `onlyUpdateRate()` if 0, right?

Yes. I could change the check to 
```diff
    modifier onlyUpdatedRate(bytes32 rateId) {
-        require(rates[rateId].accumulatedRate != 0, RateIdMissing(rateId));
+        require(rates[rateId].lastUpdated != 0, RateIdMissing(rateId));
```

> Also keep in mind that rates can be mutated, so maybe some computation results in a zero rate...

So far, mutation is only possible via `drip` and mathematically speaking, that's a strong monotone increasing series under the assumption that rates of `x%` are represented as `100% + x%`.

src/LinearAccrual.sol
```diff
@@ -0,0 +1,127 @@
...
+    }
+
+    /// @inheritdoc ILinearAccrual
+    function increaseNormalizedDebt(bytes32 rateId, uint128 prevNormalizedDebt, uint128 debtIncrease)
+        external
+        view
+        onlyUpdatedRate(rateId)
+        returns (uint128 newNormalizedDebt)
+    {
+        // TODO(@review): Discuss if precions better if we do (prev * rate + debtIncrease) / rate
```
wischli: At reviewers: Curious about your opinion of this matters.

src/LinearAccrual.sol
```diff
@@ -0,0 +1,127 @@
...
+        require(rate.accumulatedRate != 0, RateIdMissing(rateId));
+
+        // Short circuit to save gas
+        if (rate.lastUpdated == uint64(block.timestamp)) {
+            return;
+        }
+
+        // Infallible since group storage exists iff rate storage exists
+        Group memory group = groups[rateId];
+        require(group.ratePerPeriod != 0, GroupMissing(rateId));
```
wischli: This line is missing coverage because it is infallible. Curious on how to deal with such cases.

src/LinearAccrual.sol
lemunozm: I would really like to have the check in the way you've shared here. Basically, to be able to represent 0 interest rates from my side.

wischli: Done in [c551c3e](https://github.com/centrifuge/asset-pools/pull/9/commits/c551c3e336ebe0726c2f9a07e9f56354370f40cb)

src/interfaces/ILinearAccrual.sol
```diff
@@ -0,0 +1,59 @@
+// SPDX-License-Identifier: AGPL-3.0-only
+pragma solidity 0.8.28;
+
+import {CompoundingPeriod} from "../Compounding.sol";
+
+interface ILinearAccrual {
```
lemunozm, line 6: I'm pretty sure I'll need in this interface a `debt()` method to read the real value from the `Portfolio`

wischli: Since `debt` is declared as `public` in `LinearAccrual`, you can use it. Unfortunately, it cannot be part of the corresponding interface because interface defined functions are restricted to `external` whereas `LinearAccrual` needs to use it internally as well.

lemunozm: You can inherit a method and change the publicity. So the interface define it as `external` but the implementation is `public`.

I only can use `debt()` if the portfolio knows your implementation, but it only knows about the `ILinearAccrual`

src/LinearAccrual.sol
lemunozm: Adding more doubts 😆: 

Should infallible errors have custom errors that can never be tested or read? In Substrate we used `Other(string)` for those (which here would be just an string). Not sure what's the best practice here.

src/LinearAccrual.sol
lemunozm: Thanks William!

As ref, the first impl of interest-accrual also discarded 0 rates, and later, we had a lot of problems with that, because it's super easy to ending up with a 0 rate in the configurations. But you only reach that error when the loan/item is created and start managing it.

Also, note that this contract will likely be used for DFC (DiscountCashFlows) and other purposes.

src/Compounding.sol
lemunozm: Ohh, nice to know about that. Thanks for sharing!

src/interfaces/ILinearAccrual.sol
wischli: > You can inherit a method and change the publicity. So the interface define it as `external` but the implementation is `public`.

Somehow I missed this during my learning session. That's incredible good news for me.

Added `debt` to the interface in [ba87b7b](https://github.com/centrifuge/asset-pools/pull/9/commits/ba87b7bcc83f19f6cb0a9e52958b8a5ccbdf7e33)

src/LinearAccrual.sol
wischli: > Should infallible errors have custom errors that can never be tested or read? In Substrate we used `Other(string)` for those (which here would be just an string). Not sure what's the best practice here.

Probably not but will keep for now until one of the Solidity experts has reviewed.

src/LinearAccrual.sol
wischli: After discussing about the Cash Account, it became absolutely necessary. I was in the accrual bubble. Thanks for sharing that! Super curious how this contract will change once we aligned on negative interest 😅

src/LinearAccrual.sol
```diff
@@ -0,0 +1,128 @@
...
+    mapping(bytes32 rateId => Group group) public groups;
+
+    /// @inheritdoc ILinearAccrual
+    function getRateId(uint128 rate, CompoundingPeriod period) public pure returns (bytes32) {
+        Group memory group = Group(rate, period);
+
+        return keccak256(abi.encode(group));
+    }
+    /// @inheritdoc ILinearAccrual
+
```
wischli, line 58: I'd love to know why `forge fmt` adds a newline here.. CI fails if I remove it manually..

src/LinearAccrual.sol
```diff
@@ -0,0 +1,128 @@
...
+    /// @inheritdoc ILinearAccrual
+    function renormalizeDebt(bytes32 oldRateId, bytes32 newRateId, uint128 prevNormalizedDebt)
+        external
+        view
+        returns (uint128 newNormalizedDebt)
+    {
+        _requireUpdatedRateId(oldRateId);
+        _requireUpdatedRateId(newRateId);
+
+        uint256 _debt = debt(oldRateId, prevNormalizedDebt);
```
hieronx: Solidity style nit: prefix underscore is for internal variables, postfix underscore is for avoiding variable shadowing
```suggestion
        uint256 debt_ = debt(oldRateId, prevNormalizedDebt);
```

src/LinearAccrual.sol
```diff
@@ -0,0 +1,128 @@
...
+        return MathLib.toUint128(_debt / rates[newRateId].accumulatedRate);
+    }
+
+    /// @inheritdoc ILinearAccrual
+    function debt(bytes32 rateId, uint128 normalizedDebt) public view returns (uint256) {
+        _requireUpdatedRateId(rateId);
+
+        return MathLib.mulDiv(normalizedDebt, rates[rateId].accumulatedRate, MathLib.One18);
```
hieronx: Can also do:
```suggestion
                return normalizedDebt.mulDiv(rates[rateId].accumulatedRate, MathLib.One18);
```
if you add `using MathLib for uint128` at the top of the contract.

hieronx (commented):
Will take a fresh look at the compounding logic tomorrow, but from a Solidity coding perspective this looks amazing! Really nice work @wischli !

src/interfaces/ILinearAccrual.sol
lemunozm: Thanks so much! 🙌🏻

src/LinearAccrual.sol
lemunozm: Hopefully, it's just from `uint128` to `int128` 😆. I would say a negative debt should still be accrued, but in negative

src/interfaces/ILinearAccrual.sol
```diff
@@ -0,0 +1,64 @@
...
+
+    /// @notice     Returns the difference of the current normalized debt and the normalized decrement.
+    ///
+    /// @param      rateId Identifier of the rate group
+    /// @param      prevNormalizedDebt Normalized debt before decreasing
+    /// @param      debtDecrease The amount by which we decrease the debt
+    function decreaseNormalizedDebt(bytes32 rateId, uint128 prevNormalizedDebt, uint128 debtDecrease)
+        external
+        view
+        returns (uint128 newNormalizedDebt);
```
lemunozm: Once we move to support "negative" debt using `int128` types, does it make sense to unify these two methods into one `modifyNormalizedDebt` or similar, accepting an `int128` variation?

I think the only purpose of having two of them now is to differentiate if the current `uint128` means addition or subtraction. This no longer affect us using `int128`

src/LinearAccrual.sol
```diff
@@ -0,0 +1,129 @@
...
+
+        rateId = keccak256(abi.encode(group));
+        if (groups[rateId].ratePerPeriod == 0) {
+            groups[rateId] = group;
+            // TODO(@wischli): Some source stated another timestamp should be used instead of block.timestamp
+            rates[rateId] = Rate(ratePerPeriod, uint64(block.timestamp));
+            emit NewRateId(rateId, ratePerPeriod, period);
+        } else {
+            revert RateIdExists(rateId, ratePerPeriod, period);
+        }
```
hieronx: More idiomatic solidity would be

```suggestion
        require(groups[rateId].ratePerPeriod == 0, RateIdExists(rateId, ratePerPeriod, period));

        groups[rateId] = group;
         // TODO(@wischli): Some source stated another timestamp should be used instead of block.timestamp
        rates[rateId] = Rate(ratePerPeriod, uint64(block.timestamp));
        emit NewRateId(rateId, ratePerPeriod, period);
```

src/LinearAccrual.sol
```diff
@@ -0,0 +1,129 @@
...
+
+contract LinearAccrual is ILinearAccrual {
+    using MathLib for uint128;
+    using MathLib for uint256;
+
+    mapping(bytes32 rateId => Rate rate) public rates;
+    mapping(bytes32 rateId => Group group) public groups;
+
+    /// @inheritdoc ILinearAccrual
+    function getRateId(uint128 rate, CompoundingPeriod period) public pure returns (bytes32) {
```
hieronx: Generally view methods go to the bottom, below any functions that modify state.

src/LinearAccrual.sol
```diff
@@ -0,0 +1,129 @@
...
+    }
+
+    /// @inheritdoc ILinearAccrual
+    function debt(bytes32 rateId, uint128 normalizedDebt) public view returns (uint256) {
+        _requireUpdatedRateId(rateId);
+
+        return normalizedDebt.mulDiv(rates[rateId].accumulatedRate, MathLib.One18);
+    }
+
+    /// @notice     Updates the accumulated rate of the corresponding identifier based on the periods which have passed
```
hieronx: Comments should move to the interface

src/LinearAccrual.sol
```diff
@@ -0,0 +1,129 @@
...
+
+        // Infallible since group storage exists iff rate storage exists
+        Group memory group = groups[rateId];
+        require(group.ratePerPeriod != 0, "group-missing");
+
+        // Determine number of full compounding periods passed since last update
+        uint256 periodsPassed = Compounding.getPeriodsPassed(group.period, rate.lastUpdated);
+
+        if (periodsPassed > 0) {
+            rate.accumulatedRate = group.ratePerPeriod.rpow(periodsPassed, MathLib.One18).mulDiv(
```
hieronx: Should use D18 I presume

src/libraries/MathLib.sol
```diff
@@ -2,7 +2,7 @@
 pragma solidity 0.8.28;
 
 library MathLib {
-    uint256 public constant One27 = 10 ** 27;
+    uint256 public constant One18 = 10 ** 18;
     /// @notice Returns x^n with rounding precision of base
```
hieronx, line 5:
```suggestion

    /// @notice Returns x^n with rounding precision of base
```

src/Compounding.sol
```diff
@@ -0,0 +1,34 @@
+// SPDX-License-Identifier: AGPL-3.0-only
+pragma solidity 0.8.28;
```
hieronx: This file should be in `src/libraries`

src/Compounding.sol
```diff
@@ -0,0 +1,34 @@
...
+    Secondly,
+    Daily
+}
+
+library Compounding {
+    uint256 constant SECONDS_PER_DAY = 86400; // 60 * 60 * 24
+
+    /// @notice Returns the amount of seconds for the given compounding period.
+    ///
+    /// @dev    Default case is `CompoundingPeriod.Secondly`.
```
hieronx:
```suggestion
    /// @dev    Default case is `CompoundingPeriod.Daily`.
```

src/Compounding.sol
```diff
@@ -0,0 +1,34 @@
...
+    function getSeconds(CompoundingPeriod period) public pure returns (uint256) {
+        if (period == CompoundingPeriod.Daily) return SECONDS_PER_DAY;
+        else return 1;
+    }
+
+    /// @notice Returns the number of full compounding periods that have passed since a given timestamp.
+    ///
+    /// @dev    Default case is `CompoundingPeriod.Secondly` and returns 0 for any given future timestamp.
```
hieronx:
```suggestion
    /// @dev    Default case is `CompoundingPeriod.Daily` and returns 0 for any given future timestamp.
```

src/LinearAccrual.sol
```diff
@@ -0,0 +1,129 @@
...
+    {
+        _requireUpdatedRateId(oldRateId);
+        _requireUpdatedRateId(newRateId);
+
+        uint256 debt_ = debt(oldRateId, prevNormalizedDebt);
+        return (debt_ / rates[newRateId].accumulatedRate).toUint128();
+    }
+
+    /// @inheritdoc ILinearAccrual
+    function debt(bytes32 rateId, uint128 normalizedDebt) public view returns (uint256) {
```
hieronx: Why return `uint256` here rather than `uint128`? Since we use this type for normalized debt as well.

src/Compounding.sol
```diff
@@ -0,0 +1,34 @@
...
+    Daily
+}
+
+library Compounding {
+    uint256 constant SECONDS_PER_DAY = 86400; // 60 * 60 * 24
+
+    /// @notice Returns the amount of seconds for the given compounding period.
+    ///
+    /// @dev    Default case is `CompoundingPeriod.Secondly`.
+    function getSeconds(CompoundingPeriod period) public pure returns (uint256) {
```
hieronx: Why return `uint256` here? I think `uint64` should suffice (would just need to be cast)

src/LinearAccrual.sol
lemunozm: `int128` now that we can support negative debt (?)

src/LinearAccrual.sol
wischli: Would be great if the style guide were updated with function declaration order. I added this as a suggest: https://www.notion.so/Solidity-style-guide-1152eac24e17809ba9f2f81d5b30ba62?pvs=4#1152eac24e1780efa2f9cb3a1e3eca2a

src/LinearAccrual.sol
wischli: Is that the case already for this PR? Would be good to know @hieronx

lemunozm: At least from the portfolio, right now, it can call to `decreaseNormalizedDebt()` with a decrease amount higher enough to have a negative value. That casuistic exist now

hieronx: We did decide we want to support negative debt. I don't have strong feelings whether that needs to be in this PR, I would say it can easily be a followup PR as well to get this one merged (and until then integration tests between portfolio contract and linear accrual with negative debt won't work, which is not really an issue).

src/LinearAccrual.sol
hieronx: Good point, will expand on this

src/Compounding.sol
```diff
@@ -0,0 +1,34 @@
...
+        if (period == CompoundingPeriod.Daily) return SECONDS_PER_DAY;
+        else return 1;
+    }
+
+    /// @notice Returns the number of full compounding periods that have passed since a given timestamp.
+    ///
+    /// @dev    Default case is `CompoundingPeriod.Secondly` and returns 0 for any given future timestamp.
+    function getPeriodsPassed(CompoundingPeriod period, uint256 startTimestamp) public view returns (uint256) {
+        if (startTimestamp >= block.timestamp) {
```
ilinzweilin: don't we need to normalize the `startTimestamp` and the `block.timestamp` to 00:00 UTC of the corresponding day

src/LinearAccrual.sol
ilinzweilin:
```suggestion
    function getRateId(uint128 rate, CompoundingPeriod period) public view returns (bytes32) {
```

src/LinearAccrual.sol
```diff
@@ -0,0 +1,129 @@
...
+    /// @inheritdoc ILinearAccrual
+
+    function registerRateId(uint128 ratePerPeriod, CompoundingPeriod period) public returns (bytes32 rateId) {
+        Group memory group = Group(ratePerPeriod, period);
+
+        rateId = keccak256(abi.encode(group));
+        if (groups[rateId].ratePerPeriod == 0) {
+            groups[rateId] = group;
+            // TODO(@wischli): Some source stated another timestamp should be used instead of block.timestamp
+            rates[rateId] = Rate(ratePerPeriod, uint64(block.timestamp));
```
ilinzweilin: I think the timestamp also has to be normalized to UTC 00:00

src/LinearAccrual.sol
```diff
@@ -0,0 +1,129 @@
...
+            ).toUint128();
+
+            emit RateAccumulated(rateId, rate.accumulatedRate, periodsPassed);
+            rate.lastUpdated = uint64(block.timestamp);
+        }
+    }
+
+    function _requireUpdatedRateId(bytes32 rateId) internal view {
+        require(rates[rateId].lastUpdated != 0, RateIdMissing(rateId));
+        require(rates[rateId].lastUpdated == block.timestamp, RateIdOutdated(rateId, rates[rateId].lastUpdated));
```
ilinzweilin, line 139: I think we need to also work with normalized timestamps here

src/LinearAccrual.sol
```diff
@@ -0,0 +1,129 @@
...
+            // TODO(@wischli): Some source stated another timestamp should be used instead of block.timestamp
+            rates[rateId] = Rate(ratePerPeriod, uint64(block.timestamp));
+            emit NewRateId(rateId, ratePerPeriod, period);
+        } else {
+            revert RateIdExists(rateId, ratePerPeriod, period);
+        }
+    }
+
+    /// @inheritdoc ILinearAccrual
+    function increaseNormalizedDebt(bytes32 rateId, uint128 prevNormalizedDebt, uint128 debtIncrease)
```
ilinzweilin: since this function doesn't change any state, I would rename to `getIncreasedNormalizedDebt`

ilinzweilin: since this function doesn't change any state, I would rename to `getIncreasedNormalizedDebt`

src/LinearAccrual.sol
ilinzweilin: I would need to see in a test :)

src/LinearAccrual.sol
```diff
@@ -0,0 +1,129 @@
...
+        returns (uint128 newNormalizedDebt)
+    {
+        _requireUpdatedRateId(rateId);
+
+        // TODO(@review): Discuss if precions better if we do (prev * rate + debtIncrease) / rate
+        return prevNormalizedDebt + debtIncrease / rates[rateId].accumulatedRate;
+    }
+
+    /// @inheritdoc ILinearAccrual
+    function decreaseNormalizedDebt(bytes32 rateId, uint128 prevNormalizedDebt, uint128 debtDecrease)
```
ilinzweilin: same here, would also name `getDecreaseNormalizedDeb`

src/LinearAccrual.sol
```diff
@@ -0,0 +1,129 @@
...
+    function debt(bytes32 rateId, uint128 normalizedDebt) public view returns (uint256) {
+        _requireUpdatedRateId(rateId);
+
+        return normalizedDebt.mulDiv(rates[rateId].accumulatedRate, MathLib.One18);
+    }
+
+    /// @notice     Updates the accumulated rate of the corresponding identifier based on the periods which have passed
+    /// since the last update
+    /// @param      rateId the id of the interest rate group
+    function drip(bytes32 rateId) public {
```
ilinzweilin: we also need to use normalized timestamps here as well

ilinzweilin (commented):
currently `drip` has to be called from an external source before increasing / decreasing / retrieving normalized debt. 
I wonder if we should the functions shouldn't be view and we always call drip from within those functions, so that the external contracts don't need to worry about that...

src/LinearAccrual.sol
wischli: Can you elaborate on you prefer `view` over `pure` if this function is indeed `pure`?

src/Compounding.sol
wischli: To the best of my knowledge, the timestamp returned by `block.timestamp` is already in UTC (see [stackexchange](https://ethereum.stackexchange.com/a/68065)). So there is no need to renormalize it anywhere.

ilinzweilin: what I mean is shouldn't the periods always start at 00:00 UTC.

src/LinearAccrual.sol
ilinzweilin: I think it's both :D

src/Compounding.sol
```diff
@@ -0,0 +1,34 @@
...
+            return 0;
+        } else if (period == CompoundingPeriod.Daily) {
+            uint256 startDay = startTimestamp / SECONDS_PER_DAY;
```
ilinzweilin: From my understanding you either need to normalize timestamps to 00:00 of the corresponding day or use the modulo operator here.

src/LinearAccrual.sol
wischli: That should be impossible by definition? `pure` doesnt read state whereas `read` does.

src/Compounding.sol
lemunozm: I agree with William here. It should be the division, and the remainder should be discarded. Why do you say to use the modulo?

wischli: It cannot be modulo because that operation would only cycle over the remainders throughout one day, i.e. `today % SEC_PER_DAY == tomorrow_same_time % SEC_PER_DAY` because modulo forms a mathmatical group.

wischli: @ilinzweilin No need to normalize to UTC as my test proves [767c595](https://github.com/centrifuge/asset-pools/pull/9/commits/767c595cb907af43f52e050a5fe859d0f672dea9)

src/LinearAccrual.sol
wischli: I wrote one and couldn't catch any precision loss. Removed the comment without adding the test.

src/libraries/MathLib.sol
wischli: This is automatically reverted via `forge fmt`

src/Compounding.sol
```diff
@@ -0,0 +1,34 @@
...
+    /// @dev    Default case is `CompoundingPeriod.Daily`.
+    function getSeconds(CompoundingPeriod period) public pure returns (uint64) {
+        if (period == CompoundingPeriod.Daily) return SECONDS_PER_DAY;
+        else return 1;
+    }
+
+    /// @notice Returns the number of full compounding periods that have passed since a given timestamp.
+    ///
+    /// @dev    Default case is `CompoundingPeriod.Daily` and returns 0 for any given future timestamp.
+    function getPeriodsPassed(CompoundingPeriod period, uint64 startTimestamp) public view returns (uint64) {
```
mustermeiszer: Couldn't we make the PERIOD parameter? Just thinking out loud ^^

src/Compounding.sol
```diff
@@ -0,0 +1,34 @@
...
+        if (startTimestamp >= block.timestamp) {
+            return 0;
+        } else if (period == CompoundingPeriod.Daily) {
+            uint64 startDay = startTimestamp / SECONDS_PER_DAY;
+            uint64 nowDay = uint64(block.timestamp) / SECONDS_PER_DAY;
```
mustermeiszer: What is the reasoning for not doing

```
(startTimestamp - uint64(block.timestamp))/SECONDS_PER_DAY
```

src/LinearAccrual.sol
```diff
@@ -0,0 +1,127 @@
...
+
+        if (periodsPassed > 0) {
+            rate.accumulatedRate = d18(
+                rate.accumulatedRate.mulInt(
+                    uint256(group.ratePerPeriod.inner()).rpow(periodsPassed, d18(1e18).inner()).toUint128()
+                )
+            );
+
+            emit RateAccumulated(rateId, rate.accumulatedRate, periodsPassed);
+            rate.lastUpdated = uint64(block.timestamp);
```
mustermeiszer: Is this the same as 
image

where 

- A = future value of the investment
- P = principal amount
- r = interest rate per period
- d = number of full periods passed
- n = number of full compounding periods passed

I can not see how one could say, I want to have daily interest but only 180 daily compounding.

src/libraries/Compounding.sol
```diff
@@ -0,0 +1,34 @@
...
+    /// @notice Returns the number of full compounding periods that have passed since a given timestamp.
+    ///
+    /// @dev    Default case is `CompoundingPeriod.Daily` and returns 0 for any given future timestamp.
+    function getPeriodsPassed(CompoundingPeriod period, uint64 startTimestamp) public view returns (uint64) {
+        if (startTimestamp >= block.timestamp) {
+            return 0;
+        } else if (period == CompoundingPeriod.Daily) {
+            uint64 startDay = startTimestamp / SECONDS_PER_DAY;
+            uint64 nowDay = uint64(block.timestamp) / SECONDS_PER_DAY;
+            return nowDay - startDay;
```
mustermeiszer, line 29: What is the difference too 

```
(startTimestamp - block.timestamp)) / SECONDS_PER_DAY
```

src/LinearAccrual.sol
```diff
@@ -51,75 +74,51 @@ contract LinearAccrual is ILinearAccrual {
...
         external
         view
         returns (uint128 newNormalizedDebt)
     {
         _requireUpdatedRateId(oldRateId);
         _requireUpdatedRateId(newRateId);
 
-        uint256 debt_ = debt(oldRateId, prevNormalizedDebt);
-        return (debt_ / rates[newRateId].accumulatedRate).toUint128();
+        return (debt(oldRateId, prevNormalizedDebt) / rates[newRateId].accumulatedRate.inner()).toUint128();
```
lemunozm: NIT. In this call, `_requireUpdateRateId()` is called again; maybe we can directly add the multiplication here (?)

wischli: If it's okay, I will handle this as part of the refactor in a follow-up PR because none of the library functions will call `_requireUpdateRateId`.

src/Compounding.sol
wischli: Can you elaborate please? I don't get where you are going.

src/Compounding.sol
wischli: Yeah good point, that works too and is a bit less code.

src/LinearAccrual.sol
wischli: > I can not see how one could say, I want to have daily interest but only 180 daily compounding.

So even if an entire year has passed, only 180 days would account for compounding?

I think we need to discuss this at the offsite. The current design does not support this. We decided to stick with secondly and daily compounding period schedules. IIUC, we would have to split the current `Group.period` into two periods?!

```diff
struct Group {
    D18 ratePerPeriod;
    /// @dev Duration of compound period
-    CompoundingPeriod period;
+    CompoundingPeriod compoundingPeriod;
+    /// @dev Duration of interest period
+    CompoundingPeriod interestPeriod;
}

File: src/Compounding.sol (hunk 38 above)
Comment (lemunozm) [reply src/Compounding.sol]: I do not think this works. Mathematically it's not equivalent if using integers, that equality is only valid for real numbers.

Comment (lemunozm) [reply src/Compounding.sol]: i.e: 

```
180 / 100 - 90 / 100 = 1
(180 - 90) / 100 = 0
```

File: src/libraries/Compounding.sol (hunk 40 above)
Comment (lemunozm) [reply src/libraries/Compounding.sol:29]: There is a difference, see my comment above

File: src/Compounding.sol (hunk 38 above)
Comment (wischli) [reply src/Compounding.sol]: Good point thanks @lemunozm ! Now that I think about it, I first took @mustermeiszer's approach and then went into this direction thanks to unit tests.

If those numbers were `real` numbers, it would have to work out. However, we are applying floor here and `floor(x1 / y) - floor(x2 / y) != floor((x1 - x2) / y)` for all integers x1, x2 and y as it not distributive.

File: src/interfaces/ILinearAccrual.sol [hunk 42]
```diff
@@ -0,0 +1,70 @@
...
+    ///
+    /// @param      ratePerPeriod Rate per compound period
+    /// @param      period Compounding schedule
+    function registerRateId(D18 ratePerPeriod, CompoundingPeriod period) external returns (bytes32 rateId);
+
+    /// @notice     Returns the rate identifier for the given rate and compound period.
+    ///
+    /// @param      ratePerPeriod Rate per compound period
+    /// @param      period Compounding schedule
+    function getRateId(D18 ratePerPeriod, CompoundingPeriod period) external pure returns (bytes32 rateId);
```
lemunozm: Just a thought. Using `D18` here means that the consumers of this API are tied to an implementation of `LinearAccrual` that uses 18 decimals (although, i.e. the portfolio doesn't need that information). 

If we open it and use a plain `uint128` in the interface (only in the interface), the consumers of this interface can work fine if later we modify to a new implementation using 27 decimals, for example.

Just a random thought, but interested in your wisdom about this @hieronx

lemunozm: This would also affect the DebtAccrual pure lib interface, which should use a generic `uint128` as an accumulator rate parameter to not force 18 decimals forever

src/LinearAccrual.sol
```diff
@@ -0,0 +1,127 @@
...
+
+    /// @inheritdoc ILinearAccrual
+    function getIncreasedNormalizedDebt(bytes32 rateId, uint128 prevNormalizedDebt, uint128 debtIncrease)
+        external
+        view
+        returns (uint128 newNormalizedDebt)
+    {
+        _requireUpdatedRateId(rateId);
+
+        return prevNormalizedDebt + debtIncrease / rates[rateId].accumulatedRate.inner();
```
lemunozm: I'm a bit lost here @wischli, if using the `.inner()` value of the rate to perform the division, have we lost the decimal meaning?

lemunozm: Suppose we use 27 decimals. The result will be the same but 10^9 times greater 🤔

wischli: You might be right, will check later. Good catch!

src/LinearAccrual.sol
wischli: Will actually be fixed with normalized debt for signed integers in the next commit. Thanks for spotting!

src/LinearAccrual.sol
wischli: You are right! Ive started fixing it, will be part of the `int128` refactor. Eagle eyes!

lemunozm: Curious, should this have been caught in the tests?

wischli: Definitely! Unfortunately, the same issue was repeated in the test. In the future, I will always add at least one hardcoded test for which I know the result such that this should not occur anymore.

lemunozm: Yeah, I had the same feeling in some tests. Some issues could be reproduced if it's too generic.

Ok, thanks William! TBH, I thought I was not right, and somehow the normalizedDebt already took care of the decimals 😆

wischli: https://github.com/centrifuge/asset-pools/pull/9/commits/1f059597c0b0e766443a876228b7c01197f1f2e3#diff-34058dace177b7c1e503c3c1c962688fd0a113ca0442c3887e3ad436b920a0a2R92

src/LinearAccrual.sol
wischli: Support for negative debt applied in [1f05959](https://github.com/centrifuge/asset-pools/pull/9/commits/1f059597c0b0e766443a876228b7c01197f1f2e3)

src/LinearAccrual.sol
```diff
@@ -80,44 +81,48 @@ contract LinearAccrual is ILinearAccrual {
...
+    function debt(bytes32 rateId, int128 normalizedDebt) public view returns (int128) {
         _requireUpdatedRateId(rateId);
 
-        return normalizedDebt.mulDiv(rates[rateId].accumulatedRate.inner(), 1e18).toUint128();
+        // @dev Casting to int128 safe because we don't exceed number of digits of normalizedDebt
+        // @dev Casting to uint256 necessary for mulDiv
+        if (normalizedDebt >= 0) {
+            return normalizedDebt.toUint256().mulDiv(rates[rateId].accumulatedRate.inner(), 1e18).toUint128().toInt128();
+        } else {
+            return -(-normalizedDebt).toUint256().mulDiv(rates[rateId].accumulatedRate.inner(), 1e18).toUint128().toInt128();
```
lemunozm, line 129: Amazing trick to get `mulDiv` works fine with negative numbers!

src/types/D18.sol
```diff
@@ -32,9 +32,18 @@ function mulInt(D18 d, uint128 value) pure returns (uint128) {
     return MathLib.mulDiv(D18.unwrap(d), value, 1e18).toUint128();
 }
 
+/// @dev  Divides an integer by a decimal, i.e.
+/// @dev  Same as mulDiv for integers, i.e:
+/// - d (decimal):      2_000_000_000_000_000_000
+/// - value (integer):  100_000_000_000_000_000_000
+/// - result (integer): 50_000_000_000_000_000_000
+function reciprocalMulInt(D18 d, uint128 value) pure returns (uint128) {
```
lemunozm, line 40: Why not the name `divInt`?

src/LinearAccrual.sol
wischli: Yeah the double `-` is ugly but I couldn't come up with a generic wrapper because the inner one is always applied to the `int128` but the outer one after a (potentially) longer trail of calculations.

test/unit/libraries/D18.t.sol
```diff
@@ -5,7 +5,7 @@ import "forge-std/Test.sol";
 import "src/types/D18.sol";
 import "src/libraries/MathLib.sol";
 
-contract TestPortfolio is Test {
```
lemunozm, line 8: 😅

src/types/D18.sol
wischli: IMO `divInt` implies that I want to divide a D18 by some integer. But here we do the opposite. We divide some integer by a D18. Therefore, we mutiply reciprocal of the D18 with some integer.

src/interfaces/ILinearAccrual.sol
wischli: Applied in [3720dcd](https://github.com/centrifuge/asset-pools/pull/9/commits/3720dcd2c1d97556d4fbbfa3f384893bc610274e)

src/types/D18.sol
lemunozm: Makes sense! Good contra-argument

src/LinearAccrual.sol
lemunozm: Just an idea. Maybe we can add a method abstracting this into MathLib, as:

```solidity
// MathLib.sol
function mulDiv(int256 val, uint256 d) returns int256 {
   -(-val).toUint256().mulDiv(d)
}

// In LinearAccrual.sol
using MathLib for int128;

normalizedDebt.mulDiv(..)
```

wischli: In `LinearAccrual`, we use `MathLib.mulDiv` only implicitly inside `D18.reciprocalMulInt` outside of tests. I suppose we want to keep D18 decoupled from signed integers such that I don't think we **currently** win anything by adding such a wrapper.

lemunozm: Are we do not using `mulDiv` also in the line above?

wischli: Okay you got me, this is the only direct occurrence of `mulDiv` inside `LinearAccrual`.

lemunozm: Just was an idea, feel free to implement it or not 🫶, as you think it's better

src/LinearAccrual.sol
```diff
@@ -0,0 +1,133 @@
...
+        Group memory group = groups[rateId];
+        require(group.ratePerPeriod.inner() != 0, "group-missing");
+
+        // Determine number of full compounding periods passed since last update
+        uint64 periodsPassed = Compounding.getPeriodsPassed(group.period, rate.lastUpdated);
+
+        if (periodsPassed > 0) {
+            rate.accumulatedRate = d18(
+                rate.accumulatedRate.mulInt(
+                    uint256(group.ratePerPeriod.inner()).rpow(periodsPassed, d18(1e18).inner()).toUint128()
```
lemunozm: NIT
```suggestion
                    uint256(group.ratePerPeriod.inner()).rpow(periodsPassed, 1e18).toUint128()
```

Also, we can add `rpow()` to `D18` and perform this automatically:

```solidity
group.ratePerPeriod.rpow(periodsPassed)
```

src/LinearAccrual.sol
```diff
@@ -0,0 +1,133 @@
...
+    using MathLib for uint256;
+    using MathLib for int128;
+
+    mapping(bytes32 rateId => Rate rate) public rates;
+    mapping(bytes32 rateId => Group group) public groups;
+
+    /// @inheritdoc ILinearAccrual
+    function drip(bytes32 rateId) public {
+        Rate storage rate = rates[rateId];
+        require(rate.accumulatedRate.inner() != 0, RateIdMissing(rateId));
```
lemunozm: Not fully remember the output from this conversation, but for rates that represent 0%, thee acc rate will be 0 here, but they are still valid. Then, I could not use `LinearAccrual` here from the portfolio.

src/LinearAccrual.sol
```diff
@@ -0,0 +1,133 @@
...
+            rate.lastUpdated = uint64(block.timestamp);
+        }
+    }
+
+    /// @inheritdoc ILinearAccrual
+    function registerRateId(uint128 ratePerPeriod_, CompoundingPeriod period) public returns (bytes32 rateId) {
+        D18 ratePerPeriod = d18(ratePerPeriod_);
+        Group memory group = Group(ratePerPeriod, period);
+
+        rateId = keccak256(abi.encode(group));
```
lemunozm, line 65: Curious, will encoding an struct encode its attributes or just the "address" of that struct?

src/LinearAccrual.sol
```diff
@@ -0,0 +1,133 @@
...
+            return -(rates[newRateId].accumulatedRate.reciprocalMulInt((-debt_).toUint256().toUint128()).toInt128());
+        }
+    }
+
+    /// @inheritdoc ILinearAccrual
+    function debt(bytes32 rateId, int128 normalizedDebt) public view returns (int128) {
+        _requireUpdatedRateId(rateId);
+
+        // @dev Casting to int128 safe because we don't exceed number of digits of normalizedDebt
+        // @dev Casting to uint256 necessary for mulDiv
```
lemunozm: NIT. I think `@dev` is not needed here

src/LinearAccrual.sol
lemunozm: I think this check should be done against `lastUpdated` instead (more occurrences)

lemunozm (commented):
Some minor things and one "major", but beside that, I think this is ready! Thanks for all the well received feedback 🙌🏻

src/LinearAccrual.sol
wischli: Something to reconsider if there is more codepaths where we need to use `rpow` with `D18`. For now, I'd keep it out to keep contract deployment cost and maintenance lower.

src/LinearAccrual.sol
wischli: Good point! Seems like the `drip` impl was "forgotten" by me when adding support for zero rates. We can't use `lastUpdated` check though, because the purpose of `drip` is to bump `lastUpdated` up as well. Working on a fix now.

src/LinearAccrual.sol
wischli: Apparently, the encoding is only dependent on the fields of the struct. The following tests works:

```solidity
    function testEncoding() public pure {
        D18 ratePerPeriod = d18(1e18);
        CompoundingPeriod period = CompoundingPeriod.Daily;

        bytes32 groupEncoding = keccak256(abi.encode(Group(ratePerPeriod, period)));
        bytes32 tupleEncoding = keccak256(abi.encode(ratePerPeriod, CompoundingPeriod.Daily));

        assertEq(groupEncoding, tupleEncoding);
    }
```

src/LinearAccrual.sol
wischli: [212814c](https://github.com/centrifuge/asset-pools/pull/9/commits/212814caba78fc91cec893cc3f87f01b63b6378f)

lemunozm: Super great workaround. Thanks!

src/LinearAccrual.sol
lemunozm: Thanks for clarifying

src/LinearAccrual.sol
lemunozm: Ok! I will simplify at least from `d18(1e18).inner()` to `1e18`

src/LinearAccrual.sol
```diff
@@ -0,0 +1,133 @@
...
+    }
+
+    /// @inheritdoc ILinearAccrual
+    function registerRateId(uint128 ratePerPeriod_, CompoundingPeriod period) public returns (bytes32 rateId) {
+        D18 ratePerPeriod = d18(ratePerPeriod_);
+        Group memory group = Group(ratePerPeriod, period);
+
+        rateId = keccak256(abi.encode(group));
+
+        require(groups[rateId].ratePerPeriod.inner() == 0, RateIdExists(rateId, ratePerPeriod.inner(), period));
```
lemunozm: I think here we need to check against `lastUpdated` here too.

src/LinearAccrual.sol
wischli: I was too blind to see that, thanks!

src/LinearAccrual.sol
wischli: Another great catch. Apologies for the bumpy process to support zero rates truly. Now, everything should be fixed! I added short circuits for all debt calculations because in the case of a zero rate, we can just return the (normalized) debt as no normalization happens. Moreover, during renormalization, we cannot divide by `0` (i.e. `reciprocalMulInt` for a zero rate): c8745a793323c171645d66b0173a2b2e19eec73a

src/LinearAccrual.sol
wischli: Fixed in c8745a793323c171645d66b0173a2b2e19eec73a

src/LinearAccrual.sol
```diff
@@ -0,0 +1,146 @@
...
+        external
+        view
+        returns (int128 newNormalizedDebt)
+    {
+        _requireUpdatedRateId(rateId);
+
+        // Short circuit
+        if (rates[rateId].accumulatedRate.inner() == 0) {
+            return prevNormalizedDebt + debtChange;
+        }
```
lemunozm: I'm pretty sure this is:
```suggestion
        // Short circuit
        if (rates[rateId].accumulatedRate.inner() == 0) {
            return prevNormalizedDebt;
        }
```
Or just returning zero.

src/LinearAccrual.sol
lemunozm: Thanks!

src/LinearAccrual.sol
wischli: Your suggestion would mean that you can never change the debt of an item with zero rate.

lemunozm: Well, actually, it doesn't matter what to return. Will be multiplied always by 0 when getting the debt

wischli: As for the short circuits: We need them at least in two code paths where we call `reciprocalMulInt`, i.e. `getModifiedNormalizedDebt` and `getRenormalizedDebt`, otherwise `mulDiv` panics due to division by zero.

wischli: > Well, actually, it doesn't matter what to return. Will be multiplied always by 0 when getting the debt

Can you elaborate?

lemunozm: > otherwise mulDiv panics due to division by zero.

Right!

> Your suggestion would mean that you can never change the debt of an item with zero rate.

> Can you elaborate?

Mmmm... I think I'm lost with the debt computation then. From `pallet-interest-accrual` the maths is: `accumulated_rate * normalized_debt`. It means that if  acc rate is zero, the resulting debt will be zero to.

But here we're doing it differently.

lemunozm: Ohhh, wait wait. In centrifuge-chain a 0% interest per year rate is `1.0` of interest_per_seconds. Would does it mean we can never have `accumulateRate == 0`?

src/libraries/MathLib.sol
hieronx: That's wild..

src/LinearAccrual.sol
wischli: Based on the discussion on Slack ([ref](https://kflabs.slack.com/archives/C07PG2EUR9C/p1730381816260549)), I decided to require all rates to be "prefixed" by one, e.g. the `zero` accumulated interest rate shall be represented by `1.0` and 5% by `1.05`. Therefore, `debt`, `getRenormalizedDebt` and `getModifiedNormalizedDebt` (but not `drip`) will now throw if the `rate.accumulatedRate.inner()` is zero to not divide by zero: [c88982a](https://github.com/centrifuge/asset-pools/pull/9/commits/c88982a487f6bdadc8dc84a5adf857700dce1642)

lemunozm (approved):
Just a comment but LGTM!

wischli:
> Just a comment but LGTM!

Can't see the comment. Is it still pending?

lemunozm:
Missing, don't know why 😆 

It was regarding the `1.x` nomenclature expected in params. Theoretically, the `1` is not related to the rate. It comes from the interest formula: `A = P + r * P = P * (1 + r)`. I think the library should add such `1` to the input rate param,  but not the API user. If not, we're mixing concepts about "what is a rate" and "how the rate is later used".

Just a thought!

mustermeiszer:
@wischli I think we can also close for now, right?

wischli:
> @wischli I think we can also close for now, right? 

Sure!

PR #14, opened 2024-10-14
Makes CI to fail if warning

lemunozm:
Feel free to disagree with. 

I think we should force the CI to warn us about warnings.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -75,7 +75,7 @@ jobs:
       - name: Build contracts
         run: |
           forge --version
-          forge build --sizes --deny-warnings
+          forge build --sizes
 
       - name: Check formatting
         run: forge fmt --check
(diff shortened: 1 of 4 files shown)


hieronx (approved):
Maybe there are some warnings we might want to intentionally ignore, but for now, I agree this is worth adding. Can always remove it again later.

lemunozm:
Does there exist some kind of "warning lint" section, for that matter? In Rust you can enable/disable a subset of warnings

lemunozm:
I've changed it to be a `foundry.toml` property instead

hieronx:
> Does there exist some kind of "warning lint" section, for that matter? In Rust you can enable/disable a subset of warnings

https://book.getfoundry.sh/reference/config/solidity-compiler#ignored_error_codes

hieronx (approved):

peculiarity (approved):

PR #19, opened 2024-10-17
Minimal fixed point number library (D18)

lemunozm:
I extracted the `Decimal18` type from the portfolio PR (https://github.com/centrifuge/asset-pools/pull/5) into its own PR (to simplify, called now `D18`), because I will need this also in the oracle implementation.

It is still open to discussion, but the benefits of this are:
- Better readability about how an `uint128` is represented internally and what it is used for.
  i.e:
  ```solidity
      D18 rate = ...
      uint128 balance = ...
      rate * balance // will fail because it requires a special multiplication treatment.
  ```
- Avoid handling low-level maths each time an operation with fixed point numbers is required.
  i.e: 
  ```solidity
     // without explicit types
     uint128 rate = ...
     uint128 balance = ...
     Mathlib.mulDiv(rate, balance, 10**18) // easy to get confused.
     
     // using D18
     D18 rate = ...
     uint128 balance = ...
     rate.mulInt(balance)
     ```


src/libraries/D18.sol
```diff
@@ -0,0 +1,40 @@
+// SPDX-License-Identifier: UNLICENSED
+pragma solidity 0.8.28;
+
+// Small library to handle fixed point number operations with 18 decimals with static typing support.
```
hieronx: Why is it not actually structured as a library?

lemunozm: I can not if I want `D18` to be an `uint128` forcibly. If I make `library D18`, then a user can write: `using D18 as uint32`, which is not correct.

But, of course all of this can be wrong due my lack of knowledge 😄

peculiarity: They can but none of the methods will be visible on the `uint32` type because there is no signature that accepts `uint32` as a first argument.

Also a library no necessary have to be used in the way `using D18 for uint128`. You can just import it  as if the methods are "static".
`D18.add(d1,d2)`

src/libraries/D18.sol
```diff
@@ -0,0 +1,40 @@
...
+/// @dev Multiplies a decimal by an integer. i.e:
+/// - d (decimal):      1_500_000_000_000_000_000
+/// - value (integer):  4_000_000_000_000_000_000
+/// - result (integer): 6_000_000_000_000_000_000
+function mulInt(D18 d, uint128 value) pure returns (uint128) {
+    return MathLib.mulDiv(D18.unwrap(d), value, 1e18).toUint128();
+}
+
+/// @dev Easy way to construct a decimal number
+function d18(uint128 value) pure returns (D18) {
```
peculiarity:
```suggestion
function toD18(uint128 value) pure returns (D18) {
```

src/libraries/D18.sol
```diff
@@ -0,0 +1,40 @@
...
+    return D18.wrap(D18.unwrap(d1) + D18.unwrap(d2));
+}
+
+/// @dev substract two D18 types
+function sub(D18 d1, D18 d2) pure returns (D18) {
+    return D18.wrap(D18.unwrap(d1) - D18.unwrap(d2));
+}
+
+/// @dev sugar for getting the inner representation of a D18
+function inner(D18 d1) pure returns (uint128) {
```
peculiarity:
```suggestion
function fromD18(D18 d1) pure returns (uint128) {
```

src/libraries/D18.sol
lemunozm: The name was inspired by the library you recommended, `prb-math` which uses this style of naming.

I think the perspective is that you should see the function as a constructor instead of a conversion. So you build `D18` objects using `d18` constructors, and later you can access the inner value

src/libraries/D18.sol
lemunozm: If doing this with a library, then should I write `using D18 for uint128` each time I import the type?

Currently importing `D18`, you have everything ready to be used as intended.

lemunozm: Also I think `using D18 for uint128` will not differentiate types in compile time, which is the big point of doing this. They will treat both `D18` and `uint128` indistinctly, loosing the type checks.

`D18` is not a library to extend `uint128` types. `D18` is a type with specific methods that `uint128` does not have and must not have

src/libraries/D18.sol
```diff
@@ -0,0 +1,40 @@
+// SPDX-License-Identifier: UNLICENSED
+pragma solidity 0.8.28;
+
+// Small library to handle fixed point number operations with 18 decimals with static typing support.
+
+import {MathLib} from "src/libraries/MathLib.sol";
+
+type D18 is uint128;
+
+using MathLib for uint256;
```
peculiarity: I don't think this line is needed. You don't have uint256 anywhere

lemunozm: The one returning the `mulDiv` method from `Mathlib` and then casting it through `.toUint128()`

ilinzweilin (approved):
looks good!

Would put `D18.sol` into a types folder instead of libraries folder

peculiarity (approved):

PR #20, opened 2024-10-17
Oracle Contract implementation

### LINKED ISSUE #7: Oracle implementation
- Factory to deploy oracle contract, intended to have 1 per pool, although it could be shared
- On deployment, set the `feeder`/`owner`. For now just 1 account.
- When providing a new quote, also include a `quote_reference_time` param. Linear pricing will be based off of this timestamp rather than the `block.timestamp` it was provided.

```solidity
function getIndicativeQuote(..) external view returns (uint256);

// Uses linear pricing
function getIndicativeQuote(..) external view returns (uint256);

function setQuote(..) external onlyFeeder;
```

lemunozm:
Fixes #7 

Initial implementation of the Oracle contract.

Some considerations:
- Super simple, 1 feeder per contract, no medians, no time checks.
- Any non-ERC20 ID will have 18 decimals.
- Indicative quotes not implemented yet (requires more thinking).

### DIFF
--- a/src/Oracle.sol
+++ b/src/Oracle.sol
@@ -5,6 +5,7 @@ import {IERC7726, IERC7726} from "src/interfaces/IERC7726.sol";
 import {IERC20} from "forge-std/interfaces/IERC20.sol";
 import {MathLib} from "src/libraries/MathLib.sol";
 
+/// @notice Quote value representation value representation
 struct Quote {
     /// @notice Price of one base in quote denomination
     uint256 amount;
@@ -86,15 +87,16 @@ contract OracleFactory {
     /// @param salt Extra bytes to generate different address for the same feeder.
     function deploy(address feeder, bytes32 salt) external returns (Oracle) {
         Oracle deployed = new Oracle{salt: salt}(feeder);
+
         emit NewOracleDeployed(address(deployed));
 
         return deployed;
     }
 
-    /// Retuns the deterministic contract address for a feeder.
+    /// @notice Retuns the deterministic contract address for a feeder.
     /// @param feeder The account that will be able to fed values in the contract.
     /// @param salt Extra bytes to generate different address for the same feeder.
-    function oracleAddress(address feeder, bytes32 salt) public view returns (address) {
+    function getAddress(address feeder, bytes32 salt) public view returns (address) {
         bytes memory bytecode = abi.encodePacked(type(Oracle).creationCode, abi.encode(feeder));
 
         return address(
--- a/src/Oracle.sol
+++ b/src/Oracle.sol
@@ -1,11 +1,12 @@
 // SPDX-License-Identifier: AGPL-3.0-only
 pragma solidity = 0.8.28;
 
-import {IERC7726, IERC7726Ext} from "src/interfaces/IERC7726.sol";
+import {IERC7726, IERC7726} from "src/interfaces/IERC7726.sol";
 import {IERC165} from "forge-std/interfaces/IERC165.sol";
 import {IERC20} from "forge-std/interfaces/IERC20.sol";
+import {MathLib} from "src/libraries/MathLib.sol";
 
-contract Oracle is IERC7726Ext {
+contract Oracle is IERC7726 {
     error NotValidFeeder();
     error ValueNotFound();
 
@@ -45,15 +46,28 @@ contract Oracle is IERC7726Ext {
         Value storage quoteValue = values[base][quote];
         require(quoteValue.referenceTime > 0, ValueNotFound());
 
-        return baseAmount * quoteValue.amount;
+        return MathLib.mulDiv(baseAmount, quoteValue.amount, 1 ** _extractDecimals(base));
     }
 
-    /// @inheritdoc IERC7726Ext
-    function getIndicativeQuote(uint256 baseAmount, address base, address quote)
-        external
-        view
-        returns (uint256 quoteAmount)
-    {
-        //TODO
+    /// @dev extract the decimals used for the assetId
+    /// - If the asset is an ERC20, then we ask the contract for its decimals
+    /// - Otherwise we assume 18 decimals
+    function _extractDecimals(address assetId) internal view returns (uint8) {
+        if (IERC165(assetId).supportsInterface(type(IERC20).interfaceId)) {
+            IERC20 erc20 = IERC20(assetId);
+            return erc20.decimals();
+        } else {
+            return 18;
+        }
+    }
+}
+
+contract OracleFactory {
+    event NewOracle(address where);
+
+    function build(address feeder) external {
+        address deployed = address(new Oracle(feeder));
+
+        emit NewOracle(deployed);
     }
 }
--- a/src/Oracle.sol
+++ b/src/Oracle.sol
@@ -6,6 +6,8 @@ import {IERC20} from "forge-std/interfaces/IERC20.sol";
 import {MathLib} from "src/libraries/MathLib.sol";
 
 contract Oracle is IERC7726 {
+    uint8 public constant DEFAULT_DECIMALS = 18;
+
     /// @notice Dispatched when the action is not performed by the required feeder.
     error NotValidFeeder();
 
@@ -63,13 +65,13 @@ contract Oracle is IERC7726 {
     /// - Otherwise we assume 18 decimals
     function _extractDecimals(address assetId) internal view returns (uint8) {
         if (assetId.code.length == 0) {
-            return 18;
+            return DEFAULT_DECIMALS;
         } else {
             (bool ok, bytes memory data) = assetId.staticcall(abi.encodeWithSelector(IERC20.decimals.selector));
             if (ok) {
                 return abi.decode(data, (uint8));
             } else {
-                return 18;
+                return DEFAULT_DECIMALS;
             }
         }
     }
--- a/src/Oracle.sol
+++ b/src/Oracle.sol
@@ -17,7 +17,7 @@ contract Oracle is IERC7726 {
     /// @notice Emitted when the contract is feed with a new quote amount.
     event Fed(address indexed base, address indexed quote, uint256 quoteAmount, uint64 referenceTime);
 
-    struct Value {
+    struct Quote {
         /// @notice Price of one base in quote denomination
         uint256 amount;
         /// @notice Timestamp when the value was fed
@@ -25,10 +25,10 @@ contract Oracle is IERC7726 {
     }
 
     /// @notice Owner of the contract able to feed new values
-    address feeder;
+    address public feeder;
 
     /// @notice All fed values.
-    mapping(address base => mapping(address quote => Value)) public values;
+    mapping(address base => mapping(address quote => Quote)) public values;
 
     /// @dev check that only the feeder perform the action
     modifier onlyFeeder() {
@@ -47,14 +47,14 @@ contract Oracle is IERC7726 {
     /// @param quoteAmount The amount of 1 wei of base amount represented as quote units.
     function setQuote(address base, address quote, uint256 quoteAmount) external onlyFeeder {
         uint64 referenceTime = uint64(block.timestamp);
-        values[base][quote] = Value(quoteAmount, referenceTime);
+        values[base][quote] = Quote(quoteAmount, referenceTime);
 
         emit Fed(base, quote, quoteAmount, referenceTime);
     }
 
     /// @inheritdoc IERC7726
     function getQuote(uint256 baseAmount, address base, address quote) external view returns (uint256 quoteAmount) {
-        Value storage quoteValue = values[base][quote];
+        Quote storage quoteValue = values[base][quote];
         require(quoteValue.referenceTime > 0, NeverFed());
 
         return MathLib.mulDiv(baseAmount, quoteValue.amount, 10 ** _extractDecimals(base));
--- a/src/Oracle.sol
+++ b/src/Oracle.sol
@@ -55,12 +55,14 @@ contract Oracle is IOracle {
     /// - Otherwise we assume 18 decimals
     function _extractDecimals(address assetId) internal view returns (uint8) {
         if (assetId.code.length == 0) {
+            // It means, it is not an contract address
             return DEFAULT_DECIMALS;
         } else {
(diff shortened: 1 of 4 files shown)


src/Oracle.sol
```diff
@@ -0,0 +1,104 @@
...
+    error NeverFed();
+
+    /// @notice Emitted when the contract is feed with a new quote amount.
+    event Fed(address indexed base, address indexed quote, uint256 quoteAmount, uint64 referenceTime);
+
+    struct Quote {
+        /// @notice Price of one base in quote denomination
+        uint256 amount;
+        /// @notice Timestamp when the value was fed
+        uint64 referenceTime;
```
lemunozm: There is no treatment at this time right now. I guess we need this for a future improvement of this, or let feeders to read when the latest value was updated.

src/Oracle.sol
```diff
@@ -0,0 +1,104 @@
...
+        emit Fed(base, quote, quoteAmount, referenceTime);
+    }
+
+    /// @inheritdoc IERC7726
+    function getQuote(uint256 baseAmount, address base, address quote) external view returns (uint256 quoteAmount) {
+        Quote storage quoteValue = values[base][quote];
+        require(quoteValue.referenceTime > 0, NeverFed());
+
+        return MathLib.mulDiv(baseAmount, quoteValue.amount, 10 ** _extractDecimals(base));
+    }
```
lemunozm: Do we want to consider that prices at this level are symmetric? It means:

```
getQuote(1, base, quote) == 1 / getQuote(1, quote, base)
```

- If yes, then in the implementation, if the value for base -> quote, is not found, we can try quote -> base and returns the inverse value.

- If not, feeders should feed in both directions. Because both directions are needed from the Portfolio to work

src/Oracle.sol
```diff
@@ -0,0 +1,104 @@
...
+
+    /// @dev extract the decimals used for the assetId
+    /// - If the asset is an ERC20, then we ask the contract for its decimals
+    /// - Otherwise we assume 18 decimals
+    function _extractDecimals(address assetId) internal view returns (uint8) {
+        if (assetId.code.length == 0) {
+            return DEFAULT_DECIMALS;
+        } else {
+            (bool ok, bytes memory data) = assetId.staticcall(abi.encodeWithSelector(IERC20.decimals.selector));
```
peculiarity: Could you please elaborate on this? Why not simply calling:
```
uint8 decimals = IERC20(assetId).decimals();
```
Probably the name of the arg could be `target` rather than `assetId` . Wdyt?

src/Oracle.sol
```diff
@@ -0,0 +1,104 @@
...
+            if (ok) {
+                return abi.decode(data, (uint8));
+            } else {
+                return DEFAULT_DECIMALS;
+            }
+        }
+    }
+}
+
+contract OracleFactory {
```
peculiarity: Create an interface and put events and func defs there

src/Oracle.sol
```diff
@@ -0,0 +1,104 @@
...
+        Oracle deployed = new Oracle{salt: salt}(feeder);
+        emit NewOracleDeployed(address(deployed));
+
+        return deployed;
+    }
+
+    /// Retuns the deterministic contract address for a feeder.
+    /// @param feeder The account that will be able to fed values in the contract.
+    /// @param salt Extra bytes to generate different address for the same feeder.
+    function oracleAddress(address feeder, bytes32 salt) public view returns (address) {
```
peculiarity:
```suggestion
    function previewAddress(address feeder, bytes32 salt) public view returns (address) {
```
of
```suggestion
    function previewOracleAddress(address feeder, bytes32 salt) public view returns (address) {
```

peculiarity: Probably we can have a separate abstract contract that implements this function and factories can inherit it. Wdyt?

src/Oracle.sol
```diff
@@ -0,0 +1,104 @@
+// SPDX-License-Identifier: AGPL-3.0-only
+pragma solidity = 0.8.28;
+
+import {IERC7726, IERC7726} from "src/interfaces/IERC7726.sol";
+import {IERC20} from "forge-std/interfaces/IERC20.sol";
+import {MathLib} from "src/libraries/MathLib.sol";
+
+contract Oracle is IERC7726 {
```
peculiarity:
```suggestion
contract Oracle is IOracle {
```

peculiarity: The `IOracel` interface will extend the `IERC7726` and it will have all the events and structs defined there.

src/Oracle.sol
```diff
@@ -0,0 +1,104 @@
...
+
+    struct Quote {
+        /// @notice Price of one base in quote denomination
+        uint256 amount;
+        /// @notice Timestamp when the value was fed
+        uint64 referenceTime;
+    }
+
+    /// @notice Owner of the contract able to feed new values
+    address public feeder;
```
peculiarity, line 22: Follow the style guide - These are state variables so move up :)

src/Oracle.sol
```diff
@@ -0,0 +1,104 @@
...
+    function setQuote(address base, address quote, uint256 quoteAmount) external onlyFeeder {
+        uint64 referenceTime = uint64(block.timestamp);
+        values[base][quote] = Quote(quoteAmount, referenceTime);
+
+        emit Fed(base, quote, quoteAmount, referenceTime);
+    }
+
+    /// @inheritdoc IERC7726
+    function getQuote(uint256 baseAmount, address base, address quote) external view returns (uint256 quoteAmount) {
+        Quote storage quoteValue = values[base][quote];
```
peculiarity, line 47:
```suggestion
        Quote memory quoteValue = values[base][quote];
```
`quoteValue.referenceTime` and `quoteValue.amount` are 2 additional storage reads which cost 100 each.
Put them into memory and I think each will cost around 6 or smth like that.

src/Oracle.sol
```diff
@@ -0,0 +1,104 @@
...
+contract Oracle is IERC7726 {
+    uint8 public constant DEFAULT_DECIMALS = 18;
+
+    /// @notice Dispatched when the action is not performed by the required feeder.
+    error NotValidFeeder();
+
+    /// @notice Dispatched when the base/quote pair has never been feed.
+    error NeverFed();
+
+    /// @notice Emitted when the contract is feed with a new quote amount.
```
peculiarity:
```suggestion
    /// @notice Emitted when the contract is fed with a new quote amount.
```

src/Oracle.sol
```diff
@@ -0,0 +1,104 @@
...
+import {IERC20} from "forge-std/interfaces/IERC20.sol";
+import {MathLib} from "src/libraries/MathLib.sol";
+
+contract Oracle is IERC7726 {
+    uint8 public constant DEFAULT_DECIMALS = 18;
+
+    /// @notice Dispatched when the action is not performed by the required feeder.
+    error NotValidFeeder();
+
+    /// @notice Dispatched when the base/quote pair has never been feed.
```
peculiarity:
```suggestion
    /// @notice Dispatched when the base/quote pair has never been fed.
```

src/Oracle.sol
```diff
@@ -0,0 +1,104 @@
...
+    error NeverFed();
+
+    /// @notice Emitted when the contract is feed with a new quote amount.
+    event Fed(address indexed base, address indexed quote, uint256 quoteAmount, uint64 referenceTime);
```
peculiarity:
```suggestion
    event NewQuoteSet(address indexed base, address indexed quote, uint256 quoteAmount, uint64 referenceTime);
```

src/Oracle.sol
```diff
@@ -0,0 +1,104 @@
...
+import {MathLib} from "src/libraries/MathLib.sol";
+
+contract Oracle is IERC7726 {
+    uint8 public constant DEFAULT_DECIMALS = 18;
+
+    /// @notice Dispatched when the action is not performed by the required feeder.
+    error NotValidFeeder();
+
+    /// @notice Dispatched when the base/quote pair has never been feed.
+    error NeverFed();
```
peculiarity:
```suggestion
    error NoQuote();
```
or
```suggestion
    error MissingQuote();
```

src/Oracle.sol
lemunozm: It's because `assetId` could not be a ERC20. If I call as you say, I have no way of checking the success of the call to return `DEFAULT_DECIMALS` if it was not an ERC20.

Yeah, probably it can be target. But it will always represent an asset identification. Is target used for more generical representations? I lost in the common solidity namings

src/Oracle.sol
lemunozm: Could you expand on why? AFAIK, the factory is just one implementation, and will not be called from other contracts.

src/Oracle.sol
lemunozm: That's sound good! However, the problem with this is that the specific function parameters are encoded to get the address. So other factories will use other parameters, and the computation (and function signature) will differ

lemunozm: Unless we pass the bytecode as param 🤔

src/Oracle.sol
lemunozm: Interesting. I think it makes sense!

Just one question: Why also the structs? Are they not fully dependent on the implementation?

src/Oracle.sol
lemunozm: Correct!

src/Oracle.sol
lemunozm: So, IIUC, you say that put this as memory force a read of both variables at the same time as a whole?

I understood `memory` creates a copy of those values, and `storage` just references them. So if I'm just reading from them, it can be cheaper to only reference them.

src/Oracle.sol
lemunozm: Ok, let's keep consistent with the naming 👍🏻

src/Oracle.sol
peculiarity: It is just a suggestion.  It could be considered as an id because addresses are unique. Again - just a suggestion.

src/Oracle.sol
peculiarity: This is something we do. Check the[ liquidity-pools](https://github.com/centrifuge/liquidity-pools/tree/main/src/interfaces/factories).
If there is someone integrating with our protocol, it simply imports the interface instead of the whole implementation. When you import a file, you "copy" the code from the imported file. So imagine having a small interface that defines what methods an address has ( `IOracle(target)`) and just importing way way more lines of code.
Also if we want to have that factory somewhere as a dependency, same thing will apply for our code base.

Let me know if something is unclear.

src/Oracle.sol
peculiarity: Regarding the structs -> Save that for our discussion related to errors :sweat_smile:

src/Oracle.sol
peculiarity: OK, I see your point.I have to investigate it further if it makes sense.

src/Oracle.sol
peculiarity: It doesn't work like this when it comes to gas cost :sweat_smile: .
Whenever you "load" a slot for the first time, you pay gas for `COLD_SLOAD` which is 2100 gas. Every following read is 100 gas.  
If you put it into memory, it will be also 2100 gas because you have to read the value. Of course there will be some minor op code for coping the data and etc but it will be small one. Then every time you read from the memory, it will cost you I think 3 or 6 gas  ( couple of op codes involved ).

peculiarity: One thing that I just realized, is this method supposed to be call directly from EoAs only  or it will be used in contract integrations ?

peculiarity: Please ignore this. Your approach is cheaper

src/Oracle.sol
lemunozm: Ok, thanks. I think I get more of the point of interfaces! I think this solidity copy-paste style of dependencies goes in favor of over interfacing things. It remembers me to `.h` header files in c++

src/Oracle.sol
lemunozm: > is this method supposed to be call directly from EoAs only or it will be used in contract integrations ?

It will be primarily used in contract integrations.

src/Oracle.sol
lemunozm: Regarding this topic (which follows [here](https://kflabs.slack.com/archives/C07PG2EUR9C/p1729251591951409)). I'm not sure if I can get this work with try/catch because I can not differentiate between:
 - an error that occurred inside the implementation of `.decimals()`, which should throw that error.
 - the error of not existing `.decimals()` abi in that address, which should not error out and use 18 decimals.

I think I'm entirely forced to use `staticcall` (although I agree, for readability, it's would be prettier the direct call 🥲)

src/Oracle.sol
lemunozm: Finally, I've chosen "getAddress" to follow the liquidity-pool naming style, which is quite concise.

src/Oracle.sol
lemunozm: Done

src/Oracle.sol
lemunozm: Done

src/Oracle.sol
lemunozm: After some play with it, it seems that using `staticCall` does not prevent what I say. Reverting from a `staticCall` makes the `ok` from above `false` in the same way as if the abi is not found.

So I'll go with the try/catch solution: [5c98630](https://github.com/centrifuge/asset-pools/pull/20/commits/5c9863027c89522eb5382dc883d793f1f267b35a)

mustermeiszer:
@lemunozm I think we can also close fo rnow, right?

lemunozm:
Don't we expect to use real oracles by now? If that's the case, then we can close it

mustermeiszer:
I think in the EVM world for now we expect most oracles to come from other parties. We most likely will only need wrappers if they do not match our needs. I would close to keep the PR cleans. But no strong opinion.

Cc @hieronx

lemunozm:
Ok, I agree in closing it. But we will need some oracle based on ERC6909 that works with the IERC7726 interface. Not sure if that exists 🤔

PR #21, opened 2024-10-20
(Draft) feat: cash account

### LINKED ISSUE #17: Cash account contract
Spec: https://www.notion.so/Cash-Account-1212eac24e17808a85e0f9604371cbdc?pvs=4

wischli:
Fixes #17

### DIFF
--- a/src/CashAccount.sol
+++ b/src/CashAccount.sol
@@ -1,0 +1,60 @@
+// SPDX-License-Identifier: AGPL-3.0-only
+pragma solidity 0.8.28;
+
+import {IPortfolio} from "src/interfaces/IPortfolio.sol";
+import {ICashAccount} from "src/interfaces/ICashAccount.sol";
+
+// TODO: is IItem?
+contract CashAccount is ICashAccount {
+    // IShareManager public shareManager;
+    IPortfolio public portfolio;
+
+    // TODO: Should probably be replaced with querying from top level contract. However, Portfolio only stores array of
+    // items.
+    mapping(uint128 poolId => mapping(uint32 itemId => address owner)) owners;
+
+    modifier onlyOwner(uint64 poolId, uint32 itemId, address owner_) {
+        require(owner_ == owners[poolId][itemId], NotOwner(poolId, itemId));
+        _;
+    }
+    /// @inheritdoc ICashAccount
+
+    function create(uint64 poolId, address owner) external {
+        // TODO(from spec): Call portfolio.lock
+        // TODO(from spec): Call portfolio.create
+        // TODO(from spec): Set valuation to address(0)
+
+        // TODO: Get itemId from creation
+        // TODO: owners[poolId][itemId] = owner;
+    }
+
+    /// @inheritdoc ICashAccount
+    function deposit(uint64 poolId, uint32 itemId, uint128 principal) public onlyOwner(poolId, itemId, msg.sender) {
+        // TODO(from spec): Withdraw from share manager: shareManager.withdraw(poolId, itemId, principal);
+        // TODO(from spec): Increase debt: increaseDebt(poolId, itemId, principal);
+    }
+
+    /// @inheritdoc ICashAccount
+    function withdraw(uint64 poolId, uint32 itemId, uint128 principal, uint128 unscheduled)
+        public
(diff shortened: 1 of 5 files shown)


src/CashAccount.sol
```diff
@@ -0,0 +1,60 @@
...
+
+    modifier onlyOwner(uint64 poolId, uint32 itemId, address owner_) {
+        require(owner_ == owners[poolId][itemId], NotOwner(poolId, itemId));
+        _;
+    }
+    /// @inheritdoc ICashAccount
+
+    function create(uint64 poolId, address owner) external {
+        // TODO(from spec): Call portfolio.lock
+        // TODO(from spec): Call portfolio.create
```
lemunozm, line 24: I would expect this to be always the pattern? I mean, first `lock()` and just later `create()`?. I would expect the benefits of spliting these (at the cost of a new storage inside the escrow to track the associations) to open the possibility to be the one to lock different than the one who creates the asset.

If the entity who calls lock() and create() is always the same. I do not see any advantages then of splitting this 🤔. Note that we're storing a new mapping just to splitting the lock from the attach phases.

cc @hieronx @ilinzweilin

lemunozm, line 24: William, now it will be as follows:
```suggestion
        // TODO(from spec): Call escrow.lock
        // TODO(from spec): Call portfolio.create
```

src/CashAccount.sol
```diff
@@ -0,0 +1,60 @@
...
+        // TODO(from spec): Increase debt: increaseDebt(poolId, itemId, principal);
+    }
+
+    /// @inheritdoc ICashAccount
+    function withdraw(uint64 poolId, uint32 itemId, uint128 principal, uint128 unscheduled)
+        public
+        onlyOwner(poolId, itemId, msg.sender)
+    {
+        // TODO(from spec): Deposit into share manager: shareManager.deposit(poolId, itemId, principal, unscheduled);
+        // TODO(from spec): Decrease debt: decreaseDebt(poolId, itemId, principal + unscheduled);
```
lemunozm, line 43: Note. I think the unscheduled should not be added to the portfolio, only to the share manager. It's an amount that we do not want to track to the item.

src/CashAccount.sol
hieronx: If there is significant cost / additional complexity in the portfolio contract + escrow contract to splitting it, I agree it doesn't seem worth it.

lemunozm: The extra mapping (collateralId => itemId) is used to avoid using the same collateral in different items. Without that mapping someone could do:

escrow.lock(ASSET_1)
portfolio.create(ASSET_1)
portfolio.create(ASSET_2) //when shouldn't

src/CashAccount.sol
wischli: Can you elaborate? As per the TODO, I wanted to call `deposit` with the unscheduled on the share manager.

lemunozm: Yes, the schedule amount should be added to the `deposit()`, that is correct. But it should not be communicated/registered in any way to the portfolio, AFAIK.

It means:

lemunozm:
```suggestion
        // TODO(from spec): Decrease debt: decreaseDebt(poolId, itemId, principal);
```

src/CashAccount.sol
```diff
@@ -0,0 +1,60 @@
...
+    }
+
+    /// @inheritdoc ICashAccount
+    function deposit(uint64 poolId, uint32 itemId, uint128 principal) public onlyOwner(poolId, itemId, msg.sender) {
+        // TODO(from spec): Withdraw from share manager: shareManager.withdraw(poolId, itemId, principal);
+        // TODO(from spec): Increase debt: increaseDebt(poolId, itemId, principal);
+    }
+
+    /// @inheritdoc ICashAccount
+    function withdraw(uint64 poolId, uint32 itemId, uint128 principal, uint128 unscheduled)
```
lemunozm, line 38: Question. How do you pay the interest? Should we add a new param `uint128 interest`.

Just to confirm. For me unscheduled and interest are different things. Is it the same for you?

PR #22, opened 2024-10-22
Escrow contract for Nfts

lemunozm:
# Description

Implementation of an escrow to handle NFTs (collaterals from the portfolio)

This PR needs to be merged before #5

### DIFF
--- a/src/NftEscrow.sol
+++ b/src/NftEscrow.sol
@@ -26,9 +26,4 @@ contract NftEscrow is Auth, INftEscrow {
 
         emit Unlocked(source, tokenId);
     }
-
-    /// @inheritdoc INftEscrow
-    function computeNftId(IERC6909 source, uint256 tokenId) public pure returns (uint160) {
-        return uint160(uint256(keccak256(abi.encodePacked(source, tokenId))));
-    }
 }
--- a/src/NftEscrow.sol
+++ b/src/NftEscrow.sol
@@ -26,6 +26,9 @@ contract NftEscrow is Auth, INftEscrow {
 
     /// @inheritdoc INftEscrow
     function unlock(IERC6909 source, uint256 tokenId, address to) external auth {
+        uint160 nftId = computeNftId(source, tokenId);
+        require(associations[nftId] == 0, AlreadyAttached());
+
         bool ok = source.transferFrom(address(this), to, tokenId, 10 ** source.decimals(tokenId));
         require(ok, CanNotBeTransferred());
 
--- a/src/NftEscrow.sol
+++ b/src/NftEscrow.sol
@@ -5,63 +5,30 @@ import {Auth} from "src/Auth.sol";
 import {IERC6909, INftEscrow} from "src/interfaces/INftEscrow.sol";
 
 contract NftEscrow is Auth, INftEscrow {
-    /// A list of NFTs with an associated element.
-    mapping(uint160 nftId => uint256 elementId) public associations;
-
     constructor(address owner) Auth(owner) {}
 
     /// @inheritdoc INftEscrow
-    function lock(IERC6909 source, uint256 tokenId, address from) external auth returns (uint160) {
-        uint160 nftId = computeNftId(source, tokenId);
-
+    function lock(IERC6909 source, uint256 tokenId, address from) external auth {
         require(source.balanceOf(address(this), tokenId) == 0, AlreadyLocked());
 
-        bool ok = source.transferFrom(from, address(this), tokenId, 10 ** source.decimals(tokenId));
+        bool ok = source.transferFrom(from, address(this), tokenId, 1);
         require(ok, CanNotBeTransferred());
 
         emit Locked(source, tokenId);
-
-        return nftId;
     }
 
     /// @inheritdoc INftEscrow
     function unlock(IERC6909 source, uint256 tokenId, address to) external auth {
-        uint160 nftId = computeNftId(source, tokenId);
-        require(associations[nftId] == 0, AlreadyAttached());
+        require(source.balanceOf(address(this), tokenId) == 1, NotLocked());
 
-        bool ok = source.transferFrom(address(this), to, tokenId, 10 ** source.decimals(tokenId));
+        bool ok = source.transferFrom(address(this), to, tokenId, 1);
         require(ok, CanNotBeTransferred());
 
         emit Unlocked(source, tokenId);
     }
 
     /// @inheritdoc INftEscrow
-    function attach(IERC6909 source, uint256 tokenId, uint256 elementId) public auth returns (uint160) {
-        uint160 nftId = computeNftId(source, tokenId);
-
-        require(elementId != 0, InvalidElement());
-        require(associations[nftId] == 0, AlreadyAttached());
-        require(source.balanceOf(address(this), tokenId) > 0, NotLocked());
-
-        associations[nftId] = elementId;
-
-        emit Attached(nftId, elementId);
-
-        return nftId;
-    }
-
-    /// @inheritdoc INftEscrow
-    function detach(uint160 nftId) public auth {
-        uint256 elementId = associations[nftId];
-        require(elementId != 0, NotAttached());
-
-        delete associations[nftId];
-
-        emit Detached(nftId, elementId);
-    }
-
-    /// @inheritdoc INftEscrow
     function computeNftId(IERC6909 source, uint256 tokenId) public pure returns (uint160) {
-        return uint160(uint256(keccak256(abi.encode(source, tokenId))));
+        return uint160(uint256(keccak256(abi.encodePacked(source, tokenId))));
     }
 }
--- a/src/NftEscrow.sol
+++ b/src/NftEscrow.sol
@@ -1,0 +1,64 @@
+// SPDX-License-Identifier: UNLICENSED
+pragma solidity 0.8.28;
+
+import {Auth} from "src/Auth.sol";
+import {IERC6909, INftEscrow} from "src/interfaces/INftEscrow.sol";
+
+contract NftEscrow is Auth, INftEscrow {
+    /// A list of NFTs with an associated element.
+    mapping(uint160 nftId => uint256 elementId) public associations;
+
+    constructor(address owner) Auth(owner) {}
+
+    /// @inheritdoc INftEscrow
+    function lock(IERC6909 source, uint256 tokenId, address from) external auth returns (uint160) {
+        uint160 nftId = computeNftId(source, tokenId);
+
+        require(source.balanceOf(address(this), tokenId) == 0, AlreadyLocked());
+
+        bool ok = source.transferFrom(from, address(this), tokenId, 10 ** source.decimals(tokenId));
+        require(ok, CanNotBeTransferred());
+
+        emit Locked(source, tokenId);
+
+        return nftId;
+    }
+
+    /// @inheritdoc INftEscrow
+    function unlock(IERC6909 source, uint256 tokenId, address to) external auth {
+        bool ok = source.transferFrom(address(this), to, tokenId, 10 ** source.decimals(tokenId));
+        require(ok, CanNotBeTransferred());
+
+        emit Unlocked(source, tokenId);
+    }
+
+    /// @inheritdoc INftEscrow
+    function attach(IERC6909 source, uint256 tokenId, uint256 elementId) public auth returns (uint160) {
+        uint160 nftId = computeNftId(source, tokenId);
+
+        require(elementId != 0, InvalidElement());
+        require(associations[nftId] == 0, AlreadyAttached());
+        require(source.balanceOf(address(this), tokenId) > 0, NotLocked());
+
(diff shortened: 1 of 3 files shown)


src/NftEscrow.sol
```diff
@@ -0,0 +1,63 @@
+// SPDX-License-Identifier: UNLICENSED
+pragma solidity 0.8.28;
+
+import {Auth} from "src/Auth.sol";
+import {IERC6909, INftEscrow} from "src/interfaces/INftEscrow.sol";
+
+contract NftEscrow is Auth, INftEscrow {
```
lemunozm, line 7: I didn't name it `RwaEscrow` given the Jeroen's comment. I also think collateral at this level is very specific. The escrow is generic enough to handle any NFT through ERC6909, so I called it `NftEscrow`. Open to more accurate naming

src/NftEscrow.sol
```diff
@@ -0,0 +1,64 @@
...
+    /// A list of NFTs with an associated element.
+    mapping(uint160 nftId => uint256 elementId) public associations;
```
lemunozm: I left the mapping as simple as possible by now. If we finally need more information, such as the `source` and `token_id`, we can add it.

test/unit/NftEscrow.t.sol
```diff
@@ -0,0 +1,132 @@
...
+        );
+    }
+
+    function _mockTransfer(address from, address to, bool result) internal {
+        vm.mockCall(
+            address(nfts),
+            abi.encodeWithSelector(IERC6909.transferFrom.selector, from, to, TOKEN_ID, 10 ** DECIMALS),
+            abi.encode(result)
+        );
+    }
```
lemunozm: NOTE: probably for a dependency as ERC6909, it's better to use our implementation directly than mocks, but by now, I'm forced to mock it.

src/interfaces/INftEscrow.sol
```diff
@@ -0,0 +1,67 @@
+// SPDX-License-Identifier: UNLICENSED
+pragma solidity >= 0.8.0;
+
+interface IERC6909 {
```
peculiarity: I guess once, my PR is merged, you can just import this interface

lemunozm: Yeah, that's the idea

src/NftEscrow.sol
peculiarity: Not sure about NFT either. Remember that we discuss supporting fungible tokens in ERC6909 instance as well.  Why not call it just `Escrow` ? Does adding `Nft` before the name give you any benefit? Do you really care if the  item under specific erc6909 and tokenId is an Nft or not ?

src/NftEscrow.sol
```diff
@@ -0,0 +1,64 @@
...
+
+    constructor(address owner) Auth(owner) {}
+
+    /// @inheritdoc INftEscrow
+    function lock(IERC6909 source, uint256 tokenId, address from) external auth returns (uint160) {
```
peculiarity: I'm just curious if it is possible to "attach" directly when you  lock.
You  must have the item `locked` before you can attach it.
Meanwhile :
 - `nftId` is calculated twice ( in both `lock` and `attach`). 
 - In `attach` it is checked if the item's ownership is transferred to the escrow ( external call ) but you already have this information during `lock`
  - In `attach` you check if it is locked or not... well you have that information in `lock`.
  
Also you "detach" when you `close` in `Portfolio` but is this a needed behavior. 
The actual "detachment" between portfolio and and escrow happens in:
`delete items[poolId][itemId - 1];` - this is from the portfolio.


I think the logic that you have in `detach` should happen when you transfer back the token to the owner which is in the `unlock`

src/interfaces/INftEscrow.sol
```diff
@@ -0,0 +1,67 @@
...
+    function transferFrom(address sender, address receiver, uint256 id, uint256 amount)
+        external
+        returns (bool success);
+    function decimals(uint256 tokenId) external view returns (uint8);
+    function balanceOf(address owner, uint256 id) external returns (uint256 amount);
+}
+
+/// Interface to handle an escrow of nfts
+interface INftEscrow {
+    event Locked(IERC6909 source, uint256 tokenId);
```
peculiarity: I have the feeling you want to make all these  events indexed  on both args

src/NftEscrow.sol
peculiarity: How does this flow work? Who calls `lock`? 
In order for this to work either someone made the escrow an operator or gave allowance to the escrow.
Also whoever is going to call this method , they must be aware of these preconditions.

src/NftEscrow.sol
lemunozm: AFAIK, this escrow is only used for those ERC6909 that represents non-fungible tokens.

src/NftEscrow.sol
lemunozm: The flow is the following:

https://www.notion.so/Portfolio-Specs-12a2eac24e1780caa5e7ef6663512700?pvs=4#12a2eac24e17803e8365e18f199697bf

TLDR: 
- The fundAdmin can call lock()/unlock()
- The portfolio can call attach()/detach() but only if the fundAdmin called lock() first

> I'm just curious if it is possible to "attach" directly when you lock.

This is one of the most significant points discussed in the slack thread. And still opened. The idea to let it split is that different entities can call lock() and attach(), but at the cost of an extra storage for link those calls.

It's an open topic under dicussion, and maybe the Escrow is merged again into the portfolio

src/interfaces/INftEscrow.sol
lemunozm: Right

src/NftEscrow.sol
peculiarity: Sure, but if we decide to support fungile tokens then this escrow's name won't be valid and I'm not sure if it matter for NFT to be part of the name of the contract. What's wrong with just escrow ?

lemunozm: `ERC6909` can work with fungible and non-fungible. But `NftEscrow` can work only for those `ERC6909` that are non-fungible. It's like a restriction it adds to the ERC6909 used. Actually, when it locks/attach, it does checks expecting ERC6909 behaves as a non-fungible. So you can not lock 2 different amounts of the same token id from two accounts, something you will expect to do from a fungible escrow.

So it does not support ERC6909 that represents fungibles

src/NftEscrow.sol
lemunozm: This no longer applies

src/NftEscrow.sol
```diff
@@ -0,0 +1,34 @@
...
+
+contract NftEscrow is Auth, INftEscrow {
+    constructor(address owner) Auth(owner) {}
+
+    /// @inheritdoc INftEscrow
+    function lock(IERC6909 source, uint256 tokenId, address from) external auth {
+        require(source.balanceOf(address(this), tokenId) == 0, AlreadyLocked());
+
+        bool ok = source.transferFrom(from, address(this), tokenId, 1);
+        require(ok, CanNotBeTransferred());
```
lemunozm, line 15: I want to propose a new entry for our style guidelines, which is not to use mutable calls/actions inside `require()`.

Rationale. From the reader's perspective, require implies a condition, a pre-check. I would not expect to perform any action in the chain or modify some state if I'm requiring something. So I would leave such intention clear and leave the mutable call in another line.

Regarding the performance, I've checked this: https://docs.soliditylang.org/en/latest/internals/optimizer.html#commonsubexpressioneliminator

> All subexpressions that are identifiers themselves are replaced by their current value if the value is an identifier.

I think this applies here, so with some minimum optimization level in the compilation phase, the new variable should not add any cost.

Interesting in your thoughts and if you agree on this @wischli @ilinzweilin @peculiarity @hieronx @mustermeiszer

lemunozm:
Ready for review

src/interfaces/INftEscrow.sol
```diff
@@ -0,0 +1,33 @@
...
+    /// @notice Lock a nft transfering one token to the contract.
+    /// No more tokens from the same `tokenId` can be sent to the escrow until it's locked.
+    /// @param source The contract or collection where identify the `tokenId`
+    /// @param tokenId The asset identification
+    /// @param from Address allowed to transfer one token.
+    function lock(IERC6909 source, uint256 tokenId, address from) external;
+
+    /// @notice Unlock a nft transfering back the token from the contract.
+    /// The same `tokenId` can be locked again
+    /// @param source The contract or collection where identify the `tokenId`
```
wischli: Nit
```suggestion
    /// @param source The contract or collection which identifies the `tokenId`
```

src/NftEscrow.sol
```diff
@@ -0,0 +1,29 @@
...
+
+        emit Locked(source, tokenId);
+    }
+
+    /// @inheritdoc INftEscrow
+    function unlock(IERC6909 source, uint256 tokenId, address to) external auth {
+        require(source.balanceOf(address(this), tokenId) == 1, NotLocked());
```
wischli, line 22: `lock` and `unlock` are asymmetrical in their requirements: `lock` throws only if the balance of source is zero such that an NFT with supply > 1 can be locked multiple times. However, you cannot unlock it multiple times due to requiring strict balance of 1.

src/interfaces/INftEscrow.sol
```diff
@@ -0,0 +1,33 @@
...
+    error AlreadyLocked();
+
+    /// @notice NFT not locked in the escrow.
+    error NotLocked();
+
+    /// @notice The asset can not be transfered to/from this contract.
+    error CanNotBeTransferred();
+
+    /// @notice Lock a nft transfering one token to the contract.
+    /// No more tokens from the same `tokenId` can be sent to the escrow until it's locked.
```
wischli: I think this comment causes confusion and should be reworked or removed, see above in asymmetrical lock vs unlock.

wischli (commented):
Ultra nit: Some of your natspec docs are terminated with period `.` and others are not.

src/interfaces/INftEscrow.sol
lemunozm: Yeah, the comment is not ok

> ... can be sent to the escrow until it's locked.

Should be: can be sent to the escrow until it's **unlocked**.

src/NftEscrow.sol
lemunozm: > lock throws only if the balance of source is zero

lock throws only if the balance of *destination* is zero (it means, the own escrow)

That means you can only lock it once.

wischli: But both `lock` and `unlock` check the `source.balanceOf(address(this), tokenId)`?!

For instance, the following test fails with `NotLocked`:
```solidity
    function testMultipleLocks() public {
        _mockBalanceOf(0);
        _mockTransferFrom(OWNER, address(escrow), true);

        escrow.lock(nfts, TOKEN_ID, OWNER);
        escrow.lock(nfts, TOKEN_ID, OWNER);

        // FIXME: Reverts with NotLocked()
        escrow.unlock(nfts, TOKEN_ID, OWNER);
    }
 ```

File: src/interfaces/INftEscrow.sol [hunk 11]
```diff
@@ -0,0 +1,33 @@
...
+
+import {IERC6909} from "src/interfaces/ERC6909/IERC6909.sol";
+
+/// Interface to handle an escrow of nfts.
+interface INftEscrow {
+    event Locked(IERC6909 indexed source, uint256 indexed tokenId);
+    event Unlocked(IERC6909 indexed source, uint256 indexed tokenId);
+
+    /// @notice NFT already locked in the escrow.
+    error AlreadyLocked();
```
wischli, line 12: Food for thought: Maybe we want to add the `IERC6909` source as parameter to all three errors?

src/NftEscrow.sol
lemunozm: Note the first call to `lock()` will imply adding a new mock before the second call:

```solidity
_mockBalanceOf(1);
```
Because the escrow now is the owner of such token. So the second `lock()` will fail

src/interfaces/INftEscrow.sol
lemunozm: Interesting. My first thought is that if it fails, you know the ERC6909 address / tokenId used when it failed.

I would put in errors only values computed by the method or retrieved from the storage. In this case, I think both can be known because they match the parameters of the method.

But open to listening to others' thoughts regarding this too.

src/NftEscrow.sol
wischli: Ah I see, then everything is good!

src/NftEscrow.sol
```diff
@@ -0,0 +1,29 @@
...
+
+        bool ok = source.transferFrom(address(this), to, tokenId, 1);
```
lemunozm, line 24: Note for myself: This is actually:
```suggestion
        bool ok = source.transfer(to, tokenId, 1);
```

mustermeiszer:
Closing for now

PR #23, opened 2024-10-23
Deny warnings only for CI

lemunozm:
Here https://github.com/centrifuge/asset-pools/pull/14 we addws `deny-warnings` option to `foundry.toml`. Nevertheless, this option is super annoying for local development/testing and trying things.

Moved this option only to be applied by the CI but not affect local workflows

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -19,7 +19,7 @@ jobs:
         uses: foundry-rs/foundry-toolchain@v1
 
       - name: Run tests
-        run: forge test --no-match-path "test/invariant/**/*.sol"
+        run: forge test --no-match-path "test/invariant/**/*.sol" --deny-warnings
         env:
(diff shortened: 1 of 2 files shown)


peculiarity:
I'm I get what you mean but  at some point these warnings must be fixed even locally before pushed to github because CI will fail otherwise. If you have them locally, that will enforce you to fix them ASAP and not think about these.

TL;DR - You have to fix the local issues anyways so for me this change doesn't add much value.

lemunozm:
The value is that the warnings (treated as errors) are not annoying during your local development process trying things. It's quite annoying during some test, do not read a parameter of a method and having the compilation crash because of that.

You will want warnings during the try/test phase, but you will want errors once you upload the change to ensure no warnings are introduced in the codebase

peculiarity (approved):

ilinzweilin (approved):

PR #24, opened 2024-10-28
Add `IFile` interface

lemunozm:
# Description

Add a small interface to the `file()` method used in many places. 

Benefits: 
  It factorizes event, error, and documentation to avoid copy/paste them in all contracts

### DIFF
--- a/src/interfaces/IFile.sol
+++ b/src/interfaces/IFile.sol
@@ -1,0 +1,16 @@
+// SPDX-License-Identifier: AGPL-3.0-only
+pragma solidity >=0.5.0;
+
+/// Interface for contracts that requires a `file()` methods in order to update properties
+interface IFile {
+    /// Emit when the `what` parameter of `file()` is not supported by the implementation.
+    error UnrecognizedWhatParam();
+
+    /// @notice Emit when a call to `file()` was performed.
+    event Filed(bytes32 what, address addr);
+
+    /// @notice Updates a contract parameter.
+    /// @param what Name of the parameter to update.
+    /// @param data New value given to the `what` parameter.
+    function file(bytes32 what, address data) external;
+}


peculiarity:
I don't think is a good idea and let me elaborate:
Based on all the contracts so far that we had, it is really hard to generalize the file pattern.
It's not always 1 param for the data. Sometimes it is 2 or 3. They are of different types as well.
You might say that we can extend the `IFile` interface with these new variations but it won't make sense for some contracts to have signatures as part of their ABI that are not used. 

Would like to hear any other thoughts of course.

lemunozm:
Which other params are expected to be used other than the contract addresses? AFAIK, the file pattern is used to attach contract dependencies, which always imply just addresses

hieronx:
@lemunozm the `file` pattern is used for any key parameter changes for contracts, basically for any update methods that are not expected to be called during regular operations (so a setter for a token price is not using this pattern, while a setter for dependencies is). 

A few examples from liquidity pools for non dependency usages of `file`:
- Updating the time lock on the root contract: https://github.com/centrifuge/liquidity-pools/blob/87250853dd0cee6ec56a5e8ab116cfbc0b1b3fce/src/interfaces/IRoot.sol#L44
- Updating the set of adapters on the gateway: https://github.com/centrifuge/liquidity-pools/blob/87250853dd0cee6ec56a5e8ab116cfbc0b1b3fce/src/interfaces/gateway/IGateway.sol#L58
- Updating message ID handler on the gateway: https://github.com/centrifuge/liquidity-pools/blob/87250853dd0cee6ec56a5e8ab116cfbc0b1b3fce/src/interfaces/gateway/IGateway.sol#L73
- Updating token name/symbol on ERC20 token: https://github.com/centrifuge/liquidity-pools/blob/87250853dd0cee6ec56a5e8ab116cfbc0b1b3fce/src/interfaces/token/ITranche.sol#L43

Now, whether we abstract away the file interface for the most common case (filing dependencies), I am 50/50. It does remove a little bit of duplicate interface code. The downside is that the interface is more generic, i.e. it doesn't specify what the possible values for `what` parameter are.

hieronx:
To add context: the main reason we use the `file` pattern, is that we can easily write a monitoring system that alerts on any `File`, `Rely`, and `Deny` events, and through that, detect unexpected patterns. Plus it allows external integrators to do the same.

Plus it creates a consistent pattern for code reviewers and auditors.

lemunozm:
Interesting... Thanks for the explanation and examples. Let's keep this staling by now and let's go with a specific implementation in the portfolio then. I think it's too early to know if we need this "generic" IFile interface.

PR #26, opened 2024-10-31
chore: Implement generic mint/burn functions. Used them according to use case

peculiarity:
The is the have a generic `_mint` and `_burn` function in the `ERC6909` contract.
Each contract that inherits from it will determine how to these two internal functions.

As shown in this PR, _minting_ always mint a single quantity for a specific _tokenId_.
Same applies for _burning_

If we would like to have an `ERC6909` that supports fungible tokens, then we just implement the `mint` and `burn` functions in a contract that inherits from `ERC6909` and allow specifying the amounts. Also that new contract can implement that optional `Token Supply Extension` which will be applicable in that case.


src/ERC6909/ERC6909NFT.sol
```diff
@@ -0,0 +1,54 @@
...
+import {IERC6909NFT} from "src/interfaces/ERC6909/IERC6909NFT.sol";
+import {IERC6909URIExtension} from "src/interfaces/ERC6909/IERC6909URIExtension.sol";
+
+contract ERC6909NFT is IERC6909NFT, ERC6909, Auth {
+    using StringLib for string;
+    using StringLib for uint256;
+
+    uint8 constant MAX_SUPPLY = 1;
+
+    uint256 public latestTokenId;
```
wischli, line 15: Wondering if it is feasible to use `uint128` for `latestTokenId` such that `MAX_SUPPLY` and `latestTokenId` can be fit into the same storage slot.

lemunozm: I think it cannot, because the interface of `ERC6909` forces `uint256` for `tokenId` 🥲

src/ERC6909/ERC6909.sol
```diff
@@ -58,11 +61,33 @@ abstract contract ERC6909 is IERC6909, IERC165 {
         return type(IERC6909).interfaceId == interfaceId || type(IERC165).interfaceId == interfaceId;
     }
 
+    function _mint(address owner, uint256 tokenId, uint256 amount) internal {
+        require(owner != address(0), EmptyOwner());
+        require(tokenId > 0, InvalidTokenId());
```
lemunozm, line 64: Feel like an unnecessary condition for us (?). This contract can deal with tokenId == 0 without problems

src/interfaces/ERC6909/IERC6909NFT.sol
```diff
@@ -4,23 +4,14 @@ pragma solidity 0.8.28;
...
-interface IERC6909Centrifuge is IERC6909, IERC6909URIExtension {
+interface IERC6909NFT is IERC6909, IERC6909URIExtension {
     /// Errors
     error UnknownTokenId(address owner, uint256 tokenId);
-    error EmptyOwner();
-    error EmptyAmount();
-    error EmptyURI();
-    error MaxSupplyReached();
+
     error Burn_InsufficientBalance(address owner, uint256 tokenId);
```
lemunozm: Question, why `Burn_` prefix? Seems like `InsufficientBalance` is an error that can be reused from different methods for the same meaning

lemunozm (approved):
Really really like the new implementation 💯! Much simpler and very organized 🙌🏻 .

A couple of Nits but LGTM

src/ERC6909/ERC6909NFT.sol
hieronx: `MAX_SUPPLY` is a constant so it's not actually in storage, just included in the bytecode

hieronx: I like the thinking though :)

src/interfaces/ERC6909/IERC6909NFT.sol
peculiarity: That's a leftover. I haven't cleaned that error properly

peculiarity: I will fix it in the other branch

src/ERC6909/ERC6909.sol
peculiarity: How come this is an unnecessary condition ?

src/interfaces/ERC6909/IERC6909NFT.sol
peculiarity: Actually I fixed it.

src/ERC6909/ERC6909.sol
peculiarity: I don't agree that `tokenId` should be treated as a normal id. 
Default values are the 0 values. This might cause  unexpected behavior / issues.

src/ERC6909/ERC6909.sol
```diff
@@ -3,12 +3,15 @@ pragma solidity 0.8.28;
...
 import {IERC6909} from "src/interfaces/ERC6909/IERC6909.sol";
 import {IERC165} from "src/interfaces/IERC165.sol";
+import {StringLib} from "src/libraries/StringLib.sol";
 
 /// @title      Basic implementation of all properties according to the ERC6909.
 ///
 /// @dev        This implementation MUST be extended with another contract which defines how tokens are created.
 ///             Either implement mint/burn or override transfer/transferFrom.
 abstract contract ERC6909 is IERC6909, IERC165 {
+    using StringLib for string;
```
hieronx: Looks like this is unused now

src/ERC6909/ERC6909.sol
hieronx: Agreed that enforcing the token ID not being 0 makes sense

src/ERC6909/ERC6909.sol
```diff
@@ -58,11 +61,33 @@ abstract contract ERC6909 is IERC6909, IERC165 {
...
+        balanceOf[owner][tokenId] += amount;
+
+        emit Transfer(msg.sender, address(0), owner, tokenId, amount);
+    }
+
+    function _burn(address owner, uint256 tokenId, uint256 amount) internal {
+        uint256 balance = balanceOf[owner][tokenId];
+        require(balance >= amount, InsufficientBalance(msg.sender, tokenId));
+
+        /// the underflow check is handled by the require line above
```
hieronx: Ultra-ultra-nit, only double slash for inline comments:
```suggestion
        // The underflow check is handled by the require line above
```

src/ERC6909/ERC6909NFT.sol
```diff
@@ -0,0 +1,54 @@
...
+
+import {ERC6909} from "src/ERC6909/ERC6909.sol";
+import {StringLib} from "src/libraries/StringLib.sol";
+import {Auth} from "src/Auth.sol";
+import {IERC6909NFT} from "src/interfaces/ERC6909/IERC6909NFT.sol";
+import {IERC6909URIExtension} from "src/interfaces/ERC6909/IERC6909URIExtension.sol";
+
+contract ERC6909NFT is IERC6909NFT, ERC6909, Auth {
+    using StringLib for string;
+    using StringLib for uint256;
```
hieronx: Why is this needed? 🤔

peculiarity: this is rather not used any more.

peculiarity: removed

hieronx (approved):

src/ERC6909/ERC6909.sol
lemunozm: Makes sense. The non-existence behavior as the default value makes me feel yet uncomfortable

PR #31, opened 2024-11-25
PoolManager first impl

### LINKED ISSUE #28: Init `PoolManager`
* Create the `PoolManager` contract and add boilerplate for handling decoded incoming messages forwarded from the Gateway
   * Example: Gateway receives a message which decodes to `RequestDeposit { .. }` and calls `PoolManager.requestDeposit(..)`
   * NOTE: TBD if Gateway does encoding/decoding, definitely shouldn't be done by the `PoolManager` to reduce bloat
* Of course, not all parameters of the PoolManager handlers are fully clear -> decide on your own for which function declarations you want to leave out some/all parameters 
   E.g.
   ```solidity
  // If unsure
  function requestDeposit() { ... }
  
  // If somewhat clear
  function requestDeposit(uint64 poolId, bytes16 shareClassId, address investor, address accountingInvestorToken, uint128 amount)

* For now, error out with some `Unsupported` error for all incoming handler calls. We will incrementally add handlers for incoming messages in subsequent PRs
   E.g.
   ```solidity
   function requestDeposit(uint64 poolId, bytes16 shareClassId, address investor, address accountingInvestorToken, uint128 amount) {
      revert("Unsupported");
   }
   
* ~~Copy and paste messages from CV repository~~
* ~~Apply minor changes discussed during offsite: https://www.notion.so/LP-Integration-and-Changes-1342eac24e17802aa954c073a5b916e4~~

lemunozm:
fixes #28 

Just as draft to show the unlocking mechanism.

I think if we know the entry point is always `PoolManager`, we no longer need the complex `unlock(poolId, array_of_encoded_calls)`. Simply:
- `multicall`
  - `unlock(poolId)`
  - `approveDistribution()`
  - Do whatever onchain/offchain
  - `issueShares()`
  - lock() //optional
  
The caller (fund manager) is free in the multicall to compose any method from `PoolManager` as they want to perform its custom business logic.

### DIFF
--- a/src/PoolManager.sol
+++ b/src/PoolManager.sol
@@ -1,0 +1,62 @@
+// SPDX-License-Identifier: AGPL-3.0-only
+pragma solidity 0.8.28;
+
+import {Auth} from "src/Auth.sol";
+
+contract PoolRegistry is Auth {
+    constructor(address owner) Auth(owner) {}
+
+    // Can also be called isFundManager or be extracted in a Permissions contract
+    // NOTE: The gateway contract is able to unlock any poolId
+    function isUnlocker(address who, uint64 poolId) auth public view returns (bool) {}
+
+    // Associate who to be the owner/poolAdmin of the new poolId
+    function registerPool(address who) auth public returns (uint64) {}
+}
+
+contract PoolManager is Auth {
+    PoolRegistry poolRegistry;
+    uint64 transient unlockedPool;
+
+    constructor(address owner) Auth(owner) {
+        poolRegistry = PoolRegistry(address(this));
+    }
+
+    modifier poolUnlocked() {
+        require(unlockedPool != 0);
+        _;
+    }
+
+    // This method is called first in a multicall
+    function unlock(uint64 poolId) external {
+        require(poolRegistry.isUnlocker(msg.sender, poolId));
+        unlockedPool = poolId;
+    }
+
+    // In case the fundManager wants to lock in the same multicall to do other actions
+    function lock() external {
+        require(poolRegistry.isUnlocker(msg.sender, unlockedPool));
+        unlockedPool = 0;
+    }
+
+    // ---- Calls that require to unlock ----
+
+    function allowPool(uint32 chainId) external poolUnlocked() {}
+    function allowShareClass(uint32 chainId) external poolUnlocked() {}
+
+    function depositRequest() external poolUnlocked() {
+        // Retriver the unlocked poolId to use it internally
+        shareClassManager.depositRequest(unlockedPool, ..);
+    }
+    function approveSubscription(uint64 poolId) external poolUnlocked() {}
+    function issueShares(uint64 poolId) external poolUnlocked() {}
+
+    // ---- Permissionless calls ----
+
+    function registerPool() external returns (uint64) {
+        // Dispatch some event associated to PoolManager
+        return poolRegistry.registerPool(msg.sender);
+    }
+
+    function claimDistribute(uint64 poolId) external {}
+}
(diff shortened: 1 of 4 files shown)


lemunozm:
I think this scheme will work for the PoC. Considering that `PoolManager` is the entry point, I also think it will work for the use cases where the user wants to do a Uniswap swap in between if using multising. Tell me WDYT @mustermeiszer @wischli

lemunozm:
As a note, probably we should ward somehow the calls that comes from the gateway to avoid the fund manager to call them and break some invariant between CP and CV. i.e. We do not want the fund manager to emulater a depositRequest, that didn't happened in CV

src/PoolManager.sol
```diff
@@ -0,0 +1,62 @@
...
+        // Retriver the unlocked poolId to use it internally
+        shareClassManager.depositRequest(unlockedPool, ..);
+    }
+    function approveSubscription(uint64 poolId) external poolUnlocked() {}
+    function issueShares(uint64 poolId) external poolUnlocked() {}
+
+    // ---- Permissionless calls ----
+
+    function registerPool() external returns (uint64) {
+        // Dispatch some event associated to PoolManager
```
mustermeiszer: ensure its never `0`

src/PoolManager.sol
```diff
@@ -0,0 +1,62 @@
...
+    }
+
+    modifier poolUnlocked() {
+        require(unlockedPool != 0);
+        _;
+    }
+
+    // This method is called first in a multicall
+    function unlock(uint64 poolId) external {
+        require(poolRegistry.isUnlocker(msg.sender, poolId));
```
mustermeiszer, line 84: should also `acconting.unlock(poolId)`

src/PoolManager.sol
```diff
@@ -0,0 +1,62 @@
...
+        unlockedPool = poolId;
+    }
+
+    // In case the fundManager wants to lock in the same multicall to do other actions
+    function lock() external {
+        require(poolRegistry.isUnlocker(msg.sender, unlockedPool));
```
mustermeiszer: should also `acconting.lock(poolId)` 

locking the accounting will check whether debit/credit are in balance.

src/PoolManager.sol
```diff
@@ -0,0 +1,62 @@
...
+    // Associate who to be the owner/poolAdmin of the new poolId
+    function registerPool(address who) auth public returns (uint64) {}
+}
+
+contract PoolManager is Auth {
+    PoolRegistry poolRegistry;
+    uint64 transient unlockedPool;
+
+    constructor(address owner) Auth(owner) {
+        poolRegistry = PoolRegistry(address(this));
```
mustermeiszer: needs `accounting` as parameter

src/PoolManager.sol
lemunozm: Exactly what should never be 0?

src/PoolManager.sol
lemunozm: Sure! And a bunch of other contracts, will add them 👍🏻

src/PoolManager.sol
lemunozm: Curious, what `accounting.unlock` will do?

src/PoolManager.sol
wischli: The pool id which we are registering.

lemunozm: But that's comes from the pool registry, who creates it, right? It can not be zero

wischli: Sorry it was just an assumption and a poorly phrased one. I meant the CP-pool id without the chain id. But thinking about it, I might be wrong.

src/PoolManager.sol
lemunozm: cc @mustermeiszer

PR #32, opened 2024-11-25
Use BUSL license

hieronx:


lemunozm (approved):

src/interfaces/IAuth.sol
```diff
@@ -1,5 +1,4 @@
-// SPDX-License-Identifier: AGPL-3.0-only
-pragma solidity >=0.5.0;
+// SPDX-License-Identifier: BUSL-1.1
```
wischli, line 1: Reason for CI failure

wischli (commented):
LGTM apart from the unwanted solidity version removal

src/interfaces/IAuth.sol
hieronx: Good catch! Fixed

wischli (approved):

PR #33, opened 2024-11-26
PoolRegistry that stores pools' configurations.

peculiarity:


peculiarity:
Current questions that arised.
- Do we want to have methods that can change the `ShareClassManager`? ( Yes ) 
- Are we completely sure that we would like to use the `chainId` as part of the `PoolId`? `block.chainid` is `uint256`. ( yes )
- Do we want to have multiple pool managers or one is sufficient? I know that according to the general diagram, it is supposed to have multiple managers (unlockers) but just double checking. ( double mapping )
- Do we still collect a “large fee”? Is this large fee in `eth` ? Where is this fee going to be stored? Some escrow owned by us, maybe? ( n / a for now ) 
- Do we really want to manage everything through `PoolManager`? Do we have to somehow ensure that the `PoolManager` passed the correct `manager` address? ( yes )
- What is the metadata supposed to be? ( yes )

wischli:
Here are my notes from our sync: 
    
  > Do we want to have methods that can change the `ShareClassManager`?
  > 
  - Updating can be allowed
  - Migration should be part of the `IShareClassManager` impl
  - If FM screws up pool, its their fault
  
  > Are we completely sure that we would like to use the `chainId` as part of the `PoolId`? `block.chainid` is `uint256`
  > 
  - Most important: `poolId` needs to be unique across all chains
  - If cannot derive the `chainId` from the `poolId`, we need to change all/most LP messages such that CV knows where to send messages (because right now all LP messages related to pools are sent to Centrifuge Chain)
  - In short: Yes, if `chainId` exceeds `uint32`, we cannot launch there
  
  > Do we want to have multiple pool managers or one is sufficient? I know that according to the general diagram, it is supposed to have multiple managers (unlockers) but just double checking.
  > 
  - Eventually, we should use Permissions
  - Recommendation: Use double mapping `mapping(poolId -> mapping(address -> bool))`
  
  > Do we still collect a “large fee”? Is this large fee in `eth` ? Where is this fee going to be stored? Some escrow owned by us, maybe?
  > 
  - Fees won’t be in ETH
  - We can assume to have some treasury address which receives fees
  
  > Do we really want to manage everything through `PoolManager`? Do we have to somehow ensure that the `PoolManager` passed the correct `manager` address?
  > 
  - Argument for always going through `PoolManager`: Easier for UI, default `multicall` support
  
  > What is the metadata supposed to be?
  > 
  - `bytes32` are fine
  - Usually will be IPFS address
  - But can be anything: Filepath, some string

src/interfaces/IPoolRegistry.sol
```diff
@@ -0,0 +1,23 @@
+// SPDX-License-Identifier: BUSL-1.1
+pragma solidity 0.8.28;
+
+interface IPoolRegistry {
+    /// Types
+    type PoolId is uint64;
+    type CurrencyId is address;
```
lemunozm: I would move it out from `IPoolRegistry` because almost all contracts will use them without requiring `IPoolRegistry`. Probably some `types/Domain.sol` file or similar.

src/PoolRegistry.sol
```diff
@@ -0,0 +1,52 @@
...
+    mapping(PoolId => address) public poolManagers;
+    mapping(PoolId => address) public shareClassManagers;
+    mapping(PoolId => CurrencyId) public poolCurrencies;
+
+    constructor(address deployer) Auth(deployer) {}
+
+    /// @inheritdoc IPoolRegistry
+    function registerPool(CurrencyId poolCurrency, address shareClassManager)
+        external
+        payable
```
lemunozm: Why payable?

src/PoolRegistry.sol
```diff
@@ -0,0 +1,52 @@
...
+        poolManagers[poolId] = msg.sender;
+        poolCurrencies[poolId] = poolCurrency;
+        shareClassManagers[poolId] = shareClassManager;
+
+        emit NewPool(poolId, msg.sender);
+    }
+
+    /// @inheritdoc IPoolRegistry
+    function changeManager(address currentManager, PoolId poolId, address newManager) external auth {
+        require(poolManagers[poolId] == currentManager, NotManagerOrNonExistingPool());
```
lemunozm: I would say the error is just `NotExistingPool()`, same for the metadata

src/PoolRegistry.sol
```diff
@@ -0,0 +1,52 @@
...
+import {MathLib} from "src/libraries/MathLib.sol";
+import {IPoolRegistry} from "src/interfaces/IPoolRegistry.sol";
+
+contract PoolRegistry is Auth, IPoolRegistry {
+    using MathLib for uint256;
+
+    uint32 public latestId;
+
+    mapping(PoolId => bytes) public poolMetadata;
+    mapping(PoolId => address) public poolManagers;
```
lemunozm: The correct name here is "unlockers" or "poolAdmins"

lemunozm (commented):
Just a minor details, but as I said, looks really nice!

src/PoolRegistry.sol
wischli: I suppose this refers to the fee for registering a pool. I am curious about the overhead of requiring some ERC20 (most ideally (w)CFG) as payment. IIUC, this requires adding an ERC20 contract and requiring the `msg.sender` to have appropriate allowance for transferring the fee via the
`PoolRegistry` to some treasury address?

lemunozm: > I suppose this refers to the fee for registering a pool.

Yes, I think if we go with cfg fees in `PoolManager` we do not need payable here. I think ERC20 transfers are quite cheap, just writing in a couple of storages inside the ERC20 implementation

src/interfaces/IPoolRegistry.sol
peculiarity: Yeah, I was thinking about having a separate folder for types .
So something like `types/PoolRegistry.sol` you mean or ?

src/PoolRegistry.sol
peculiarity: Yeah... that needs to be refactored. I just put it there considering the "fee" but since it will be most likely part of the `PoolManager` then this is not needed.

src/PoolRegistry.sol
peculiarity: I definitely don't like `unlockers`  and I'm ready to debate.  I can go with `poolAdmins`.

src/PoolRegistry.sol
peculiarity: In general this must be changed. I assumed all these methods will be called directly so yeah.. :)

src/interfaces/IPoolRegistry.sol
lemunozm: But I would not call it PoolRegistry, I think it should be a general file that almost all of our contracts import it

src/PoolRegistry.sol
lemunozm: I like more `poolAdmins` too, but it all depends on whether we have a permission layer whose names can collide with this one. But pool admin seems reasonable by now

src/PoolRegistry.sol
wischli: Not sure if the fee payment should indeed be moved to the PoolManager. I still see it as part of the PoolRegistry but maybe the allowance target should rather be the manager? What do you think is cleaner @peculiarity?

lemunozm: I think one of the main reason to be in the `PoolManager`, is that the pool manager has access to a future Treasury contract or some ERC6909 escrow. Also, how the fee is computed can be easily upgraded just changing the `PoolManager` contract.

peculiarity: I also think that fee collection should be at a higher level.

peculiarity:
- [Uniswap-v4](https://github.com/Uniswap/v4-core/tree/main/src/types) is used as an inspiration for the types structure
- We check for the `shareClassManager` on the assumption that this cannot be empty. There must be a way see if the  a given `poolId` exists and the way to do so is to check if for a given `poolId` there is a non-zero `shareClassManager`
- We check the `admin` if it’s non-empty address because there is an assumption that only `admin` can call `modifyAdmin`  That check most likely happen in on `PoolManager` level.
- There is no way to update the currency ( this isn’t a behavior we would like to have ) so we would like to make sure it is non-empty.
- I know that `0x0` philosophically doesn’t differ from any address that the is not the fund manager but what is special about this case is that all of these methods are called within a contract so when they are called, some non-initialized variable will be passed which will have a default empty value. These checks will display that something is wrong. You don’t rely on the caller to provide the right values.
- Do we want to pass the address of the admin who is calling the `update` / `modify` ? Reasoning: So that an admin can’t delete themselves from the poolAdmins and also it will be included in the event.

src/libraries/PoolIdLib.sol
```diff
@@ -0,0 +1,10 @@
+// SPDX-License-Identifier: AGPL-3.0-only
+pragma solidity 0.8.28;
+
+import {PoolId} from "src/types/PoolId.sol";
+
+library PoolIdLib {
+    function chainId(PoolId poolId) internal pure returns (uint64) {
```
lemunozm: I think `chainId` can be `uint32`. Just a thought: should `chainId` also be a type like `PoolId`?

src/PoolRegistry.sol
```diff
@@ -0,0 +1,70 @@
...
+        external
+        auth
+        returns (PoolId poolId)
+    {
+        require(admin != address(0), EmptyAdmin());
+        require(Currency.unwrap(currency) != address(0), EmptyCurrency());
+        require(shareClassManager != address(0), EmptyShareClassManager());
+
+        // TODO: Make this part of the library. Something like PoolId.generate();
+        poolId = PoolId.wrap((uint64(block.chainid.toUint32()) << 32) | uint64(latestId++));
```
lemunozm: I think we should use `nextId` instead of `latestId`
```suggestion
        poolId = PoolId.wrap((uint64(block.chainid.toUint32()) << 32) | uint64(++nextId));
```

We do not want to generate a `poolId == 0` because that pool can not pass some nullity checks

src/PoolRegistry.sol
```diff
@@ -0,0 +1,70 @@
...
+
+        poolAdmins[poolId][admin] = true;
+        poolCurrencies[poolId] = currency;
+        shareClassManagers[poolId] = shareClassManager;
+
+        emit NewPool(poolId, msg.sender);
```
lemunozm: I think it should be `admin` instead of `msg.sender`. We don't care who creates the pool but we care about who can control the pool

lemunozm: Not sure if we need to also add `currency` a `shareClassManager` to the event 🤔

src/PoolRegistry.sol
```diff
@@ -0,0 +1,70 @@
...
+    }
+
+    /// @inheritdoc IPoolRegistry
+    function updateAdmin(PoolId poolId, address admin, bool canManage) external auth {
+        require(admin != address(0), EmptyAdmin());
+        require(shareClassManagers[poolId] != address(0), NonExistingPool(poolId));
```
lemunozm, line 45: NIT. Given that `shareClassManagers[poolId]` is updatable and change but `poolCurrencies[poolId]` doesn't, it feels like `poolCurrency[poolId]` is a better candidate to check the non-existance. But nothing wrong with the current implementation

src/interfaces/IPoolRegistry.sol
```diff
@@ -0,0 +1,28 @@
...
+    event NewPoolMetadata(PoolId indexed poolId, bytes metadata);
+    event NewShareClassManager(PoolId indexed poolId, address shareClassManager);
+
+    /// Errors
+    error NonExistingPool(PoolId id);
+    error EmptyAdmin();
+    error EmptyCurrency();
+    error EmptyShareClassManager();
+
+    /// @notice TODO
```
lemunozm, line 21: A reminder here to not forget this before merging it

lemunozm (commented):
Looks pretty good. Some NITs and one comment regarding how poolId starts incrementing. But beside that, ready to be merge. Good job!

src/libraries/PoolIdLib.sol
```diff
@@ -0,0 +1,10 @@
+// SPDX-License-Identifier: AGPL-3.0-only
+pragma solidity 0.8.28;
+
+import {PoolId} from "src/types/PoolId.sol";
+
+library PoolIdLib {
```
lemunozm, line 6: Question, why is this a lib instead of just adding this method to the `PoolId` type?

test/unit/PoolRegistry.t.sol
```diff
@@ -0,0 +1,129 @@
...
+    using MathLib for uint256;
+
+    PoolRegistry registry;
+    Currency USD = Currency.wrap(address(840));
+    address shareClassManager = makeAddr("shareClassManager");
+
+    modifier nonZero(address addr) {
+        vm.assume(addr != address(0));
+        _;
+    }
```
lemunozm, line 21: That's cool for preconditions! I will probably use it

src/PoolRegistry.sol
```diff
@@ -0,0 +1,70 @@
...
+    using MathLib for uint256;
+
+    uint32 public latestId;
+
+    mapping(PoolId => bytes) public metadata;
+    mapping(PoolId => mapping(address => bool)) public poolAdmins;
+    mapping(PoolId => address) public shareClassManagers;
+    mapping(PoolId => Currency) public poolCurrencies;
+
+    constructor(address deployer) Auth(deployer) {}
```
lemunozm, line 20: NIT. It does not need to be the deployer, it can be `root`, or any contract with auth access like the `PoolManager`. I would rename this to `owner` to be more generic here

src/types/Currency.sol
```diff
@@ -0,0 +1,4 @@
+// SPDX-License-Identifier: BUSL-1.1
+pragma solidity 0.8.28;
+
+type Currency is address;
```
mustermeiszer, line 4: @peculiarity don't we need to ensure that we can get at least the decimals from the currency or is this not needed at this level?

src/PoolRegistry.sol
mustermeiszer: Plus one for that!

src/PoolRegistry.sol
mustermeiszer: We definitely should add the `ShareClassManager` then the indexing logic can dynamically start indexing that contract If it is not already indexed.

Regarding currency, I would also use it, but no strong feelings.

src/PoolRegistry.sol
mustermeiszer: We should also allow to update the currency! I would say, empty admin = non existing pool.

mustermeiszer (commented):
Looks great! 

Change request: 
- allow to update currency of a pool

src/PoolRegistry.sol
lemunozm: Actually `latestId` is indeed the correct name, just the preincrement

src/PoolRegistry.sol
lemunozm: > We should also allow to update the currency!

Really? Can not that break all the invariants regarding how investor tokens were converted into the pool currency by the oracles?

mustermeiszer: We already changed the currency for `Anemoy` in a migration. I guess there will be multiple instances. If they wanna da that AND change the decimals of the base currency, they need to handle the changes in valuation IMO. 

If we deem it to dangerous at one point, we can always just not expose it through the `PoolManager` right?

src/PoolRegistry.sol
```diff
@@ -0,0 +1,70 @@
+// SPDX-License-Identifier: BUSL-1.1
+pragma solidity 0.8.28;
+
+import {Auth} from "src/Auth.sol";
+import {PoolId} from "src/types/PoolId.sol";
+import {Currency} from "src/types/Currency.sol";
+import {MathLib} from "src/libraries/MathLib.sol";
+import {IPoolRegistry} from "src/interfaces/IPoolRegistry.sol";
+
+import "forge-std/console.sol";
```
wischli: Reminder to remove the console logger dep

src/PoolRegistry.sol
```diff
@@ -0,0 +1,70 @@
+// SPDX-License-Identifier: BUSL-1.1
+pragma solidity 0.8.28;
+
+import {Auth} from "src/Auth.sol";
+import {PoolId} from "src/types/PoolId.sol";
+import {Currency} from "src/types/Currency.sol";
```
wischli, line 6: I would like to start a discussion if we still want to keep `Currency` and `CurrencyId` or rather refer to it as `Asset` and `AssetId`? WDYT @lemunozm and @peculiarity

src/PoolRegistry.sol
wischli: +1 for including the `Currency` as well. Apps need to know the decimals of the pool denomination

src/PoolRegistry.sol
```diff
@@ -0,0 +1,70 @@
...
+    }
+
+    /// @inheritdoc IPoolRegistry
+    function updateAdmin(PoolId poolId, address admin, bool canManage) external auth {
+        require(admin != address(0), EmptyAdmin());
+        require(shareClassManagers[poolId] != address(0), NonExistingPool(poolId));
+
+        poolAdmins[poolId][admin] = canManage;
+
+        emit NewPoolManager(admin);
```
wischli: Maybe change request: Missing `poolId` here IMO.

src/interfaces/IPoolRegistry.sol
```diff
@@ -0,0 +1,28 @@
...
+    event NewPool(PoolId indexed poolId, address indexed manager);
+    event NewPoolManager(address indexed manager);
+    event NewPoolMetadata(PoolId indexed poolId, bytes metadata);
+    event NewShareClassManager(PoolId indexed poolId, address shareClassManager);
+
+    /// Errors
+    error NonExistingPool(PoolId id);
+    error EmptyAdmin();
+    error EmptyCurrency();
+    error EmptyShareClassManager();
```
wischli, line 19: Curious why we exclude the `poolId` here. I thought it would be valuable information. Imagine a client approaches us with some errors from calls they did. I envision the debugging much easier if errors include some of the input arguments for calls in which they are thrown. But maybe this is considered bad practise in solidity?

src/libraries/MathLib.sol
```diff
@@ -57,7 +57,8 @@ library MathLib {
     function mulDiv(uint256 x, uint256 y, uint256 denominator) internal pure returns (uint256 result) {
         unchecked {
             // 512-bit multiply [prod1 prod0] = x * y. Compute the product mod 2^256 and mod 2^256 - 1, then use
-            // use the Chinese Remainder Theorem to reconstruct the 512 bit result. The result is stored in two 256
+            // use the Chinese Remainder Theorem to reconstruct the 512 bit result. The result is stored in two
+            // 256https://www.alchemy.com/overviews/solidity-call
```
wischli: Change request: Accidental paste spotted 😄 
```suggestion
            // use the Chinese Remainder Theorem to reconstruct the 512 bit result. The result is stored in two 256
```

src/libraries/MathLib.sol
```diff
@@ -155,6 +156,13 @@ library MathLib {
         return uint8(value);
     }
 
+    function toUint32(uint256 value) internal pure returns (uint32) {
+        if (value > type(uint32).max) {
+            revert("MathLib/uint32-overflow");
+        }
+        return uint32(value);
```
wischli, line 162: Q: Is there any reason to continue the seemingly outdated `if (condition) { revert(..) }` pattern over `require(...)`? Suppose you just wanted to stay consistent with the rest of the file? Just asking if there is any Solidity reason. 

Otherwise, I will open A PR at a later stage to refactor using `require` as I did that in #9 which is stale for now.

src/libraries/PoolIdLib.sol
wischli: I agree `uint32` should suffice.

wischli (commented):
Looks really good! Adding my few cents to the centrifuge

src/PoolRegistry.sol
lemunozm: > If we deem it to dangerous at one point, we can always just not expose it through the PoolManager right?

Yes. Actually, I would not expose this for the PoC.

There are some scenarios that I'm not sure how they would work, i.e. The fund manager changes the currency in the middle of some investment processes. That would break a lot of invariants in `ShareClassManager` regarding pending/approval amounts.

src/libraries/MathLib.sol
lemunozm: BTW, from where did you bring the code from Mathlib @peculiarity ? I would like to bring the associated tests too

wischli: > BTW, from where did you bring the code from Mathlib @peculiarity ? I would like to bring the associated tests too

From here: https://github.com/centrifuge/liquidity-pools/blob/main/src/libraries/MathLib.sol

src/PoolRegistry.sol
lemunozm: Mmm... The good thing about using `Currency` is that you know it's a fungible token, and you can ask for decimals and you should also take care about decimals. `AssetId` would not imply that. `AssetId` for me is a higher level abstraction that just represents the UUID of such currency and can be more things than currencies

src/interfaces/IPoolRegistry.sol
```diff
@@ -0,0 +1,28 @@
...
+
+import {PoolId} from "src/types/PoolId.sol";
+import {Currency} from "src/types/Currency.sol";
+
+interface IPoolRegistry {
+    /// Events
+    event NewPool(PoolId indexed poolId, address indexed manager);
+    event NewPoolManager(address indexed manager);
+    event NewPoolMetadata(PoolId indexed poolId, bytes metadata);
+    event NewShareClassManager(PoolId indexed poolId, address shareClassManager);
```
wischli: Curious if we should also index the `shareClassManager`.

lemunozm: We should, yeah

src/libraries/MathLib.sol
lemunozm: Ok, which seems a subset of: https://github.com/OpenZeppelin/openzeppelin-contracts/blob/master/contracts/utils/math/Math.sol

And the related tests should be https://github.com/OpenZeppelin/openzeppelin-contracts/blob/master/test/utils/math/Math.t.sol

src/PoolRegistry.sol
lemunozm: > If we deem it to dangerous at one point, we can always just not expose it through the PoolManager right?

So you mean, that the `PoolManager` have some method like "migrateCurrency" that performs all required checks and actions in `ShareClassManager`, to perform a correct currency pool change, right?

lemunozm (changes requested):
I almost forget that we need to also store the `IItemManager` per Pool in the registry.

@mustermeiszer AFAIK, you can have several `IItemManager` per pool right? Still not sure how to store them. i.e. if `Tokens` is a `ItemManager`, then the `PoolManager` should know that just that stored `ItemManager` is the `Tokens` to call to `updateHolding` over it and not over others itemsManagers. Or should that address be passed by parameter by the FM?

src/libraries/PoolIdLib.sol
peculiarity: Are you talking about the return value, if so then indeed it must be `uint32` No idea why it's `uint64`

peculiarity: @lemunozm  To answer your question - I don't think `chainId` should be `chainId` or at least not at this stage. It is not used anywhere ( prod code ). Current usage is for tests and I assumed that we might need the `chainId` somewhere. I guess if we see heavy usage of `chainId` as dependency to methods, then we can refactor. If you agree, please resolve.

src/PoolRegistry.sol
peculiarity: My assumption is that the `PoolId` cannot be `0` even if the counter is `0`. You cannot have `chainId` which is 0 so you can have for example pool with id `1 | 0`. What nullity checks are we talking about?

src/PoolRegistry.sol
peculiarity: I can include `Currency` but it won't be indexed because you can have at most 3 indexed params.

peculiarity: Not sure if it makes sense for the `poolid` to be indexed though. The poolId is generated  but if we index the `Currency` then we can see all pools created for a given currency.  What are you thoughts?

src/PoolRegistry.sol
peculiarity: @lemunozm 
> NIT. Given that shareClassManagers[poolId] is updatable and change but poolCurrencies[poolId] doesn't, it feels like poolCurrency[poolId] is a better candidate to check the non-existance. But nothing wrong with the current implementation

Even though it is upgradable it cannot be  `address(0)` so this is fine IMO.

@lemunozm @mustermeiszer  
I'm a lot a little bit. Do I make a change to allow the currency of the pool and leave it as it is for now?

@mustermeiszer 
> empty admin = non existing pool.

In theory this should be the check indeed but if you see the double mapping, you cannot achieve that.
` mapping(PoolId => mapping(address => bool)) public poolAdmins;`
You can get only the `bool` value and you can't differentiate if it is `false` because the `address` doesn't exist of if for the correct address, someone explicitly set to `false`.

src/interfaces/IPoolRegistry.sol
peculiarity: In LP we did that right before the audit. :sweat_smile:   These signatures might change  so I won't go into that much details for the PoC. It's just  a waste of time IMO

src/libraries/PoolIdLib.sol
peculiarity: In general this is the way you do it in solidity.  I guess you are referring to the `use.... global`, right ?

src/PoolRegistry.sol
peculiarity: Not really. The reason for the deploy is something else.  This will be always the deployers and it is related to how `CREATE2` works. The assumption here is taht we would like to have core contracts deployed with the same address across different chains. We use a deployment script and we have to pass explicitly the deployer address ( It has something to do how scripts with in foundry )

i.e [Escrow](https://github.com/centrifuge/liquidity-pools/blob/main/src/Escrow.sol) in LP

Also whoever calls the constructor is the address that is the "deployer" of the contracts. If `Root` deploys it, then `Root` is the deployer.

src/types/Currency.sol
peculiarity: You can get the currency decimals from the `AssetRegistry` where the asset is registered by provide the address, right ?

src/PoolRegistry.sol
peculiarity: good point

src/interfaces/IPoolRegistry.sol
peculiarity: 1. You throw that very same error during `registerPool` where you dont' have any id there at all.
2. Having some data in these error will cost a little bit more gas being consumed and less funds for gas returned on revert.
3. I have debugged many transactions and this will be a pretty simple call that can be debugged pretty easily.

I don't see any benefit of using it.

src/libraries/MathLib.sol
peculiarity: DAmn .. that's a good one :D

src/libraries/MathLib.sol
peculiarity: We need to migrate but I would do that in a separate PR. Same goes for the `Auth` contract which also uses that old pattern. I think I can create a separate PR for these 2 , once this is merged.

src/interfaces/IPoolRegistry.sol
wischli: Fine with me for methods which are sufficiently descriptive from the naming.

src/PoolRegistry.sol
wischli: > What nullity checks are we talking about?

I suppose at least the transient storage one. But your argument holds true that `poolId != 0` since there is no `chainId` which is zero. So fine with me.

wischli (commented):
No blockers from my side, looks good!

src/PoolRegistry.sol
lemunozm: No strong opinion 👍🏻

src/libraries/PoolIdLib.sol
lemunozm: Similar to how `D18` works, yes. Basically, to avoid two files handling the same thing.

src/interfaces/IPoolRegistry.sol
lemunozm: I agree, but do not lose the point about docs; it helps us to reason about the code to ourselves.

src/PoolRegistry.sol
lemunozm: I do not know which is more important here, maybe `msg.sender` is something we can avoid to index?

src/PoolRegistry.sol
lemunozm: There could be parts where we only want to use the `poolId` without the `chainId` for some process (some cross pool-related stuff in the future), I would keep 0 for nullity. It does not add any cost to reserve the number 0, and we gain in scalability.

src/libraries/PoolIdLib.sol
lemunozm: I think it will be used, because the `chainId` is used for allow things to vaults and for transfer assets, and for all incoming messages. So, it definitely will be used. But I agree it is not required right now; we can modify that later 👍🏻

src/PoolRegistry.sol
lemunozm: The outcome of this thread is:
- We want the change that allows to update the currency in the poolRegistry
- We do not use it by now in the poolManager, only if some feature or migration is required, but because poolRegistry will be immutable, we need that "door" there to modify the currency.

src/PoolRegistry.sol
wischli: I would also favor indexing the `Currency` over the pool manager

src/PoolRegistry.sol
peculiarity: Made the change. Some time ago we agreed that we shouldn't rely on default values so lets' stick to that.

src/PoolRegistry.sol
peculiarity: Done.

src/interfaces/IPoolRegistry.sol
```diff
@@ -10,14 +10,13 @@ interface IPoolRegistry {
     event NewPoolManager(PoolId indexed poolId, address indexed manager);
     event NewPoolMetadata(PoolId indexed poolId, bytes metadata);
     event NewShareClassManager(PoolId indexed poolId, address indexed shareClassManager);
-    event NewItemManager(PoolId indexed poolId, address indexed itemManager);
+    event NewPoolCurrency(PoolId indexed poolId, Currency currency);
```
lemunozm: Nit. I think these events should start with `UpdatedXYZ` (?) Only `NewPool` means a new entry. Others rewrite an existing entry

lemunozm (approved):
Still some threads opened that should be resolved but LGTM!

src/types/Currency.sol
lemunozm: What I think Frederiks is pointing out here is that if the address is not a currency, we want to fail at this point, not later when the decimals are retrieved to perform some stuff

src/PoolRegistry.sol
peculiarity: @lemunozm  there is no msg.sender that is being indexed.

peculiarity: Ok, I index the currency  instead of the Poolid

wischli (approved):

wischli:
AFAICT, there are two leftovers from #33:

1. Add wrapper contract for ERC7726 pool currency to support view `.decimals()` ([ref](https://kflabs.slack.com/archives/C07PG2EUR9C/p1733734897629139?thread_ts=1733234073.846029&cid=C07PG2EUR9C))
2. `ItemManager` mapping ([ref](https://kflabs.slack.com/archives/C07PG2EUR9C/p1733405319988019?thread_ts=1733234073.846029&cid=C07PG2EUR9C))

The former is rather urgent due to being required at least for #35. The latter can be implemented at a later point in time.

Do you agree @peculiarity? Am I missing something?

lemunozm:
I think by now we don't really need the `ItemManager` storage right now, let's see how this evolves.

But agree with the first point. Although it is not an ERC7726, the ERC7726 is the oracle. We need something that can be a parameter to an ERC7726.

As a draft:
```solidity
interface IPoolCurrency {
    function decimals() external returns (uint8);
    function name() external returns (string calldata);
    function symbol() external returns (string calldata);
}
```

EDIT: As discussed in the sync, I think for now, we can just return and store an IERC20 for the pool currency and avoid the above interface. The USD, Euro representation would be some implementors of ERC20 without funds, just with name/symbol/decimal information

peculiarity:
Sure thing

PR #34, opened 2024-11-28
PoolLocker implementation

lemunozm:
### A brief description

`PoolLocker` is the superclass of `PoolManager` that contains and abstracts the lock/unlock and multicall logic that `PoolManager` will use. The idea is to make `PoolManager` as simple as possible to have only business-related methods.

The pool manager will implement the `_lock()` / `_unlock()` virtual methods to perform extra things/checks as ensuring the correct fund manager address or lock the accounting. At the same time, the implementation of `PoolManager` can not write in the `unlocked` transient value, which invariants are ensured by `PoolLocker`.

To perform the call aggregation, `PoolManager` calls to a `Multicall` contract that will call each user contract. We want to avoid `PoolManager` directly do those calls because `PoolManager` is auth for other contracts.


lemunozm:
Ready to review @mustermeiszer @wischli @peculiarity. 

I will create a PR on top of this with the `PoolManager` base implementation

lemunozm:
Given the CI error regarding formatting with transient: https://github.com/foundry-rs/foundry/issues/9088. In the meantime, I propose just removing the transient keyword

mustermeiszer (commented):
Security related: 
@hieronx and @peculiarity how safe is it that the `msg.sender` of the multicalls is the `PoolManager`? If I understand it correctly, that would also mean, that a user could simply call into the `AssetManager` and the `AssetManager` would accept it as the `PoolManager` has `auth`, right? 
Seems like an exploit to me, but maybe I am missing something,.

Just out of interest. 
@peculiarity and @hieronx  `PoolLocker.execute()` is this a composable pattern in solidity? I..e something that others COULD integrate easily into their smart contracts?

lemunozm:
> a user could simply call into the AssetManager and the AssetManager would accept it as the PoolManager has auth, right?

Good point here @mustermeiszer! I think you're right. Probably, we can fix this by doing the caller to `.call()` methods other entity than `PoolManager`.

lemunozm:
Ok, updated. Now `PoolManager` calls to a generic `Multicall` who then calls each separated method. So, for all methods called in the multicall, the `msg.sender` will be that `Multicall` contract.

BTW, does it make sense to use `IMulticall3` (it's in forge-std)? 🤔

src/Multicall.sol
```diff
@@ -0,0 +1,29 @@
...
+
+        results = new bytes[](datas.length);
+
+        for (uint32 i; i < targets.length; i++) {
+            (bool success, bytes memory result) = targets[i].call(datas[i]);
+            if (!success) {
+                // Forward the error happened in target.call().
+                if (!success) {
+                    assembly {
+                        let ptr := mload(0x40)
```
wischli: I am curious about the magic number of `0x40`. Can you add more information about it, ideally as a comment?

src/interfaces/IMulticall.sol
```diff
@@ -0,0 +1,11 @@
+// SPDX-License-Identifier: AGPL-3.0-only
+pragma solidity 0.8.28;
+
+interface IMulticall {
+    /// @notice Dispatched when the `targets` and `datas` length parameters in `execute()` do not matched.
+    error WrongExecutionParams();
+
+    /// @notice Execute a generic multicall.
+    /// In one call fails, it reverts the whole transaction.
```
wischli: Nit:
```suggestion
    /// If one call fails, it reverts the whole transaction.
```

src/Multicall.sol
```diff
@@ -0,0 +1,29 @@
...
+    function aggregate(address[] calldata targets, bytes[] calldata datas) external returns (bytes[] memory results) {
+        require(targets.length == datas.length, WrongExecutionParams());
+
+        results = new bytes[](datas.length);
+
+        for (uint32 i; i < targets.length; i++) {
+            (bool success, bytes memory result) = targets[i].call(datas[i]);
+            if (!success) {
+                // Forward the error happened in target.call().
+                if (!success) {
```
wischli: One of the `if` conditions can be removed

src/interfaces/IPoolLocker.sol
```diff
@@ -0,0 +1,23 @@
...
+
+    /// @notice Dispatched when the `targets` and `datas` length parameters in `execute()` do not matched.
+    error WrongExecutionParams();
+
+    /// @notice Dispatched when the pool is not unlocked to interact with.
+    error PoolLocked();
+
+    /// @notice Returns the unlocked poolId.
+    /// In only will contain a non-zero value if called inside `execute()`
+    function unlockedPoolId() external returns (uint64);
```
wischli: In your implementation, you have declared it as `view`. Suppose that's missing here.

Also, based on [this styleguide](https://www.rareskills.io/post/solidity-style-guide), which @peculiarity shared with us a few weeks ago, view functions should be ordered below non-view ones.

I.e., functions should be sorted by "externality" first, then "state-changingness" second:

1. If applicable: receive > fallback
2. Non-view: external > public > internal > private
3. view: external > public > internal > private
4. pure: external > public > internal > private

test/unit/Multicall.t.sol
```diff
@@ -0,0 +1,65 @@
...
+        bytes[] memory methods = new bytes[](2);
+        methods[0] = abi.encodeWithSelector(userContract.setState.selector);
+        methods[1] = abi.encodeWithSelector(userContract.setState.selector);
+
+        multicall.aggregate(targets, methods);
+
+        assertEq(userContract.state(), 200);
+    }
+
+    function testRevertAtError() public {
```
wischli, line 41: Nit: I suppose eventually we should permutate/fuzz the order in which the error occurs

test/unit/PoolLocker.t.sol
```diff
@@ -0,0 +1,70 @@
...
+
+import "forge-std/Test.sol";
+import {PoolLocker} from "src/PoolLocker.sol";
+import {IPoolLocker} from "src/interfaces/IPoolLocker.sol";
+import {IMulticall} from "src/interfaces/IMulticall.sol";
+import {Multicall} from "src/Multicall.sol";
+
+contract PoolManagerMock is PoolLocker {
+    uint64 public wasUnlock;
+    bool public waslock;
```
wischli: Nit: The naming is misleading as both imply to be `bool` but `wasUnlock` refers to a pool id.

test/unit/PoolLocker.t.sol
```diff
@@ -0,0 +1,70 @@
...
+}
+
+contract PoolLockerTest is Test {
+    uint64 constant POOL_A = 42;
+
+    Multicall multicall = new Multicall();
+    PoolManagerMock poolManager = new PoolManagerMock(multicall);
+
+    function testWithPoolUnlockerMethod() public {
+        address[] memory targets = new address[](2);
```
wischli: Is the array length of `2` intentional or copy-paste-mistake?

test/unit/PoolLocker.t.sol
```diff
@@ -0,0 +1,70 @@
...
+        innerTargets[0] = address(poolManager);
+
+        bytes[] memory innerMethods = new bytes[](1);
+        innerMethods[0] = abi.encodeWithSelector(poolManager.poolRelatedMethod.selector);
+
+        address[] memory targets = new address[](1);
+        targets[0] = address(poolManager);
+
+        bytes[] memory methods = new bytes[](1);
+        methods[0] = abi.encodeWithSelector(poolManager.execute.selector, POOL_A, innerTargets, innerMethods);
```
wischli, line 61: I was just trying to understand why this test was expected to revert beforehand. Now it all makes sense!

src/Multicall.sol
lemunozm: Makes total sense 😅 , will do

src/Multicall.sol
lemunozm: Right! Wrong refactor

src/interfaces/IPoolLocker.sol
lemunozm: Thanks for the reminder and good resume. Will do the check of all methods

It should be `view` but AFAIR, ~~in the interface you can not specify `view` or `pure`~~ (ok, it seems you can), just in the impl, which it is.

test/unit/Multicall.t.sol
lemunozm: What it checks is that when the first error appears (no matter where), it breaks the multicall, and returns, without checking more things or accumulating more errors. How will you fuzz this?

test/unit/PoolLocker.t.sol
lemunozm: Right, they were both `bool` at the beginning 😆, will change!

test/unit/PoolLocker.t.sol
lemunozm: It was copy-paste-mistake, good view!

test/unit/PoolLocker.t.sol
lemunozm: For new reviewers: we want to forbid: `poolManager.execute(POOL_A, poolManager.execute(POOL_B, user.method))`. 

And force all calls to be on one level. If not, `POOL_A` will not call `.lock()` at the end of the execution

mustermeiszer (commented):
NIT: Now that we anyways have a "separate" multicall contract, we can also go with the callback option IMO, then the caller of `execute` can decide which contract to call. 

But we can also refactor that later. 

Also, then with this patter, we could make the `Multicall is Callback.execute(..)` method payable in case we want to pay eth and so on. IIUC that would not be possible with the current design. 

Just some thoughts. Besides that happy.

src/Multicall.sol
lemunozm: Searching and investigating about this, I've found another simpler solution. Just:

```yul
revert(add(result, 32), mload(result))
```

Basically, the `revert(ptr, size)` opcode will throw an exception with some content located in `ptr`.
- `ptr`: `add(result, 32)` which means skipping the size value of the `bytes memory result` array. 
- `size`: just loads the value of the address of the array, which corresponds to the size of the array.

Note the bytes array is encoded as: size (32bytes) + data

src/types/Domain.sol
```diff
@@ -0,0 +1,24 @@
+// SPDX-License-Identifier: BUSL-1.1
+pragma solidity 0.8.28;
+
+type PoolId is uint64;
```
wischli, line 4: Thanks for including my temporary types file 😄

wischli (approved):
LGTM!

lemunozm:
@mustermeiszer @wischli @peculiarity can we merge this?

> Given the CI error regarding formatting with transient: https://github.com/foundry-rs/foundry/issues/9088. In the meantime, I propose just removing the transient keyword

Do you agree?

src/PoolLocker.sol
```diff
@@ -0,0 +1,52 @@
...
+
+    /// @inheritdoc IPoolLocker
+    /// @dev All calls with `poolUnlocked` modifier are able to be called inside this method
+    function execute(PoolId poolId, address[] calldata targets, bytes[] calldata datas)
+        external
+        returns (bytes[] memory results)
+    {
+        require(PoolId.unwrap(unlocked) == 0, PoolAlreadyUnlocked());
+        _unlock(poolId);
+        unlocked = poolId;
```
peculiarity: Isn't the idea of the `_unlock()` function to do this?
You call the internal one but also manually unlock the `poolId`. What is the reason behind this ?

src/PoolLocker.sol
```diff
@@ -0,0 +1,52 @@
...
+
+        results = multicall.aggregate(targets, datas);
+
+        unlocked = PoolId.wrap(0);
```
peculiarity: Isn't this supposed to be done by the `lock()` ?

src/PoolLocker.sol
lemunozm: The `_unlock()` will perform the business logic of the implementor of `PoolLocker` see: https://github.com/centrifuge/pools-internal/pull/36/files#diff-84227e9327ad53c4ad37657de3ed4ee30143b70e038ddabd7e6214eba4f0865cR148-R155

src/PoolLocker.sol
lemunozm: Same reason as above

src/PoolLocker.sol
lemunozm: Basically, the `PoolLocker` does not know about `PoolRegistry` or `Accounting` to perform some locking/unlocking duties. It delegates business control to the implementor

src/PoolLocker.sol
```diff
@@ -0,0 +1,52 @@
...
+import {PoolId} from "src/types/Domain.sol";
+import {IPoolLocker} from "src/interfaces/IPoolLocker.sol";
+import {IMulticall} from "src/interfaces/IMulticall.sol";
+
+abstract contract PoolLocker is IPoolLocker {
+    /// Contract for the multicall
+    IMulticall private immutable multicall;
+
+    /// @dev Represents the unlocked pool Id
+    PoolId private /*TODO: transient*/ unlocked;
```
peculiarity: If this is a transient variable is there a reason why this would be marked as `private` ?

lemunozm: Actually not 👍🏻. Will change that. I still think indirectly that public storage can be modified externally when not xD

peculiarity: This will help you with testing a lot. I saw you created some helper variable in your mock that are not needed.

lemunozm: @peculiarity (ping test)

lemunozm: Oh, I remember why I did it `private`. I do not want to let the implementor of this to modify this variable. Only `PoolLocker` can modify it, not `PoolManager`, if I let public, any method of PoolManager can just do `unlocked = otherPool`

lemunozm: > I saw you created some helper variable in your mock that are not needed.

The helper variable `wasLock` was just to check the hooks are correctly called @peculiarity

peculiarity: Do you expect PoolLocker to be implemented by multiple contract?

peculiarity: If this variable is indeed needed to be private ,please use prefix underscore. 
[Solidity styleguide](https://docs.soliditylang.org/en/latest/style-guide.html#underscore-prefix-for-non-external-functions-and-variables)

peculiarity: Also I would rename it to `_poolId` because you assing a poolID. `unlocked` , `locked` suggests a boolean type. `_unlockedPoolId` would work as well.

src/PoolLocker.sol
```diff
@@ -0,0 +1,52 @@
...
+        unlocked = poolId;
+
+        results = multicall.aggregate(targets, datas);
+
+        unlocked = PoolId.wrap(0);
+        _afterLock();
+    }
+
+    /// @inheritdoc IPoolLocker
+    function unlockedPoolId() public view returns (PoolId) {
```
peculiarity, line 43: Once you rename the variable, this can be actually called with the variable's name without the prefix

lemunozm: Not if it is private, right?

src/PoolLocker.sol
lemunozm: > Do you expect PoolLocker to be implemented by multiple contract?

By different versions of PoolManagers

src/PoolLocker.sol
peculiarity: I understand where the confusion comes from. My fault. Right now it's fine, what I meant was that there was difference between the method and the variable name. 
So the variable will be with underscore prefix and the method will be the same as the variable name but without the prefix.

lemunozm: I left as you said 👍🏻, seems reasonable to me

peculiarity (approved):

PR #36, opened 2024-12-03
PoolManager impl

### LINKED ISSUE #28: Init `PoolManager`

lemunozm:
Based on https://github.com/centrifuge/pools-internal/pull/34

PoolManager implementation

### Updated diagram

image


src/PoolManager.sol
```diff
@@ -0,0 +1,150 @@
...
+
+        uint128 totalApproved = scm.approveDeposit(poolId, scId, assetId, approvalRatio, valuation);
+
+        address pendingShareClassEscrow = holdings.pendingShareClassEscrow(poolId, scId);
+        address shareClassEscrow = holdings.shareClassEscrow(poolId, scId);
+        uint256 erc6909Id = uint256(uint160(AssetId.unwrap(assetId)));
+        assetManager.transferFrom(pendingShareClassEscrow, shareClassEscrow, erc6909Id, totalApproved);
+    }
+
+    function updateHoldings(ShareClassId scId, AssetId assetId, uint128 approvedAmount) external poolUnlocked {
```
lemunozm: Note. This method could actually be `increaseDebt(IItemManager im, ShareClassId scId, ItemId, amount)`, where `IItemManager == Tokens` (aka `Holdings`) and `ItemId == AssetId`.

Slack thread: https://kflabs.slack.com/archives/C07PG2EUR9C/p1733731612273679?thread_ts=1732700295.700449&cid=C07PG2EUR9C

src/PoolManager.sol
```diff
@@ -0,0 +1,195 @@
...
+    }
+
+    function decreaseItemValue(IItemManager im, ItemId itemId, uint128 amount) external poolUnlocked {
+        PoolId poolId = unlockedPoolId();
+
+        IAccountingItemManager.Accounts memory accounts = im.itemAccounts(poolId, itemId);
+        uint128 valueChange = im.decrease(poolId, itemId, amount);
+
+        accounting.updateEntry(accounts.asset, accounts.equity, valueChange);
+    }
```
lemunozm: This will be the equivalent to `updateHoldings(PoolId, ShareClassId, AssetId, amount)`

lemunozm (commented):
@mustermeiszer The PR still requires some polishing but now you can review this. Thanks! 

The investment flow, `updateHoldings()` (now called `increaseItemValue()`, `decreaseItemValue()`) and `moveOut()` should be completed (except some missing check)

src/PoolManager.sol
```diff
@@ -0,0 +1,218 @@
...
+
+        int128 diff = im.update(poolId, itemId);
+
+        IAccountingItemManager.Accounts memory accounts = im.itemAccounts(poolId, itemId);
+        if (diff > 0) {
+            accounting.updateEntry(accounts.asset, accounts.loss, uint128(diff));
+        } else if (diff < 0) {
+            accounting.updateEntry(accounts.gain, accounts.asset, uint128(-diff));
+        }
+    }
```
lemunozm: This is the `updateValueOfHoldings` using the `IItemManager` interfaces

lemunozm: I do not really sure if `diff > 0` means loss or gain. `diff > 0` means that now the valuation has been greater than before. Does it mean that they have lost money because they increased when the value was lower?

lemunozm: @mustermeiszer

src/PoolManager.sol
```diff
@@ -0,0 +1,218 @@
...
+        IGateway gateway_
+    ) Auth(owner) PoolLocker(multicall) {
+        poolRegistry = poolRegistry_;
+        assetManager = assetManager_;
+        accounting = accounting_;
+        holdings = holdings_;
+        gateway = gateway_;
+    }
+
+    function createPool(IERC20Metadata currency, IShareClassManager shareClassManager)
```
mustermeiszer: NIT

I think it is always great if knowledge of following arguments does not depend on previous calls. 

What do you think about providing the `PoolId` as an argument and the registry errors out if already taken?

src/PoolManager.sol
```diff
@@ -0,0 +1,218 @@
...
+    function allowPool(ChainId chainId) external poolUnlocked {
+        // TODO: store somewhere the allowed info?
+        gateway.sendAllowPool(chainId, unlockedPoolId());
+    }
+
+    function allowShareClass(ChainId chainId, ShareClassId scId) external poolUnlocked {
+        // TODO: store somewhere the allowed info?
+        gateway.sendAllowShareClass(chainId, unlockedPoolId(), scId);
+    }
+
```
mustermeiszer, line 166: Missing

* `function allowHoldings(ShareClassId scId, AssetId) external poolUnlocked{..}` -> allows holding asset
* `function disallowHoldings(ShareClassId scId, AssetId) external poolUnlocked{..}` -> disallows holding assets
* `function allowAsset(ShareClasId, scId, AssetId assetId) external poolUnlocked {..}` -> allows subscriptions/redemptions with asset, requires allowance of holdings first
* `function disallowAsset(ShareClasId, scId, AssetId assetId) external poolUnlocked {..}` -> disallows subscriptions/redemptions with asset

src/PoolManager.sol
```diff
@@ -0,0 +1,218 @@
...
+        PoolId poolId = unlockedPoolId();
+
+        IShareClassManager scm = poolRegistry.shareClassManagers(poolId);
+        ItemId itemId = holdings.itemIdFromAsset(poolId, scId, assetId);
+        IERC7726 valuation = holdings.valuation(poolId, itemId);
+
+        scm.approveDeposit(poolId, scId, assetId, approvalRatio, valuation);
+    }
+
+    function issueShares(ShareClassId scId, uint128 nav) external poolUnlocked {
```
mustermeiszer: @wischli and I spoke about that. Needs to be with `AssetId` and instead of `nav` we use `pricePerShare`/`NavPerShare`

src/PoolManager.sol
```diff
@@ -0,0 +1,218 @@
...
+        PoolId poolId = unlockedPoolId();
+
+        IShareClassManager scm = poolRegistry.shareClassManagers(poolId);
+
+        scm.issueShares(poolId, scId, nav);
+    }
+
+    function revokeShares(ShareClassId scId, uint128 nav) external poolUnlocked {
```
mustermeiszer: @wischli and I spoke about that. Needs to be with AssetId and instead of nav we use pricePerShare/NavPerShare

src/PoolManager.sol
```diff
@@ -0,0 +1,218 @@
...
+        PoolId poolId = unlockedPoolId();
+
+        IShareClassManager scm = poolRegistry.shareClassManagers(poolId);
+
+        scm.revokeShares(poolId, scId, nav);
+    }
+
+    function increaseItem(IItemManager im, ItemId itemId, uint128 amount) external poolUnlocked {
```
mustermeiszer: Needs valuation, see holdings pr

src/PoolManager.sol
```diff
@@ -0,0 +1,218 @@
...
+        PoolId poolId = unlockedPoolId();
+
+        IAccountingItemManager.Accounts memory accounts = im.itemAccounts(poolId, itemId);
+        uint128 valueChange = im.increase(poolId, itemId, amount);
+
+        accounting.updateEntry(accounts.equity, accounts.asset, valueChange);
+    }
+
+    function decreaseItem(IItemManager im, ItemId itemId, uint128 amount) external poolUnlocked {
```
mustermeiszer: Needs valuation, see holdings pr

src/PoolManager.sol
mustermeiszer: >  means that now the valuation has been greater than before. 

if valuation greater than before, than we have a gain. Otherwise, we have a loss. At least for items representing asset side of the balance sheet. That is fine for now.

src/PoolManager.sol
```diff
@@ -0,0 +1,218 @@
...
+
+        scm.issueShares(poolId, scId, nav);
+    }
+
+    function revokeShares(ShareClassId scId, uint128 nav) external poolUnlocked {
+        PoolId poolId = unlockedPoolId();
+
+        IShareClassManager scm = poolRegistry.shareClassManagers(poolId);
+
+        scm.revokeShares(poolId, scId, nav);
```
mustermeiszer: Missing

* move revoked amounts of `AssetId` to `pendingShareClassEscrow`

src/PoolManager.sol
```diff
@@ -0,0 +1,218 @@
...
+
+        gateway.sendUnlockTokens(chainId, assetId, receiver, poolAmount);
+    }
+
+    function requestDeposit(PoolId poolId, ShareClassId scId, AssetId assetId, address investor, uint128 amount)
+        external
+        onlyGateway
+    {
+        address pendingShareClassEscrow = _escrow(poolId, scId, Escrow.PENDING_SHARE_CLASS);
+        assetManager.mint(pendingShareClassEscrow, assetId, amount);
```
mustermeiszer: I think that should be a transfer. All mint and burn should happen in the `AssetManager`, done by the gateway IMO. Let me know what you think @ilinzweilin @wischli @hieronx . I do not know Georgies handle ^^

mustermeiszer: But maybe easier here for upgradability.

src/PoolManager.sol
```diff
@@ -0,0 +1,218 @@
...
+
+    function requestRedemption(PoolId poolId, ShareClassId scId, AssetId assetId, address investor, uint128 amount)
+        external
+        onlyGateway
+    {
+        IShareClassManager scm = poolRegistry.shareClassManagers(poolId);
+        scm.requestRedemption(poolId, scId, assetId, investor, amount);
+    }
+
+    function lockTokens(AssetId assetId, address recvAddr, uint128 amount) external {
```
mustermeiszer: Remove. Locking tokens ALWAYS happens on the CV side. CP only handles accounting representation of those.

src/PoolManager.sol
```diff
@@ -0,0 +1,218 @@
...
+        (uint128 shares, uint128 tokens) = scm.claimShares(poolId, scId, assetId, investor);
+        gateway.sendFulfilledDepositRequest(poolId, scId, assetId, investor, shares, tokens);
+    }
+
+    function claimTokens(PoolId poolId, ShareClassId scId, AssetId assetId, address investor) external {
+        IShareClassManager scm = poolRegistry.shareClassManagers(poolId);
+
+        (uint128 shares, uint128 tokens) = scm.claimShares(poolId, scId, assetId, investor);
+
+        assetManager.burn(_escrow(poolId, scId, Escrow.SHARE_CLASS), assetId, tokens);
```
mustermeiszer: should be pending escrow. Otherwise, we need to update the bookeeping.

mustermeiszer (commented):
Love it! I feel the current approach is really well suited wrt. to how we can compose stuff. At least my feeling. 

One thing I am not sure about is the following. 

* all calls are really composable ONLY `moveOut` does update the bookkeeping and also triggers a transfer out. 
* in contrast: `approveSubscription` and `revokeShares` do not automatically update the relevant holdings of the asset and of the accounting side

IMO I would rather split up `moveOut` too. What do you think @lemunozm ? I think I have been going back and forth with this. So sorry for that.

src/PoolManager.sol
lemunozm: Ok, then I wrote the other way around! Because I'm updating the loss account when `diff > 0` 😅

src/PoolManager.sol
lemunozm: Mmm.... 🤔 

I think `createPool()` is special because it's not a method you need to call inside `unlock()`, so it's not batched in the same way as others. Also, because it's incremental, you know what `poolId` you will need in future calls. In case you want to set up a pool in some consecuent calls. So you can "avoid" reading the returned value.

I would leave this method as it is for now, but keep your comment in mind if we see any issues.

src/PoolManager.sol
lemunozm: Sure! thanks for the reminder 🙌🏻

src/PoolManager.sol
lemunozm: Makes sense! I need to update the investment part with the latest changes from the ShareClassManager PR. Like that idea!

src/PoolManager.sol
lemunozm: Open a thread in the other PR regarding my unknowledge of this

src/PoolManager.sol
lemunozm: To explain the whys of this:

As I envision it, `Gateway` only knows about `PoolManager` and `PoolManager` about the rest of the contacts, so the `Gateway` lets the `PoolManager` to do that call. Note that the call is marked as `onlyGateway`, which means that it does things related to what a gateway should do, not what a FM should do.

The idea is to simplify dependencies if we upgrade some low-level contract. Only PoolManager should be updated again (or use `file()`)

src/PoolManager.sol
lemunozm: In the on-chain settlement flow diagram, CP is minting those, because they reflect the CV locked amount. Is it wrong there?

src/PoolManager.sol
mustermeiszer: Is it still possible to batch `create` with other methods of the pool manager? That would be perfect for UX. 

E.g. user creating a pool and setting up items at the same run would be great.

mustermeiszer: Only reason, why I said we can have `poolId` known upfront.

src/PoolManager.sol
mustermeiszer: For referecne. That is for `DebitNormal` accounts. 

e.g. 
* Bond valued at 90
* Bond next value at 100
* Diff 10 -> 10 more debit on the asset

mustermeiszer: @lemunozm 

just for me to get the full picture. Do you think, we could also track items that have a negative impact through the holdings implementation? 

E.g. if we wanna track liabilities that are represented as some kind of token

src/PoolManager.sol
mustermeiszer: Okay. Yes, fine for me.

src/PoolManager.sol
mustermeiszer: Then I would name it `lockedTokens`, because they have been locked at the CV side.

Sorry for the confusion. And it mus be `onlyGateway`.

src/PoolManager.sol
lemunozm: I think we could, if we use `int128` for values instead of `uint128`. If the value after evaluating is negative it means it has a negative impact later when we sum up all valuations to get the NAV. Does it make sense?

src/PoolManager.sol
lemunozm: I think so. 

You can do:
- multidelegateCall
  - createPool
  - execute(PoolA) // You can know this because you know what the next poolId will be in the poolRegistry storage.
    - FM methodA
    - FM methodB
    
I think by now, we can go with this solution until we find a general solution for passing output -> inputs everywhere. I think the chance of having an error guessing the next poolId is pretty low because it's not super common to create new pools.

src/PoolManager.sol
lemunozm: I agree with both things. 👍🏻 The name change changes everything!

lemunozm:
> all calls are really composable ONLY moveOut does update the bookkeeping and also triggers a transfer out.

Sorry @mustermeiszer , I didn't answer to this topic. First, to understand well your concern:

What do you mean by "update the bookkeping"? I thought it meant calling to `updateEntry()`, but that is not happening already inside `moveOut()`. `moveOut() `right now only does: `mint()` and `sendUnlockTokens()` but do not modify the bookkeeping in `Accounting`

EDIT: solved here: https://kflabs.slack.com/archives/C07PG2EUR9C/p1734612004966299?thread_ts=1734353300.996649&cid=C07PG2EUR9C

src/PoolManager.sol
```diff
@@ -0,0 +1,218 @@
...
+    function createPool(IERC20Metadata currency, IShareClassManager shareClassManager)
+        external
+        returns (PoolId poolId)
+    {
+        // TODO: add fees
+        return poolRegistry.registerPool(msg.sender, currency, shareClassManager);
+    }
+
+    function allowPool(ChainId chainId) external poolUnlocked {
+        // TODO: store somewhere the allowed info?
```
wischli: I think we need to. Otherwise, we cannot disallow a previously allowed pool on some CV. Moreover, without storing this information, it appears to me that the Gateway would be the source of truth whether an incoming CV message for a pool is valid. Or am I missing something?

src/PoolManager.sol
wischli: FYI: It's still TBD if `issueShares` will be coupled with `depositAssetId`.

src/interfaces/IPoolManager.sol
```diff
@@ -0,0 +1,61 @@
...
+    function requestDeposit(PoolId poolId, ShareClassId scId, AssetId assetId, address investor, uint128 amount)
+        external;
+
+    function requestRedemption(PoolId poolId, ShareClassId scId, AssetId assetId, address investor, uint128 amount)
+        external;
+
+    function lockTokens(AssetId assetId, address recvAddr, uint128 amount) external;
+}
+
+interface IPoolManager is IPoolUnlockedMethods, IFromGatewayMethods {
```
wischli: I really like how you compose this!

src/PoolManager.sol
lemunozm: Yes, I also think it should be there. It's per chainId, which is what the Gateway knows about.

src/interfaces/IPoolManager.sol
lemunozm: Thanks! I'm open too to change their names because I am still not super happy with them

src/PoolManager.sol
lemunozm: Discussion in the holdings PR

src/PoolManager.sol
mustermeiszer: @lemunozm I rethought about this. For the migration, I would say, we MUST allow the `PoolId` to be choosable externally. So that Anemoy can migrate to the same `PoolId`. EVen

mustermeiszer: Of course affects the registry too.

lemunozm: Ok, IIUC, the current solution supports what you say. You can choose any PoolId for `execute()` externally

mustermeiszer: Not sure I follow. We need to create the pool with a specific pool id

src/PoolManager.sol
```diff
@@ -0,0 +1,324 @@
...
+        require(msg.sender == address(gateway), NotGateway());
+        _;
+    }
+
+    constructor(
+        IMulticall multicall,
+        IPoolRegistry poolRegistry_,
+        IAssetManager assetManager_,
+        IAccounting accounting_,
+        IHoldings holdings_,
```
mustermeiszer, line 56: Why do we need holdings here?

mustermeiszer (commented):
Looks great alrady! Just am missing the IERC7726 implementation for manager itself or did I overlook that? And the associated `setCurrentValuation` method for the transient storage.

lemunozm:
> Just am missing the IERC7726 implementation for manager itself

Could you refer to this: https://github.com/centrifuge/pools-internal/pull/53

src/PoolManager.sol
lemunozm: I understood you wrongly here. So we no longer generate a `PoolId`, instead we set from the outside 👍🏻

src/PoolManager.sol
lemunozm: There are some methods that requires to access directly to the Holdings contract, without let the FM to choose an `IItemManager` as other actions (as icrease/decrease/update).

i.e. `approveDeposit` needs to act directly over the holdings.

lemunozm: Anyways, I wanted to refactor all this due your last requirements about not having now an `IITemManager`. Probably, all methods from now on should access the holdings directly instead of passing the holding contract each time by parameter as an item manager.

src/PoolManager.sol
```diff
@@ -0,0 +1,368 @@
...
+            scm.cancelRedeemRequest(poolId, scId.toBytes(), investor.addr(), payoutAssetId.addr());
+
+        gateway.sendFulfilledRedeemRequest(
+            poolId, scId, payoutAssetId, investor, cancelledAssetAmount, cancelledAssetAmount
+        );
+    }
+
+    function handleLockedTokens(ShareClassId scId, AssetId assetId, uint128 amount) external onlyGateway {
+        assetManager.mint(_escrow(unlockedPoolId(), scId, Escrow.SHARE_CLASS), assetId, amount);
+    }
```
lemunozm: @mustermeiszer should we automatically `increaseHoldings` here? Basically to avoid a time gap between receiving the locked tokens and register them in the Holdings by the FM.

src/PoolManager.sol
```diff
@@ -0,0 +1,368 @@
...
+        (uint128 shares, uint128 tokens) = scm.claimDeposit(poolId, scId.toBytes(), assetId.addr(), investor.addr());
+        gateway.sendFulfilledDepositRequest(poolId, scId, assetId, investor, shares, tokens);
+    }
+
+    function claimRedeem(PoolId poolId, ShareClassId scId, AssetId assetId, GlobalAddress investor) external {
+        IShareClassManager scm = poolRegistry.shareClassManager(poolId);
+
+        (uint128 shares, uint128 tokens) = scm.claimRedeem(poolId, scId.toBytes(), assetId.addr(), investor.addr());
+
+        assetManager.burn(_escrow(poolId, scId, Escrow.PENDING_SHARE_CLASS), assetId, tokens);
```
wischli: While going over the flows, I realized this should be the non pending escrow. The flow also included the pending escrow but since we liquiditate deployed capital here, it should be the normal escrow?

lemunozm: Mmmm... when you revoke shares, you transfer from the normal escrow to the pending escrow. So later, when you claim the redeem, it should burn from the pending one.

Am I wrong with this reasoning?

lemunozm: Basically the asset flow is:
- requestDeposit -> mint pending
- approveDeposits -> pending -> normal
- revokeShares -> normal -> pending
- claimRedeem -> burn pending

wischli: Thanks for summarizing that! It makes sense to me now that I see it in total!

src/PoolManager.sol
mustermeiszer: The CP side will have no knowledge about pool nor share class id as part of this message. 

We will deprecate `LockTokens` and `UnlockTokens` once CC is shutdown in favor of `IncreaseHoldings` and `DecreaseHoldings`.

Once we have this, adapting the holdings on the CP side can only be done through the CAL side (naming is wip ^^)

src/PoolManager.sol
```diff
@@ -0,0 +1,388 @@
...
+
+    function setPoolAdmin(address newAdmin, bool canManage) external {
+        poolRegistry.updateAdmin(unlockedPoolId(), newAdmin, canManage);
+    }
+
+    function allowInvestorAsset(AssetId assetId, bool allow) external {
+        PoolId poolId = unlockedPoolId();
+
+        require(assetManager.isRegistered(assetId), IAssetManager.AssetNotFound());
+        require(holdings.isAssetAllowed(poolId, assetId), IHoldings.AssetNotAllowed());
```
lemunozm: @hieronx interested in your opinion.

Here, I'm reusing a couple of errors from interfaces that this contract uses but does not implement. Is it something considered ok, or could it be later tricky to find where that error was originated? i.e. somebody looking inside the AssetManager when it was originated here.

src/PoolManager.sol
```diff
@@ -0,0 +1,384 @@
...
+
+    function approveDeposits(ShareClassId scId, AssetId paymentAssetId, D18 approvalRatio) external poolUnlocked {
+        PoolId poolId = unlockedPoolId();
+
+        IShareClassManager scm = poolRegistry.shareClassManager(poolId);
+        IERC7726 valuation = holdings.valuation(poolId, scId, paymentAssetId);
+
+        (, uint128 approvedAssetAmount) = scm.approveDeposits(
+            poolId, scId.toBytes(), approvalRatio, paymentAssetId.addr(), IERC7726Ext(address(valuation))
+        );
```
lemunozm: @wischli here I'm only using the `approvedAssetAmount` (note that the pool amount is first in the tuple and asset is the second)

src/PoolManager.sol
```diff
@@ -0,0 +1,384 @@
...
+
+        scm.issueShares(poolId, scId.toBytes(), depositAssetId.addr(), navPerShare);
+    }
+
+    function revokeShares(ShareClassId scId, AssetId payoutAssetId, D18 navPerShare) external poolUnlocked {
+        PoolId poolId = unlockedPoolId();
+
+        IShareClassManager scm = poolRegistry.shareClassManager(poolId);
+
+        (uint128 payoutAssetAmount,) = scm.revokeShares(poolId, scId.toBytes(), payoutAssetId.addr(), navPerShare);
```
lemunozm: And here, I'm using again only the asset amount, but in this case it appears as first parameter and pool amount is the second parameter.

I would propose, if you agree. Just returning the asset amount in both cases or if not, having the same order in the tuple.

lemunozm (commented):
@wischli This is a comment for your ShareClass PR but done here for showing the usage.

src/PoolManager.sol
wischli: According to the [flow diagram](https://www.notion.so/CentrifugePools-Architecture-1372eac24e1780809ffbd92f3366eb51?pvs=4#c837c05947c14bd89d5a19fb69b10b65), we also need to update holdings with the `approvedPoolAmount` which we are not doing here. Is that part of the flow diagram deprecated at this point?

src/PoolManager.sol
wischli: Same as [above](https://github.com/centrifuge/pools-internal/pull/36/files#r1942443690): According to the [Revoke Shares flow](https://www.notion.so/CentrifugePools-Architecture-1372eac24e1780809ffbd92f3366eb51?pvs=4#df01804be59e4aaf8f6b195169a8d598), the pool amount is necessary in order to update holdings.

lemunozm: You're right @wischli! That was an excellent catch I've entirely missed!

NIT. Could we then have both methods return the amounts in the same order?

wischli: > NIT. Could we then have both methods return the amounts in the same order?

Yes of course, sorry about that. I will go with `(assetAmount, poolAmount)`.

test/unit/Accounting.t.sol
```diff
@@ -19,13 +19,13 @@ PoolId constant POOL_A = PoolId.wrap(1);
...
-    AccountId immutable CASH_ACCOUNT = accountId(1, uint8(AccountType.ASSET));
-    AccountId immutable BOND1_INVESTMENT_ACCOUNT = accountId(101, uint8(AccountType.ASSET));
-    AccountId immutable FEES_EXPENSE_ACCOUNT = accountId(401, uint8(AccountType.ASSET));
-    AccountId immutable FEES_PAYABLE_ACCOUNT = accountId(201, uint8(AccountType.LOSS));
-    AccountId immutable EQUITY_ACCOUNT = accountId(301, uint8(AccountType.EQUITY));
-    AccountId immutable GAIN_ACCOUNT = accountId(302, uint8(AccountType.GAIN));
-    AccountId immutable NON_INITIALIZED_ACCOUNT = accountId(999, uint8(AccountType.ASSET));
+    AccountId immutable CASH_ACCOUNT = newAccountId(1, uint8(AccountType.ASSET));
+    AccountId immutable BOND1_INVESTMENT_ACCOUNT = newAccountId(101, uint8(AccountType.ASSET));
+    AccountId immutable FEES_EXPENSE_ACCOUNT = newAccountId(401, uint8(AccountType.ASSET));
```
peculiarity, line 24: Super nit: Can you order them by their Id, please?

src/interfaces/IPoolManager.sol
```diff
@@ -0,0 +1,213 @@
...
+    /// @notice Claim a redemption for an investor address located in the chain where the asset belongs
+    function claimRedeem(PoolId poolId, ShareClassId scId, AssetId assetId, bytes32 investor) external;
+
+    /// @notice Compute the escrow address used for a share class
+    /// @return The escrow address
+    function escrow(PoolId poolId, ShareClassId scId, EscrowId escrow_) external returns (address);
+}
+
+/// @notice Interface for methods called by the gateway
+interface IPoolManagerHandler {
```
peculiarity, line 173: I think all these methods should be just named `requestDeposit` , `requestRedeem` ....
You can have 1 `handle` message that handle the incoming encoded message, decode it and just calls the appropriate method.

src/PoolManager.sol
```diff
@@ -0,0 +1,442 @@
...
+        external
+        onlyGateway
+    {
+        IShareClassManager scm = poolRegistry.shareClassManager(poolId);
+        (uint128 cancelledAssetAmount) = scm.cancelDepositRequest(poolId, scId, investor, depositAssetId);
+
+        address pendingShareClassEscrow = escrow(poolId, scId, EscrowId.PENDING_SHARE_CLASS);
+        assetManager.burn(pendingShareClassEscrow, depositAssetId.raw(), cancelledAssetAmount);
+
+        gateway.sendFulfilledCancelDepositRequest(poolId, scId, depositAssetId, investor, cancelledAssetAmount);
```
peculiarity, line 395: This comment applies for all the `send*` method. So each param is always padded so that it take 32 bytes. Function params are 1 word length. That costs gas.  You can simply create the message here and call `gateway.send(message)`. The responsibility of t he `gateway` will be to send it to different channels, pay fees or smth like that. It won't be responsible for creating messages. Also the `Gateway`  implementation will be smaller.

lemunozm: But `PoolManager` implementation will be bigger. I explain my reason above. I think gas cost here is depreciable, taking into account that a message is going through routers/adapters & axelar. 

Serializing/deserializing messages is a duty more related to the `Gateway` than the `PoolManager`.

src/interfaces/IPoolManager.sol
lemunozm: As I explained here: https://github.com/centrifuge/pools-internal/pull/81#discussion_r1944253296 I would prefer not to do that and instead leave the `PoolManager` free from any serialization/deserialization stuff. That's something the gateway should do.

I know that the interface becomes wide, because instead of send/handle, we have sendXYZ and handleXYZ, but I think that helps to reason what's happening in each stage. Also, for testing, it helps for mocking these methods

wischli (commented):
I went through all the flows and could not spot any logical error! Incredibly surgical work you have done here @lemunozm 🚀 🚀 🚀  I bow for you!

There are a few super minor nits that I will postpone until next week when others have performed a full logical review as well to reduce noise.

src/PoolManager.sol
```diff
@@ -0,0 +1,442 @@
...
+
+    /// @inheritdoc IPoolManagerAdminMethods
+    function addDebit(AccountId account, uint128 amount) external poolUnlocked {
+        accounting.addDebit(account, amount);
+    }
+
+    /// @inheritdoc IPoolManagerAdminMethods
+    function addCredit(AccountId account, uint128 amount) external poolUnlocked {
+        accounting.addCredit(account, amount);
+    }
```
lemunozm: I think we can get rid of `updateEntry()` and just let the FM compose `addDebit()` and `addCredit()` in the multicall, avoiding bloating the `PoolManager` with a code that it does not really need to be on-chain, WDYT?

mustermeiszer: Agree!

src/PoolManager.sol
```diff
@@ -0,0 +1,435 @@
...
+        accounts[2] = newAccountId(rawAccountId, uint8(AccountType.LOSS));
+        accounts[3] = newAccountId(rawAccountId, uint8(AccountType.GAIN));
+
+        createAccount(accounts[0], true);
+        createAccount(accounts[1], false);
+        createAccount(accounts[2], false);
+        createAccount(accounts[3], false);
+
+        holdings.create(unlockedPoolId(), scId, assetId, valuation, accounts);
+    }
```
lemunozm: I modified a bit this method in [f7adc6b](https://github.com/centrifuge/pools-internal/pull/36/commits/f7adc6b4d867b584f1757d66c464f9ccea18762a)

It's a bit painful to do this manually each time you need to create a new holding, and it is also error-prone to set up all calls in the execute.

The point is that you surely want/need to create at least those accounts each time you create a new holding, so everything can be abstracted to the FM.

NOTE. The FM still have all the control if they later want to modify some accounts or add new ones using `setHoldingAccountId`

lemunozm: @mustermeiszer could you double check that:
- asset is debit normal
- equity is credit normal
- loss is credit normal
- gain is credit normal

src/libraries/CastLib.sol
```diff
@@ -0,0 +1,44 @@
+// SPDX-License-Identifier: BUSL-1.1
+pragma solidity 0.8.28;
+
+/// @title  CastLib
+library CastLib {
```
peculiarity, line 5: Tests missing :)

src/PoolManager.sol
```diff
@@ -0,0 +1,435 @@
...
+        emit File(what, data);
+    }
+
+    //----------------------------------------------------------------------------------------------
+    // Permisionless methods
+    //----------------------------------------------------------------------------------------------
+
+    /// @inheritdoc IPoolManager
+    function createPool(AssetId currency, IShareClassManager shareClassManager) external returns (PoolId poolId) {
+        // TODO: add fees
```
wischli, line 89: FYI: I have created [an issue](https://github.com/centrifuge/pools-internal/issues/83) to handle this later. Definitely OOS.

src/PoolManager.sol
```diff
@@ -0,0 +1,435 @@
...
+        require(scm.exists(poolId, scId), IShareClassManager.ShareClassNotFound());
+
+        gateway.sendNotifyShareClass(
+            chainId,
+            poolId,
+            scId,
+            string("TODO"),
+            string("TODO"),
+            assetManager.decimals(poolRegistry.currency(poolId).raw()),
+            bytes32("TODO")
```
wischli, line 135: FYI: I have created [an issue](https://github.com/centrifuge/pools-internal/issues/84) to deal with this in a later PR.

src/PoolManager.sol
```diff
@@ -0,0 +1,435 @@
...
+        accounting.addDebit(holdings.accountId(poolId, scId, assetId, uint8(AccountType.EQUITY)), valueChange);
+    }
+
+    /// @inheritdoc IPoolManagerAdminMethods
+    function updateHolding(ShareClassId scId, AssetId assetId) external poolUnlocked {
+        PoolId poolId = unlockedPoolId();
+
+        int128 diff = holdings.update(poolId, scId, assetId);
+
+        if (diff > 0) {
```
wischli, line 296: @mustermeiszer Just want to make sure we can skip updating with `diff == 0`. I.e., just wondering if events with timestamps/blocks will still be beneficial from a product perspective, even if no diff occurred.

src/SingleShareClass.sol
```diff
@@ -517,6 +517,11 @@ contract SingleShareClass is Auth, ISingleShareClass {
         );
     }
 
+    /// @inheritdoc IShareClassManager
+    function exists(PoolId poolId, ShareClassId shareClassId_) public view returns (bool) {
```
wischli, line 519: Nit: We could use this check for the `ShareClassNotFound` error. I will apply this as part of some later ShareClassManager refactor.

src/interfaces/IAssetManager.sol
```diff
@@ -6,7 +6,7 @@ import {AssetId} from "src/types/AssetId.sol";
 import {IERC6909Fungible} from "src/interfaces/ERC6909/IERC6909Fungible.sol";
 import {IERC6909MetadataExt} from "src/interfaces/ERC6909/IERC6909MetadataExt.sol";
 
-// TODO
+/// @notice Interface for register and handling assets
```
wischli: Nit:
```suggestion
/// @notice Interface for registering and handling assets
```

src/interfaces/IPoolManager.sol
```diff
@@ -0,0 +1,207 @@
...
+    /// @notice Credit normal account for tracking losses
+    LOSS,
+    /// @notice Credit normal account for tracking profits
+    GAIN
+}
+
+/// @notice Interface for methods that requires the pool to be unlocked
+/// They do not require a poolId parameter, all acts over the unlocked pool
+interface IPoolManagerAdminMethods {
+    /// @notice Dispatched whem a holding asset is disallowed but the asset is still allowed for investor usage.
```
wischli: Nit:
```suggestion
    /// @notice Dispatched when a holding asset is disallowed but the asset is still allowed for investor usage.
```

src/interfaces/IPoolManager.sol
```diff
@@ -0,0 +1,207 @@
...
+    /// @notice Credit normal account for tracking equities
+    EQUITY,
+    /// @notice Credit normal account for tracking losses
+    LOSS,
+    /// @notice Credit normal account for tracking profits
+    GAIN
+}
+
+/// @notice Interface for methods that requires the pool to be unlocked
+/// They do not require a poolId parameter, all acts over the unlocked pool
```
wischli: Nit
```suggestion
/// They do not require a poolId parameter, all act over the unlocked pool
```

src/libraries/CastLib.sol
lemunozm: I copied this file from CV, where it is missing too.

wischli (approved):
As mentioned, a few very minor nits that you can ignore, if you want to. Apart from that, I have created tickets for the two leftover TODOs which are surely OOS of this great PR.

src/PoolManager.sol
```diff
@@ -0,0 +1,435 @@
...
+
+    /// @inheritdoc IPoolManagerAdminMethods
+    function notifyAllowedAsset(ShareClassId scId, AssetId assetId) external poolUnlocked {
+        PoolId poolId = unlockedPoolId();
+
+        gateway.sendNotifyAllowedAsset(poolId, scId, assetId, poolRegistry.isInvestorAssetAllowed(poolId, assetId));
+    }
+
+    /// @inheritdoc IPoolManagerAdminMethods
+    function setPoolMetadata(bytes calldata metadata) external {
```
mustermeiszer: CR

`poolUnlocked`

src/PoolManager.sol
```diff
@@ -0,0 +1,435 @@
...
+        poolRegistry.setMetadata(unlockedPoolId(), metadata);
+    }
+
+    /// @inheritdoc IPoolManagerAdminMethods
+    function allowPoolAdmin(address account, bool allow) external {
```
mustermeiszer: CR

`poolUnlocked`

src/PoolManager.sol
```diff
@@ -0,0 +1,435 @@
...
+        poolRegistry.updateAdmin(unlockedPoolId(), account, allow);
+    }
+
+    /// @inheritdoc IPoolManagerAdminMethods
+    function allowHoldingAsset(AssetId assetId, bool allow) external {
```
mustermeiszer: CR

`poolUnlocked`

src/PoolManager.sol
```diff
@@ -0,0 +1,435 @@
...
+        PoolId poolId = unlockedPoolId();
+
+        if (!allow) {
+            require(!poolRegistry.isInvestorAssetAllowed(poolId, assetId), InvestorAssetStillAllowed());
+        }
+
+        holdings.allowAsset(poolId, assetId, allow);
```
mustermeiszer, line 164: QUESTION

We allow holdings without a registered asset, right? I think this is fine, just wanted to ask whether this is intended.

src/PoolManager.sol
```diff
@@ -0,0 +1,435 @@
...
+    }
+
+    /// @inheritdoc IPoolManagerAdminMethods
+    function allowInvestorAsset(AssetId assetId, bool allow) external {
```
mustermeiszer: CR

poolUnlocked

src/PoolManager.sol
```diff
@@ -0,0 +1,435 @@
...
+        PoolId poolId = unlockedPoolId();
+
+        require(assetManager.isRegistered(assetId), IAssetManager.AssetNotFound());
+        require(holdings.isAssetAllowed(poolId, assetId), IHoldings.AssetNotAllowed());
+
+        poolRegistry.allowInvestorAsset(poolId, assetId, allow);
+    }
+
+    /// @inheritdoc IPoolManagerAdminMethods
+    function addShareClass(bytes calldata data) external returns (ShareClassId) {
```
mustermeiszer: CR

poolUnlocked

src/PoolManager.sol
```diff
@@ -0,0 +1,435 @@
...
+
+        increaseHolding(scId, paymentAssetId, valuation, approvedAssetAmount);
+    }
+
+    /// @inheritdoc IPoolManagerAdminMethods
+    function approveRedeems(ShareClassId scId, AssetId payoutAssetId, D18 approvalRatio) external poolUnlocked {
+        PoolId poolId = unlockedPoolId();
+
+        IShareClassManager scm = poolRegistry.shareClassManager(poolId);
+        IERC7726 valuation = holdings.valuation(poolId, scId, payoutAssetId); // TODO: remove
```
mustermeiszer: TODO

src/PoolManager.sol
mustermeiszer: Good point. not sure. depends on indexer. would update.

src/PoolManager.sol
mustermeiszer: QUESTION

IIUC one account now is related to all `AccountTypes`. If yes, that is definitely an issue. If those are 4 different accounts, all good. 

Usually a super simple accounting setup would be to have for ALL holdings 
* 1 ASSET account
* 1 EQUITY account
* 1 GAIN account
* 1 LOSS account

but those would be all separate accounts.

src/PoolManager.sol
lemunozm: > We allow holdings without a registered asset, right?

Yes, I think we decided that. But it's true that the asset comes from another chain, and you should wait for that message to know the `AssetId` generated by the CV side.

Basically, without registerAsset message you do not know the AssetId value to allow holdings.

No strong opinion if you want to add the check

src/PoolManager.sol
lemunozm: I skiped them! That's correct!

src/PoolManager.sol
lemunozm: Yes, this is related to this [slack thread](https://kflabs.slack.com/archives/C07PG2EUR9C/p1738840178173459?thread_ts=1738258418.630189&cid=C07PG2EUR9C) where I asked to remove the valuation input from either `approveRedeems` or `revokeShares`.

@wischli, if you agree, I can just remove it from the `ShareClassManager::approveDeposit()` interface until we decide the exact required changes of the CAL messages addition

src/PoolManager.sol
lemunozm: > IIUC one account now is related to all AccountTypes. If yes, that is definitely an issue. If those are 4 different accounts, all good.

They are 4 different accounts, but using the same prefix "rawAccoutId" (maybe I should call it prefix better).

But for sure they are treated as different accounts

src/interfaces/IPoolManager.sol
lemunozm: Thanks for always double-checking docs!

src/PoolManager.sol
lemunozm: Renamed it to prefix

lemunozm:
@wischli @mustermeiszer review changes here: [74ba550](https://github.com/centrifuge/pools-internal/pull/36/commits/74ba550ae61e4250033c0bea30c9b7cfc9101ed9)

src/PoolManager.sol
wischli: Yes, we can remove that. I can open a quick PR after lunch.

lemunozm: I did in this PR! No worries

src/PoolManager.sol
hieronx: I think it's fine. From testing perspective it's anyways not an issue. On Etherscan, `IHoldings` will just also be pulled into the set of verified source code files related to the contract. In terms of indexers, it can just also add this other interface.

src/PoolManager.sol
```diff
@@ -244,15 +243,15 @@ contract PoolManager is Auth, PoolLocker, IPoolManager, IPoolManagerHandler {
...
+    function createHolding(ShareClassId scId, AssetId assetId, IERC7726 valuation, uint24 prefix)
         external
         poolUnlocked
     {
         AccountId[] memory accounts = new AccountId[](4);
-        accounts[0] = newAccountId(rawAccountId, uint8(AccountType.ASSET));
-        accounts[1] = newAccountId(rawAccountId, uint8(AccountType.EQUITY));
-        accounts[2] = newAccountId(rawAccountId, uint8(AccountType.LOSS));
-        accounts[3] = newAccountId(rawAccountId, uint8(AccountType.GAIN));
+        accounts[0] = newAccountId(prefix, uint8(AccountType.ASSET));
```
mustermeiszer, line 251: @lemunozm get it now. 

Curious, about whether using a prefix is not confusing. This is due to our encoding for sure. But probably fine. 

All accounts MUST be created first, so there is always an event with the full accountId first, right?

lemunozm: Yes, the event is generated inside `createAccount()` with the whole id, so the event values are different

lemunozm: > Curious, about whether using a prefix is not confusing. This is due to our encoding for sure. But probably fine.

If we do not generate the account ID, we need some way to obtain some value for creating different accounts. 

I think, as a side effect of this. that having some part of the encoding to be the same could be useful for UX to filter/sort accounts related to the same holding. Just seeing the ID you know that they are related in some way regarding the holding usage. But of course this is not a mandatory relation

mustermeiszer: True! I think with the given account events it is all fine!

wischli (approved):
Once again, amazing work tying this all together so nicely! 👏 Good to go from my perspective!

peculiarity (approved):

PR #37, opened 2024-12-09
chore: Migrate hardcoded string errors on revert to custom errors for…

peculiarity:
… gas optimization


lemunozm (approved):
LGTM!! Thanks for taking care of this change 🙌🏻

wischli (approved):
Thanks so much for temporarily acting as Mr. Clean 🧹

PR #38, opened 2024-12-10
Make sure the caller can unlock in the locker tiself

peculiarity:


src/PoolLocker.sol
```diff
@@ -30,6 +30,7 @@ abstract contract PoolLocker is IPoolLocker {
         returns (bytes[] memory results)
     {
         require(PoolId.unwrap(unlocked) == 0, PoolAlreadyUnlocked());
+        require(_canUnlock(poolId), "User cannot unlock");
         _unlock(poolId);
```
lemunozm, line 34: I think this is more complex than just letting to perform the check in the `_unlock()` method. `canUnlock()` and `unlock()` are being called at the same moment to the same implementor.

What's the benefit of having `_canUnlock()`?

PR #39, opened 2024-12-10
(Draft: Proposal) Fiat currency registry

peculiarity:


src/types/Currency.sol
```diff
@@ -1,4 +1,11 @@
...
 
-type Currency is address;
+
+
+struct Currency {
+    address addr;
+    uint8 decimals;
+    string name;
+    string symbol;
+}
```
lemunozm, line 11: I think we need the currency to be an `address`, more specifically an `IERC20Metadata`, and from where to ask for the info. That way, we treat that address as an ID and we can move it easily and cheaply everywhere in the system without moving all the data, strings, etc.

peculiarity: And if you need to get the decimals you will be doing external calls so not sure if this is going to be cheaper than passing around the memory of the currency which contains everything

peculiarity: Actually this is the other reason why I decided to go with a struct. Every time you would like to get  name, symbol and decimals you will have to make external calls.

peculiarity: Probably I don't understand how this is going to be used exactly

lemunozm: Only in some cases, you need the data. In those cases you can call to get the data you want (I think a simple call is not very expensive). Also, take into account that this currency "address" needs to be used as an address parameter in an ERC7726

peculiarity: The way I understand you is that you want to deploy new cotnract which will have their addresses which will break the idea in 7726 about using ISO code for fiat currencies.

lemunozm: Why that?

Suppose you want to code the Usd real fiat. You can implement an specific fiat contract implementation of `IERC20Metadata`, right?

PR #35, opened 2024-11-28
feat: init `IShareClassShareManager` & impl `SingleShareClass`

wischli:
Closes #29

**Before reviewing the PR for the first time, I recommend reading next two sections.**

### Update 09.01.2025

Based on [this Slack conversation](https://kflabs.slack.com/archives/C07PG2EUR9C/p1736246262483589), investor permission checks have been removed entirely. We can assume they are performed on CV such that we can ignore them on CP. 

### From 2024

All of the following decisions are completely up to discussion. I wanted to highlight them outside the code so we are all aligned. The list is currently not exhaustive.

0. We use the concept of epochs solely as a result of a lightweight/efficient lazy/pull-based design in which we don't need to deal (i.e. read/write) with pending deposit amounts of separate users when approving and issuing/revoking but only when the user claims. I don't think we can get around this Solidity/EVM limitation because the number of pending user requests during approval/issuance/revocation could be endless. Therefore, the **purpose of epochs is to block users from mutating their pending amounts such that it suffices to update user storage during claiming but not before that**.
1. The pool global epoch id `epoch[poolId]` is counter a starting at 1 (initialized when setting a `shareClassId` to a pool via `addShareClass`). It is only incremented, whenever deposits or redemptions are approved for the first time inside the same `multiCall`, see `_advanceEpoch`. The `epoch[poolId]` represents the current epoch and marks the upper bound for more specific epoch pointers to the last {deposit, redeem} approval and shares {issuance, revocation}. I will provide a flowchart for epoch advancing later. **For more details, see [this comment ](https://github.com/centrifuge/pools-internal/pull/35#issuecomment-2554659507)referencing two deeper explanations. I recommend everyone to read those!**
2. Investor permission checks are done on CV side, CP assumes that any inbound message is valid, including the amounts. I.e. we should not receive messages for deposit requests with zero amounts.
3. An investor can update their pending {deposit, redeem} amount iff its their first request (`i.e. userOrder.lastUpdated == 0`) or this update does not mutate a pending amount which has not been claimed yet but can be claimed. This is the case iff the last update did not happen after the last epoch in which an approval happened for the same deposit asset id (i.e. `userOrder.lastUpdate > lastDepositApproval[poolId][depositAssetId]`). Reasoning why `>` instead of `>=`, see [this comment](https://github.com/centrifuge/pools-internal/pull/35#discussion_r1892472920).
4. Same for redemptions. Here, we track the last redeem approval and last revocation as counterparts of the deposit flow.
[...]

PR #40, opened 2024-12-10
Add ERC7726 interface

lemunozm:
- Add ERC7726 to be used in several PRs
- Remove ERC165 and use the file from `forge-std`


src/interfaces/IERC7726.sol
```diff
@@ -0,0 +1,12 @@
+// SPDX-License-Identifier: BUSL-1.1
+pragma solidity >=0.6.2;
+
+/// [ERC-7726](https://eips.ethereum.org/EIPS/eip-7726): Common Quote Oracle
+/// Interface for asset conversions.
+interface IERC7726 {
+    /// @notice Returns the value of `baseAmount` of `base` in `quote` terms.
+    /// @param base The asset the user provides the amount.
+    /// @param quote The asset that the user needs to know the value for.
+    /// @param baseAmount An amount of `base` in `base` terms.
```
wischli: Nit: Only if you do any other changes. Otherwise I will propose in #35

```suggestion
    /// @notice Returns the value of baseAmount of base in quote terms, e.g. 10 ETH (base) in USDC (quote). It's rounded
    /// towards 0 and reverts if it overflows.
    /// @param base The asset in which the baseAmount is denominated in
    /// @param quote The asset in which the user needs to value the baseAmount
    /// @param baseAmount An amount of base in quote terms
```

wischli (approved):
Awesome, thanks! I improved docs for IERC7726 on my branch and proposed as ultra nit. Happy to merge this way!

src/interfaces/IERC7726.sol
lemunozm: I would change line 11, which I think is not well defined (introduced by myself, I think)

lemunozm: Another thing introduced by me "It's rounded towards 0 and reverts if it overflows", I think it's unnecessary and it depends on the implementation. From the consumer of this API he just trusts in the retrieved amount.

lemunozm: Done!

lemunozm:
@hieronx @ilinzweilin @mustermeiszer can I have one approval from your side to merge this?

hieronx (approved):

PR #42, opened 2024-12-12
Holdings contract implementation

lemunozm:
## Architecture

The first implementation of an `IItemManager` where `ItemId` represents` (ShareClassId, AssetId)`


mustermeiszer:
@lemunozm can you explain in more detail what the overall flow will be with the additional managers and? What will `AccountingItemManager` do?

lemunozm:
`AccountingItemManager` contract is just a superclass of `Holdings`, so it is the "same" contract (at least by now) than `Holdings`. I split just because that logic will be shared among all `ItemManager`. 

I think everything will be more understandable when I finish with the `PoolManager`

src/interfaces/IAccountingItemManager.sol
```diff
@@ -0,0 +1,19 @@
...
+// used in an IAccounting contract.
+interface IAccountingItemManager {
+    struct Accounts {
+        AccountId assetAccount;
+        AccountId equityAccount;
+        AccountId lossAccount;
+        AccountId gainAccount;
+    }
+
+    function setItemAccounts(PoolId poolId, ItemId itemId, Accounts calldata itemAccounts) external;
```
wischli: Is it guaranteed that each ItemManager has at most one `Accounts` or could they have multiple?

src/Holdings.sol
```diff
@@ -0,0 +1,87 @@
...
+        ItemId itemId = ItemId.wrap(++lastItemId[poolId]);
+        itemIds[poolId][scId][assetId] = itemId;
+        items[poolId][itemId] = Item(assetId, valuation_, 0, 0);
+    }
+
+    function close(PoolId poolId, ItemId itemId, bytes calldata /*data*/ ) external auth {
+        delete items[poolId][itemId];
+    }
+
+    function increase(PoolId poolId, ItemId itemId, uint128 amount) external auth returns (uint128 deltaValue) {
```
wischli: Nit: You can name this `amountValue` to not have to initialize that later and potentially skip the `return ...` line.
```suggestion
    function increase(PoolId poolId, ItemId itemId, uint128 amount) external auth returns (uint128) {
```

wischli (commented):
Thanks for starting with this already!

src/interfaces/IAccountingItemManager.sol
lemunozm: AFAIK, each item can have one `Accounts`, but one `Accounts` implies 4 `AccountId`s: equity, asset, loss, gain accounts

lemunozm: Each item of each `ItemManager` should be identical in this regard if I'm not mistaken. cc @mustermeiszer

mustermeiszer: No, you can not generalize that for all `ItemManagers`. That was just the first draft for `Tokens`. It should be sufficient for most use-cases though. 

But if I understand the interface correctly, all that it is doing here, is to allow to set an arbitrary amount of accounts, right?

mustermeiszer: What is `ItemId` in these cases?

lemunozm: I understood that all implementations of `IItemManager` should have (at least) those 4 accounts: equity, asset, loss, gain. For example, an item from `Bonds`, should also have those because you can track in the `Accounting` that information regarding those accounts. i.e. adding a loss for a bond item.

Is my assumption false?

> But if I understand the interface correctly, all that it is doing here, is to allow to set an arbitrary amount of accounts, right?

`Accounts` is:
```solidity
struct Accounts {
    AccountId asset;
    AccountId equity;
    AccountId loss;
    AccountId gain;
}
```

So you can specify those accounts per item.

> What is ItemId in these cases?

For `Tokens` (aka `Holdings`), an `ItemId` can be mapped to `(ShareClassId, AssetId)`. I think this is the correct output from the thread regarding this (see `PoolManager` impl to check the usage)

src/Holdings.sol
lemunozm: Done, good recommendation!

lemunozm:
Ready to review (pending tests, events and possibly some errors). Let's focus by now in the logic itself

Happy to merge the AccountingElementManager with the Holdings implementation as long as I can have those 4 accounts in the IItemManager interface.

The usage of this accounts can be easier seen in the PoolManager PR: https://github.com/centrifuge/pools-internal/pull/36

src/AccountingItemManager.sol
```diff
@@ -0,0 +1,21 @@
+// SPDX-License-Identifier: BUSL-1.1
+pragma solidity 0.8.28;
+
+import {PoolId} from "src/types/PoolId.sol";
+import {ItemId, AccountId} from "src/types/Domain.sol";
+import {IAccountingItemManager} from "src/interfaces/IAccountingItemManager.sol";
+import {Auth} from "src/Auth.sol";
+
+contract AccountingItemManager is Auth, IAccountingItemManager {
```
mustermeiszer: abstract

src/Holdings.sol
```diff
@@ -0,0 +1,114 @@
...
+        (ShareClassId scId, AssetId assetId) = abi.decode(data, (ShareClassId, AssetId));
+
+        ItemId itemId = ItemId.wrap(++lastItemId[poolId]);
+        itemIds[poolId][scId][assetId] = itemId;
+        items[poolId][itemId] = Item(assetId, valuation_, 0, 0);
+    }
+
+    /// @inheritdoc IItemManager
+    function close(PoolId poolId, ItemId itemId, bytes calldata /*data*/ ) external auth {
+        delete items[poolId][itemId];
```
mustermeiszer: I think we should never allow that. Reusing ids seems always like a bad idea.

src/Holdings.sol
```diff
@@ -0,0 +1,114 @@
...
+    }
+
+    /// @inheritdoc IItemManager
+    function increase(PoolId poolId, ItemId itemId, uint128 amount) external auth returns (uint128 amountValue) {
```
mustermeiszer:
```suggestion
    function increase(PoolId poolId, ItemId itemId, uint128 amount, uint128 valuation) external auth returns (uint128 amountValue) {
```

We need to support increasing and decreasing with a value in order to note valuations correctly.

src/Holdings.sol
```diff
@@ -0,0 +1,114 @@
...
+        Item storage item = items[poolId][itemId];
+        address poolCurrency = address(poolRegistry.poolCurrencies(poolId));
+
+        amountValue = uint128(item.valuation.getQuote(amount, AssetId.unwrap(item.assetId), poolCurrency));
```
mustermeiszer: Remove valuation here and use from input.

src/Holdings.sol
```diff
@@ -0,0 +1,114 @@
...
+    }
+
+    /// @inheritdoc IItemManager
+    function updateValuation(PoolId poolId, ItemId itemId, IERC7726 valuation_) external auth {
+        items[poolId][itemId].valuation = valuation_;
+    }
+
+    function itemIdFromAsset(PoolId poolId, ShareClassId scId, AssetId assetId) external view returns (ItemId) {
+        return itemIds[poolId][scId][assetId];
+    }
```
mustermeiszer: Can we also have it the other way around?

src/Holdings.sol
```diff
@@ -0,0 +1,114 @@
...
+            ? uint256(currentAmountValue - item.assetAmountValue).toInt128()
+            : -uint256(item.assetAmountValue - currentAmountValue).toInt128();
+
+        item.assetAmountValue = currentAmountValue;
+    }
+
+    /// @inheritdoc IItemManager
+    function decreaseInterest(PoolId, /*poolId*/ ItemId, /*itemId*/ uint128 /*amount*/ ) external pure {
+        revert("unsupported");
+    }
```
mustermeiszer: we are missing the counterpart here, right? `increaseInterest`

src/Holdings.sol
```diff
@@ -0,0 +1,114 @@
...
+
+    mapping(PoolId => mapping(ItemId => Item)) public items;
+    mapping(PoolId => mapping(ShareClassId => mapping(AssetId => ItemId))) public itemIds;
+    mapping(PoolId => uint32) lastItemId;
+
+    IPoolRegistry immutable poolRegistry;
+
+    constructor(address deployer, IPoolRegistry poolRegistry_) AccountingItemManager(deployer) {
+        poolRegistry = poolRegistry_;
+        // TODO: should we initialize the accounts from AccountingItemManager here?
```
mustermeiszer: We have accounts per `item` not for all holdings of a pool or share class

src/interfaces/IAccountingItemManager.sol
mustermeiszer: So the UI would need to do that during creation after calling `create`, right?

src/interfaces/IItemManager.sol
```diff
@@ -0,0 +1,41 @@
...
+    /// @notice Increments the amount of an item and updates the value for that increment.
+    /// @return value The value the item has increment.
+    function increase(PoolId poolId, ItemId itemId, uint128 amount) external returns (uint128 value);
+
+    /// @notice Decrements the amount of an item and updates the value for that decrement.
+    /// @return value The value the item has decrement.
+    function decrease(PoolId poolId, ItemId itemId, uint128 amount) external returns (uint128 value);
+
+    /// @notice Decrements the amount of an item as interest.
+    function decreaseInterest(PoolId poolId, ItemId itemId, uint128 amount) external;
```
mustermeiszer: missing counterpart.

src/interfaces/IItemManager.sol
```diff
@@ -0,0 +1,41 @@
...
+
+    /// @notice Returns the item value of this item.
+    function itemValue(PoolId poolId, ItemId itemId) external view returns (uint128 value);
+
+    /// @notice Returns the valuation method used for this item.
+    function valuation(PoolId poolId, ItemId itemid) external view returns (IERC7726);
+
+    /// @notice Returns the valuation method used for this item.
+    function updateValuation(PoolId poolId, ItemId itemid, IERC7726) external;
+}
```
mustermeiszer: Doc off. 

rather does update the stored `valuation` of an item using the valuation method

src/interfaces/IItemManager.sol
```diff
@@ -0,0 +1,41 @@
+// SPDX-License-Identifier: BUSL-1.1
+pragma solidity 0.8.28;
+
+import {PoolId} from "src/types/PoolId.sol";
+import {ItemId} from "src/types/Domain.sol";
+
+import {IERC7726} from "src/interfaces/IERC7726.sol";
+import {IAccountingItemManager} from "src/interfaces/IAccountingItemManager.sol";
+
+interface IItemManager is IAccountingItemManager {
```
mustermeiszer: Can we also have methods that allow updating the valuation method of an item? Not sure, maybe to flexible...

mustermeiszer: I bet we will need that ^^

mustermeiszer (commented):
Looks great already! Really astonighing how quickly you and William picked up speed!

src/Holdings.sol
lemunozm: Do you want to mean something like a "settlementPrice" that should be optional, right?

lemunozm: Could a workaround be:

```solidity
prev = getValuation()
updateValuation(customFixedValuation)
increase()
updateValuation(prev)
```

src/Holdings.sol
lemunozm: For sure, just by now I didn't need it in the code base

src/Holdings.sol
lemunozm: I understood the interest always increased internally for some IItemManager implementations. Do we also have a way to set that interest directly?

src/interfaces/IItemManager.sol
lemunozm: Yeah, it is `updateValuation()`, but as you correctly pointed out, the doc was wrong.

src/Holdings.sol
lemunozm: > Reusing ids

That's at least not the intention. Where is the ID reused? Here, we just remove the item from the storage (I would also need to remove it from `itemsIds` storage, BTW). But we never reuse an id, that's for sure

src/Holdings.sol
lemunozm: I'm not sure if I understand what you mean here.

The accounts are per `ItemId`, which means the accounts are also per `PoolId => ShareClassId => AssetId`. Is that not correct?

src/Holdings.sol
lemunozm: I do not understand this. I thought the valuation used should be the one attached to the item. Could you explain a bit more about the whys?

src/Holdings.sol
mustermeiszer: We should defintely allow to set it manually in case there are errors.

src/Holdings.sol
mustermeiszer: But that enforces to always have a second contract that actually impelemnts `IERC7726 `. I think that is not really feasible. IMO, `increase`/`decrease` are flexible with given "`settlementPrice`" parameter. `update` actual uses the set valuation method.

src/interfaces/IItemManager.sol
mustermeiszer: Haha. Sorry, I already forgot that when I arrived at that point here. Perfect!

src/Holdings.sol
lemunozm: Ok, now I understand you.

I already had some thoughts about this. My idea is to offer a IERC7726 implementation that the FM can access (not the feeders), which just sets the values to what they want.

The above should be:

```solidity
prev = getValuation()
settlementPriceOracle.set(assetId, value)
updateValuation(settlementPriceOracle)
increase()
updateValuation(prev)
```

My intention with this is to let the `IItemManager.increase()` interface as simply as possible because different implementations should behave differently. i.e. The bonds could not use a custom settlement price and could use the valuation computed on-chain. That should happen each time we want to compute the NAV for them, so we should be able to not give a list of automated prices each time.

My point is, let this valuation be as much automated as possible unless you want to force some value manually

src/interfaces/IAccountingItemManager.sol
lemunozm: Yes, or in the same `create()`, we can add a new parameter to set the accounts. Does it make more sense?

src/Holdings.sol
lemunozm: Ok, thinking out loud and making a note for myself:

In bonds, if we compute the debt with `LinearAccrual`, and the FM calls `increaseInterest()`, I think internally we should modify the `normalizedDebt` to achieve a debt proportional to the increased interest. Maybe we need some method to perform this computation in `LinearAccrual` contract or in the associated library.

src/Holdings.sol
lemunozm: Discussion in the above thread

src/Holdings.sol
wischli: >  TODO: should we initialize the accounts from AccountingItemManager here?

I am curious how you would initialize the accounts in the constructor then which doesn't know about any `poolId` etc?

src/Holdings.sol
```diff
@@ -0,0 +1,128 @@
...
+
+        ItemId itemId = ItemId.wrap(++lastItemId[poolId]);
+        itemIds[poolId][scId][assetId] = itemId;
+        items[poolId][itemId] = Item(scId, assetId, valuation_, 0, 0);
+    }
+
+    /// @inheritdoc IItemManager
+    function close(PoolId poolId, ItemId itemId, bytes calldata /*data*/ ) external auth {
+        Item storage item = items[poolId][itemId];
+        itemIds[poolId][item.scId][item.assetId] = ItemId.wrap(0);
```
wischli: Nit: Couldn't we just delete this entry as well instead of setting it to 0? Probably more or less same gas. Just think it looks cleaner.

src/interfaces/IItemManager.sol
```diff
@@ -0,0 +1,44 @@
...
+    /// @notice Creates a new item in a pool using a valuation
+    function create(PoolId poolId, IERC7726 valuation, bytes calldata data) external;
+
+    /// @notice Closes an item in a pool using a valuation
+    /// An item can only be closed if their value is 0.
+    function close(PoolId poolId, ItemId itemId, bytes calldata data) external;
+
+    /// @notice Increments the amount of an item and updates the value for that increment.
+    /// @return value The value the item has increment.
```
wischli, line 54: Nit: I suppose `value` is the updated value and not only the incremented part?

```suggestion
    /// @return value The value of the item after incrementing.
```

src/interfaces/IItemManager.sol
```diff
@@ -0,0 +1,44 @@
...
+    function increase(PoolId poolId, ItemId itemId, uint128 amount) external returns (uint128 value);
+
+    /// @notice Decrements the amount of an item and updates the value for that decrement.
+    /// @return value The value the item has decrement.
```
wischli, line 60: Nit: I suppose `value` is the updated value and not only the incremented part?

```suggestion
Nit: I suppose value is the updated value and not only the incremented part?

Suggested change
    /// @return value The value of the item after decrementing.
```

src/interfaces/IItemManager.sol
```diff
@@ -0,0 +1,44 @@
...
+    function decrease(PoolId poolId, ItemId itemId, uint128 amount) external returns (uint128 value);
+
+    /// @notice Decrements the amount of an item as interest.
```
wischli: Nit: This comment throws me of. IIUC, it decreases the interest of the item and implicitly its value due to renormalizing based on the decreased interest?

wischli (commented):
Great work so far! Looking forward to the tests. A few questions and nits from my side.

src/Holdings.sol
lemunozm: Ouch! sorry I've just realized I wrote that comment in the wrong method! I wanted to write it down in the `create()` method. The accounts should be another parameter of `create()`

src/Holdings.sol
lemunozm: I tried that before, but solidity does let me. Or I'm missing something, or I can only delete stuff when it's an struct

lemunozm: Or at least not with a type wrapper as `ItemId`:

```
Error:
Compiler run failed:
Error (9767): Built-in unary operator delete cannot be applied to type ItemId.
  --> src/Holdings.sol:52:9:
   |
52 |         delete itemId[poolId][item_.scId][item_.assetId];
```

lemunozm: Seems you can not with user defined types: https://ethereum.stackexchange.com/questions/112266/how-do-i-delete-entries-from-mappings-with-user-defined-types

src/interfaces/IItemManager.sol
lemunozm: As I envisioned this, it should be the incremented part, because that would be the amount used later in `updateEntry()`

lemunozm: The value of the item can be retrieved calling `itemValue()` after `increase()`, but the only way to know what was de difference of that increment is returning it in that moment

src/interfaces/IItemManager.sol
lemunozm: Good question, and I do not have a good answer.

It really would depend on how the `ItemManager` implementation handles the interest. If valuing an item for such implementation is affected by the interest, then yes. 

I'm gonna write a more generic doc like: "Decrements the interest". Whatever the interest is, it is up to the implementation.

src/Holdings.sol
```diff
@@ -0,0 +1,131 @@
...
+        poolRegistry = poolRegistry_;
+    }
+
+    /// @inheritdoc IItemManager
+    function create(PoolId poolId, IERC7726 valuation_, bytes calldata data) external auth {
+        (ShareClassId scId, AssetId assetId) = abi.decode(data, (ShareClassId, AssetId));
+
+        ItemId itemId_ = ItemId.wrap(++lastItemId[poolId]);
+        itemId[poolId][scId][assetId] = itemId_;
+        item[poolId][itemId_] = Item(scId, assetId, valuation_, 0, 0);
```
mustermeiszer: One optimization and logic proposal: 

* make all interfaces of `IItemManager` use `(PoolId poolId, ShareClassId scId, ItemId itemid, ...)` 
* For `Holdings` -> `ItemId = AssetId`

lemunozm: I thought about that. That would mean that other items, such as Bonds, should also be per `scId`, are we ok with that I guess. 

Could be any issue if later we want to move, i.e. a Bond from one share class to another?

src/AccountingItemManager.sol
```diff
@@ -0,0 +1,21 @@
...
+
+import {PoolId} from "src/types/PoolId.sol";
+import {ItemId, AccountId} from "src/types/Domain.sol";
+import {IAccountingItemManager} from "src/interfaces/IAccountingItemManager.sol";
+import {Auth} from "src/Auth.sol";
+
+abstract contract AccountingItemManager is Auth, IAccountingItemManager {
+    mapping(PoolId => mapping(ItemId => Accounts)) public accounts;
+
+    constructor(address deployer) Auth(deployer) {}
```
ilinzweilin: ``
```suggestion
    constructor(address ward) Auth(ward) {}
```

ilinzweilin: deployer can pass a different address as ward

src/AccountingItemManager.sol
```diff
@@ -0,0 +1,21 @@
+// SPDX-License-Identifier: BUSL-1.1
+pragma solidity 0.8.28;
```
ilinzweilin:
```suggestion
pragma solidity ^0.8.28; 
```

ilinzweilin: I would always use caret to make sure that this certain version is used **or any higher**

src/interfaces/IAccountingItemManager.sol
```diff
@@ -0,0 +1,19 @@
...
+
+// IAccountingItemManager is an interface all ItemManager should implement (mostly in the same way) to be able to be
+// used in an IAccounting contract.
+interface IAccountingItemManager {
+    struct Accounts {
+        AccountId asset;
+        AccountId equity;
+        AccountId loss;
+        AccountId gain;
+    }
```
ilinzweilin: struct is a very unflexible structure to go with imo, based on the discussion below.
Why not just having an array
 containing `AccountIds`. 
In that way we can stay flexible and provide different type & number of accounts for each item. 

We just need to make sure AccountId is unique across all account types. 
alternatively we can go with the following struct: 
```
struct Account {
        AccountId id;
        AccountType type; (can be an enum)
    }

```
and `Accounts` would be an array of the type `Account`

src/AccountingItemManager.sol
```diff
@@ -0,0 +1,21 @@
...
+abstract contract AccountingItemManager is Auth, IAccountingItemManager {
+    mapping(PoolId => mapping(ItemId => Accounts)) public accounts;
+
+    constructor(address deployer) Auth(deployer) {}
+
+    function setItemAccounts(PoolId poolId, ItemId itemId, Accounts calldata itemAccounts_) external auth {
+        accounts[poolId][itemId] = itemAccounts_;
+    }
+
+    function itemAccounts(PoolId poolId, ItemId itemId) external view returns (Accounts memory) {
```
ilinzweilin: in theory this function is not required, as `accounts` is a public mapping and can be used as a getter

src/AccountingItemManager.sol
ilinzweilin, line 1: having a separate contract and interface for `AccountingItemManager` feels a bit overengineered to me.
The logic is very tightly coupled to ItemManager and I would just merge that code into the ItemManager.

src/interfaces/IItemManager.sol
```diff
@@ -0,0 +1,46 @@
...
+    /// @notice Decrements the interest of an item.
+    /// @param interestAmount The amount of interest to be decremented.
+    function decreaseInterest(PoolId poolId, ItemId itemId, uint128 interestAmount) external;
+
+    /// @notice Reset the value of an item using the current valuation.
+    /// @return diff The difference in value after the new valuation.
+    function update(PoolId poolId, ItemId itemId) external returns (int128 diff);
+
+    /// @notice Returns the item value of this item.
+    function itemValue(PoolId poolId, ItemId itemId) external view returns (uint128 value);
```
ilinzweilin, line 84:
```suggestion
    function valuation(PoolId poolId, ItemId itemId) external view returns (uint128 value);
```

src/interfaces/IItemManager.sol
```diff
@@ -0,0 +1,46 @@
...
+
+    /// @notice Returns the valuation method used for this item.
+    function valuation(PoolId poolId, ItemId itemid) external view returns (IERC7726);
```
ilinzweilin:
```suggestion
    function oracle(PoolId poolId, ItemId itemid) external view returns (IERC7726);
```

src/interfaces/IItemManager.sol
```diff
@@ -0,0 +1,46 @@
...
+
+    /// @notice Updates the valuation method used for this item.
+    function updateValuation(PoolId poolId, ItemId itemid, IERC7726 valuation) external;
```
ilinzweilin:
```suggestion
    function updateOracle(PoolId poolId, ItemId itemid, IERC7726 valuation) external;
```

src/interfaces/IItemManager.sol
```diff
@@ -0,0 +1,46 @@
...
+    /// @param interestAmount The amount of interest to be incremented.
+    function increaseInterest(PoolId poolId, ItemId itemId, uint128 interestAmount) external;
+
+    /// @notice Decrements the interest of an item.
+    /// @param interestAmount The amount of interest to be decremented.
+    function decreaseInterest(PoolId poolId, ItemId itemId, uint128 interestAmount) external;
+
+    /// @notice Reset the value of an item using the current valuation.
+    /// @return diff The difference in value after the new valuation.
+    function update(PoolId poolId, ItemId itemId) external returns (int128 diff);
```
ilinzweilin:
```suggestion
    function updateValuation(PoolId poolId, ItemId itemId) external returns (int128 diff);
```

ilinzweilin: I am not sure we need this function. 
Wouldn't the most recent valuation getter return the most recent valuation by calling the oracle?

src/interfaces/IItemManager.sol
```diff
@@ -0,0 +1,46 @@
...
+
+import {PoolId} from "src/types/PoolId.sol";
+import {ItemId} from "src/types/Domain.sol";
+
+import {IERC7726} from "src/interfaces/IERC7726.sol";
+import {IAccountingItemManager} from "src/interfaces/IAccountingItemManager.sol";
+
+interface IItemManager is IAccountingItemManager {
+    /// @notice Creates a new item in a pool using a valuation
+    function create(PoolId poolId, IERC7726 valuation, bytes calldata data) external;
```
ilinzweilin:
```suggestion
    function create(PoolId poolId, IERC7726 oracle, bytes calldata data) external;
```

src/Holdings.sol
```diff
@@ -0,0 +1,131 @@
...
+contract Holdings is AccountingItemManager, IHoldings {
+    using MathLib for uint256;
+
+    mapping(PoolId => mapping(ItemId => Item)) public item;
+    mapping(PoolId => mapping(ShareClassId => mapping(AssetId => ItemId))) public itemId;
+    mapping(PoolId => uint32) lastItemId;
+
+    IPoolRegistry immutable poolRegistry;
+
+    constructor(address deployer, IPoolRegistry poolRegistry_) AccountingItemManager(deployer) {
```
ilinzweilin:
```suggestion
    constructor(address ward, IPoolRegistry poolRegistry_) AccountingItemManager(deployer) {
```

src/Holdings.sol
```diff
@@ -0,0 +1,131 @@
...
+    AssetId assetId;
+    IERC7726 valuation;
+    uint128 assetAmount;
+    uint128 assetAmountValue;
+}
+
+contract Holdings is AccountingItemManager, IHoldings {
+    using MathLib for uint256;
+
+    mapping(PoolId => mapping(ItemId => Item)) public item;
```
ilinzweilin: the more I look at the mappings with that naming structure, giving it a singular name instead of plural feels not intuitive to me.  
```
Item item = items[poolId][ItemId] -> intuitive
Item item = item[poolId][ItemId] -> not intuitive
```

I have never noticed any other solidity contracts using the singular naming & and we haven't in the past. Is there any reference?

I know I said in slack thatI don't have hard feelings, but now that I am reading through the code, I notice the difference...

src/Holdings.sol
```diff
@@ -0,0 +1,131 @@
...
+    mapping(PoolId => uint32) lastItemId;
+
+    IPoolRegistry immutable poolRegistry;
+
+    constructor(address deployer, IPoolRegistry poolRegistry_) AccountingItemManager(deployer) {
+        poolRegistry = poolRegistry_;
+    }
+
+    /// @inheritdoc IItemManager
+    function create(PoolId poolId, IERC7726 valuation_, bytes calldata data) external auth {
```
ilinzweilin: I would make sure that the poolId exists  & the oracle is a valid address

src/Holdings.sol
```diff
@@ -0,0 +1,131 @@
...
+        (ShareClassId scId, AssetId assetId) = abi.decode(data, (ShareClassId, AssetId));
+
+        ItemId itemId_ = ItemId.wrap(++lastItemId[poolId]);
+        itemId[poolId][scId][assetId] = itemId_;
```
ilinzweilin, line 74: hmmm, maybe I am missing something, but why would an item be bound to a share class. Shouldn't the items be per pool and the share class NAVs etc are calculated on top.

src/Holdings.sol
```diff
@@ -0,0 +1,131 @@
...
+
+import {AccountingItemManager} from "src/AccountingItemManager.sol";
+import {Auth} from "src/Auth.sol";
+
+struct Item {
+    ShareClassId scId;
+    AssetId assetId;
+    IERC7726 valuation;
+    uint128 assetAmount;
+    uint128 assetAmountValue;
```
ilinzweilin: I think it wouldn't hurt to also store `poolId` and `itemId` info here. Since we don't have a mapping from `item -> poolId`. 

I am don't fully understand why we need `scId` here

src/Holdings.sol
ilinzweilin: since we are using increment for `itemId` of each pool here, we could simplify the mapping:
`mapping(PoolId => Item[]) public items;`

each poolId would return an array of items, where the index of each item is the itemId

src/Holdings.sol
```diff
@@ -0,0 +1,131 @@
...
+
+        // TODO: should we initialize the accounts from AccountingItemManager here using an Accounts parameter?
+        // I think so.
+    }
+
+    /// @inheritdoc IItemManager
+    function close(PoolId poolId, ItemId itemId_, bytes calldata /*data*/ ) external auth {
+        Item storage item_ = item[poolId][itemId_];
+        itemId[poolId][item_.scId][item_.assetId] = ItemId.wrap(0);
+        delete item[poolId][itemId_];
```
ilinzweilin: are we sure that we want to delete item and not just flag closed

src/Holdings.sol
```diff
@@ -0,0 +1,131 @@
...
+        itemId[poolId][scId][assetId] = itemId_;
+        item[poolId][itemId_] = Item(scId, assetId, valuation_, 0, 0);
+
+        // TODO: should we initialize the accounts from AccountingItemManager here using an Accounts parameter?
+        // I think so.
+    }
+
+    /// @inheritdoc IItemManager
+    function close(PoolId poolId, ItemId itemId_, bytes calldata /*data*/ ) external auth {
+        Item storage item_ = item[poolId][itemId_];
```
ilinzweilin: same here would make sure pool & item exist. Same applies for all the following functions

src/Holdings.sol
ilinzweilin: why are we storing `assetAmountValue` that needs to be updated.
I think the valuation getter should just always return the most recent valuation from the oracle to avoid outdated data...

src/Holdings.sol
```diff
@@ -0,0 +1,131 @@
...
+
+    /// @inheritdoc IItemManager
+    function increase(PoolId poolId, ItemId itemId_, uint128 amount) external auth returns (uint128 amountValue) {
+        Item storage item_ = item[poolId][itemId_];
+        address poolCurrency = address(poolRegistry.currency(poolId));
+
+        amountValue = uint128(item_.valuation.getQuote(amount, AssetId.unwrap(item_.assetId), poolCurrency));
+
+        item_.assetAmount += amount;
+        item_.assetAmountValue += amountValue;
```
ilinzweilin, line 99: I think only the `assetAmount` needs to be increased and the valuation just always retrieved from the oracle based on the `assetAmount`

src/Holdings.sol
```diff
@@ -0,0 +1,131 @@
...
+    }
+
+    /// @inheritdoc IItemManager
+    function itemValue(PoolId poolId, ItemId itemId_) external view returns (uint128 value) {
+        return item[poolId][itemId_].assetAmountValue;
+    }
+
+    /// @inheritdoc IItemManager
+    function valuation(PoolId poolId, ItemId itemId_) external view returns (IERC7726) {
+        return item[poolId][itemId_].valuation;
```
ilinzweilin: I would always return the exact valuation from the oracle here based on the asset amount of course: 
pseudo code:
`        return uint128(item_.valuation.getQuote(item_.assetAmount, item_.assetId), poolCurrency));`

src/Holdings.sol
```diff
@@ -0,0 +1,131 @@
...
+        address poolCurrency = address(poolRegistry.currency(poolId));
+
+        amountValue = uint128(item_.valuation.getQuote(amount, AssetId.unwrap(item_.assetId), poolCurrency));
+
+        item_.assetAmount -= amount;
+        item_.assetAmountValue -= amountValue;
+    }
+
+    /// @inheritdoc IItemManager
+    function update(PoolId poolId, ItemId itemId_) external auth returns (int128 diff) {
```
ilinzweilin: can be removed imo

src/Holdings.sol
```diff
@@ -0,0 +1,131 @@
...
+
+    /// @inheritdoc IItemManager
+    function decrease(PoolId poolId, ItemId itemId_, uint128 amount) external auth returns (uint128 amountValue) {
+        Item storage item_ = item[poolId][itemId_];
+        address poolCurrency = address(poolRegistry.currency(poolId));
+
+        amountValue = uint128(item_.valuation.getQuote(amount, AssetId.unwrap(item_.assetId), poolCurrency));
+
+        item_.assetAmount -= amount;
+        item_.assetAmountValue -= amountValue;
```
ilinzweilin, line 119: same here: think only the quantity needs to be increased and the valuation just always retrieved from the oracle based on the quantity

ilinzweilin (changes requested):
🚀

src/interfaces/IItemManager.sol
mustermeiszer: We need this definitely from a business perspective. This is actually updating the holdings if there has been a los or gain between now and the last update.

src/Holdings.sol
mustermeiszer: That is a business requirement. See PRD for details.

src/Holdings.sol
mustermeiszer: We need that for knowing intermediate gains/loss.

src/Holdings.sol
mustermeiszer: Second that. Deleting seems dangerous to me ^^

src/Holdings.sol
mustermeiszer: That is going to change. Increase will take into account an externally provided value for the given amount.  So that we can note how the item was priced when we got it. Note: This is in most cases not a synchronous composition, so that oracle value and purchase value can differ significantly.

src/Holdings.sol
mustermeiszer: See above.

src/Holdings.sol
mustermeiszer: Need that for sure for tracking loss/gains.

src/Holdings.sol
ilinzweilin: yes, I am just suggesting having this wrapped into a getter function that always uses the most recent price from the oracle. 
The `assetAmountValue` can be outdated & if you need to update the value first it comes with a write / storage update operation which costs gas

src/Holdings.sol
ilinzweilin: can you point me to the section please

src/Holdings.sol
lemunozm: Curious, what's the security problem with deleting items from a map? (if deleting implies that the variable used to check their existence becomes 0). We'll never reuse that itemId

src/Holdings.sol
mustermeiszer: Yes, all items should be per share class. we can rather have a special share class that is the global one later.

src/Holdings.sol
mustermeiszer: https://www.notion.so/Centrifuge-Protocol-V1-1112eac24e178040bdfafb6a8e9b090c?pvs=4#1122eac24e178069aed7f3469d920c78

src/Holdings.sol
mustermeiszer: How do you wanna calculate the gain/loss between last update and now, if you don't now the past value?

src/Holdings.sol
lemunozm: Ok

> For Holdings -> ItemId = AssetId

Just one thing regarding this. `ItemId` could be something different to `AssetId` in other `IItemManager` implementations, so it should be generic enough.  I think we will still need a mapping from `ItemId` => `AssetId`. So, changing this to use a `ShareClassId`, we're not saving us from the mapping. 

Anyways, I'm ok with using the `ShareClassId` param in `IItemManager` interface

src/Holdings.sol
lemunozm: Regarding this, let's start with your solution of having an extra parameter 👍🏻

src/AccountingItemManager.sol
lemunozm: @georgi had a strong opinion in favor of `deployer`. I'm ok with both names, but we should reach an agreement to use the same name for all contracts.

src/AccountingItemManager.sol
lemunozm: I'm using the fixed version because, IIRC, @hieronx argued in favor of it for non-interfaces, but I do not remember now the exact reason. 

Again, I'm open to any approach as long as we have a consensus in all contracts to do the same

src/interfaces/IAccountingItemManager.sol
lemunozm: Yes, I think this makes sense. What we need to be sure of is that at least those 4 accounts exist for all `IItemManager` implementations. But we can model them as you say.

src/AccountingItemManager.sol
lemunozm: The intention with it is that all `IItemManager` should share that logic. But I agree that maybe at this point feels like overengineering, so I agree, by now, to merge this with the `IItemManager` and put the implementation in the `Holdings` to let things simple. But probably we should factorize this part again later

src/interfaces/IItemManager.sol
lemunozm: The `valuation` name is the name we're using for `IERC7726`, which is different from the item value

src/interfaces/IItemManager.sol
lemunozm: We already have "reached" consensus about these namings. In this case, valuation does not need to be an oracle; could be another contract that compute the value on-chain

src/interfaces/IItemManager.sol
lemunozm: Same as above

src/interfaces/IItemManager.sol
lemunozm: Same as above

src/Holdings.sol
lemunozm: Thread: https://github.com/centrifuge/pools-internal/pull/42#discussion_r1892983647

src/Holdings.sol
lemunozm: Agree with the mapping => array stuff.

Regarding the naming, I think it's something we need to discuss again in Slack, given we already reached a consensus about that

src/Holdings.sol
lemunozm: Sure

src/Holdings.sol
lemunozm: @mustermeiszer

src/Holdings.sol
lemunozm: Those are different values.`valuation.getQuote` returns the current value, here we're returning the valuation address used

src/Holdings.sol
lemunozm: Refs:
- slack thread: https://kflabs.slack.com/archives/C07PG2EUR9C/p1734423525896669
- PR where this was changed for poolRegistry: https://github.com/centrifuge/pools-internal/pull/45

src/Holdings.sol
lemunozm: @mustermeiszer I think this optimization can not be applied because we still need the mapping anyways. Also, because we still need the mapping, we can avoid passing the `ShareClassId` in all interface methods. I think this can be noisy for other `IItemManager` implementations, such as `Bonds`, where we would not need a `ShareClassId` parameter.

src/Holdings.sol
mustermeiszer: Never know what external integrations depend on this to be stable. Just safer and easier. Once defined, never used for something different than in the beginning.

lemunozm: I think we're talking about different things. Sure, a removed item would never be reused. Here, I'm just removing the item to save storage, but that ID will never be reused.

src/Holdings.sol
mustermeiszer: We would also need it for Bonds. All assets are either
* associated with ONE share class
* associated with ALL share classes

lemunozm: But don't we decide that the asset side could be agnostic from the `ShareClassId` here?

https://kflabs.slack.com/archives/C07PG2EUR9C/p1733859638266489?thread_ts=1733859377.513819&cid=C07PG2EUR9C

I'm not against adding this, but I want to be sure that adding the ShareClassId to the IItemInterface is not an extra layer that does not really need it.

src/Holdings.sol
```diff
@@ -0,0 +1,190 @@
...
+    IERC7726 valuation;
+    uint128 assetAmount;
+    uint128 assetAmountValue;
+    bool alive;
+}
+
+contract Holdings is Auth, IHoldings {
+    using MathLib for uint256; // toInt128()
+
+    mapping(PoolId => Item[]) public item;
```
peculiarity, line 32: **Note**: I would rename this to catalog. Catalog itself is a singular but it implies a collection of items. So for each `PoolId`, you have a list of items. Also not to mention that for a given `PoolId`  an array of items  is returned so in this case I don't think usage of `item` is the correct one.

src/Holdings.sol
```diff
@@ -0,0 +1,190 @@
...
+    mapping(PoolId => mapping(ShareClassId => mapping(AssetId => ItemId))) public itemId;
```
peculiarity, line 33: **Question**: 
Does this mapping exist for the sole purpose that you have a O(1) read  for getting a specific item instead of iterating over the array returned from the `item` variable and looking for one specific item?

src/Holdings.sol
```diff
@@ -0,0 +1,190 @@
...
+    mapping(PoolId => mapping(ItemId => mapping(uint8 kind => AccountId))) public accountId;
```
peculiarity, line 34: **Question**: 
I'm trying to understand how this is going to be used. I mean you populate the kind from the `accountId` so you know the accountId in advance. I assume you are going to read this mapping passing some sort of a `kind` but if that's the case can't you just pass directly the `accountId`?

src/Holdings.sol
```diff
@@ -0,0 +1,190 @@
...
+
+    IPoolRegistry immutable poolRegistry;
+
+    constructor(address deployer, IPoolRegistry poolRegistry_) Auth(deployer) {
```
peculiarity: ****Ultra NIT**:** Following CV side, the deployer address is always the last param.

src/Holdings.sol
```diff
@@ -0,0 +1,190 @@
...
+        poolRegistry = poolRegistry_;
+    }
+
+    /// @inheritdoc IItemManager
+    function create(PoolId poolId, IERC7726 valuation_, AccountId[] memory accounts, bytes calldata data)
```
peculiarity, line 61: **Question:**
Any particular reason why `data` is not passed directly as `ShareClassId` , `AssetId` and instead it is passed as encoded calldata?

src/Holdings.sol
```diff
@@ -0,0 +1,190 @@
...
+        (ShareClassId scId, AssetId assetId) = abi.decode(data, (ShareClassId, AssetId));
+
+        require(!scId.isNull(), WrongShareClassId());
+        require(!assetId.isNull(), WrongAssetId());
+
+        ItemId itemId_ = newItemId(item[poolId].length);
+        item[poolId].push(Item(scId, assetId, valuation_, 0, 0, true));
+        itemId[poolId][scId][assetId] = itemId_;
+
+        for (uint256 i = 0; i < accounts.length; i++) {
```
peculiarity:
```suggestion
        for (uint256 i ; i < accounts.length; i++) {
```
It is a bad practice to initialize to 0. `uint256 i` is always 0 initially. Even during audits, this gets flagged.

src/Holdings.sol
```diff
@@ -0,0 +1,190 @@
...
+
+import "forge-std/Test.sol";
+
+struct Item {
+    ShareClassId scId;
+    AssetId assetId;
+    IERC7726 valuation;
+    uint128 assetAmount;
+    uint128 assetAmountValue;
+    bool alive;
```
wischli: Q: Wondering if keeping inactive items will rather be an edge-case or common practice in case we have an understanding of that currently.

src/Holdings.sol
```diff
@@ -0,0 +1,190 @@
...
+    using MathLib for uint256; // toInt128()
+
+    mapping(PoolId => Item[]) public item;
+    mapping(PoolId => mapping(ShareClassId => mapping(AssetId => ItemId))) public itemId;
+    mapping(PoolId => mapping(ItemId => mapping(uint8 kind => AccountId))) public accountId;
+
+    IPoolRegistry immutable poolRegistry;
+
+    constructor(address deployer, IPoolRegistry poolRegistry_) Auth(deployer) {
+        poolRegistry = poolRegistry_;
```
wischli, line 39: CR: Missing check that `poolRegistry_ != address(0)`.

src/Holdings.sol
```diff
@@ -0,0 +1,190 @@
...
+
+        emit CreatedItem(poolId, itemId_, valuation_);
+    }
+
+    /// @inheritdoc IItemManager
+    function close(PoolId poolId, ItemId itemId_, bytes calldata /*data*/ ) external auth {
+        Item storage item_ = item[poolId][itemId_.index()];
+        require(item_.alive, ItemNotFound());
+
+        item_.alive = false;
```
wischli: Q: Curious on why we don't remove closed items from storage.

src/interfaces/IItemManager.sol
```diff
@@ -0,0 +1,89 @@
...
+
+interface IItemManager {
+    /// @notice Emitted when an item is created
+    event CreatedItem(PoolId indexed, ItemId indexed, IERC7726 valuation);
+
+    /// @notice Emitted when an item is closed
+    event ClosedItem(PoolId indexed, ItemId indexed);
+
+    /// @notice Emitted when an item is increased
+    event ItemIncreased(PoolId indexed, ItemId indexed, IERC7726 valuation, uint128 amount, uint128 increasedValue);
```
wischli, line 18: Nit: Maybe make this more explicit? 

```suggestion
    event ItemIncreasedAmount(PoolId indexed, ItemId indexed, IERC7726 valuation, uint128 amount, uint128 increasedValue);
```

src/types/ItemId.sol
```diff
@@ -0,0 +1,20 @@
...
+import {MathLib} from "src/libraries/MathLib.sol";
+
+type ItemId is uint32;
+
+function newItemId(uint256 arrayLength) pure returns (ItemId) {
+    return ItemId.wrap(MathLib.toUint32(arrayLength) + 1);
+}
+
+function index(ItemId itemId) pure returns (uint32) {
+    return ItemId.unwrap(itemId) - 1;
```
wischli, line 17: Nit: Maybe add a comment line that an item id is always >= 1

src/Holdings.sol
peculiarity: By default the value for the pooLRegistry is 0x0. Setting it to the zero address doesn't affect it that much.  Also usually in CV we avoid checking against the zero address ( there are some cases where you'd like to check ). The reason is that 0x0 address doesn't different much from any other address that is not the correct one.  What if I provide 0x1? Why 0x0 is so special and must be checked and 0x1 should not be ?

peculiarity: **Suggestion**.  Actually @wischli  made me thinking. I think it's a good idea to add a file method that can update the `poolRegistry` in here of wrong configuration / redeployment of the registry ( because of some reason ).

src/Holdings.sol
peculiarity: @wischli  If you refer to using `delete`, then you can't use it with Custom Types. Also it doesn't "Delete" anything per se. It sets all non-mapping variables of the struct to their default value ( 0, "", false ... )

src/Holdings.sol
lemunozm: That's what I think at first, and I would expect:

```solidity
holdings.items(POOL_A)[index];
```

but from the test, I needed to call:

```solidity
holdings.item(POOL_A, index);
```
to retrieve the item. So it seems that when the element of the map is an array, the generated call retrieve the final element and we should follow the decision made. Unless I'm missing something

src/Holdings.sol
lemunozm: The mapping exists because given a `(ShareclassId, AssetId)` pair you should identify an `ItemId`. This is needed, for example, in approveDeposit/Redemption methods

src/Holdings.sol
lemunozm: The point here is that you must support different accounts for an item. So basically each item has a set of accounts, and each account of that set is identified by the `kind`

src/Holdings.sol
lemunozm: Yes. `ShareClassId` and `AssetId` are specific to Holdings implementation. Other IItemManager will not use them. The IItemManager interface should be generic enough.

i.e. `Bonds` will use a collateral instead of those

src/Holdings.sol
lemunozm: Oh, didn't now that. 👍🏻 

> It is a bad practice to initialize to 0.

Is there any reason? Bloating code or so?

src/Holdings.sol
lemunozm: Here is the story about this. I'm still not super sure about array vs mapping for store items. For now I decided using an array.

The point is, if right now I'm using an array to store them, then I can not remove from the item array the closed items. I could clear them, but I'm not achieving nothing with that. So basically I can keep the old data as history for the UI (something similar we did in pallet-loans).

src/Holdings.sol
lemunozm: > The reason is that 0x0 address doesn't different much from any other address that is not the correct one.

Mostly agree with this reasoning.

From my point of view, we should check against 0 if something is treated as optional. So I would say 0x0 is special because it's always used as optional::none in a lot of cases. And that value could emerge automatically from reading a empty slot storage.

> add a file method

Correct, will add it 👍🏻

src/Holdings.sol
lemunozm: Yes. More details here: https://github.com/centrifuge/pools-internal/pull/42#discussion_r1908328856

src/interfaces/IItemManager.sol
lemunozm: You hint this to differentiate from the Interest, right?

Not sure about that 🤔. There is no something like IncreasedAmount vs IncreasedInterest. `ItemIncreased` is the general way, and `ItemIncreasedInterest` is a specific way of increasing it. It's more like top-down instead of being at the same level. 

But I will keep this idea in mind if I feel it's getting more confusing

src/types/ItemId.sol
lemunozm: Sounds good!

src/Holdings.sol
lemunozm: > add a file method

I'm rethinking this, and not sure if we should. We decided poolRegistry to be an immutable contract. It's supposed never to change, and if it finally changes, we should upload again the Holdings contract.

I will open a thread to discuss this among all of us. Not sure the decision to take.

src/types/ItemId.sol
lemunozm: Done

src/Holdings.sol
lemunozm: Done

src/interfaces/IItemManager.sol
```diff
@@ -0,0 +1,89 @@
...
+    function valuation(PoolId poolId, ItemId itemId) external view returns (IERC7726);
+
+    /// @notice Updates the valuation method used for this item.
+    function updateValuation(PoolId poolId, ItemId itemId, IERC7726 valuation) external;
+
+    /// @notice Sets an account id for an specific kind
+    function setAccountId(PoolId poolId, ItemId itemId, AccountId accountId) external;
+
+    /// @notice Returns an account id for an specific kind
+    function accountId(PoolId poolId, ItemId itemId, uint8 kind) external view returns (AccountId);
```
peculiarity, line 93: Not sure where to put it but  you are missing the getter definions for the `item` and the `itemId` public variables.

src/Holdings.sol
peculiarity: Ah, right! Automatically generated getters don't return dynamic arrays.  Same like they don't return structs but rather tuples.  The way you call it is the correct one.  I would still rename it to a catalog but that's more of a **Suggestion**.  Given this comment and the one about missing getter definition in the interface, remember that it should return a tuple ( explicitly you have to define all members ). It cannot return the struct directly.r

src/interfaces/IItemManager.sol
lemunozm: I think we do not need it, at list in the `IItemManager`, because they are related to the custom `Holding` implementation. Other `IItemManager` will have other structs to represent their items.

Maybe it makes more sense in `IHoldings`. Will think about it

lemunozm: Maybe they should be private instead. At least the `item` array

src/Holdings.sol
peculiarity: I see. For me this is more like a code smell. I see you have written documentation but you haven't provided explanation for each parameter. Here comes the tricky part. Lets say explain in the documentation if it's implemented by the Holding then the data is encoded `ShareClassId` and `AssetId`. 
Then there is another contract that implements it and you have to extend that documentation because let's face it, first `data` as param name is very generic and second the person who uses this interface must know what this param is used for.

My suggestion is to define this method as part of the final interface . i.e. IHoldings, IBonds with its specific definition. 

One more thing. **Super Nit** `create` somehow implies that something must be returned from the method. Suggestion would be to rename it to `addItemEntry` or `initializeItem`.

src/interfaces/IItemManager.sol
peculiarity: I agree that they are not needed in the `ItemManager`  but rather in the `IHoldings`. I commented it here because `accountId` was alos a state variable in the `Holdings` contract but if they are public, definitely must be defined in the interface.

src/libraries/MathLib.sol
```diff
@@ -169,6 +170,12 @@ library MathLib {
         return uint128(value);
     }
 
+    /// @notice Safe type conversion from uint256 to int128.
+    function toInt128(uint256 value) internal pure returns (int128) {
```
peculiarity, line 174: Nit: Test is missing

src/Holdings.sol
lemunozm: > remember that it should return a tuple ( explicitly you have to define all members )

Yeah, that sucks. Will think about the naming

src/Holdings.sol
lemunozm: > My suggestion is to define this method as part of the final interface . i.e. IHoldings, IBonds with its specific definition

Do you mean define as part as `IHolding` interface, and later let the generic `IItemManager` method to be implemented calling the `IHolding` interface? I like that 👍🏻 .

> create somehow implies that something must be returned from the method.

I didn't thought about that implicit behavior through the `create()` name. But it could make sense. I'll think about it 👍🏻

src/Holdings.sol
lemunozm: Added

src/Holdings.sol
```diff
@@ -0,0 +1,185 @@
...
+
+    /// @inheritdoc IItemManager
+    function decrease(PoolId poolId, ItemId itemId_, IERC7726 valuation_, uint128 amount)
+        external
+        auth
+        returns (uint128 amountValue)
+    {
+        require(address(valuation_) != address(0), WrongValuation());
+
+        Item storage item_ = item[poolId][itemId_.index()];
```
lemunozm, line 113: @georgi, maybe you can help me with this.

I do not want these methods to return an indexOutOfBounds if the index is not in the array. Instead. I want them to return a `ItemNotFound` as the `IItemManager` dictates.

My first approach is:

```solidity
try item[poolId][itemId_.index()] returns (Item memory item_) {
    // Stuff
} catch Panic(uint236 0x32 /*oob*/) {
    revert(ItemNotFound())
}
```

Nevertheless, solidity forces me not to use `storage` and use `memory` instead. Any workaround you suggest?

src/Holdings.sol
lemunozm: Added the `IHolding::create()` method. 

By now, I'm gonna let the `create()` name, given our inheritance of this name and the current understanding from all of us. I think if the method still creates something internally, even though it does not return stuff, the name is still valid IMO.

src/Holdings.sol
peculiarity: `try / catch` doesn't support storage pointers as returned values

peculiarity: Possible workaround IMO could be 
```
Item[] storage items = item[poolId];
require(items.length >= itemId_.index(), "NotFound");
Item storage item_ = items[itemId_.index()];
 ```

File: src/Holdings.sol (hunk 42 above)
Comment (peculiarity) [reply src/Holdings.sol:61]: Yeah, there are cases where `create` not necessarily have to return anything. I'm fine with how it is . Couldn't think of a better name anyways.

File: src/Holdings.sol (hunk 51 above)
Comment (lemunozm) [reply src/Holdings.sol:113]: Thanks!

File: src/libraries/MathLib.sol (hunk 50 above)
Comment (lemunozm) [reply src/libraries/MathLib.sol:174]: Added

File: src/Holdings.sol (hunk 46 above)
Comment (lemunozm) [reply src/Holdings.sol]: After a conversation with Frederik, we do not need to remove items from Holdings, so I just remove the `alive` field. Everything is alive now 🌳

File: src/interfaces/IItemManager.sol (hunk 49 above)
Comment (lemunozm) [reply src/interfaces/IItemManager.sol:93]: Could be the case that they are public because you need to access them from the test, but they are not exposed through the interface?

File: test/unit/libraries/MathLib.t.sol [hunk 52]
```diff
@@ -65,6 +65,18 @@ contract MathLibTest is Test {
         MathLib.toUint128(uint256(type(uint128).max) + x);
     }
 
+    function testToInt128(uint256 x) public pure {
+        x = bound(x, 0, uint256(uint128(type(int128).max)));
+
+        assertEq(x, uint256(uint128(MathLib.toInt128(x))));
+    }
+
+    function testToInt128Overflow(int128 x) public {
```
peculiarity: I find this one:
```
 function testToInt128Overflow(uint256 x) public {
        x= bound(x, uint256(uint128(type(int128).max)) + 1 , type(uint256).max);
        vm.expectRevert(MathLib.Int128_Overflow.selector);
        MathLib.toInt128(x);
   }
```
simpler. 
Also the passed param is of type uint256  and you are testing (uint256 -> int128 ) conversion.

src/interfaces/IItemManager.sol
lemunozm: Right now, I added `itemId()` getter through the interface with the autogenerated code, and I also have an `itemProperties(),` which only returns the relevant properties. Other properties can be considered internal at this point (could not be so in the future).

Does that make sense by now?

test/unit/libraries/MathLib.t.sol
lemunozm: Makes sense! added your solution

src/interfaces/IItemManager.sol
peculiarity: Well here is what bothers me. You have 3 separate view methods `itemProperties` , `itemValue`, `valuation` that basically return a value of specific property. 
If you define the getter in the interface for the `item` and return a specific `Item` then  I don't see the need of all these methods. 
I have the feeling that this is back to the discussion of having multiple getters and bloating the code vs just 1 getter that returns the `Item` and you can read any value you want. Wdyt?

lemunozm: The problem here is that `itemValue()` and `valuation()` are generic methods from `IItemManager`. Each implementation should store or compute those values as they need. In different ways.

`itemProperties()` is specific from `IHoldings`.

The attributes from `Item` struct are private to how `Holdings` decide internally to store its data and should not be public. It can change, and we do not want to break the `Holdings` API. And definitelly, they are not shared across other `IItemManager` implementation

src/Holdings.sol
peculiarity: It's more of a gas thingy. It's already 0 so the assigning explicitly 0 adds some opcodes.

src/Holdings.sol
lemunozm: No longer applies because we do not support close()

src/Holdings.sol
lemunozm: Ok, I thought the optimizer automatically discarded these kind of things

src/Holdings.sol
```diff
@@ -0,0 +1,188 @@
...
+        (ShareClassId scId, AssetId assetId) = abi.decode(data, (ShareClassId, AssetId));
+
+        require(poolRegistry.exists(poolId), IPoolRegistry.NonExistingPool(poolId));
+        require(address(valuation_) != address(0), WrongValuation());
+        require(!scId.isNull(), WrongShareClassId());
+        require(!assetId.isNull(), WrongAssetId());
+
+        itemId_ = newItemId(item[poolId].length);
+        item[poolId].push(Item(scId, assetId, valuation_, 0, 0));
+        itemId[poolId][scId][assetId] = itemId_;
```
peculiarity, line 74: **Question:** Isn't it supposed to have a check whether the given `poolId` has assigned the  `scId` in `ShareClassManager` ? Same goes with hte `assetID`

lemunozm: From the Holdings perspective, it's independent. The pool manager will check that probably before call `create()`, because the PoolManager has access to the `ShareClassId`

src/Holdings.sol
lemunozm: I prefer by now to leave `item`, let's see how to name this more accurately when we have other `IItemManager`

src/Holdings.sol
```diff
@@ -0,0 +1,188 @@
...
+    /// @inheritdoc IItemManager
+    /// @param data Expect a (ShareClassId, AssetId) encoded.
+    function create(PoolId poolId, IERC7726 valuation_, AccountId[] memory accounts, bytes calldata data)
+        external
+        auth
+        returns (ItemId itemId_)
+    {
+        (ShareClassId scId, AssetId assetId) = abi.decode(data, (ShareClassId, AssetId));
+
+        require(poolRegistry.exists(poolId), IPoolRegistry.NonExistingPool(poolId));
```
lemunozm: Thinking about this... given `Holdings.create()` needs to be called under the unlocked pool. We do not need to check for pool existence because it always exists. Probably, we do not need to check for pool existence anywhere given the current architecture precondition.

Would you remove the check? It seems like a basic check, but really it's not needed. It's the `PoolManager` duty from my PoV.

src/Holdings.sol
lemunozm: Discussed in the review thread

wischli (approved):
Stellar work 🚀🚀🚀 Thanks for answering all my nitty-gritty questions! Nothing to complain from my side.

peculiarity (approved):

src/Holdings.sol
lemunozm: Finally, I decided to remove the check. I think we should not add these checks in other contracts if by architecture they're not required right now

src/Holdings.sol
```diff
@@ -0,0 +1,187 @@
...
+        ShareClassId scId;
+        AssetId assetId;
+        IERC7726 valuation;
+        uint128 assetAmount;
+        uint128 assetAmountValue;
+    }
+
+    mapping(PoolId => Item[]) public item;
+    mapping(PoolId => mapping(ShareClassId => mapping(AssetId => ItemId))) public itemId;
+    mapping(PoolId => mapping(ItemId => mapping(uint8 kind => AccountId))) public accountId;
```
mustermeiszer, line 34: QUESTION

This replaces the fixed types (Gains, Loss, etc) with a variable amount, right?

src/Holdings.sol
```diff
@@ -0,0 +1,187 @@
...
+        auth
+        returns (uint128 amountValue)
+    {
+        require(address(valuation_) != address(0), WrongValuation());
+        require(itemId_.index() < item[poolId].length, ItemNotFound());
+
+        Item storage item_ = item[poolId][itemId_.index()];
+        address poolCurrency = address(poolRegistry.currency(poolId));
+
+        amountValue = uint128(valuation_.getQuote(amount, AssetId.unwrap(item_.assetId), poolCurrency));
```
mustermeiszer, line 96: CR

We can not rely on the valuation here. This would need a separate contract to deployed that gives the right past value. Lets just use a fixed `D18` here.

src/Holdings.sol
```diff
@@ -0,0 +1,187 @@
...
```
mustermeiszer, line 116: CR

See above.

ilinzweilin (approved):

src/Holdings.sol
```diff
@@ -0,0 +1,187 @@
...
+        emit ValuationUpdated(poolId, itemId_, valuation_);
+    }
+
+    /// @inheritdoc IItemManager
+    function setAccountId(PoolId poolId, ItemId itemId_, AccountId accountId_) external auth {
+        require(itemId_.index() < item[poolId].length, ItemNotFound());
+
+        accountId[poolId][itemId_][accountId_.kind()] = accountId_;
+
+        emit AccountIdSet(poolId, itemId_, accountId_);
```
mustermeiszer: NIT

Should we allow to retreive the `kind` more easily in the event by making it a separate field?

src/Holdings.sol
```diff
@@ -0,0 +1,187 @@
...
+    }
+
+    /// @inheritdoc IItemManager
+    function itemValue(PoolId poolId, ItemId itemId_) external view returns (uint128 value) {
+        require(itemId_.index() < item[poolId].length, ItemNotFound());
+
+        return item[poolId][itemId_.index()].assetAmountValue;
```
mustermeiszer, line 165: NOTE

Can we also add one that is `itemAmount` so that if one would want to compute the latest value they could 

`let currentValue = IItemManager.itemAmount() * IItemManager.valuation()`

mustermeiszer (commented):
Have a change request, and one note. Otherwise looks great!

src/Holdings.sol
lemunozm: No strong opinion. Maybe we're wasting an extra duplicated byte if we do so?

src/Holdings.sol
lemunozm: Solved above.

src/Holdings.sol
lemunozm: Solved in the sync. Outcome:
- We can have an IERC7726 implementation with transient storage to let the FM to set the price they want.
- Using IERC7726 everywhere allows also to use onChain computed values in the interface easily.

src/Holdings.sol
lemunozm: Yeah you have 256 accounts per ItemId

src/Holdings.sol
lemunozm: Makes sense! Will do

src/Holdings.sol
lemunozm: Maybe it makes sense if we want to index the kind

lemunozm: Added

PR #43, opened 2024-12-12
PoolRegistry changes

lemunozm:
Some additions that are required for the `PoolRegistry` to be used by the `PoolManager`

- Getter methods in the `IPoolRegistry` interface 
- Adding and using `IERC20Metadata` instead of `Currency`
- Using `IShareClassManager` instead of a plain `address`
- Added `addressFor` method to the registry
  - By now, they will be used to store the escrows attached to a pool
- `PoolLocker` using the correct `PoolId` type.
- Fixed some wrong licenses


test/unit/PoolRegistry.t.sol
```diff
@@ -30,37 +31,26 @@ contract PoolRegistryTest is Test {
...
-    function testAuthorization(address additionalWard) public nonZero(additionalWard) notThisContract(additionalWard) {
-        assertEq(registry.wards(address(this)), 1);
-
-        assertEq(registry.wards(additionalWard), 0);
-        registry.rely(additionalWard);
-        assertEq(registry.wards(additionalWard), 1);
-
-        registry.deny(additionalWard);
-        assertEq(registry.wards(additionalWard), 0);
-    }
```
lemunozm, line 42: tech debt to move this into a future `Auth.t.sol` file

src/interfaces/IPoolRegistry.sol
```diff
@@ -19,13 +25,26 @@ interface IPoolRegistry {
...
+    function setAddressFor(PoolId poolId, bytes32 key, address value) external;
+
+    /// @notice TODO
+    function currency(PoolId poolId) external view returns (IERC20Metadata);
+    /// @notice TODO
+    function isAdmin(PoolId poolId, address admin) external view returns (bool);
+    /// @notice TODO
+    function shareClassManager(PoolId poolId) external view returns (IShareClassManager);
+    /// @notice TODO
+    function addressFor(PoolId poolId, bytes32 key) external view returns (address);
```
lemunozm: This doesn't mean we end up this way, but for now, let's use the `poolId` to see if everything fits well. If we need a more general way, we should refactor this into a more general non-pool related `Registry`. Thread: https://kflabs.slack.com/archives/C07PG2EUR9C/p1733943322989829

lemunozm:
@wischli @mustermeiszer @hieronx @ilinzweilin it would be super great if we can add these changes ASAP to unlock other PRs that require them 🫶🙏🏻

test/unit/PoolRegistry.t.sol
```diff
@@ -151,10 +143,19 @@ contract PoolRegistryTest is Test {
         registry.updateCurrency(nonExistingPool, currency);
 
         vm.expectRevert(IPoolRegistry.EmptyCurrency.selector);
-        registry.updateCurrency(poolId, Currency.wrap(address(0)));
+        registry.updateCurrency(poolId, IERC20Metadata(address(0)));
 
-        assertTrue(Currency.unwrap(registry.poolCurrencies(poolId)) != Currency.unwrap(currency));
+        vm.assume(address(registry.poolCurrencies(poolId)) != address(currency));
```
wischli, line 148: Did you really intend to replace an assert with a mock? Now, we don't perform this check anymore. If we truly want to assume, then that should happen earlier in the scope of the test such that foundry doesn't perform computations up to this point if the assumption cannot be fulfilled.

wischli (commented):
AFAICT, one `assert` was incorrectly replaced with an `assume`. Otherwise LGTM!

Curious if it is best-practice to store implementers of interfaces as the interface instead of the plain address, i.e. as `IShareClassManager` instead of plain `address`.

test/unit/PoolRegistry.t.sol
lemunozm: I think this was already an error we had in that test IIUC. 

The `assertTrue` was checked before calling the method `registry.updateCurrency(poolId, currency);`. I understand this check is to avoid registering the currency with the same currency it previously had.

This happens because the test is a fuzzy test (which I do not really understand why, @peculiarity could you explain the reason?). So basically, the previous `assertTrue` could make the test fail if the fuzzer chooses the same value used to register the pool.

lemunozm:
> Curious if it is best-practice to store implementers of interfaces as the interface instead of the plain address, i.e. as IShareClassManager instead of plain address

IMO, it all depends on what kind of information you want to "transfer" through that interface. If all addresses used for share class managers *must always* implement `IShareClassManager`, then that should be reflected in the interface.

Also, I think there are two levels of interfaces: a project-related interface as `IPoolRegistry`, and a general cross-project interface, as `IERC7726`. The first one is intended to be used by consumers who know about this project. The second one can be used by consumers who didn't know about the underlying centrifuge project, and then they should use more general types as `address`.

EDIT. This is my inexperienced way of seeing this; maybe the solidity community is used to another approach.

test/unit/PoolRegistry.t.sol
peculiarity: This should be moved to a separate Auth.t.sol file. Here we test that the registry inherits the Auth interface. We have the same thing done in LP.

src/PoolRegistry.sol
```diff
@@ -51,33 +53,55 @@ contract PoolRegistry is Auth, IPoolRegistry {
...
+        addresses[poolId][key] = value;
+    }
+
+    /// @inheritdoc IPoolRegistry
+    function currency(PoolId poolId) external view returns (IERC20Metadata) {
+        return poolCurrencies[poolId];
+    }
+
+    /// @inheritdoc IPoolRegistry
+    function isAdmin(PoolId poolId, address admin) external view returns (bool) {
```
peculiarity: This is not required. You can get `poolAdmins[poolId][admi]` directly.

src/PoolRegistry.sol
```diff
@@ -51,33 +53,55 @@ contract PoolRegistry is Auth, IPoolRegistry {
...
+        return poolAdmins[poolId][admin];
+    }
+
+    /// @inheritdoc IPoolRegistry
+    function shareClassManager(PoolId poolId) external view returns (IShareClassManager) {
```
peculiarity: Not needed. , you can take `shareClassManagers` directly.

test/unit/PoolRegistry.t.sol
lemunozm: > Here we test that the registry inherits the Auth interface.

This is already tested in each test case when we check:

```solidity
vm.expectRevert(IAuth.NotAuthorized.selector);
```

right?

wischli (approved):

hieronx (approved):

PR #44, opened 2024-12-14
Define getters in the PoolRegistry interface

peculiarity:


peculiarity:
@lemunozm  This is how the getters of the public variables are defined in order to use them. No need to define new methods just to wrap the calls to the public variables.  I forgot to add them in the interface so it is my bad.

lemunozm:
> No need to define new methods just to wrap the calls to the public variables

Oh nice, I didn't know that 👍🏻

src/PoolRegistry.sol
```diff
@@ -80,28 +80,9 @@ contract PoolRegistry is Auth, IPoolRegistry {
...
-    function shareClassManager(PoolId poolId) external view returns (IShareClassManager) {
-        return shareClassManagers[poolId];
-    }
-
-    /// @inheritdoc IPoolRegistry
-    function addressFor(PoolId poolId, bytes32 key) external view returns (address) {
-        return addresses[poolId][key];
+    function updateAddress(PoolId poolId, bytes32 key, address addr) external auth {
+        require(address(shareClassManagers[poolId]) != address(0), NonExistingPool(poolId));
+        require(addr != address(0), EmptyAddress());
```
lemunozm, line 85: The first check, I think it makes sense. But the second one, I think the fund manager should be able to set address(0) in order to "deactivate" that address.

Anyway we'll really see how this will evolve in the coming days. I will merge this. 

Thanks! And enjoy your time off!

lemunozm (approved):

src/PoolRegistry.sol
peculiarity: Yeah, I just put there and didn't think much around it. If you think it is not needed just go ahead and delete it ;)

PR #45, opened 2024-12-17
PoolRegistry renamings

lemunozm:
After some discussions, we've decided that storage should represent the name of the associated getter to later have a good naming when interacting with those getters.

This PR reflects that output.

I've also removed the "pool" from `poolCurrencies` and `poolAdmin` storage names because everything in the `PoolRegistry` is always "per pool".

I've also modified `updateAddress()` to `setAddressFor()`, due [this thread](https://kflabs.slack.com/archives/C07PG2EUR9C/p1734003089245559?thread_ts=1733941850.823459&cid=C07PG2EUR9C). The address can be 0 (as null representation). The `For` part is because it's *for* a key, and because we need to add something to the name to avoid collisions with the `address` keyword

### DIFF
--- a/src/PoolRegistry.sol
+++ b/src/PoolRegistry.sol
@@ -14,46 +14,46 @@ contract PoolRegistry is Auth, IPoolRegistry {
     uint32 public latestId;
 
     mapping(PoolId => bytes) public metadata;
-    mapping(PoolId => IERC20Metadata) public poolCurrencies;
-    mapping(PoolId => IShareClassManager) public shareClassManagers;
-    mapping(PoolId => mapping(address => bool)) public poolAdmins;
-    mapping(PoolId => mapping(bytes32 key => address)) public addresses;
+    mapping(PoolId => IERC20Metadata) public currency;
+    mapping(PoolId => IShareClassManager) public shareClassManager;
+    mapping(PoolId => mapping(address => bool)) public isAdmin;
+    mapping(PoolId => mapping(bytes32 key => address)) public addressFor;
 
     constructor(address deployer) Auth(deployer) {}
 
     /// @inheritdoc IPoolRegistry
-    function registerPool(address admin, IERC20Metadata currency_, IShareClassManager shareClassManager_)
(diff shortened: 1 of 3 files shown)


src/PoolRegistry.sol
```diff
@@ -14,46 +14,46 @@ contract PoolRegistry is Auth, IPoolRegistry {
     uint32 public latestId;
 
     mapping(PoolId => bytes) public metadata;
-    mapping(PoolId => IERC20Metadata) public poolCurrencies;
-    mapping(PoolId => IShareClassManager) public shareClassManagers;
-    mapping(PoolId => mapping(address => bool)) public poolAdmins;
-    mapping(PoolId => mapping(bytes32 key => address)) public addresses;
+    mapping(PoolId => IERC20Metadata) public currency;
```
lemunozm, line 17: I was thinking about our conversation @wischli about changing from `address` to `IERC20Metadata`, and maybe, due to the fact that only ERC7726 actually will really use `.decimals()`, we can just use an `address` as type here.

I'm pretty sure the `shareClassManager` should be an `IShareClassManager`, but not sure about the best approach for `currency`

lemunozm: Let's use by now IERC20Metadata. I just remember you need to call `.decimals()` manually to get the unit of a currency

mustermeiszer (approved):
Like the renaming!

src/PoolRegistry.sol
wischli: > Let's use by now IERC20Metadata. I just remember you need to call `.decimals()` manually to get the unit of a currency

Actually, with the `IERC7726Ext` idea which exposes `getFactor`, I don't need `.decimals()` inside the `SingleShareClass` anymore IF we agree that it's feasible to feed a quote for each `shareClassid` to/from its corresponding `poolCurrency `(poolCurrency, shareClassId)`

wischli (approved):
I prefer this over plural storage names which return single values 🚀

PR #47, opened 2025-01-07
Add exists() to PoolRegistry

lemunozm:
It's true that we can do:

```solidity
require(address(poolRegistry.shareClassManager(poolId)) != address(0))
```

But it seems too complex for a check that we should do everywhere, so I would prefer for the reader to have:
```solidity
require(poolRegistry.exists(poolId));
```

### DIFF
--- a/src/PoolRegistry.sol
+++ b/src/PoolRegistry.sol
@@ -43,8 +43,8 @@ contract PoolRegistry is Auth, IPoolRegistry {
 
     /// @inheritdoc IPoolRegistry
     function updateAdmin(PoolId poolId, address admin_, bool canManage) external auth {
+        require(exists(poolId), NonExistingPool(poolId));
         require(admin_ != address(0), EmptyAdmin());
-        require(address(shareClassManager[poolId]) != address(0), NonExistingPool(poolId));
 
         isAdmin[poolId][admin_] = canManage;
 
@@ -53,7 +53,7 @@ contract PoolRegistry is Auth, IPoolRegistry {
 
     /// @inheritdoc IPoolRegistry
     function updateMetadata(PoolId poolId, bytes calldata metadata_) external auth {
-        require(address(shareClassManager[poolId]) != address(0), NonExistingPool(poolId));
+        require(exists(poolId), NonExistingPool(poolId));
 
         metadata[poolId] = metadata_;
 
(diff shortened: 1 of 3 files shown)


mustermeiszer (approved):

src/PoolRegistry.sol
```diff
@@ -84,4 +84,8 @@ contract PoolRegistry is Auth, IPoolRegistry {
         require(address(shareClassManager[poolId]) != address(0), NonExistingPool(poolId));
         addressFor[poolId][key] = addr;
     }
+
+    function exists(PoolId poolId) external view returns (bool) {
```
peculiarity: **NIT:** I assume you need to use this in some external contract. You can also make the method public so that you can call it within this contract.

You can replace all `require(address(shareClassManager[poolId]) != address(0), NonExistingPool(poolId));`
with: `require(exists(poolId), NonExistingPool(poolId))`

Since this check is a duplicate in all of the `update*` methods, it can be easily extracted into a single modifier.

```
modifier onlyOnExistingPools(PoolId poolId) {
  require(exists(poolId), NonExistingPool(poolId); 
}
```
And an update method would look like smth like:
```
  function updateCurrency(PoolId poolId, IERC20Metadata currency_) external auth onlyOnExistingPools {
        require(address(currency_) != address(0), EmptyCurrency());
        currency[poolId] = currency_;
        emit UpdatedPoolCurrency(poolId, currency_);
    }
  ```

lemunozm: Good idea regarding `require(exists(poolId), NonExistingPool(poolId))`.

I do not remember exactly where, but I remember @hieronx suggesting that modifiers should be used for permission-related stuff and that other checks are preferred to be explicit in order to not overuse modifiers.

I'll go with your first suggestion, but I'm happy to use the modifier later if we decide it fits in a "permission-related" context.

Thanks for the suggestion!

lemunozm: Done

wischli: > I do not remember exactly where, but I remember @hieronx suggesting that modifiers should be used for permission-related stuff and that other checks are preferred to be explicit in order to not overuse modifiers.

I also remember that.

wischli (approved):

lemunozm:
@mustermeiszer @wischli @peculiarity sorry for the noise regarding this 🙏🏻 , but now I'm thinking that maybe it makes more sense to have instead of `exists()` the following:

poolRegistry.ensureExistance(poolId);

Which already dispatch a `NonExistingPool`, avoiding to create that error in each consumer of `PoolRegistry` contract. Is it a solidity pattern @peculiarity?

See https://github.com/centrifuge/pools-internal/pull/48

PR #48, opened 2025-01-07
PoolRegistry: exists() by ensureExistence()

lemunozm:
Minor change to avoid creating a NonExistingPool error each time a contract call `poolRegistry.exists()`

Based on: https://github.com/centrifuge/pools-internal/pull/47

Is this a correct solidity pattern?


lemunozm:
I'm gonna let this unmerged, unless we see this is really needed. Lets use `exists()` by now. It seems not to be a solidity pattern. Also having the error in the interfaces, allows to reuse inner errors, which here is a good point in favor of that pattern

PR #49, opened 2025-01-07
Add `ShareClassId` to `IItemManager`

lemunozm:
Addition of a `ShareClassId` extra param to all methods of IItemManager. A continuation from: https://github.com/centrifuge/pools-internal/pull/42#discussion_r1891806102

### DIFF
--- a/src/Holdings.sol
+++ b/src/Holdings.sol
@@ -28,10 +28,10 @@ struct Item {
 contract Holdings is Auth, IHoldings {
     using MathLib for uint256;
 
-    mapping(PoolId => mapping(ItemId => Item)) public item;
+    mapping(PoolId => mapping(ShareClassId => mapping(ItemId => Item))) public item;
     mapping(PoolId => mapping(ShareClassId => mapping(AssetId => ItemId))) public itemId;
     mapping(PoolId => uint32) lastItemId;
-    mapping(PoolId => mapping(ItemId => mapping(uint8 kind => AccountId))) public accountId;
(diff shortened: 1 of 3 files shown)


lemunozm:
I feel like it increases the complexity @mustermeiszer, not sure if we want to add this. 

We anyways need the mapping you want to avoid from `ItemId` to `AssetId`, so we can carry the `ShareClassId` information there and just avoid the parameter everywhere.

mustermeiszer (approved):
Details clarified on slack! Thanks

src/Holdings.sol
```diff
@@ -129,46 +132,52 @@ contract Holdings is Auth, IHoldings {
...
         require(!item_.assetId.isNull(), ItemNotFound());
 
         item_.valuation = valuation_;
     }
 
     /// @inheritdoc IItemManager
-    function setAccountId(PoolId poolId, ItemId itemId_, AccountId accountId_) external auth {
-        require(!item[poolId][itemId_].assetId.isNull(), ItemNotFound());
+    function setAccountId(PoolId poolId, ShareClassId scId, ItemId itemId_, AccountId accountId_) external auth {
+        require(!item[poolId][scId][itemId_].assetId.isNull(), ItemNotFound());
```
mustermeiszer, line 176: Why only a specific `AccountId` for an item? Should be an array of `AccountId`s

mustermeiszer: @lemunozm

lemunozm: Because each account has encoded the `kind` of the account inside of it, different kinds of accounts are not overwritten:

```solidity
accountId[poolId][itemId_][accountId_.kind()] = accountId_;
```

lemunozm:
Closed by now

PR #53, opened 2025-01-14
Add TransientValuation contract helper

lemunozm:
The `TransientValuation` is an `IERC7726` valuation that can be set by anyone with low cost as a temporal value to be used in methods that require an `IERC7726 valuation` parameter, but the user wants to provide a fixed value instead.

Example:

```solidity
transientValuation.setPrice(0.5);
holdings.increase(transientValuation); // It will call to getQuote() using the provided price.
```

closes POOLS-43

### DIFF
--- a/src/TransientValuation.sol
+++ b/src/TransientValuation.sol
@@ -1,17 +1,16 @@
 // SPDX-License-Identifier: BUSL-1.1
 pragma solidity 0.8.28;
 
-import {IERC7726} from "src/interfaces/IERC7726.sol";
 import {AssetId} from "src/types/AssetId.sol";
 import {D18} from "src/types/D18.sol";
-import {MathLib} from "src/libraries/MathLib.sol";
+
+import {Conversion} from "src/libraries/Conversion.sol";
+
+import {IERC7726} from "src/interfaces/IERC7726.sol";
 import {ITransientValuation} from "src/interfaces/ITransientValuation.sol";
-import {IERC20Metadata} from "src/interfaces/IERC20Metadata.sol";
 
 contract TransientValuation is ITransientValuation {
-    using MathLib for uint256;
-
-    /// Temporal price set and used to obtain the quote.
+    /// @notice Temporal price set and used to obtain the quote.
     D18 public /*TODO: transient*/ price;
 
     /// @inheritdoc ITransientValuation
@@ -21,13 +20,6 @@ contract TransientValuation is ITransientValuation {
 
     /// @inheritdoc IERC7726
     function getQuote(uint256 baseAmount, address base, address quote) external view returns (uint256 quoteAmount) {
-        uint8 baseDecimals = IERC20Metadata(base).decimals();
-        uint8 quoteDecimals = IERC20Metadata(quote).decimals();
-
-        if (baseDecimals == quoteDecimals) {
-            return price.mulUint256(baseAmount);
-        }
-
-        return price.mulUint256(MathLib.mulDiv(baseAmount, 10 ** quoteDecimals, 10 ** baseDecimals));
+        return Conversion.convertWithPrice(baseAmount, base, quote, price);
     }
 }
--- a/src/TransientValuation.sol
+++ b/src/TransientValuation.sol
@@ -1,0 +1,33 @@
+// SPDX-License-Identifier: BUSL-1.1
+pragma solidity 0.8.28;
+
+import {IERC7726} from "src/interfaces/IERC7726.sol";
+import {AssetId} from "src/types/AssetId.sol";
+import {D18} from "src/types/D18.sol";
+import {MathLib} from "src/libraries/MathLib.sol";
+import {ITransientValuation} from "src/interfaces/ITransientValuation.sol";
+import {IERC20Metadata} from "src/interfaces/IERC20Metadata.sol";
+
+contract TransientValuation is ITransientValuation {
+    using MathLib for uint256;
+
+    /// Temporal price set and used to obtain the quote.
+    D18 public /*TODO: transient*/ price;
+
+    /// @inheritdoc ITransientValuation
+    function setPrice(D18 price_) external {
+        price = price_;
+    }
+
+    /// @inheritdoc IERC7726
+    function getQuote(uint256 baseAmount, address base, address quote) external view returns (uint256 quoteAmount) {
+        uint8 baseDecimals = IERC20Metadata(base).decimals();
+        uint8 quoteDecimals = IERC20Metadata(quote).decimals();
+
+        if (baseDecimals == quoteDecimals) {
+            return price.mulUint256(baseAmount);
+        }
+
+        return price.mulUint256(MathLib.mulDiv(baseAmount, 10 ** quoteDecimals, 10 ** baseDecimals));
+    }
+}
--- a/src/interfaces/ITransientValuation.sol
+++ b/src/interfaces/ITransientValuation.sol
@@ -4,7 +4,8 @@ pragma solidity 0.8.28;
 import {D18} from "src/types/D18.sol";
 import {IERC7726} from "src/interfaces/IERC7726.sol";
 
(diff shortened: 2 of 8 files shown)


src/TransientValuation.sol
```diff
@@ -0,0 +1,33 @@
...
+    /// @inheritdoc IERC7726
+    function getQuote(uint256 baseAmount, address base, address quote) external view returns (uint256 quoteAmount) {
+        uint8 baseDecimals = IERC20Metadata(base).decimals();
+        uint8 quoteDecimals = IERC20Metadata(quote).decimals();
+
+        if (baseDecimals == quoteDecimals) {
+            return price.mulUint256(baseAmount);
+        }
+
+        return price.mulUint256(MathLib.mulDiv(baseAmount, 10 ** quoteDecimals, 10 ** baseDecimals));
```
lemunozm: FYI. Given a price, we need to do these maths, which are not super simple and error-prone. This is why I would use `getQuote` as much as possible instead of handling prices directly to keep this complexity apart from the business logic

lemunozm: If we mandatorily need a price (i.e. `navPerShare` or similar @wischli), maybe we can move this inner logic into a lib and have a method like:

```solidity
Conversion.convert(D18 price, baseAmount, base, quote) => quoteAmount;
```

I can do it in this PR if this helps you William, with the `ShareClassManager`'s prices computation

src/TransientValuation.sol
```diff
@@ -0,0 +1,33 @@
...
+pragma solidity 0.8.28;
+
+import {IERC7726} from "src/interfaces/IERC7726.sol";
+import {AssetId} from "src/types/AssetId.sol";
+import {D18} from "src/types/D18.sol";
+import {MathLib} from "src/libraries/MathLib.sol";
+import {ITransientValuation} from "src/interfaces/ITransientValuation.sol";
+import {IERC20Metadata} from "src/interfaces/IERC20Metadata.sol";
+
+contract TransientValuation is ITransientValuation {
```
mustermeiszer, line 12: NIT

Abstract?

src/TransientValuation.sol
mustermeiszer: I like that for the prices!

mustermeiszer (commented):
@lemunozm will you make the PoolManger inherent from this contract in the main PoolManager PR?

src/TransientValuation.sol
lemunozm: As I envisioned this. It is a separate contract whose address you choose to pass through the IERC7726 parameter. We only need to deploy it once ever. And if we find some bug in the price calculation, we can upload it again without modifying the `PoolManager` entry point.

mustermeiszer: I am not a friend of that. We need to track these contracts across all deployments. Much easier for us to just use the pool manager itself. 

@hieronx and @peculiarity interested in your take on this.

lemunozm: This contract is a utility. But it's not "mandatory" for the system. 

I feel like we will have many more implementations of `IERC7726`, such as linearInterest, onChain computations, oracles, ... This would be one more. Also, it feels weird that the own `PoolManager` is a `IERC7726` implementation, just for offering this utility. 

Anyways listen to your thoughts

src/TransientValuation.sol
wischli: Great proposal!

src/TransientValuation.sol
lemunozm: Also, I feel the usage will be difficult to understand:

```solidity
poolManager.setPrice(0.5) // what does it mean to set a price for the PoolManager?
poolManager.increase(poolManager) // why passing the own poolManager?
```

Instead, if other contract:

```solidity
transientValuation.setPrice(0.5)
poolManager.increase(transientValuation)
```

src/interfaces/ITransientValuation.sol
```diff
@@ -0,0 +1,15 @@
...
+
+import {D18} from "src/types/D18.sol";
+import {IERC7726} from "src/interfaces/IERC7726.sol";
+
+/// @notice An IERC7726 valuation that allows to set a price that is only valid for the current transaction.
+/// NOTE: Do not use it if the valuation lifetime is longer than the transaction.
+interface ITransientValuation is IERC7726 {
+    /// @notice Set the price for the valuation.
+    /// The price is the 1 amount of base denominated in quote.
+    /// i.e: if base BTC and quote USDC, the price is 90_000 USDC per BTC
```
wischli, line 12: Nit: Actually it would be `90_000_000_000` USDC since USDC denomination is 6 decimals, i.e. `1_000_000`

test/unit/TransientValuation.t.sol
```diff
@@ -0,0 +1,58 @@
...
+    }
+}
+
+contract TestPortfolio is Test {
+    address c18 = address(new C18());
+    address c6 = address(new C6());
+    TransientValuation valuation = new TransientValuation();
+
+    function testSameDecimals() public {
+        valuation.setPrice(d18(2, 1)); //2.0
```
wischli, line 29: Question: How did this compile?

wischli (commented):
Much in favor of this approach! Thanks for working on this already

test/unit/TransientValuation.t.sol
lemunozm: I added a new method to `d18` to support fractions.

src/interfaces/ITransientValuation.sol
lemunozm: With the example, I'm trying to show the direction the price is set, in "human" representation. The internal representation of that price, in `d18`, should it should be 90_000 for the integer part and 0 for the fractional part. But that should depend on how d18 works.

src/TransientValuation.sol
peculiarity: Usually protocols have Core contracts and Periphery contract so I think we can create a separate repo with these.
Check  [Uniswap V3 Periphery](https://github.com/Uniswap/v3-periphery)

src/TransientValuation.sol
lemunozm: Added the conversion library

src/TransientValuation.sol
mustermeiszer: Doesn't really solve how we track the contracts over deployments. That is a real nightmare as far as I know, but if it is not an issue. Fine for me. 

We could still multicall that, right?

lemunozm: I think this contract, as Georgi says, can be conceptually placed out from the core contracts required for us to track the deployment. 

IMO, I feel like we need to pass through that process to understand the implications and the scope well (at least for me). 

If you agree, we can merge this as it is, to keep concepts split as much as possible, at least for the first iterations, and if we find deployment issues, I can modify it to merge it into `PoolManager`. WDYT?

lemunozm: To summarize the PROS/CONS of having this `TransientValuation` as a new contract instead of being part of the `PoolManager`.

- PROs
  - Split behaviors. This `TransientValuation` is just another valuation used as a utility (check example above).
  - If we find a bug in the conversion, we do not need to re-upload the whole PoolManager
  - PoolManager will already be too big. We have a max bytecode size we can attach to it.
  - Same gas cost when it's used.
- CONS
  - New contract to deploy once.
  - Say to the FM the address of the `TransientContract` to let use it.
 
What are your other CONS @mustermeiszer? Why do you say it's more aesthetic to have it inside the `PoolManager` when they are different things?

lemunozm: Decision made in the sync: Leave the contract to be a separate contract. At least during the PoC developing stage, until we decide more in the future how all contracts will be organized.

Using `create2` allows deterministic address deployments

peculiarity (approved):

mustermeiszer (approved):

PR #58, opened 2025-01-15
PoolRegistry required changes

lemunozm:
Requires changes to `PoolRegistry`

- We need `createPool()` to allow us to set a `PoolId` from the outside: [thread](https://github.com/centrifuge/pools-internal/pull/36#discussion_r1905330508)
- Added `allowInvestorAsset()`, previous `allowAsset()` in `ShareClassManager`
  - @peculiarity I was thinking about the name, and I really prefer "allow" because it means a binary answer: yes or no. All configure methods out there configure the content of the element. They receive a bunch of parameters or a `Config` struct. Here we are not configuring the Asset (that would happen in `AssetRegistry` probably), instead allowing its usage.
- Added tests for events.
- Rename `updateCurrency` (previously suggested by myself in the arch diagram) by `setCurrency` following this [rule](https://kflabs.slack.com/archives/C07PG2EUR9C/p1734003089245559?thread_ts=1733941850.823459&cid=C07PG2EUR9C)

### DIFF
--- a/src/PoolRegistry.sol
+++ b/src/PoolRegistry.sol
@@ -2,7 +2,7 @@
 pragma solidity 0.8.28;
 
 import {Auth} from "src/Auth.sol";
-import {PoolId} from "src/types/PoolId.sol";
+import {PoolId, PoolIdLib} from "src/types/PoolId.sol";
 import {AssetId} from "src/types/AssetId.sol";
 import {MathLib} from "src/libraries/MathLib.sol";
 import {IPoolRegistry} from "src/interfaces/IPoolRegistry.sol";
@@ -11,8 +11,6 @@ import {IShareClassManager} from "src/interfaces/IShareClassManager.sol";
 
 contract PoolRegistry is Auth, IPoolRegistry {
     using MathLib for uint256;
-
-    uint32 public latestId;
 
     mapping(PoolId => bytes) public metadata;
     mapping(PoolId => IERC20Metadata) public currency;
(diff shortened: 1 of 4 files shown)


src/PoolRegistry.sol
```diff
@@ -11,29 +12,29 @@ import {IShareClassManager} from "src/interfaces/IShareClassManager.sol";
...
     constructor(address deployer) Auth(deployer) {}
 
     /// @inheritdoc IPoolRegistry
-    function registerPool(address admin_, IERC20Metadata currency_, IShareClassManager shareClassManager_)
-        external
-        auth
-        returns (PoolId poolId)
-    {
+    function registerPool(
+        uint32 localPoolId,
```
peculiarity: Question: Where do you get this `id` from ? Is this some random value that the pool creator provides?

lemunozm: Yes, exactly

src/PoolRegistry.sol
```diff
@@ -48,16 +49,26 @@ contract PoolRegistry is Auth, IPoolRegistry {
 
         isAdmin[poolId][admin_] = canManage;
 
-        emit UpdatedPoolAdmin(poolId, admin_);
+        emit UpdatedAdmin(poolId, admin_, canManage);
     }
 
     /// @inheritdoc IPoolRegistry
-    function updateMetadata(PoolId poolId, bytes calldata metadata_) external auth {
+    function allowInvestorAsset(PoolId poolId, AssetId assetId, bool isAllowed) external auth {
```
peculiarity, line 56: Note: IMO `allowInvestorAsset` is not a good name. You are calling a method that is supposedly allowing an asset but if the `isAllowed` value is false you are actually disallowing it which is completely the opposite of what the name method implies.  `configureAsset`, `modifyAsset` 

Imagine having a door that opens when you press the handle down and closes when you press the handle up ( actually I've seen such doors ) and you want to represent that programmatically. So you will create a method like:
`open(HandlePosition.UP)` and the door will actually closes. A better name would be `moveDoor(..)` because it is neutral and the behavior depends on the passed arguments.

lemunozm: It is precisely the same when you call a method as `enableX(bool)`. When you pass a `false`, you're disabling. This pattern is used everywhere. Basically, it is a binary verb with a binary parameter.

lemunozm: If you still are against that. I'm ok with calling the method `setInvestorAssetAllowance()`. But as I explained in the description, `configure` or `modify` are things that should probably exist in the `AssetRegistry` not here.

mustermeiszer: Is there any reason to not be specific and have 2 methods instead of relying on boolean-parameters? Not a fan of boolean anyways ^^

peculiarity: OK, let's leave it `allowInvestorAsset`. 

@mustermeiszer  Another method = more bytecode during deployment, more expensive but this is something that we don't care. According to some it might bloat the code, I like it because it shows the intention without getting into details the value of the passed arguments.
Using a single method with a switch param, increase the calldata on each call which is more expensive a little bit. Even if `bool` is  a single byte the arguments are always padded to 32. For each zero byte you pay 4 gas and for non-zero byte you pay 16 gas. 
Frankly I can go with @lemunozm suggestion and close the topic.

lemunozm: I agree here with @peculiarity, two methods means duplicate all paths everywhere, here, in holdings, in poolManager... When from the user point of view it is a binary/opposite action.

lemunozm: I would go with this approach for the PoC which is faster to factorize and handle (even for the UX/SDK). If we decide to be more explicit once we have everything well tied. I'm not against that.

mustermeiszer: All good arguments. Fine for me.

peculiarity (approved):

mustermeiszer (approved):

PR #62, opened 2025-01-20
ShareClassManager: use prices instead of factors

lemunozm:
This PR comes from this [thread](https://github.com/centrifuge/pools-internal/pull/35#discussion_r1916273277)

## Description 

There are currently two minor issues if we handle prices as factors that we ideally do not want to have them:

1. Values used to convert from `assetToPool` are no prices, they are factors. For example, for CFG -> USDC we will have a price of `0.3` but a factor of `0.0000000000003` because CFG is 18 decimals and usdc 6. This can be difficult to handle from a UX perspective if they want to show the "human" readable prices as `0.3`, or perform inverses.

2. The `D18` type treated as a factor does not have enough resolution to convert a 27-decimal currency into a 6-decimal currency. Even for a 18 dec -> 6 dec conversion, we can lose significant precision. `D18` should be used only to hold prices, no factors.

This PR avoids factor usage to use prices instead.

### DIFF
--- a/src/SingleShareClass.sol
+++ b/src/SingleShareClass.sol
@@ -8,6 +8,7 @@ import {IERC7726Ext} from "src/interfaces/IERC7726.sol";
 import {IShareClassManager} from "src/interfaces/IShareClassManager.sol";
 import {ISingleShareClass} from "src/interfaces/ISingleShareClass.sol";
 import {MathLib} from "src/libraries/MathLib.sol";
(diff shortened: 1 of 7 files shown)


lemunozm:
Probably not required if going with #70

lemunozm:
Closing this in favor of: https://github.com/centrifuge/pools-internal/pull/70

PR #65, opened 2025-01-21
OneToOneValuation contract

lemunozm:
This small valuation utility will help us in those cases where we assume the conversion prices are 1:1. Highly assumed for the PoC.

Instead of using `TransientValuation` before each call, we can use this contract address in those cases where we know the price is 1:1. Avoiding transient issues and improving the usage.

### DIFF
--- a/src/IdentityValuation.sol
+++ b/src/IdentityValuation.sol
@@ -7,10 +7,15 @@ import {Conversion} from "src/libraries/Conversion.sol";
 
 import {IERC7726, IERC7726Ext} from "src/interfaces/IERC7726.sol";
 import {IIdentityValuation} from "src/interfaces/IIdentityValuation.sol";
+import {IAssetManager} from "src/interfaces/IAssetManager.sol";
 
-contract IdentityValuation is IIdentityValuation {
+import {BaseValuation} from "src/BaseValuation.sol";
+
+contract IdentityValuation is BaseValuation, IIdentityValuation {
+    constructor(IAssetManager assetManager, address deployer) BaseValuation(assetManager, deployer) {}
+
     /// @inheritdoc IERC7726
     function getQuote(uint256 baseAmount, address base, address quote) external view returns (uint256 quoteAmount) {
-        return Conversion.convertWithPrice(baseAmount, base, quote, d18(1e18));
+        return Conversion.convertWithPrice(baseAmount, _getDecimals(base), _getDecimals(quote), d18(1e18));
     }
 }
--- a/src/IdentityValuation.sol
+++ b/src/IdentityValuation.sol
@@ -1,0 +1,16 @@
+// SPDX-License-Identifier: BUSL-1.1
+pragma solidity 0.8.28;
+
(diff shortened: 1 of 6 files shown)


test/unit/IdentityValuation.t.sol
```diff
@@ -0,0 +1,39 @@
...
+    }
+}
+
+contract C6 {
+    function decimals() external pure returns (uint8) {
+        return 6;
+    }
+}
+
+contract TestIden is Test {
```
peculiarity: Nit: 
```suggestion
contract TestOneToOneConversion is Test {
```

src/IdentityValuation.sol
```diff
@@ -0,0 +1,16 @@
...
+pragma solidity 0.8.28;
+
+import {D18, d18} from "src/types/D18.sol";
+
+import {Conversion} from "src/libraries/Conversion.sol";
+
+import {IERC7726, IERC7726Ext} from "src/interfaces/IERC7726.sol";
+import {IIdentityValuation} from "src/interfaces/IIdentityValuation.sol";
+
+contract IdentityValuation is IIdentityValuation {
```
peculiarity: Nit: I completely agree where you come with the naming of `identity`  f(x) -> x  meaning that given the `baseAmount`, you will get the same value but I find it like a mental exercise to figure it out.
I propoes ` OneToOneValuation` or `ParityValuation` or `OneToOnePegValuation`. It's more explicit IMO about the intention.

test/unit/IdentityValuation.t.sol
lemunozm: I left wrote unfinished xD

src/IdentityValuation.sol
lemunozm: I really think the identity word from the mathematical world fits pretty well here. I saw this pattern in a lot of places, indicating that the effect of applying this valuation does not change the final result. Which is exactly the behavior I want to emphasize.

peculiarity: That's the thing. It gets too mathematical :). Would love to hear thoughts from @mustermeiszer , @hieronx , @wischli .

wischli: I get your point @peculiarity. AFAIK, it is common to use math terms in finance, so that should also hold true for DeFi. As a mathematician, I am biased and favor @lemunozm's "Identity".

onnovisser: My 2 cents, I understand the reasoning and it makes sense to want to name it "identity", but I do like the explicit obviousness of "OneToOneValuation"

mustermeiszer: You fight the out. ^^ 

No opinion

lemunozm: Do not be neutral and give a tiebreaker 😛.

lemunozm: In the sync OneToOne was the winner

mustermeiszer (approved):

wischli (approved):

PR #67, opened 2025-01-22
Refactor: Adapt Holdings to new requirements

lemunozm:
PR opened again from: https://github.com/centrifuge/pools-internal/pull/63


mustermeiszer (approved):
Re approve

src/Holdings.sol
```diff
@@ -56,149 +51,126 @@ contract Holdings is Auth, IHoldings {
...
-        return item[poolId][itemId_.index()].assetAmount;
+        emit AccountIdSet(poolId, scId, assetId, accountId_);
     }
 
-    /// @inheritdoc IItemManager
-    function valuation(PoolId poolId, ItemId itemId_) external view returns (IERC7726) {
-        require(itemId_.index() < item[poolId].length, ItemNotFound());
+    /// @inheritdoc IHoldings
+    function value(PoolId poolId, ShareClassId scId, AssetId assetId) external view returns (uint128 value_) {
+        Holding storage holding_ = holding[poolId][scId][assetId];
```
wischli, line 155: Comment: I initially wanted to flag this to rather use memory but apparently that is more costly if not accessed multiple times due to  the overhead of copying data ([ref](https://ethereum.stackexchange.com/a/66413)).

Just wanted to raise this here as a knowledge share in case it was not only news to me.

wischli (approved):

src/Holdings.sol
lemunozm: Thanks!!

PR #68, opened 2025-01-23
BaseValuation: abstracting the way decimals are getting from AssetRegistry

lemunozm:
To summarize an issue we have with the ERC7726. [slack thread](https://kflabs.slack.com/archives/C04GJQAM9P0/p1737624389397889?thread_ts=1737454926.871929&cid=C04GJQAM9P0)

We're expecting that the address passed as base and quote are `ERC20`. At least in the current `TransientValuation` , `IdentityValuation` & `Conversion` follows that. Nevertheless, given that the `AssetManager` is an `ERC6909`, those addresses should actually be the identification of the tokens in `ERC6909`. All of our `ERC7726` implementations that work with tokens from the AssetManager should keep an AssetManager dependency to ask for the decimals registered there.

Nevertheless, we use the pool currency as `ERC20`, so we need mix currency types in the `ERC7726` representation.

Solution inspired in https://github.com/euler-xyz/euler-price-oracle/blob/master/src/adapter/BaseAdapter.sol

Latest approach:
- Using ERC6909 for any asset in the system. See: https://kflabs.slack.com/archives/C07PG2EUR9C/p1737630902558699

### DIFF
--- a/src/BaseERC7726.sol
+++ b/src/BaseERC7726.sol
@@ -6,21 +6,34 @@ import {D18} from "src/types/D18.sol";
 
 import {Conversion} from "src/libraries/Conversion.sol";
 
-import {IERC7726} from "src/interfaces/IERC7726.sol";
+import {IBaseERC7726} from "src/interfaces/IBaseERC7726.sol";
 import {IERC20Metadata} from "src/interfaces/IERC20Metadata.sol";
+import {IAssetManager} from "src/interfaces/IAssetManager.sol";
 
-interface IAssetManager {
-    function decimals(uint256 tokenId) external view returns (uint8);
-}
+import {Auth} from "src/Auth.sol";
 
-abstract contract BaseERC7726 is IERC7726 {
-    uint160 private constant ASSET_MANAGER_TOKEN = type(uint64).max;
+abstract contract BaseERC7726 is Auth, IBaseERC7726 {
+    /// @notice Max raw value an AssetId can be.
+    uint160 private constant MAX_ASSET_MANAGER_TOKEN_REPRESENTATION = type(uint64).max;
 
-    /// @notice Temporal price set and used to obtain the quote.
+    /// @notice AssetManager dependency.
     IAssetManager public assetManager;
 
+    constructor(IAssetManager assetManager_, address deployer) Auth(deployer) {
+        assetManager = assetManager_;
+    }
+
+    /// @inheritdoc IBaseERC7726
+    function file(bytes32 what, address data) external auth {
+        if (what == "assetManager") assetManager = IAssetManager(data);
+        else revert FileUnrecognizedWhat();
+
+        emit File(what, data);
+    }
+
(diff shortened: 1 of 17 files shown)


src/BaseERC7726.sol
```diff
@@ -0,0 +1,43 @@
...
+
+import {Conversion} from "src/libraries/Conversion.sol";
+
+import {IBaseERC7726} from "src/interfaces/IBaseERC7726.sol";
+import {IERC20Metadata} from "src/interfaces/IERC20Metadata.sol";
+import {IAssetManager} from "src/interfaces/IAssetManager.sol";
+
+import {Auth} from "src/Auth.sol";
+
+abstract contract BaseERC7726 is Auth, IBaseERC7726 {
```
peculiarity: IMO might be good to name our abstract contract / interfaces that gets directly implemented something more agnostic. `BaseOracle` for example because you don't care if it implements the `IBaseERC7726` or if tomorrow it will implement some other interface that will have more methods.

lemunozm: Agree. Although the name "oracle" is too specific, we have valuations that are not oracles. I will go with `BaseValuation`, the most generic I can find.

peculiarity (approved):

lemunozm:
How did it get merged with just one review?

test/unit/BaseValuation.t.sol
```diff
@@ -0,0 +1,41 @@
...
+import {BaseValuation} from "src/BaseValuation.sol";
+import {IAuth} from "src/interfaces/IAuth.sol";
+import {IBaseValuation} from "src/interfaces/IBaseValuation.sol";
+import {IAssetManager} from "src/interfaces/IAssetManager.sol";
+
+contract BaseValuationImpl is BaseValuation {
+    constructor(IAssetManager assetManager, address deployer) BaseValuation(assetManager, deployer) {}
+
+    function getQuote(uint256 baseAmount, address base, address quote) external view returns (uint256 quoteAmount) {}
+}
```
wischli, line 18: Am I missing something or is `BaseValuationImpl` redundant?

test/unit/BaseValuation.t.sol
```diff
@@ -0,0 +1,41 @@
...
+
+contract TestFile is Test {
```
wischli, line 20: I am sure you are already aware of the missing `getQuote` tests due to the unexpected early merge @lemunozm ?!

test/unit/BaseValuation.t.sol
lemunozm: This is needed to have an instantiable contract. `BaseValuation` is abstract

test/unit/BaseValuation.t.sol
lemunozm: It's fine. `getQuote` behavior should be implemented by another valuation child of `BaseValuation`. `BaseValuation` does not touch that method at all.

PR #70, opened 2025-01-23
Draft Idea. Avoid storing prices in SCM

lemunozm:
## Context of the problem

Handling prices directly is difficult. (i.e What precision they need? Should be prices or factors? How they interact with plain uint128 of unknown representation?). In fact, the own [ERC7726](https://eips.ethereum.org/EIPS/eip-7726#rationale) spec hides the price concept and gives a rationale of why:

> The spec doesn’t include a getPrice function because it is rarely needed on-chain, and it would be a decimal number of difficult representation.

So we should avoid calling to a `getPrice` or `getFactor` as much as we can, given their bad side-effects

The current implementation has a couple of "issues" we should fix:

- Values used to convert from assetToPool are no prices, they are factors. For example, for CFG -> USDC we will have a price of 0.3 but a factor of 0.0000000000003 because CFG is 18 decimals and usdc 6. This can be difficult to handle from a UX perspective if they want to show the "human" readable prices as 0.3, or perform inverses.

- The D18 type treated as a factor does not have enough resolution to convert a 27-decimal currency into a 6-decimal currency. Even for a 18 dec -> 6 dec conversion, we can lose significant precision. D18 should be used only to hold prices, no factors.

## Current approach

Instead of finding a way of handling the price as smoother as possible, we can just get rid of the problem if we avoid storing prices/factors and saving other information to be able to claim correctly later. See the code (currently compiling the contract). Still WIP the redemption flow.

### DIFF
--- a/src/SingleShareClass.sol
+++ b/src/SingleShareClass.sol
@@ -18,12 +18,10 @@ struct Epoch {
 }
 
 struct EpochRatio {
-    /// @dev Percentage of approved deposits
-    D18 depositRatio;
+    uint128 approvedDepositAssetAmount;
+    uint128 approvedDepositPoolAmount;
     /// @dev Percentage of approved redemptions
     D18 redeemRatio;
-    /// @dev Price of one pool per asset token set during redeem approval
-    D18 depositAssetToPoolQuote;
     /// @dev Price of one pool per asset token set during redeem approval
     D18 redeemAssetToPoolQuote;
     /// @dev Price of one pool per share class token set during deposit flow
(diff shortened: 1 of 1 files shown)


mustermeiszer:
> The D18 type treated as a factor does not have enough resolution to convert a 27-decimal currency into a 6-decimal currency. Even for a 18 dec -> 6 dec conversion, we can lose significant precision. D18 should be used only to hold prices, no factors.

We allow for at most 18 decimals on the CV side. We could remove that, but I doubt there will be any relevant token with more than 18 decimals.

wischli:
Closing due to superseding of #76

PR #72, opened 2025-01-30
feat: Introduce AssetManager

peculiarity:
Implementation of https://github.com/centrifuge/pools-internal/issues/30

- How do we handle the duplicate errors? ( Not same name but rather different name for the same thing, defined in different interfaces ).
    
    i.e. `AssetManager.EmptyAsset`, `IPoolRegistry.EmptyAsset` , `IHoldings.WrongAssetId`
    
    Probably we can rename the errors to be more context specific. 
    
    `IHoldings.WrongAssetId` → `IHoldings.AssetNotAllowedAsHolding`
    
    `IPoolRegistry.EmptyAsset`  → `IPoolRegsitry.AssetNotAllowedAsInvestment`
    
- `Burn` has simple functionality based on this [answer](https://kflabs.slack.com/archives/C07PG2EUR9C/p1738075508058179) . Only `auth` caller can call the method  and they don't necessarily have to be the owner or has allowance or be an operator.
- `authTransferFrom` - not particularly sure if it should be part of the `IERC6909Fungible` interface or just a separate interface.
- `test/mock/Mock*` - used in CV that way - consistency reasoning
- **Question**: Do we want to allow `mint` with `amount=0` ? For some reason when implementing the ERC6909 for the NFT part, we have this check.
- **Idea**: Refactor the tests so that can test ERC6909 abstract contract in isolation ( that means that probably we need some Test contract that will just inherit from it. That way we will remove the duplication of testing `transfer`  `approve` and so on for both `ERC6909fungible` and `ERC6909NFT`. IF some custom logic for either of these is implemented, then we test that custom logic only.


src/AssetManager.sol
```diff
@@ -0,0 +1,39 @@
...
+        if (asset_.decimals == 0) {
+            asset_.decimals = decimals_;
+        }
+
+        emit NewAssetEntry(assetId_, name_, symbol_, asset_.decimals);
+    }
+
+    function isRegistered(AssetId assetId) external view returns (bool) {
+        return asset[assetId].decimals > 0;
+    }
```
lemunozm, line 39: Question. What would happen if we had assets with no decimals? Would this even exist?

src/ERC6909/ERC6909Fungible.sol
```diff
@@ -0,0 +1,39 @@
...
+        }
+    }
+
+    /// @inheritdoc IERC6909Fungible
+    function authTransferFrom(address spender, address sender, address receiver, uint256 tokenId, uint256 amount)
+        external
+        auth
+        returns (bool)
+    {
+        return _transferFrom(spender, sender, receiver, tokenId, amount);
```
lemunozm: I think it should be `_transfer()` instead because the call is already auth. We do not need approval/owner protections done in `_transferFrom()`

src/interfaces/ERC6909/IERC6909TotalSupplyExt.sol
```diff
@@ -0,0 +1,11 @@
+// SPDX-License-Identifier: AGPL-3.0-only
+pragma solidity 0.8.28;
+
+/// @notice Extension of ERC6909 Standard for tracking total supply
+interface IERC6909TotalSupplyExt {
```
lemunozm, line 5: Does the standard separate this from `ERC6909`? If not, why not inside `IERC6909Fungible` interface?

lemunozm (commented):
Fast review, just few comments but looks good!

src/AssetManager.sol
peculiarity: You cannot have an asset with decimal less than 2. If you check the changes for the CV, we support only tokens between 2 and 18.

lemunozm:
> Question: Do we want to allow mint with amount=0 ? For some reason when implementing the ERC6909 for the NFT part, we have this check.

The standard says something in this regard?

> Idea ...

I like the idea!

> How do we handle the duplicate errors?

I think there is no need to handle. The own interface acts as namespace, right?

src/interfaces/ERC6909/IERC6909TotalSupplyExt.sol
peculiarity: According to the standard this is a separate [Ext](https://eips.ethereum.org/EIPS/eip-6909#token-supply-extension). The only reason for putting it as a separate interface.  When you generate `interfaceId` it takes into account all the method. So you can easily check if only this method is supported.  I think I'm missing modification of `supportInterface`. That should be updated.

peculiarity:
@lemunozm 
> The standard says something in this regard?

I checked it out and couldn't see anywhere that being mentioned. Actually `mint` is not defined by the standard at all. The only thing that is mentioned is that `mint` can happen either through `mint` method or via `Transfer` from `address(0)`

> I think there is no need to handle. The own interface acts as namespace, right?

What bothers me is that we are handling the same thing. I have the gut feeling that either this should be handled at a higher level or I dont know :)

src/AssetManager.sol
lemunozm: Yes, but note that this `AssetManager` could contain more kinds of assets in the future, other than CV investor assets.

Right now, for example, it also contains national currencies (not an issue either).

Let's keep it as it is! just a thought about scalability

peculiarity: Yeah, I thought about the  national currency  but no matter what you use, you need decimals.  We used `MIN_DECIMAL=1` in CV and it was flagged by auditors. The lower the number, the higher the precision loss is. Even for currencies, we would like to have at least 2.

lemunozm:
I understand your concern about errors. But I would leave the errors as they are, scoped by interfaces, at least by now, until they have a solid structure.

wischli:
> * **Idea**: Refactor the tests so that can test ERC6909 abstract contract in isolation

I'm also in favor of this. You can decide on your own of you want to include it in this PR or a follow-up one!

src/AssetManager.sol
wischli: Maybe it makes sense to change the above requirement to `> 1` then?

wischli:
> * **Question**: Do we want to allow `mint` with `amount=0` ? For some reason when implementing the ERC6909 for the NFT part, we have this check.

IMO it doesn't really matter but both impls should be in sync of course. So I propose to add this requriement here as well.

src/AssetManager.sol
```diff
@@ -0,0 +1,39 @@
...
+        require(!assetId_.isNull(), IncorrectAssetId());
+        Asset storage asset_ = asset[assetId_];
+        asset_.name = name_;
+        asset_.symbol = symbol_;
+
+        if (asset_.decimals == 0) {
+            asset_.decimals = decimals_;
+        }
+
+        emit NewAssetEntry(assetId_, name_, symbol_, asset_.decimals);
```
wischli, line 33: Nit: Maybe we want to emit two different events: One for updating an existing asset and one for a newly created one?

src/AssetManager.sol
```diff
@@ -0,0 +1,39 @@
+// SPDX-License-Identifier: AGPL-3.0-only
```
wischli: CR

```suggestion
// SPDX-License-Identifier: BUSL-1.1
```

src/ERC6909/ERC6909Fungible.sol
```diff
@@ -0,0 +1,39 @@
+// SPDX-License-Identifier: AGPL-3.0-only
```
wischli: CR: 
```suggestion
// SPDX-License-Identifier: BUSL-1.1
```

src/interfaces/ERC6909/IERC6909Fungible.sol
```diff
@@ -0,0 +1,36 @@
+// SPDX-License-Identifier: AGPL-3.0-only
```
wischli: CR
```suggestion
// SPDX-License-Identifier: BUSL-1.1
```

src/interfaces/ERC6909/IERC6909TotalSupplyExt.sol
```diff
@@ -0,0 +1,11 @@
+// SPDX-License-Identifier: AGPL-3.0-only
```
wischli: CR
```suggestion
// SPDX-License-Identifier: BUSL-1.1
```

src/types/AssetId.sol
```diff
@@ -11,8 +15,12 @@ function addr(AssetId assetId) pure returns (address) {
     return address(uint160(AssetId.unwrap(assetId)));
 }
 
-function raw(AssetId assetId) pure returns (uint64) {
+function raw(AssetId assetId) pure returns (uint128) {
     return AssetId.unwrap(assetId);
 }
 
-using {isNull, addr, raw} for AssetId global;
+function toUint256(AssetId assetId) pure returns (uint256) {
```
wischli: Nit: Wondering if this wrapper is really necessary or whether we can just `assetId.raw().toUint256()`.

test/mock/MockAssetManager.sol
```diff
@@ -0,0 +1,8 @@
+// SPDX-License-Identifier: AGPL-3.0-only
```
wischli: CR:
```suggestion
// SPDX-License-Identifier: BUSL-1.1
```

test/unit/AssetManager.t.sol
```diff
@@ -0,0 +1,110 @@
+// SPDX-License-Identifier: AGPL-3.0-only
```
wischli: CR: 
```suggestion
// SPDX-License-Identifier: BUSL-1.1
```

test/unit/ERC6909Fungible.t.sol
```diff
@@ -0,0 +1,165 @@
+// SPDX-License-Identifier: AGPL-3.0-only
```
wischli: CR
```suggestion
// SPDX-License-Identifier: BUSL-1.1
```

test/unit/libraries/AssetIdLib.t.sol
```diff
@@ -0,0 +1,22 @@
+// SPDX-License-Identifier: AGPL-3.0-only
```
wischli: CR: 
```suggestion
// SPDX-License-Identifier: BUSL-1.1
```

wischli (changes requested):
Really solid PR! My only complaint are the incorrect license headers which is why I am blocking the PR for now.

src/AssetManager.sol
lemunozm: This is related to [this thread](https://kflabs.slack.com/archives/C07PG2EUR9C/p1738099449058139?thread_ts=1738051827.456969&cid=C07PG2EUR9C) about the idempotent property of this method. Linked just for reference.

If we finally go with the idempotent implementation, I would call it just `AssetRegistered` that can work for both: creating and updating.

src/libraries/AssetIdLib.sol
```diff
@@ -0,0 +1,11 @@
...
+pragma solidity 0.8.28;
+
+import {AssetId} from "src/types/AssetId.sol";
+import {MathLib} from "src/libraries/MathLib.sol";
+
+library AssetIdLib {
+    function asAssetId(address asset) internal pure returns (AssetId) {
+        return AssetId.wrap(MathLib.toUint128(uint256(uint160(asset))));
+    }
+}
```
lemunozm: NOT. Why not have this method on the opposite side: `AssetId.fromAddress(address)` in order to avoid importing and handling a new library?

src/types/AssetId.sol
lemunozm: I also think method is not needed, right? `raw()` is already interpreted as `uint256` without casting it

src/AssetManager.sol
peculiarity: I'm just checking if the decimals are set. It does't really matter if it's 1 or 0

src/libraries/AssetIdLib.sol
peculiarity: You cannot do that. First argument MUST be of type `assetId` if you want to attach it to the `AssetId` type.

lemunozm: That's right!

And why not then `assetFromAddress(address)` leaving in the same `AssetId.sol` file?

My point is that I think it's a bit annoying to have, for each type, its counterpart `TypeLib.sol` file. We basically have a super small logic split in two files, which makes things a bit harder to read for reviewers of the codebase.

src/ERC6909/ERC6909Fungible.sol
peculiarity: This is the way how it is done in CV. Even though the method is Auth, it checks all of these prerequisites. Probably I have to clarify this with Frederik. 
[This](https://github.com/centrifuge/vaults-internal/blob/main/src/token/Tranche.sol#L117) is the `authTransferFrom` and [this](https://github.com/centrifuge/vaults-internal/blob/main/src/token/ERC20.sol#L100) is the _transferFrom defined in the ERC20

lemunozm: Note that the `AssetManager` could not be an ERC6909; it is just an ERC6909 to be able to inspect them later in the UX, but it should not be affected too much by it.

The CV `autTransferFrom` is used for ERC20s, but here we own the `AssetManager` and we can simplify some things.

Given that, I would say we do not need a `spender` parameter. If this breaks the ERC6909Fungible "rules", I would move this method then to `AssetManager.sol` file.

src/interfaces/IAssetManager.sol
```diff
@@ -1,7 +1,34 @@
...
+
+    /// @notice             A getter function to get an Asset based on AssetId
+    function asset(AssetId assetId) external view returns (bytes memory name, bytes32 symbol, uint8 decimals);
+
+    /// @notice             Checks whether an asset is registered or nto
+    function isRegistered(AssetId assetId) external view returns (bool);
+
+    /// @notice             Used to retrieve the decimals for an asset
+    /// @dev                address is used but the value corresponds to a AssetId
+    function decimals(address assetId) external view returns (uint8);
```
lemunozm: Should `decimals(`) not be implemented as part of some ERC6909 interface? along with `name()` and `symbol()`?

peculiarity: You are actually correct. There is the `Metadata Extension` so this will be an additional [interface](https://eips.ethereum.org/EIPS/eip-6909#metadata-extension). Will change

lemunozm (approved):
Thanks for solving my two latest comments!

LGTM! I think we have everything we need 🚀

src/interfaces/ERC6909/IERC6909MetadataExt.sol
```diff
@@ -0,0 +1,16 @@
...
+pragma solidity 0.8.28;
+
+interface IERC6909MetadataExt {
+    /// @notice             Used to retrieve the decimals of an asset
+    /// @dev                address is used but the value corresponds to a AssetId
+    function decimals(address assetId) external view returns (uint8);
+
+    /// @notice             Used to retrieve the name of an asset
+    /// @dev                address is used but the value corresponds to a AssetId
+    function name(address assetId) external view returns (bytes memory);
```
lemunozm: Should be `string` instead of `bytes` following the standard. And `symbol` is a `string` too.

lemunozm (changes requested):
Sorry for the blocker! Just want to ensure this change is not missing before merging it

peculiarity:
@lemunozm  It's not only that.. the input params are wrong type as well. I totally forgot to check the precise definition :sweat_smile: . Fixing it ..

src/ERC6909/ERC6909Fungible.sol
lemunozm: @peculiarity if we leave the `transferFrom` with the allowance & owner checks, then we're not able to call this method from the `PoolManager`, because we do not have an escrow to approve an infinite balance as we did in CV.

We decided not to go with the escrow by now, which means we need to bypass those checks

src/BaseValuation.sol
```diff
@@ -30,6 +30,6 @@ abstract contract BaseValuation is Auth, IBaseValuation {
 
     /// @notice Obtain the correct decimals given an asset address
     function _getDecimals(address asset) internal view returns (uint8) {
-        return assetManager.decimals(uint160(asset));
+        return assetManager.decimals(uint256(uint160(asset)));
```
wischli: Can you elaborate on why this change was necessary? I'm curios :)

src/ERC6909/ERC6909Fungible.sol
wischli: Seems like this topic is still unresolved? Let's discuss in the sync!

src/AssetManager.sol
```diff
@@ -2,18 +2,24 @@
...
 import {Auth} from "src/Auth.sol";
-import {AssetId, addrToAssetId} from "src/types/AssetId.sol";
+import {MathLib} from "src/libraries/MathLib.sol";
+import {AssetId} from "src/types/AssetId.sol";
 import {IAssetManager} from "src/interfaces/IAssetManager.sol";
 import {ERC6909Fungible} from "src/ERC6909/ERC6909Fungible.sol";
 import {IERC6909MetadataExt} from "src/interfaces/ERC6909/IERC6909MetadataExt.sol";
 
 contract AssetManager is ERC6909Fungible, IAssetManager {
+    using MathLib for uint256;
```
lemunozm, line 13: Should not `AssetManager` implement from `IERC6909MetadataExt`?

wischli (approved):
Soft approving despite the pending `transferFrom` vs `transfer` discussion which can also be resolved in a follow-up PR IMO. Great job!

src/BaseValuation.sol
peculiarity: It wasn't compliant with the standard. The input [param](https://eips.ethereum.org/EIPS/eip-6909#decimals) must be uint256.

wischli: Awesome thanks for the quick reply!

src/AssetManager.sol
peculiarity: Here is the thing. 
1. IMO the implementation should implement its interface explicitly and nothing else. You use the interface to interact with the contract, right . Of course  there are some scenarios where you can extend existing implementations ( like the abstract classes we currently have ).
2. In your `BaseValuation` contract  you import the `IAssetManager`. If the `AssetManager` extends directly the `IERC6909MetadataExt` that means, the methods won't be seen. Either you have to import the `AssetManager` itself which will grow the size of the `BaseValuation` contract or import the `IERC6909MEtadataExt` and use something like `IERC6909MetadataExt(manager)` but in that case you expose the implementation...something that uses IERC6909.

src/BaseValuation.sol
lemunozm: But an smaller number automatically fits in a bigger one, right?

peculiarity: Absolutely. There isn't anything to worry about.

lemunozm: But, then is not needed or I'm missing something? Just curious

src/AssetManager.sol
lemunozm: Ah ok, it extends from the IAssetManager.

Then I need the same for `IERC6909Fungible`. I need `IAssetManager` implementing `IERC6909Fungible` to be able to call `mint/burn/authTransfer` from `PoolManager`

src/BaseValuation.sol
peculiarity: Well It will cast it anyways but that was it is more explicit. When you read you know that a uint256 is passed. Otherwise it will look like you are passing `uint160`. I don't think there is a penalty and IMO it improvesd reading,.

src/AssetManager.sol
peculiarity: Absolutely. Let me push. You are right!

lemunozm (approved):
Let's merge this!!

PR #73, opened 2025-01-30
Accounting contract

### LINKED ISSUE #71: feat: Contract `Accounting`
# First Draft

```solidity
// SPDX-License-Identifier: MIT
pragma solidity ^0.8.24;

contract DoubleEntryBookkeeping {
    struct Account {
        uint256 totalDebit;
        uint256 totalCredit;
        bool isDebitNormal;
        uint8 accountKind;
        bytes metadata;
        bool lockedDebit;
        bool lockedCredit;
        u64 lastUpdated;
    }

    mapping(PoolId => mapping(AccountId => Account)) public accounts;

    // Transient storage slots
    uint256 private constant ACTIVE_FLAG_SLOT = 0;
    uint256 private constant TOTAL_DEBITS_SLOT = 1;
    uint256 private constant TOTAL_CREDITS_SLOT = 2;
    uint256 private constant LOCKED_POOL_ID_SLOT  = 3;


    modifier onlyActiveTransaction() {
        require(tload(ACTIVE_FLAG_SLOT) == 1, "Transaction not active");
        _;
    }

    event TransactionStarted(PoolId poolId);
    event TransactionCompleted(PoolId poolId, bool success);
    event AccountUpdated(AccountId indexed account, uint256 amount);

   function create(AccountId id, bool isDebitNormal, bytes metadata) external {
         // ensure account is not init already
         
        // init account with timestamp and method arguments
   }

    function lock(PoolId poolId) external {
        require(tload(ACTIVE_FLAG_SLOT) == 0, "Transaction already active");
        tstore(ACTIVE_FLAG_SLOT, 1);
        tstore(TOTAL_DEBITS_SLOT, 0);
        tstore(TOTAL_CREDITS_SLOT, 0);
        tstore(LOCKED_POOL_ID_SLOT, poolId)

        emit TransactionStarted(poolId);
    }

   function updateEntry(AccountId debitAccount, AccountId creditAccount, uint256 amount) {
         debit(debitAccount, amount);
         credit(creditAccount, amount):
   }

    function debit(AccountId account, uint256 amount) external onlyActiveTransaction {
        Account storage acc = accounts[account];
        require(!acc.lockedDebit, "Debit operations locked");
        
        acc.totalDebit += amount;
        tstore(TOTAL_DEBITS_SLOT, tload(TOTAL_DEBITS_SLOT) + amount);
        emit AccountUpdated(AccountId, true, amount);
    }

    function credit(AccountId account, uint256 amount) external onlyActiveTransaction {
        Account storage acc = accounts[account];
        require(!acc.lockedCredit, "Credit operations locked");
        
        acc.totalCredit += amount;
        tstore(TOTAL_CREDITS_SLOT, tload(TOTAL_CREDITS_SLOT) + amount);
        emit AccountUpdated(AccountId, false, amount);
    }

    function unlock() external onlyActiveTransaction {
        bool balanced = tload(TOTAL_DEBITS_SLOT) == tload(TOTAL_CREDITS_SLOT);
        
        // Clear transient storage
        tstore(ACTIVE_FLAG_SLOT, 0);
        tstore(TOTAL_DEBITS_SLOT, 0);
        tstore(TOTAL_CREDITS_SLOT, 0);
        tstore(LOCKED_POOL_ID_SLOT, 0);

        if (!balanced) {
            revert("Unbalanced transaction");
        }
        emit TransactionCompleted(tload(LOCKED_POOL_ID_SLOT));
    }

    // Account management functions
    function lockDebit(address account) external {
        accounts[account].lockedDebit = true;
    }

    function unlockDebit(address account) external {
        accounts[account].lockedDebit = false;
    }

    function lockCredit(address account) external {
        accounts[account].lockedCredit = true;
    }

    function unlockCredit(address account) external {
        accounts[account].lockedCredit = false;
    }

    function updateAccountMetadata(address account, bytes calldata metadata) external {
        accounts[account].metadata = metadata;
    }

    function setNormalBalanceType(address account, bool isDebitNormal) external {
        accounts[account].isDebitNormal = isDebitNormal;
    }

    function getAccountValue(address account) public view returns (int256) {
        Account storage acc = accounts[account];
        
        if (acc.isDebitNormal) {
            // For debit-normal accounts: Value = Total Debit - Total Credit
            return int256(acc.totalDebit) - int256(acc.totalCredit);
        } else {
            // For credit-normal accounts: Value = Total Credit - Total Debit
            return int256(acc.totalCredit) - int256(acc.totalDebit);
        }
    }

    // Helper functions for transient storage access
    function tload(uint256 slot) private pure returns (uint256 value) {
        assembly {
            value := tload(slot)
        }
    }

    function tstore(uint256 slot, uint256 value) private pure {
        assembly {
            tstore(slot, value)
        }
    }
}
```

onnovisser:
Starting point / first draft for the Accounting contract. Tests are still to do. Wanted to open a draft PR as early as possible to focus the discussion around it. 

Questions from my side:

- @mustermeiszer In your draft you had a `totalDebit` and `totalCredit` for each account. I turned into a single `balance` property, but maybe you had a specific reason for it that I missed.
- I made the `balance` of an account an `int256`, because I'm assuming that a balance can theoretically be negative. Is that right, or should that not be allowed and can it be a uint128 instead?
- `getAccountValue` returns negative values for positive balances on credit-normal accounts. Should it?

closes: #71

### DIFF
--- a/src/Accounting.sol
+++ b/src/Accounting.sol
@@ -10,8 +10,8 @@ import {Auth} from "src/Auth.sol";
 contract Accounting is Auth, IAccounting {
     mapping(PoolId => mapping(AccountId => Account)) public accounts;
 
-    uint128 private /*TODO: transient*/ _debited;
-    uint128 private /*TODO: transient*/ _credited;
+    uint128 public /*TODO: transient*/ debited;
+    uint128 public /*TODO: transient*/ credited;
     uint256 private /*TODO: transient*/ _transactionId;
     PoolId private /*TODO: transient*/ _currentPoolId;
 
@@ -19,17 +19,41 @@ contract Accounting is Auth, IAccounting {
 
     /// @inheritdoc IAccounting
     function updateEntry(AccountId creditAccount, AccountId debitAccount, uint128 value) external auth {
-        require(PoolId.unwrap(_currentPoolId) != 0, PoolLocked());
-        _debit(debitAccount, value);
-        _credit(creditAccount, value);
-        emit Entry(_currentPoolId, _transactionId, creditAccount, debitAccount, value);
+        addDebit(debitAccount, value);
+        addCredit(creditAccount, value);
+    }
+
+    /// @dev Debits an account. Increase the value of debit-normal accounts, decrease for credit-normal ones.
+    function addDebit(AccountId account, uint128 value) public auth {
+        require(!_currentPoolId.isNull(), AccountingLocked());
+
+        Account storage acc = accounts[_currentPoolId][account];
+        require(acc.lastUpdated != 0, AccountDoesNotExists());
+
+        acc.totalDebit += value;
+        debited += value;
+        acc.lastUpdated = uint64(block.timestamp);
+        emit Debit(_currentPoolId, _transactionId, account, value);
+    }
+
+    /// @dev Credits an account. Decrease the value of debit-normal accounts, increase for credit-normal ones.
+    function addCredit(AccountId account, uint128 value) public auth {
+        require(!_currentPoolId.isNull(), AccountingLocked());
+
+        Account storage acc = accounts[_currentPoolId][account];
+        require(acc.lastUpdated != 0, AccountDoesNotExists());
+
+        acc.totalCredit += value;
+        credited += value;
+        acc.lastUpdated = uint64(block.timestamp);
+        emit Credit(_currentPoolId, _transactionId, account, value);
     }
 
     /// @inheritdoc IAccounting
     function unlock(PoolId poolId) external auth {
-        require(PoolId.unwrap(_currentPoolId) == 0, PoolAlreadyUnlocked());
-        _debited = 0;
-        _credited = 0;
+        require(PoolId.unwrap(_currentPoolId) == 0, AccountingAlreadyUnlocked());
+        debited = 0;
+        credited = 0;
         /// @dev Include the previous transactionId in case there's multiple transactions in one block
         _transactionId = uint256(keccak256(abi.encodePacked(poolId, block.timestamp, _transactionId)));
         _currentPoolId = poolId;
@@ -37,53 +61,31 @@ contract Accounting is Auth, IAccounting {
 
     /// @inheritdoc IAccounting
     function lock() external auth {
-        require(_debited == _credited, Unbalanced());
+        require(debited == credited, Unbalanced());
         _currentPoolId = PoolId.wrap(0);
     }
 
     /// @inheritdoc IAccounting
-    function createAccount(PoolId poolId, AccountId account, bool isDebitNormal, bytes calldata metadata)
-        external
-        auth
-    {
+    function createAccount(PoolId poolId, AccountId account, bool isDebitNormal) external auth {
         require(accounts[poolId][account].lastUpdated == 0, AccountExists());
-        accounts[poolId][account] = Account(0, metadata, isDebitNormal, uint64(block.timestamp));
+        accounts[poolId][account] = Account(0, 0, isDebitNormal, uint64(block.timestamp), "");
     }
 
-    /// @dev Debits an account. Increase the balance of debit-normal accounts, decrease for credit-normal ones.
-    function _debit(AccountId account, uint128 value) internal {
-        Account storage acc = accounts[_currentPoolId][account];
-
-        if (acc.isDebitNormal) {
-            acc.balance += int256(uint256(value));
-        } else {
-            acc.balance -= int256(uint256(value));
-        }
-        _debited += value;
-        acc.lastUpdated = uint64(block.timestamp);
-    }
-
-    /// @dev Credits an account. Decrease the balance of debit-normal accounts, increase for credit-normal ones.
-    function _credit(AccountId account, uint128 value) internal {
-        Account storage acc = accounts[_currentPoolId][account];
-
-        if (acc.isDebitNormal) {
-            acc.balance -= int256(uint256(value));
-        } else {
-            acc.balance += int256(uint256(value));
-        }
-        _credited += value;
-        acc.lastUpdated = uint64(block.timestamp);
+    function updateAccountMetadata(PoolId poolId, AccountId account, bytes calldata metadata) external auth {
+        Account storage acc = accounts[poolId][account];
+        acc.metadata = metadata;
     }
 
     /// @inheritdoc IAccounting
-    function getAccountValue(PoolId poolId, AccountId account) public view returns (int256) {
+    function getAccountValue(PoolId poolId, AccountId account) public view returns (int128) {
         Account storage acc = accounts[poolId][account];
 
         if (acc.isDebitNormal) {
-            return acc.balance;
+            // For debit-normal accounts: Value = Total Debit - Total Credit
+            return int128(acc.totalDebit) - int128(acc.totalCredit);
         } else {
-            return -acc.balance;
+            // For credit-normal accounts: Value = Total Credit - Total Debit
+            return int128(acc.totalCredit) - int128(acc.totalDebit);
         }
     }
 }
--- a/src/Accounting.sol
+++ b/src/Accounting.sol
@@ -9,6 +9,7 @@ import {Auth} from "src/Auth.sol";
 
 contract Accounting is Auth, IAccounting {
     mapping(PoolId => mapping(AccountId => Account)) public accounts;
+    mapping(PoolId => uint128) private _transactionCounter;
 
     uint128 public /*TODO: transient*/ debited;
     uint128 public /*TODO: transient*/ credited;
@@ -48,8 +49,8 @@ contract Accounting is Auth, IAccounting {
         require(PoolId.unwrap(_currentPoolId) == 0, AccountingAlreadyUnlocked());
         debited = 0;
         credited = 0;
-        // Include the previous transactionId in case there's multiple transactions in one block
-        _transactionId = uint256(keccak256(abi.encodePacked(poolId, block.timestamp, _transactionId)));
+        _transactionCounter[poolId]++;
+        _transactionId = uint256(keccak256(abi.encodePacked(poolId, block.timestamp, _transactionCounter[poolId])));
         _currentPoolId = poolId;
     }
 
--- a/src/Accounting.sol
+++ b/src/Accounting.sol
@@ -15,7 +15,7 @@ contract Accounting is Auth, IAccounting {
     bytes32 private /*TODO: transient*/ _transactionId;
     PoolId private /*TODO: transient*/ _currentPoolId;
 
-    constructor(address deployer) Auth(deployer) {    }
+    constructor(address deployer) Auth(deployer) {}
 
     /// @inheritdoc IAccounting
     function addDebit(AccountId account, uint128 value) public auth {
--- a/src/Accounting.sol
+++ b/src/Accounting.sol
@@ -9,14 +9,13 @@ import {Auth} from "src/Auth.sol";
 
 contract Accounting is Auth, IAccounting {
     mapping(PoolId => mapping(AccountId => Account)) public accounts;
-    mapping(PoolId => uint128) private _transactionCounter;
 
     uint128 public /*TODO: transient*/ debited;
     uint128 public /*TODO: transient*/ credited;
-    uint256 private /*TODO: transient*/ _transactionId;
+    bytes32 private /*TODO: transient*/ _transactionId;
     PoolId private /*TODO: transient*/ _currentPoolId;
 
-    constructor(address deployer) Auth(deployer) {}
+    constructor(address deployer) Auth(deployer) {    }
 
     /// @inheritdoc IAccounting
     function addDebit(AccountId account, uint128 value) public auth {
@@ -45,12 +44,11 @@ contract Accounting is Auth, IAccounting {
     }
 
     /// @inheritdoc IAccounting
-    function unlock(PoolId poolId) external auth {
+    function unlock(PoolId poolId, bytes32 transactionId) external auth {
         require(PoolId.unwrap(_currentPoolId) == 0, AccountingAlreadyUnlocked());
         debited = 0;
         credited = 0;
-        _transactionCounter[poolId]++;
-        _transactionId = uint256(keccak256(abi.encodePacked(poolId, block.timestamp, _transactionCounter[poolId])));
+        _transactionId = transactionId;
         _currentPoolId = poolId;
     }
 
--- a/src/Accounting.sol
+++ b/src/Accounting.sol
@@ -1,0 +1,89 @@
+// SPDX-License-Identifier: BUSL-1.1
+pragma solidity 0.8.28;
+
+import {IAccounting} from "src/interfaces/IAccounting.sol";
+import {AccountId} from "src/types/AccountId.sol";
+import {PoolId} from "src/types/PoolId.sol";
+import {PoolLocker} from "src/PoolLocker.sol";
+import {Auth} from "src/Auth.sol";
+
+contract Accounting is Auth, IAccounting {
+    mapping(PoolId => mapping(AccountId => Account)) public accounts;
+
+    uint128 private /*TODO: transient*/ _debited;
+    uint128 private /*TODO: transient*/ _credited;
+    uint256 private /*TODO: transient*/ _transactionId;
+    PoolId private /*TODO: transient*/ _currentPoolId;
+
+    constructor(address deployer) Auth(deployer) {}
+
+    /// @inheritdoc IAccounting
+    function updateEntry(AccountId creditAccount, AccountId debitAccount, uint128 value) external auth {
+        require(PoolId.unwrap(_currentPoolId) != 0, PoolLocked());
+        _debit(debitAccount, value);
+        _credit(creditAccount, value);
+        emit Entry(_currentPoolId, _transactionId, creditAccount, debitAccount, value);
+    }
+
+    /// @inheritdoc IAccounting
+    function unlock(PoolId poolId) external auth {
+        require(PoolId.unwrap(_currentPoolId) == 0, PoolAlreadyUnlocked());
+        _debited = 0;
+        _credited = 0;
+        /// @dev Include the previous transactionId in case there's multiple transactions in one block
+        _transactionId = uint256(keccak256(abi.encodePacked(poolId, block.timestamp, _transactionId)));
+        _currentPoolId = poolId;
+    }
+
+    /// @inheritdoc IAccounting
+    function lock() external auth {
+        require(_debited == _credited, Unbalanced());
+        _currentPoolId = PoolId.wrap(0);
+    }
+
+    /// @inheritdoc IAccounting
+    function createAccount(PoolId poolId, AccountId account, bool isDebitNormal, bytes calldata metadata)
+        external
+        auth
+    {
+        require(accounts[poolId][account].lastUpdated == 0, AccountExists());
+        accounts[poolId][account] = Account(0, metadata, isDebitNormal, uint64(block.timestamp));
+    }
+
+    /// @dev Debits an account. Increase the balance of debit-normal accounts, decrease for credit-normal ones.
+    function _debit(AccountId account, uint128 value) internal {
+        Account storage acc = accounts[_currentPoolId][account];
+
+        if (acc.isDebitNormal) {
+            acc.balance += int256(uint256(value));
+        } else {
+            acc.balance -= int256(uint256(value));
+        }
+        _debited += value;
+        acc.lastUpdated = uint64(block.timestamp);
+    }
+
+    /// @dev Credits an account. Decrease the balance of debit-normal accounts, increase for credit-normal ones.
+    function _credit(AccountId account, uint128 value) internal {
+        Account storage acc = accounts[_currentPoolId][account];
+
+        if (acc.isDebitNormal) {
+            acc.balance -= int256(uint256(value));
+        } else {
+            acc.balance += int256(uint256(value));
+        }
+        _credited += value;
+        acc.lastUpdated = uint64(block.timestamp);
+    }
+
+    /// @inheritdoc IAccounting
+    function getAccountValue(PoolId poolId, AccountId account) public view returns (int256) {
+        Account storage acc = accounts[poolId][account];
+
+        if (acc.isDebitNormal) {
+            return acc.balance;
+        } else {
+            return -acc.balance;
+        }
+    }
+}
--- a/src/Accounting.sol
+++ b/src/Accounting.sol
@@ -65,8 +65,7 @@ contract Accounting is Auth, IAccounting {
     }
 
     function setAccountMetadata(PoolId poolId, AccountId account, bytes calldata metadata) external auth {
-        Account storage acc = accounts[poolId][account];
-        acc.metadata = metadata;
+        accounts[poolId][account].metadata = metadata;
     }
 
     /// @inheritdoc IAccounting
--- a/src/Accounting.sol
+++ b/src/Accounting.sol
@@ -18,7 +18,7 @@ contract Accounting is Auth, IAccounting {
 
     constructor(address deployer) Auth(deployer) {}
 
-/// @inheritdoc IAccounting
+    /// @inheritdoc IAccounting
     function addDebit(AccountId account, uint128 value) public auth {
         require(!_currentPoolId.isNull(), AccountingLocked());
 
--- a/src/Accounting.sol
+++ b/src/Accounting.sol
@@ -17,18 +17,12 @@ contract Accounting is Auth, IAccounting {
 
     constructor(address deployer) Auth(deployer) {}
 
-    /// @inheritdoc IAccounting
-    function updateEntry(AccountId creditAccount, AccountId debitAccount, uint128 value) external auth {
-        addDebit(debitAccount, value);
-        addCredit(creditAccount, value);
-    }
-
-    /// @dev Debits an account. Increase the value of debit-normal accounts, decrease for credit-normal ones.
+/// @inheritdoc IAccounting
     function addDebit(AccountId account, uint128 value) public auth {
         require(!_currentPoolId.isNull(), AccountingLocked());
 
         Account storage acc = accounts[_currentPoolId][account];
-        require(acc.lastUpdated != 0, AccountDoesNotExists());
+        require(acc.lastUpdated != 0, AccountDoesNotExist());
 
         acc.totalDebit += value;
         debited += value;
@@ -36,12 +30,12 @@ contract Accounting is Auth, IAccounting {
         emit Debit(_currentPoolId, _transactionId, account, value);
     }
 
-    /// @dev Credits an account. Decrease the value of debit-normal accounts, increase for credit-normal ones.
+    /// @inheritdoc IAccounting
     function addCredit(AccountId account, uint128 value) public auth {
         require(!_currentPoolId.isNull(), AccountingLocked());
 
         Account storage acc = accounts[_currentPoolId][account];
-        require(acc.lastUpdated != 0, AccountDoesNotExists());
+        require(acc.lastUpdated != 0, AccountDoesNotExist());
 
         acc.totalCredit += value;
         credited += value;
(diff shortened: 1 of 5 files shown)


src/types/AccountId.sol
```diff
@@ -4,8 +4,12 @@ pragma solidity 0.8.28;
...
 
-function kind(AccountId accountId) pure returns (uint8) {
-    return uint8(AccountId.unwrap(accountId) & 0x000000FF);
+function kind(AccountId accountId_) pure returns (uint8) {
+    return uint8(AccountId.unwrap(accountId_) & 0x000000FF);
+}
+
+function accountId(uint24 id, uint8 kind_) pure returns (AccountId) {
+    return AccountId.wrap(id << 8 | kind_);
 }
```
lemunozm: Note for checking that `id << 8` does not discard the highest 8 bytes before convert this into a `uint32`. I remember having a issue with that, but maybe I'm wrong.

If exists, we need change this to `uint32(id) << 8` instead.

src/Accounting.sol
```diff
@@ -0,0 +1,89 @@
...
+        /// @dev Include the previous transactionId in case there's multiple transactions in one block
+        _transactionId = uint256(keccak256(abi.encodePacked(poolId, block.timestamp, _transactionId)));
+        _currentPoolId = poolId;
+    }
+
+    /// @inheritdoc IAccounting
+    function lock() external auth {
+        require(_debited == _credited, Unbalanced());
+        _currentPoolId = PoolId.wrap(0);
+    }
```
lemunozm: @mustermeiszer could you reminds us in which cases we can have this check failing in the system? AFAIK, from the current implementation, it's impossible to obtain a mismatch here.

Should `credit()` or `debit()` be public instead to allow the FM to have full control?

src/interfaces/IAccounting.sol
```diff
@@ -0,0 +1,52 @@
...
+    function updateEntry(AccountId creditAccount, AccountId debitAccount, uint128 value) external;
+
+    /// @notice Sets the pool for the coming transaction.
+    function unlock(PoolId poolId) external;
+
+    /// @notice Closes the transaction and checks if the entries are balanced.
+    function lock() external;
+
+    /// @notice Creates an account.
+    function createAccount(PoolId poolId, AccountId account, bool isDebitNormal, bytes calldata metadata) external;
```
lemunozm: Probably here should be:

```solidity
function createAccount(PoolId poolId, AccountId account, bool isDebitNormal) external;

function setMetadata(PoolId poolId, AccountId account, bytes calldata metadata) external;
```

Making the metadata optional and allowing to update it even after the account creation.

lemunozm (commented):
Amazing first PR @onnovisser 🚀, really good job for your first contract!

I will do a proper review later

src/Accounting.sol
```diff
@@ -0,0 +1,89 @@
...
+
+    /// @dev Debits an account. Increase the balance of debit-normal accounts, decrease for credit-normal ones.
+    function _debit(AccountId account, uint128 value) internal {
+        Account storage acc = accounts[_currentPoolId][account];
+
+        if (acc.isDebitNormal) {
+            acc.balance += int256(uint256(value));
+        } else {
+            acc.balance -= int256(uint256(value));
+        }
```
lemunozm: ~~Regarding your question about this. I think balance should be always positive, and we do not need this if/else here. Later the `getAccountValue()` will return a negative if `isDebitNormal`.~~

~~Another impl could be leaving the if/else here but removing it from `getAccountValue()`. IIUC, when adding debit to a credit account, `balance` will be negative, and later `getAccountValue()` will transform that to positive again when it shouldn'~~

src/Accounting.sol
```diff
@@ -0,0 +1,89 @@
...
+import {IAccounting} from "src/interfaces/IAccounting.sol";
+import {AccountId} from "src/types/AccountId.sol";
+import {PoolId} from "src/types/PoolId.sol";
+import {PoolLocker} from "src/PoolLocker.sol";
+import {Auth} from "src/Auth.sol";
+
+contract Accounting is Auth, IAccounting {
+    mapping(PoolId => mapping(AccountId => Account)) public accounts;
+
+    uint128 private /*TODO: transient*/ _debited;
```
peculiarity: Why private? I think if you have `uint128 transient  debited` then it's fine.

src/Accounting.sol
```diff
@@ -0,0 +1,89 @@
...
+    uint128 private /*TODO: transient*/ _credited;
+    uint256 private /*TODO: transient*/ _transactionId;
+    PoolId private /*TODO: transient*/ _currentPoolId;
+
+    constructor(address deployer) Auth(deployer) {}
+
+    /// @inheritdoc IAccounting
+    function updateEntry(AccountId creditAccount, AccountId debitAccount, uint128 value) external auth {
+        require(PoolId.unwrap(_currentPoolId) != 0, PoolLocked());
```
peculiarity: You can check `ShareClassId` or `AssetId` and there is `isNull` method assigned to each type. We can have similar for `PoolId`. You will end up with something like:
```require(!currentPoolid.isNull, PoolLocked)```

File: src/Accounting.sol [hunk 7]
```diff
@@ -0,0 +1,89 @@
...
+            acc.balance -= int256(uint256(value));
+        } else {
+            acc.balance += int256(uint256(value));
+        }
+        _credited += value;
+        acc.lastUpdated = uint64(block.timestamp);
+    }
+
+    /// @inheritdoc IAccounting
+    function getAccountValue(PoolId poolId, AccountId account) public view returns (int256) {
```
wischli: Nit: Since we decided on `uint128` values, this should be `int128` or `uint128` depending on the decision of negative vs non-negative.

src/Accounting.sol
```diff
@@ -0,0 +1,89 @@
...
+
+    /// @dev Credits an account. Decrease the balance of debit-normal accounts, increase for credit-normal ones.
+    function _credit(AccountId account, uint128 value) internal {
+        Account storage acc = accounts[_currentPoolId][account];
+
+        if (acc.isDebitNormal) {
+            acc.balance -= int256(uint256(value));
+        } else {
+            acc.balance += int256(uint256(value));
+        }
```
wischli: Nit: You can actually use the ternary operator known from JS for in such cases:
```suggestion
        acc.balance += acc.isDebitNormal ? -int256(uint256(value) : int256(uint256(value);
```

src/interfaces/IAccounting.sol
```diff
@@ -0,0 +1,52 @@
...
+    /// @notice Dispatched when the debit and credit side do not match at the end of a transaction.
+    error Unbalanced();
+
+    /// @notice Dispatched when trying to create an account that already exists.
+    error AccountExists();
+
+    /// @notice
+    struct Account {
+        int256 balance;
+        bytes metadata;
```
wischli: I want to open the discussion of providing some upper bound on the number of metadata bytes for an account since `Account` is used as storage value. Or is this irrelevant @peculiarity? 

I can imagine `bytes32` to suffice as it corresponds to 32 ASCII characters.

wischli: On another nitpicky note: By rearranging the fields of this struct, we can reduce the storage footprint. Since Solidity lays out struct fields in 256-bit “slots” in the order they are declared. Each slot can hold up to 32 bytes. Fields that fit into the remaining space of a slot will be packed together until the slot is full or the next field is too large. A more efficient layout would be 

```solidity
struct Account {
    int256 balance;
    bool isDebitNormal;     // 1 byte
    uint64 lastUpdated;     // 8 bytes
    bytes metadata;         // dynamic
}
```

wischli (commented):
Awesome work @onnovisser :rocket: :rocket: ! It's truly incredible how quickly you pulled this out with such an excellent quality already. I am bowing for you! A few unimportant comments from me:

src/Accounting.sol
onnovisser: yes that was on my mind as well

src/Accounting.sol
lemunozm: Yes, having a PoolId == 0 means the Accounting is lock in this context. So the returned error name can express that. Maybe instead of `PoolLocked`, I would name it: `AccountingLocked`

src/Accounting.sol
lemunozm: I think just in this method, the returned type could hold negative values.

I mean the idea of this method I think is to track the debit/credit information in the sign, isn't?

lemunozm:
> @mustermeiszer In your draft you had a totalDebit and totalCredit for each account. I turned into a single balance property, but maybe you had a specific reason for it that I missed.

I think @onnovisser we need to keep both, the idea is to track both amounts for each account. The account balance is the substraction of both but definitely we need both to not lose that information.

src/Accounting.sol
lemunozm: Check my latest message. We need both `totalCredit`, `totalDebit`, and in `getAccountValue()` perform the substraction.

src/Accounting.sol
lemunozm: I think 99.9% `credit()` and d`ebit()` should be public, to allow the FM to compose things that does not really need to be symetrical.

With `updateEntry()` you increase & reduce both accounts the same amount. But the following scenario can not be reachable with just `uddateEntry()`, so the FM needs individual methods:

Account A => credit 90
Account B => debit 30
Account C => debit 60

mustermeiszer: Yes, lets expose all methods for fure.

mustermeiszer:
> I think @onnovisser we need to keep both, the idea is to track both amounts for each account. The account balance is the substraction of both but definitely we need both to not lose that information.

Yes, we need to both. That is standard for accounting and the easiest way for both adding credit and debit as also seeing a total value. All company auditors will see these accounts and now what has happened. THat will make the fund admins life easier hopfully.

src/Accounting.sol
```diff
@@ -0,0 +1,89 @@
...
+        emit Entry(_currentPoolId, _transactionId, creditAccount, debitAccount, value);
+    }
+
+    /// @inheritdoc IAccounting
+    function unlock(PoolId poolId) external auth {
+        require(PoolId.unwrap(_currentPoolId) == 0, PoolAlreadyUnlocked());
+        _debited = 0;
+        _credited = 0;
+        /// @dev Include the previous transactionId in case there's multiple transactions in one block
+        _transactionId = uint256(keccak256(abi.encodePacked(poolId, block.timestamp, _transactionId)));
```
mustermeiszer: Regarding this, I am still not sure whether to have it at this level or setting this at the pool manager level.
It is essential that changes in the holdings can be tied to these ids. 

Also, look at the idea for this in the spec [here](https://www.notion.so/Async-CP-w-Sync-Assets-17e2eac24e1780c1ac8efa50662dbbf3?pvs=4#18c2eac24e1780f39c30db50cb9364f4)

mustermeiszer: Best practise seems to be 
* date + alphanumeric counter

Maybe we do a `u64` for the date, `u64` for the `poolId`, `u128` counter per pool? and concat them?

src/Accounting.sol
onnovisser: should we then also split up the events into Credit and Debit events, as opposed to JourneyEntry events?

lemunozm: I think it makes sense!

src/types/AccountId.sol
onnovisser: ahh, got it, yes that would make sense

src/Accounting.sol
```diff
@@ -0,0 +1,89 @@
...
+    function createAccount(PoolId poolId, AccountId account, bool isDebitNormal, bytes calldata metadata)
+        external
+        auth
+    {
+        require(accounts[poolId][account].lastUpdated == 0, AccountExists());
+        accounts[poolId][account] = Account(0, metadata, isDebitNormal, uint64(block.timestamp));
+    }
+
+    /// @dev Debits an account. Increase the balance of debit-normal accounts, decrease for credit-normal ones.
+    function _debit(AccountId account, uint128 value) internal {
```
lemunozm: @onnovisser do you aggree in renaming `credit()` & `debit()` to `addDebit()` and `addCredit()` to express mutability and distinguish them from possible getters?

onnovisser: yes that makes sense

src/Accounting.sol
lemunozm: @peculiarity curious about why public 👀

src/Accounting.sol
```diff
@@ -0,0 +1,91 @@
...
+        require(!_currentPoolId.isNull(), AccountingLocked());
+
+        Account storage acc = accounts[_currentPoolId][account];
+        require(acc.lastUpdated != 0, AccountDoesNotExists());
+
+        acc.totalDebit += value;
+        debited += value;
+        acc.lastUpdated = uint64(block.timestamp);
+        emit Debit(_currentPoolId, _transactionId, account, value);
+    }
```
lemunozm: What happen if I add debit to a non `isDebitNormal` account?

src/Accounting.sol
```diff
@@ -0,0 +1,91 @@
...
+    uint256 private /*TODO: transient*/ _transactionId;
+    PoolId private /*TODO: transient*/ _currentPoolId;
+
+    constructor(address deployer) Auth(deployer) {}
+
+    /// @inheritdoc IAccounting
+    function updateEntry(AccountId creditAccount, AccountId debitAccount, uint128 value) external auth {
+        addDebit(debitAccount, value);
+        addCredit(creditAccount, value);
+    }
```
lemunozm: Given that this is fully composible from outside, I think we can just remove it and let the `PoolManager` to compose that.

src/Accounting.sol
```diff
@@ -0,0 +1,91 @@
...
+        emit Credit(_currentPoolId, _transactionId, account, value);
+    }
+
+    /// @inheritdoc IAccounting
+    function unlock(PoolId poolId) external auth {
+        require(PoolId.unwrap(_currentPoolId) == 0, AccountingAlreadyUnlocked());
+        debited = 0;
+        credited = 0;
+        /// @dev Include the previous transactionId in case there's multiple transactions in one block
+        _transactionId = uint256(keccak256(abi.encodePacked(poolId, block.timestamp, _transactionId)));
```
lemunozm: Love the idea! Although I think there is one missing case:

- In the same block, the same FM for the same pool, perform two different transactions. It means:
  - First transaction => block: 42, pool: 1, _transactionId = resultingHash
  - Transaction ends => _transactionId = 0
  - Second transaction => block: 42, pool: 1, _transactionId = same_as_above
  - Transaction ends => _transactionId = 0
  
Note that the scope of the transient storage is the transaction, not the block.

NIT. /// @devs comments can only be at function level.

src/interfaces/IAccounting.sol
```diff
@@ -0,0 +1,54 @@
...
+    struct Account {
+        uint128 totalDebit;
+        uint128 totalCredit;
+        bool isDebitNormal;
+        uint64 lastUpdated;
+        bytes metadata;
+    }
+
+    /// @notice Logs an entry where one account is debited and another is credited.
+    function updateEntry(AccountId creditAccount, AccountId debitAccount, uint128 value) external;
```
lemunozm: `addCredit` and `addDebit` should also be in the interface.

src/interfaces/IAccounting.sol
```diff
@@ -0,0 +1,54 @@
...
+
+    /// @notice Creates an account.
+    function createAccount(PoolId poolId, AccountId account, bool isDebitNormal) external;
+
+    /// @notice Creates an account.
+    function updateAccountMetadata(PoolId poolId, AccountId account, bytes calldata metadata) external;
+
+    /// @notice Returns the value of an account, returns a negative value for positive balances of credt-normal
+    /// accounts.
+    function getAccountValue(PoolId poolId, AccountId account) external returns (int128);
```
lemunozm: We decided to not use "get" for getters because most of the getter are automatically generated with the storage name which has no "get" name. So for alignment, we try to avoid "get" prefix.

src/interfaces/IAccounting.sol
```diff
@@ -0,0 +1,54 @@
...
+    function unlock(PoolId poolId) external;
+
+    /// @notice Closes the transaction and checks if the entries are balanced.
+    function lock() external;
+
+    /// @notice Creates an account.
+    function createAccount(PoolId poolId, AccountId account, bool isDebitNormal) external;
+
+    /// @notice Creates an account.
+    function updateAccountMetadata(PoolId poolId, AccountId account, bytes calldata metadata) external;
```
lemunozm: NIT. We took another decision about `set` vs `update` usage. In this case, it should be `set...`, to notify that it could never been has set. We only use `update...` if the value was already initialized somehow. Ref [slack thread](https://kflabs.slack.com/archives/C07PG2EUR9C/p1733941850823459)

lemunozm (commented):
Some comments but it looks quite ready to be merged!

src/Accounting.sol
onnovisser: It still just increases the totalDebit of the account, but to get the value of the account in `getAccountValue` it does `totalCredit - totalDebit` instead of `totalDebit - totalCredit`, so credit-normal accounts have a positive value when credits outweigh debits.

src/Accounting.sol
onnovisser: good catch! do you think we should have a counter per pool like Frederik suggests? Or externalize the transaction ID?

src/Accounting.sol
onnovisser: do you or anyone else (@peculiarity) think the gas savings could be worth keeping it?

src/Accounting.sol
lemunozm: Ok, makes sense!

src/Accounting.sol
lemunozm: For us, if exists, I would say depreciable.

src/Accounting.sol
lemunozm: 🤔 I do not know the correct answer

I feel like having a `mapping(PoolId => counter)` should work fine, but it's true that maybe we're bloating this contract with a more general logic, and should be another contract in change of this.

Let's try the most simple for you (maybe the mapping?). If that doesn't work later we can refactorize it.

wischli: > Let's try the most simple for you (maybe the mapping?). If that doesn't work later we can refactorize it.

I agree. Added this to the ["Revise Later" section](https://www.notion.so/Decision-Home-1892eac24e17809b9193d8ea6c555907?pvs=4#1892eac24e178000bd6bcddad1d505b4) of our Decision Home doc.

src/Accounting.sol
```diff
@@ -0,0 +1,91 @@
...
+    function lock() external auth {
+        require(debited == credited, Unbalanced());
+        _currentPoolId = PoolId.wrap(0);
+    }
+
+    /// @inheritdoc IAccounting
+    function createAccount(PoolId poolId, AccountId account, bool isDebitNormal) external auth {
+        require(accounts[poolId][account].lastUpdated == 0, AccountExists());
+        accounts[poolId][account] = Account(0, 0, isDebitNormal, uint64(block.timestamp), "");
+    }
```
lemunozm, line 65: @onnovisser is a requirement to pass the account as a parameter? Or can we generate it?

If we do not generate it, the UX should do it, which maybe is not easy. WDYT as frontend developer?

lemunozm (approved):
Awesome and super fast work @onnovisser!! Happy to have you with us 🚀 

I left a comment above to know your opinion and change it in case you think it would improve the usage of this. But not a blocker

src/Accounting.sol
onnovisser: I suppose it's chosen by the FM, because the ID also includes the account type (Asset, Gain, Loss, Equity) and the accounting contract doesn't know anything about those

onnovisser: I guess the same question applies to `poolManager.setHoldingAccountId()`

onnovisser: The UI could generate them too when the FM creates pool I suppose

lemunozm: Mmmm... 🤔 

I was thinking in something like:

```
function createAccount(PoolId poolId, uint8 kind, bool isDebitNormal) {}
```

So the FM only needs to specify the kind, but not the ID. The `Accounting` contract could have a `mapping(PoolId => kind => counter)` to generate the ids.

The problem I want to solve is the following:
1. The FM has already a lot of accounts created.
1. The FM wants to create another account. How does it know which ID is free to use?

mustermeiszer: @lemunozm I really really do not like automatically generated ids. It is almost always easier for the UI to create them. 
That has been a huge pain point in Substrate and I think in general, if there is no strong requirement for onchain creation due to some logic/security concern, the caller should always decide.

mustermeiszer: > The FM wants to create another account. How does it know which ID is free to use?

They can always query the storage, right? 

> The FM has already a lot of accounts created.

Are you worried for keeping track of them?

lemunozm: Yes, my point is that the UI, needs to download all the current accounts to know the next account.

lemunozm: But I do not have an strong opinion of that besides UX usability

onnovisser: I think a list of already created accounts can also be stored in the pool metadata on ipfs

src/Accounting.sol
```diff
@@ -0,0 +1,84 @@
...
+
+    function setAccountMetadata(PoolId poolId, AccountId account, bytes calldata metadata) external auth {
+        Account storage acc = accounts[poolId][account];
+        acc.metadata = metadata;
```
wischli: CR: As long as you only access/write a single field of a storage struct, it is cheaper to do that directly instead of copying the value with `memory` or `storage` and then working with that

```suggestion
        accounts[poolId][account].metadata = metadata;
```

src/Accounting.sol
```diff
@@ -0,0 +1,84 @@
...
+    }
+
+    /// @inheritdoc IAccounting
+    function accountValue(PoolId poolId, AccountId account) public view returns (int128) {
+        Account storage acc = accounts[poolId][account];
```
wischli, line 73: CR: Since we are not mutating the account, it is sufficient to initialize it as `memory`
```suggestion
        Account memory acc = accounts[poolId][account];
```

src/interfaces/IAccounting.sol
```diff
@@ -0,0 +1,57 @@
...
+    /// @notice
+    struct Account {
+        uint128 totalDebit;
+        uint128 totalCredit;
+        bool isDebitNormal;
+        uint64 lastUpdated;
+        bytes metadata;
+    }
+
+    /// @notice Debits an account. Increase the value of debit-normal accounts, decrease for credit-normal ones.
```
wischli, line 36: Nit: All of these functions are missing proper natspec docs for the arguments and return values. Could you fix this in a follow-up PR? Definitely not a blocker. I know we have skipped quite some docs so far but it appears to me that `Accounting` is quite solid and could very well be fully documented.

test/unit/Accounting.t.sol
```diff
@@ -0,0 +1,163 @@
...
+    }
+
+    function testDebitsAndCredits() public {
+        accounting.unlock(POOL_A, bytes32("1"));
+
+        vm.expectEmit();
+        emit IAccounting.Debit(POOL_A, bytes32("1"), CASH_ACCOUNT, 500);
+        accounting.addDebit(CASH_ACCOUNT, 500);
+
+        vm.expectEmit(true, false, true, true);
```
wischli: Curious why you are skipping the check for the second indexed parameter here. Shouldn't it be

```suggestion
        vm.expectEmit(true, true, true, true);
```

test/unit/Accounting.t.sol
```diff
@@ -0,0 +1,163 @@
...
+
+        assertEq(accounting.accountValue(POOL_A, CASH_ACCOUNT), 500);
+        assertEq(accounting.accountValue(POOL_A, EQUITY_ACCOUNT), 500);
+        assertEq(accounting.accountValue(POOL_B, CASH_ACCOUNT), 120);
+        assertEq(accounting.accountValue(POOL_B, EQUITY_ACCOUNT), 120);
+    }
+
+    function testUnequalDebitsAndCredits(uint128 v) public {
+        vm.assume(v != 5);
+        vm.assume(v < type(uint128).max - 250);
```
wischli, line 85: Comment: In general, `bound` is preferred but if we bound to `(6, type(uint128).max - 251)` we are missing checks for `(0, 4)`. Just wanted to bring it up in case you didn't know. Else forget my comment :)

wischli (commented):
I have two/three minor CRs but don't want to block the PR as it can be fixed in a follow-up. Again, stellar work 🚀

src/Accounting.sol
lemunozm: Was not a discussion where we realize that even in that case `storage` was still more efficient?

test/unit/Accounting.t.sol
onnovisser: ah yes, that was leftover from when the contract generated the transaction ID. updated

test/unit/Accounting.t.sol
onnovisser: I didn't know about `bound` thanks!

src/Accounting.sol
onnovisser: just tried it and yes, the `testDebitsAndCredits()`, for example, uses 10k more gas when using `memory`. interesting

lemunozm: I think memory performs a copy even if you only read. But storage uses the value comming from the chain, without copying it.

AFAIK we only need `memory` if we need to forcebly perform a copy, because either the data is no longer available or you want to mutate without altering the chain data.

src/Accounting.sol
```diff
@@ -0,0 +1,83 @@
...
+import {IAccounting} from "src/interfaces/IAccounting.sol";
+import {AccountId} from "src/types/AccountId.sol";
+import {PoolId} from "src/types/PoolId.sol";
+import {PoolLocker} from "src/PoolLocker.sol";
+import {Auth} from "src/Auth.sol";
+
+contract Accounting is Auth, IAccounting {
+    mapping(PoolId => mapping(AccountId => Account)) public accounts;
+
+    uint128 public /*TODO: transient*/ debited;
```
peculiarity, line 13: Question: Any reason for these to be public or the other ones private?

src/Accounting.sol
peculiarity: Yeah, that thing with storage was a news to me as well. After analyzing it, turned out using `storage` was cheaper..

src/Accounting.sol
```diff
@@ -0,0 +1,83 @@
...
+        acc.totalCredit += value;
+        credited += value;
+        acc.lastUpdated = uint64(block.timestamp);
+        emit Credit(_currentPoolId, _transactionId, account, value);
+    }
+
+    /// @inheritdoc IAccounting
+    function unlock(PoolId poolId, bytes32 transactionId) external auth {
+        require(PoolId.unwrap(_currentPoolId) == 0, AccountingAlreadyUnlocked());
+        debited = 0;
```
peculiarity, line 49: Question:
You did this "reset" because currently the `transient` is not used or for some other reason ?

peculiarity (approved):

src/Accounting.sol
wischli: Oh wow thanks for checking, good to know!

wischli (approved):
I heard it's time to get squashed!

src/Accounting.sol
onnovisser: also because the PoolManager may unlock/lock the accounting more than once in a transaction

src/Accounting.sol
onnovisser: I don't think they'll see much use by other contracts, but I made them public after your comment 😅

src/Accounting.sol
onnovisser: though now that I think about it, that wouldn't even matter, because they are guaranteed to be equal at unlock time and as long as they are equal it's fine.

src/Accounting.sol
mustermeiszer: > Yes, my point is that the UI, needs to download all the current accounts to know the next account.

They can use a hash from the timestamp or just a counter in the metadata.

PR #74, opened 2025-02-03
Adding Integration tests

lemunozm:
This PR adds:
- A deployer script that connects everything.
- An integration test file that uses the deploy script. 

To have this working correctly still missing:
- [x] `AssetManager` impl
- [x] `Accounting` impl
- [x] `Gateway` impl

### DIFF
--- a/src/SingleShareClass.sol
+++ b/src/SingleShareClass.sol
@@ -47,6 +47,7 @@ struct EpochPointers {
     uint32 latestRevocation;
 }
 
+/// Utility method to determine the ShareClassId for a PoolId
 function shareClassIdFor(PoolId poolId) pure returns (ShareClassId) {
     return ShareClassId.wrap(bytes16(uint128(PoolId.unwrap(poolId))));
 }
--- a/src/SingleShareClass.sol
+++ b/src/SingleShareClass.sol
@@ -48,12 +48,10 @@ struct EpochPointers {
 }
 
 /// Utility method to determine the ShareClassId for a PoolId
-function shareClassIdFor(PoolId poolId) pure returns (ShareClassId) {
+function previewShareClassId(PoolId poolId) pure returns (ShareClassId) {
     return ShareClassId.wrap(bytes16(uint128(PoolId.unwrap(poolId))));
 }
 
-// Assumptions:
-// * ShareClassId is unique and derived from pool, i.e. bytes16(keccak256(poolId + salt))
 contract SingleShareClass is Auth, ISingleShareClass {
     using MathLib for D18;
     using MathLib for uint128;
@@ -95,7 +93,7 @@ contract SingleShareClass is Auth, ISingleShareClass {
     {
         require(shareClassId[poolId].isNull(), MaxShareClassNumberExceeded(1));
 
-        shareClassId_ = shareClassIdFor(poolId);
+        shareClassId_ = previewShareClassId(poolId);
 
         shareClassId[poolId] = shareClassId_;
         epochId[poolId] = 1;
--- a/src/SingleShareClass.sol
+++ b/src/SingleShareClass.sol
@@ -47,6 +47,12 @@ struct EpochPointers {
     uint32 latestRevocation;
 }
 
+function shareClassIdFor(PoolId poolId) pure returns (ShareClassId) {
+    return ShareClassId.wrap(bytes16(uint128(PoolId.unwrap(poolId))));
+}
+
+// Assumptions:
+// * ShareClassId is unique and derived from pool, i.e. bytes16(keccak256(poolId + salt))
 contract SingleShareClass is Auth, ISingleShareClass {
     using MathLib for D18;
     using MathLib for uint128;
@@ -70,8 +76,8 @@ contract SingleShareClass is Auth, ISingleShareClass {
     mapping(ShareClassId scId => mapping(AssetId paymentAssetId => mapping(bytes32 investor => UserOrder pending)))
         public depositRequest;
 
-    constructor(address poolRegistry_, address deployer) Auth(deployer) {
-        poolRegistry = IPoolRegistry(poolRegistry_);
+    constructor(IPoolRegistry poolRegistry_, address deployer) Auth(deployer) {
+        poolRegistry = poolRegistry_;
     }
 
     function file(bytes32 what, address data) external auth {
@@ -88,7 +94,7 @@ contract SingleShareClass is Auth, ISingleShareClass {
     {
         require(shareClassId[poolId].isNull(), MaxShareClassNumberExceeded(1));
 
-        shareClassId_ = ShareClassId.wrap(bytes16(uint128(PoolId.unwrap(poolId))));
+        shareClassId_ = shareClassIdFor(poolId);
 
         shareClassId[poolId] = shareClassId_;
         epochId[poolId] = 1;
--- a/test/integration/Cases.t.sol
+++ b/test/integration/Cases.t.sol
@@ -10,6 +10,7 @@ import {D18, d18} from "src/types/D18.sol";
 import {shareClassIdFor} from "src/SingleShareClass.sol";
 
 import {AccountType} from "src/interfaces/IPoolManager.sol";
+import {IMulticall} from "src/interfaces/IMulticall.sol";
 
 import {Deployer} from "script/Deployer.s.sol";
 
@@ -21,23 +22,25 @@ contract TestCommon is Deployer, Test {
     uint32 constant CHAIN_1 = 1;
     uint32 constant CHAIN_2 = 2;
 
-    address immutable DEPLOYER = makeAddr("Deployer");
-    address immutable FM = makeAddr("Fund manager");
-    address immutable ANY = makeAddr("Anyone without role");
-    bytes32 immutable INVESTOR = bytes32(uint256(uint160(makeAddr("Investor"))));
+    address immutable FM = makeAddr("FM");
+    address immutable ANY = makeAddr("Anyone");
+    bytes32 immutable INVESTOR = bytes32("Investor");
 
-    AssetId immutable USD = newAssetIdFromISO4217(840);
     AssetId immutable USDC_C2 = newAssetId(CHAIN_2, 1);
 
-    MockCentrifugeVaults cv = new MockCentrifugeVaults(poolManager);
+    MockCentrifugeVaults cv;
 
     function setUp() public {
-        deploy(DEPLOYER);
+        // Deployment
+        deploy();
 
-        // TODO: remove this line when Gateway is implemented
-        poolManager.rely(address(cv));
+        cv = new MockCentrifugeVaults(poolManager);
 
-        // Label contracts (for debugging)
+        poolManager.rely(address(cv)); // TODO: remove this the Gateway is implemented.
+
+        removeDeployerAccess();
+
+        // Label contracts & actors (for debugging)
         vm.label(address(transientValuation), "TransientValuation");
         vm.label(address(oneToOneValuation), "OneToOneValuation");
         vm.label(address(multicall), "Multicall");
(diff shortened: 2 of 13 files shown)


mustermeiszer (commented):
Nice to see the it being wired together!

src/SingleShareClass.sol
```diff
@@ -47,6 +47,13 @@ struct EpochPointers {
     uint32 latestRevocation;
 }
 
+/// Utility method to determine the ShareClassId for a PoolId
+function shareClassIdFor(PoolId poolId) pure returns (ShareClassId) {
```
wischli: Nit: For every attached share class id, this method should be superseded by the `shareClassId(poolId)` getter.

src/SingleShareClass.sol
```diff
@@ -47,6 +47,13 @@ struct EpochPointers {
...
+    return ShareClassId.wrap(bytes16(uint128(PoolId.unwrap(poolId))));
+}
+
+// Assumptions:
+// * ShareClassId is unique and derived from pool, i.e. bytes16(keccak256(poolId + salt))
```
wischli: Nit: Re-added deprecated artifact
```suggestion
```

wischli (commented):
Need to grab some popcorn for the following commits, super psyched to see all of it coming together so nicely!

src/SingleShareClass.sol
lemunozm: Yeah, that's correct. The problem is that I can not call `shareClassId(poolId)` unless I previously call `addShareClass()`, which for some multicalls is late.

I wanted to have some way of precomputing the scId for some pool. Maybe the name should be `previewShareClassId(poolId)` instead?

src/SingleShareClass.sol
lemunozm: Some git conflict bad done! Thanks

src/SingleShareClass.sol
mustermeiszer: Should we maybe just make the `ShareClassId` a parameter then? Just like the `PoolId` that would solve a lot of these problems, right?

lemunozm: I think there is not the exact problem. I know without problems which `shareClassId` will be used in advance.

What @wischli says is that instead of precomputing it, I can use the `shareClassId()` getter, but if I use that, I would need to split the `execute()` call into two to be able to 1. perform the creation and 2. be able to read from the getter.

The above is just a preview of the expected share class to use just one single `execute()`, something the UX can also use

mustermeiszer: But then using it as arguments would exactly solve that problem, right? All in one execute. That is what I mean or why can't it all be in one call if you can determine the id upfront? 

Missing something I guess. ^^

lemunozm: > why can't it all be in one call if you can determine the id upfront?

I can for sure, and I'm doing it. Right now, there is only one call. I just added a "preview" method to `SingleShareClass` to get the expected Id, as a utility.

But I can not use the getter because the value is still not stored before calling to `execute()`. But nothing wrong with it.

wischli: > Should we maybe just make the `ShareClassId` a parameter then? Just like the `PoolId` that would solve a lot of these problems, right?

I would like to not go that path because this complicates ensuring uniqueness of all share class id across multiple CP instances. 

IMO, @lemunozm alteration of adding a preview method perfectly solves the issue of deterministically previewing a non-existent share class id.

test/integration/Cases.t.sol
```diff
@@ -0,0 +1,164 @@
...
+
+    function testGeneralConfigurationPool() public returns (PoolId poolId, ShareClassId scId) {
+        cv.registerAsset(USDC_C2, "USD Coin", "USDC", 6);
+
+        vm.prank(FM);
+        poolId = poolManager.createPool(USD, singleShareClass);
+
+        scId = previewShareClassId(poolId);
+
+        (bytes[] memory cs, uint256 c) = (new bytes[](7), 0);
```
peculiarity, line 126: Frankly IMO this is not a common pattern and it makes the whole thing hard to read. No idea what's the problem if having  `cs[0]` , `cs[1]` and so on...

test/integration/Cases.t.sol
```diff
@@ -0,0 +1,164 @@
...
+
+import {MockCentrifugeVaults} from "test/mock/MockCentrifugeVaults.sol";
+
+import "forge-std/Test.sol";
+
+contract TestCommon is Deployer, Test {
+    uint32 constant CHAIN_1 = 1;
+    uint32 constant CHAIN_2 = 2;
+
+    address immutable FM = makeAddr("FM");
```
peculiarity, line 29: Nit: When you don't use the constructor to change a storage variable, `constant` is used. Evaluation is during compile time.

test/integration/Cases.t.sol
lemunozm: It's very easy to mess it up by copying and indexing twice or leaving an index empty. We will have tens of lists of calls.

Imagine a refactor that removes a call from the `PoolManager`, you need to remove 20 times that call from different tests, and later fix hundreds of indexes to align them... you miss one, and you have a failing test with no reason and difficult to debug. Note that encoding method are not checked by the compiler, making things harder

test/integration/Cases.t.sol
lemunozm: Not sure if I get what you say. In that case, I can not use constant because `makeAddr` is not constant

test/integration/Cases.t.sol
peculiarity: 1. I don't you will be testing the same thing in 20 times. not even in  5. Even if you have to build the same multicall array, you can extract it into a helper function and have it in a single place. 
2. if you I understand correctly, you are talking about  rewriting the same index. I assume if that is the case, then something won't execute, which means your assertions will fail.
IMO we shouldn't sacrifice readability to "cover" some edge cases in terms of writing tests.

If the rest of the team agree  on this approach, then I also can adopt it.
@wischli  @onnovisser  @hieronx

lemunozm: > 1. I don't you will be testing the same thing in 20 times. not even in 5. Even if you have to build the same multicall array, you can extract it into a helper function and have it in a single place.

I'm not saying "testing the same thing 20 times"; I'm seeing that 20 different tests will contain similar calls, i.e., `addShareClass()` or `approveDeposit()`. I can not extract the "addition to the array".

> 2. if you I understand correctly, you are talking about rewriting the same index. I assume if that is the case, then something won't execute, which means your assertions will fail.

It's really easy to ending up with something like:

```solidity
cs[0] = abi.encodeWithSelector(poolManager.addShareClass.selector, bytes(""));
cs[1] = abi.encodeWithSelector(poolManager.notifyPool.selector, CHAIN_2);
cs[1] = abi.encodeWithSelector(poolManager.notifyShareClass.selector, CHAIN_2, scId);
```
by error, or just pass from `0` to `2`. I'm reducing the friction of interchaning the order of the calls by using that "index trick". Not sure exactly where the sacrifice of readability exists.

peculiarity: I'm rereading my comments and I thinking ... what went so terribly wrong. Wording and phrasing  - English Level  0 :D

peculiarity: > by error, or just pass from 0 to 2. I'm reducing the friction of interchaning the order of the calls by using that "index trick". Not sure exactly where the sacrifice of readability exists.

Yes! so I understood you correctly. In that case some method won't be called meaning that your asserts will fail.

> Not sure exactly where the sacrifice of readability exists.

`(bytes[] memory cs, uint256 c) = (new bytes[](7), 0);`

As I commented earlier, this isn't something that's commonly used. When you look at it, at first glance you have no idea what's happenings.

Also  this `[c++]` requires you to do some mental math if you need for some reason to debug  which call, comes in at what sequence.

Let's say we switched the order of the methods being called, test will fail.
```
cs[++] = abi.encodeWithSelector(poolManager.addShareClass.selector, bytes(""));
cs[++] = abi.encodeWithSelector(poolManager.notifyShareClass.selector, CHAIN_2, scId);
cs[++] = abi.encodeWithSelector(poolManager.notifyPool.selector, CHAIN_2);
```

At first glance might be easy to detect but if we have 6 ,7 rows then it might be trickier.  You have to count which row but since it starts from 0 then you have to subtract 1 and so on.


Again, this is my person opinion. I would like to hear the others. If majority is on the same page, then I can adapt.

mustermeiszer (approved):
Will most likely change quite a bit after the CV merge, right?

lemunozm:
I would like to have two levels of ITs: 
- CP-IT where CV is mocked and focused in CP behavior. With more checks, accounting checks, etc.
- CP-CV-IT, similar to what we had in CC, focused on having the whole flow working and being able to connect successfully in the happy paths with CV.

The point is, after a change, realize quickly where the error is, whether in CP or CV.

In this case, these ITs are the first ones, so I think they should not change to much. Anyways let's see how this evolves!

test/integration/Cases.t.sol
```diff
@@ -0,0 +1,191 @@
...
+        vm.label(address(poolRegistry), "PoolRegistry");
+        vm.label(address(assetManager), "AssetManager");
+        vm.label(address(accounting), "Accounting");
+        vm.label(address(holdings), "Holdings");
+        vm.label(address(singleShareClass), "SingleShareClass");
+        vm.label(address(poolManager), "PoolManager");
+        vm.label(address(gateway), "Gateway");
+        vm.label(address(cv), "CV");
+
+        // We decide CP is located at CHAIN_1 for messaging
```
wischli: Nit: Maybe we should call it `CHAIN_CP` then?

test/integration/Cases.t.sol
```diff
@@ -0,0 +1,191 @@
...
+        cs[c++] = abi.encodeWithSelector(poolManager.issueShares.selector, scId, USDC_C2, NAV_PER_SHARE);
+
+        vm.prank(FM);
+        poolManager.execute(poolId, _fromPoolManager(cs));
+
+        vm.prank(ANY);
+        poolManager.claimDeposit(poolId, scId, USDC_C2, INVESTOR);
+
+        // TODO: checks
+        // claimed amount == SHARE_AMOUNT
```
wischli, line 168: Reminder about the TODOs. Did you plan on merging without handling those?

test/integration/Cases.t.sol
```diff
@@ -0,0 +1,191 @@
...
+        cs[c++] = abi.encodeWithSelector(poolManager.revokeShares.selector, scId, USDC_C2, NAV_PER_SHARE, valuation);
+
+        vm.prank(FM);
+        poolManager.execute(poolId, _fromPoolManager(cs));
+
+        vm.prank(ANY);
+        poolManager.claimRedeem(poolId, scId, USDC_C2, INVESTOR);
+
+        // TODO: checks
+        // claimed amount == INVESTOR_AMOUNT
```
wischli, line 189: Reminder about the TODOs. Did you plan on merging without handling those?

wischli (approved):
LGTM! Though I think the two leftover TODOs were meant to be completed as part of this PR?

test/integration/Cases.t.sol
lemunozm: Will do in the follow up PR. Once I have a serialziation/deserialization library, it will be much easier to get the values from the messages to check that. Now is a bit painful

test/integration/Cases.t.sol
lemunozm: I think it makes sense. I wasn't sure because maybe some tests could be in the same chain, but I'm mocking that part, so yeah. Let's call it `CHAIN_CP`!

lemunozm: Done also with CV: [b8f5868](https://github.com/centrifuge/pools-internal/pull/74/commits/b8f5868c726758505470b78e01ec8f56b9a21c9b)

PR #75, opened 2025-02-03
Refactor: Multicall using Call struct

lemunozm:
### Problem

When a multicall grows in size, it becomes very verbose and error-prone when configuring the multicall, because `target` and `data` are split per call. See [this](https://github.com/centrifuge/pools-internal/pull/74)

### Solution

Using a `Call` struct to join both fields, it becomes more readable and less error-prone. Also, it's easier to refactor test cases.

### DIFF
--- a/src/interfaces/IMulticall.sol
+++ b/src/interfaces/IMulticall.sol
@@ -3,8 +3,11 @@ pragma solidity 0.8.28;
 
 /// @notice Allows to call several calls in the same transactions
 interface IMulticall {
+    /// @notice Identify a call method.
     struct Call {
+        /// @notice Contract from where to perform the call
         address target;
+        /// @notice Encoding of selector + parameters of the method
         bytes data;
     }
 
--- a/src/interfaces/IMulticall.sol
+++ b/src/interfaces/IMulticall.sol
@@ -3,10 +3,12 @@ pragma solidity 0.8.28;
 
 /// @notice Allows to call several calls in the same transactions
 interface IMulticall {
-    /// @notice Dispatched when the `targets` and `datas` length parameters in `execute()` do not matched.
-    error WrongExecutionParams();
+    struct Call {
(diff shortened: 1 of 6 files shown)


src/interfaces/IMulticall.sol
```diff
@@ -3,10 +3,12 @@ pragma solidity 0.8.28;
 
 /// @notice Allows to call several calls in the same transactions
 interface IMulticall {
-    /// @notice Dispatched when the `targets` and `datas` length parameters in `execute()` do not matched.
-    error WrongExecutionParams();
+    struct Call {
```
wischli, line 7: Nit: Missing docs

test/unit/Multicall.t.sol
```diff
@@ -25,41 +25,25 @@ contract MulticallTest is Test {
...
-        bytes[] memory methods = new bytes[](2);
-        methods[0] = abi.encodeWithSelector(userContract.setState.selector);
-        methods[1] = abi.encodeWithSelector(userContract.userFailMethod.selector);
+        IMulticall.Call[] memory calls = new IMulticall.Call[](2);
+        calls[0] = IMulticall.Call(address(userContract), abi.encodeWithSelector(userContract.updateState.selector));
+        calls[1] = IMulticall.Call(address(userContract), abi.encodeWithSelector(userContract.userFailMethod.selector));
 
         vm.expectRevert("user error");
-        multicall.aggregate(targets, methods);
+        multicall.aggregate(calls);
```
wischli, line 45: Comment: This diff showcases very well why this change is a great improvement!

wischli (approved):
Great improvement!

src/interfaces/IMulticall.sol
lemunozm: solved!

peculiarity (approved):

PR #76, opened 2025-02-04
refactor: Remove price ratio storage from SingleShareClass

wischli:
Supersedes #70

## Background

> Handling prices directly is difficult. (i.e What precision they need? Should be prices or factors? How they interact with plain uint128 of unknown representation?). In fact, the own [ERC7726](https://eips.ethereum.org/EIPS/eip-7726#rationale) spec hides the price concept and gives a rationale of why:
> 
> > The spec doesn’t include a getPrice function because it is rarely needed on-chain, and it would be a decimal number of difficult representation.
> 
> So we should avoid calling to a `getPrice` or `getFactor` as much as we can, given their bad side-effects

## Changes 

* Replaces storage of price ratios during approval, issuance and revocation with derivation via approved amounts
  * Base idea during claiming: 
  ```solidity
  claimableAmount = approvalRate * pendingUserAmount / approvedAmount * {issued, revoked}Amount
  ```
* Removes `IERC7726Ext` which was only required for price ratios in SCM
* Renames `AssetEpochState` to `EpochPointers`: Apart form the `EpochId`, both epoch related storages operate on asset-granularity (among other keys) such that I feel like the `Asset` prefix is redundant. Moreover, now that `EpochRatio` was replaced by `EpochAmounts`, `EpochState` is not descriptive and conflicting with `EpochAmounts` (i.e. it is not clear what an epoch state refers to)

### DIFF
--- a/src/SingleShareClass.sol
+++ b/src/SingleShareClass.sol
@@ -4,32 +4,27 @@ pragma solidity 0.8.28;
 import {Auth} from "src/Auth.sol";
 import {D18, d18} from "src/types/D18.sol";
 import {IPoolRegistry} from "src/interfaces/IPoolRegistry.sol";
-import {IERC7726Ext} from "src/interfaces/IERC7726.sol";
+import {IERC7726} from "src/interfaces/IERC7726.sol";
 import {IShareClassManager} from "src/interfaces/IShareClassManager.sol";
 import {ISingleShareClass} from "src/interfaces/ISingleShareClass.sol";
 import {MathLib} from "src/libraries/MathLib.sol";
 import {PoolId} from "src/types/PoolId.sol";
 
-struct Epoch {
-    /// @dev Amount of approved deposits (in pool denomination)
-    uint128 approvedDepositAmount;
-    /// @dev Amount of approved shares (in share denomination)
-    uint128 approvedShareAmount;
-}
-
-struct EpochRatio {
+struct EpochAmounts {
     /// @dev Percentage of approved deposits
-    D18 depositRatio;
+    D18 depositApprovalRate;
     /// @dev Percentage of approved redemptions
-    D18 redeemRatio;
-    /// @dev Price of one pool per asset token set during redeem approval
-    D18 depositAssetToPoolQuote;
-    /// @dev Price of one pool per asset token set during redeem approval
-    D18 redeemAssetToPoolQuote;
-    /// @dev Price of one pool per share class token set during deposit flow
-    D18 depositShareToPoolQuote;
-    /// @dev Price of one pool per share class token set during redeem flow
-    D18 redeemShareToPoolQuote;
+    D18 redeemApprovalRate;
+    /// @dev Total approved asset amount of deposit asset
+    uint128 depositAssetAmount;
+    /// @dev Total approved pool amount of deposit asset
+    uint128 depositPoolAmount;
+    /// @dev Total number of share class tokens issued
+    uint128 depositSharesIssued;
+    /// @dev Total asset amount of revoked share class tokens
+    uint128 redeemAssetAmount;
+    /// @dev Total approved amount of redeemed share class tokens
+    uint128 redeemSharesRevoked;
 }
 
 struct UserOrder {
@@ -55,6 +50,7 @@ struct AssetEpochState {
 contract SingleShareClass is Auth, ISingleShareClass {
     using MathLib for D18;
     using MathLib for uint128;
+    using MathLib for uint256;
 
     /// Storage
     // uint32 private transient _epochIncrement;
@@ -64,11 +60,11 @@ contract SingleShareClass is Auth, ISingleShareClass {
     mapping(bytes16 scId => uint128) public totalIssuance;
     mapping(PoolId poolId => uint32 epochId_) public epochId;
     mapping(bytes16 scId => D18 navPerShare) private _shareClassNavPerShare;
-    mapping(bytes16 scId => mapping(uint32 epochId_ => Epoch epoch)) public epoch;
     mapping(bytes16 scId => mapping(address assetId => AssetEpochState)) public assetEpochState;
     mapping(bytes16 scId => mapping(address payoutAssetId => uint128 pending)) public pendingRedeem;
     mapping(bytes16 scId => mapping(address paymentAssetId => uint128 pending)) public pendingDeposit;
-    mapping(bytes16 scId => mapping(address assetId => mapping(uint32 epochId_ => EpochRatio epoch))) public epochRatio;
+    mapping(bytes16 scId => mapping(address assetId => mapping(uint32 epochId_ => EpochAmounts epoch))) public
+        epochAmounts;
     mapping(bytes16 scId => mapping(address payoutAssetId => mapping(address investor => UserOrder pending))) public
         redeemRequest;
     mapping(bytes16 scId => mapping(address paymentAssetId => mapping(address investor => UserOrder pending))) public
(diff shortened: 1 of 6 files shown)


src/SingleShareClass.sol
```diff
@@ -4,32 +4,27 @@ pragma solidity 0.8.28;
...
 import {ISingleShareClass} from "src/interfaces/ISingleShareClass.sol";
 import {MathLib} from "src/libraries/MathLib.sol";
 import {PoolId} from "src/types/PoolId.sol";
 
-struct Epoch {
-    /// @dev Amount of approved deposits (in pool denomination)
-    uint128 approvedDepositAmount;
-    /// @dev Amount of approved shares (in share denomination)
-    uint128 approvedShareAmount;
-}
```
wischli, line 18: This struct lost its purpose over time and has not been used so far AFAICT. The only downside of replacing this with `EpochAmounts` is that now we don't know the total approved amount of an epoch for multiple assets. We can derive them via `EpochAmounts` but need to query that for each asset and sum up.

If you believe that's necessary, please let me know @mustermeiszer

src/SingleShareClass.sol
```diff
@@ -319,54 +304,94 @@ contract SingleShareClass is Auth, ISingleShareClass {
...
-            uint128 epochPoolAmount = navPerShare.mulUint128(revokedShareAmount);
-
-            epochRatio_.redeemShareToPoolQuote = navPerShare;
-            payoutPoolAmount += epochPoolAmount;
-            payoutAssetAmount += epochRatio_.redeemAssetToPoolQuote.reciprocalMulUint128(epochPoolAmount);
-            totalIssuance_ -= revokedShareAmount;
-            uint128 nav = navPerShare.mulUint128(totalIssuance_);
-
-            emit IShareClassManager.RevokedShares(poolId, shareClassId_, epochId_, navPerShare, nav, revokedShareAmount);
+            payoutPoolAmount += revokeEpochShares(
```
wischli: Unfortunately, I did not get around reintroducing an internal epoch revocation function due to the `Stack too deep` error.

src/interfaces/IERC7726.sol
```diff
@@ -12,10 +12,3 @@ interface IERC7726 {
     /// @param baseAmount The amount of base in base terms.
     function getQuote(uint256 baseAmount, address base, address quote) external view returns (uint256 quoteAmount);
 }
-
-interface IERC7726Ext is IERC7726 {
-    /// @notice Returns the internal ratio used to convert base into quote, e.g. ETH to USDC
-    /// @param base The numerator asset for the desired ratio
-    /// @param quote The denominator asset
-    function getFactor(address base, address quote) external view returns (D18 factor);
-}
```
lemunozm, line 21: I'm super happy with this change. It means that now, any ERC7726 out there that is based on ERC6909 is valid for us.

src/SingleShareClass.sol
```diff
@@ -319,54 +304,94 @@ contract SingleShareClass is Auth, ISingleShareClass {
...
 
         for (uint32 epochId_ = startEpochId; epochId_ <= endEpochId; epochId_++) {
-            EpochRatio storage epochRatio_ = epochRatio[shareClassId_][payoutAssetId][epochId_];
+            EpochAmounts storage epochAmounts_ = epochAmounts[shareClassId_][payoutAssetId][epochId_];
 
             // Skip deposit epochs
-            if (epochRatio_.redeemRatio.inner() == 0) {
+            if (epochAmounts_.redeemApprovalRate.inner() == 0) {
                 continue;
             }
```
lemunozm, line 325: I think nothing prevents the FM from setting an approval rate of 0%. If that's the case, we're fine skipping it anyway here, right?

src/SingleShareClass.sol
lemunozm: Yeah, that's sucks. Thanks!

src/SingleShareClass.sol
```diff
@@ -319,54 +304,94 @@ contract SingleShareClass is Auth, ISingleShareClass {
...
+        address poolCurrency,
+        EpochAmounts storage epochAmounts_,
+        uint128 totalIssuance_,
+        uint32 epochId_
+    ) internal returns (uint128 payoutPoolAmount) {
+        payoutPoolAmount = navPerShare.mulUint128(epochAmounts_.redeemSharesRevoked);
+        epochAmounts_.redeemAssetAmount =
+            IERC7726(valuation).getQuote(payoutPoolAmount, poolCurrency, payoutAssetId).toUint128();
+
+        uint128 nav = navPerShare.mulUint128(totalIssuance_ - epochAmounts_.redeemSharesRevoked);
```
lemunozm: Question. I should understand this is as the resulting NAV after revoking right?, because if I revoke everything, the resulting nav is zero.

lemunozm (approved):
Real amazing work that is far from being easy. Thanks for putting your mind on this. 🧠 

I'm really happy with this change!

Left a comment here: https://github.com/centrifuge/pools-internal/pull/36#pullrequestreview-2593390042 in case you can add it to this PR. Not blocking

src/SingleShareClass.sol
wischli: Yes. There is no point in continuing in case nothing was approved as this implies nothing will be issued/revoked in that epoch.

src/SingleShareClass.sol
wischli: Yes, that's the idea. Initially, the signature of issuing and revoking included the `uint128 nav`. This had to be replaced with `D18 navPerShare` in order to be able to apply the same share class token price (= nav/totalIssuance) when revoking for multiple assets in the same multicall ([ref](https://kflabs.slack.com/archives/C07PG2EUR9C/p1734455629743079?thread_ts=1734449474.830269&cid=C07PG2EUR9C)). 

Since `NAV` is the relevant data for the FM and investors, we still need to emit it.

src/SingleShareClass.sol
lemunozm: Thanks for answering my concerns!

src/SingleShareClass.sol
```diff
@@ -319,54 +304,94 @@ contract SingleShareClass is Auth, ISingleShareClass {
...
+    /// @param navPerShare Total value of assets of the pool and share class per share
+    /// @param valuation Source of truth for quotas, e.g. the price of a pool amount in payout asset
+    /// @param poolCurrency The address of the pool currency
+    /// @param epochAmounts_ Epoch ratio storage for the amount of revoked share class tokens and the corresponding
+    /// amount
+    /// in payout asset
+    /// @param totalIssuance_ Total issuance of share class tokens before revoking
+    /// @param epochId_ Identifier of the epoch for which we revoke
+    /// @return payoutPoolAmount Converted amount of pool currency based on number of revoked shares
+    function revokeEpochShares(
```
lemunozm: Ultra NIT. I thought that internal methods also should have an underscore as the private ones. Our solidity style guide is not clear about that regard 🤔, although CV uses underscore for internal ones.

wischli: Good catch, you are right, it should be prefixed with an underscore!

wischli: [f418213](https://github.com/centrifuge/pools-internal/pull/76/commits/f4182132dee38bc835bd1024dd161536779c43a0)

lemunozm (approved):
LGTM!!

peculiarity (approved):

src/SingleShareClass.sol
mustermeiszer: @onnovisser would be better suited to judge that.

If the UI can do that, let's remove it.

PR #77, opened 2025-02-04
Refactor/cleaning. Use ERC6909 as much as possible

lemunozm:
### Description

- Valuation contracts no longer know about `AssetManager`, instead they know about generic ERC6909
- Remove the deprecated `IERC20Metadata` interface

### DIFF
--- a/src/BaseValuation.sol
+++ b/src/BaseValuation.sol
@@ -7,22 +7,21 @@ import {D18} from "src/types/D18.sol";
 import {Conversion} from "src/libraries/Conversion.sol";
 
 import {IBaseValuation} from "src/interfaces/IBaseValuation.sol";
-import {IERC20Metadata} from "src/interfaces/IERC20Metadata.sol";
-import {IAssetManager} from "src/interfaces/IAssetManager.sol";
+import {IERC6909MetadataExt} from "src/interfaces/ERC6909/IERC6909MetadataExt.sol";
 
 import {Auth} from "src/Auth.sol";
 
 abstract contract BaseValuation is Auth, IBaseValuation {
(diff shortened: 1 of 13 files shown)


wischli (approved):
:broom: Watch out, cleaning Luis in da house!

lemunozm:
I know you missed Mr. propper! :P

src/BaseValuation.sol
```diff
@@ -7,29 +7,28 @@ import {D18} from "src/types/D18.sol";
...
-import {IAssetManager} from "src/interfaces/IAssetManager.sol";
+import {IERC6909MetadataExt} from "src/interfaces/ERC6909/IERC6909MetadataExt.sol";
 
 import {Auth} from "src/Auth.sol";
 
 abstract contract BaseValuation is Auth, IBaseValuation {
-    /// @notice AssetManager dependency.
-    IAssetManager public assetManager;
+    /// @notice ERC6909 dependency.
+    IERC6909MetadataExt public erc6909;
```
peculiarity, line 16: This is exactly what I was trying to avoid by putting `IERC6909MetadataExt` being inherited by `IAssetManager`.  As a dependency you are passing an AssetManager instance anyways. Also this  base eval becomes to couples with the idea of ERC6909. It doesn't need to know about this implementation detail., Heck... even the implementation has nothing to do with `ERC6909` because it takes the asset from a mapping `mapping(AssetId => Asset )`.

What's the benefit of this change?

lemunozm: The benefit is that you can now use any valuation with any ERC6909, makes them much more generic for using in other contexts

peculiarity (approved):

PR #78, opened 2025-02-04
refactor: apply custom types to ShareClassManager

### LINKED ISSUE #56: Add custom `ShareClassId` type in ShareClassManager
Ref: https://github.com/centrifuge/pools-internal/pull/35#discussion_r1913065081

* Might include refactoring `ShareClassId` generation in `SingleShareClassManager.addShareClass`

### LINKED ISSUE #57: Add custom `InvestorAddress` type
Ref: https://github.com/centrifuge/pools-internal/pull/35#discussion_r1913782249

* Needs to be future-proof for Solana bytes32 account addresses
* TBD if it can be abstracted away from lower level contracts such as `SingleShareClass`

### LINKED ISSUE #60: Custom `AssetId` type
In `SingleShareClass` and potentially other contracts use a custom `AssetId` instead of the primitive `address` for assets. This is necessary in order to know the source chain of the particular asset. Moreover, we can make use of attached methods such as `.address()` or `.erc20()`.

Ref: https://github.com/centrifuge/pools-internal/pull/35/files#r1917086320

wischli:
Fixes #56, #57 and #60


test/unit/SingleShareClass.t.sol
```diff
@@ -13,14 +13,18 @@ import {MathLib} from "src/libraries/MathLib.sol";
...
 uint64 constant POOL_ID = 42;
-bytes16 constant SHARE_CLASS_ID = bytes16(uint128(POOL_ID));
+ShareClassId constant SHARE_CLASS_ID = ShareClassId.wrap(uint128(POOL_ID));
 address constant POOL_CURRENCY = address(840);
-address constant USDC = address(0x0123456);
-address constant OTHER_STABLE = address(0x01234567);
+AssetId constant USDC = AssetId.wrap(69);
+address constant USDC_ADDR = address(uint160(AssetId.unwrap(USDC)));
+AssetId constant OTHER_STABLE = AssetId.wrap(1337);
+address constant OTHER_STABLE_ADDR = address(uint160(AssetId.unwrap(OTHER_STABLE)));
```
lemunozm: Untra Nit. `AssetId` has a `.addr()`  method you can call if needed.

test/unit/SingleShareClass.t.sol
```diff
@@ -284,8 +294,8 @@ contract SingleShareClassDepositsNonTransientTest is SingleShareClassBaseTest {
         uint128 depositAmountOther = uint128(bound(depositAmount, 1e8, MAX_REQUEST_AMOUNT));
         D18 approvalRatioUsdc = d18(uint128(bound(approvalRatio, 1e14, 1e18)));
         D18 approvalRatioOther = d18(uint128(bound(approvalRatio, 1e14, 1e18)));
-        address investorUsdc = makeAddr("investorUsdc");
-        address investorOther = makeAddr("investorOther");
+        bytes32 investorUsdc = bytes32(uint256(uint160(makeAddr("investorUsdc"))));
```
lemunozm: Ultra NIT. Today I've learn that exists a `byte20` for this cases: `bytes32(bytes20(makeAddr("investorUsdc")))`

lemunozm (approved):
Thanks a lot! Everything will fit super smoothly after merging this

test/unit/SingleShareClass.t.sol
peculiarity: @lemunozm And that is completely different value. 
`bytes32(bytes20(makeAddr("investorUsdc"))) != bytes32(uint256(uint160(makeAddr("investorUsdc"))))` 

Padding happens at the end if you use the bytes when padding happens in the beginning when you use the uint

lemunozm: Interesting, good point here!

test/unit/SingleShareClass.t.sol
wischli: I am actually using that but forgot to remove the constant, good catch!

test/unit/SingleShareClass.t.sol
wischli: @peculiarity What's your recommendation: Current implementation or `bytes32(bytes20(...))`? Here, I didn't care because the 32 bytes are arbitrary.

peculiarity: Actually that's the thing. Probably here it doesn't matter but in reality we pass from the CV side the controller's address using the way @lemunozm suggested. I will go with @lemunozm  suggestion because it will reflect the actual flow..

wischli: Awesome thanks for raising that. I will apply the same transition to the unit tests here as well such that we have a consistent mapping from bytes20- to  bytes32-addresses.

lemunozm: I would say, for debugging, the best way could be `bytes32("investorUsdc")` directly. To see in the test traces the name. `makeAddr()` I think lost its purpose if the address is used as `bytes32`

src/SingleShareClass.sol
```diff
@@ -344,8 +352,49 @@ contract SingleShareClass is Auth, ISingleShareClass {
...
+    /// @param navPerShare Total value of assets of the pool and share class per share
+    /// @param valuation Source of truth for quotas, e.g. the price of a pool amount in payout asset
+    /// @param poolCurrency The address of the pool currency
+    /// @param epochAmounts_ Epoch ratio storage for the amount of revoked share class tokens and the corresponding
+    /// amount
+    /// in payout asset
+    /// @param totalIssuance_ Total issuance of share class tokens before revoking
+    /// @param epochId_ Identifier of the epoch for which we revoke
+    /// @return payoutPoolAmount Converted amount of pool currency based on number of revoked shares
+    function revokeEpochShares(
```
peculiarity: Since this is an internal function , please use the prefix "_"

wischli: [9d05452](https://github.com/centrifuge/pools-internal/pull/78/commits/9d05452b91869cd5a87f295df7465aad184b21d7)

src/SingleShareClass.sol
```diff
@@ -629,7 +636,10 @@ contract SingleShareClass is Auth, ISingleShareClass {
     ///
     /// @param poolId Identifier of the pool.
     /// @param shareClassId_ Identifier of the share class to be checked.
-    function _ensureShareClassExists(PoolId poolId, bytes16 shareClassId_) private view {
-        require(shareClassId[poolId] == shareClassId_, IShareClassManager.ShareClassNotFound());
+    function _ensureShareClassExists(PoolId poolId, ShareClassId shareClassId_) private view {
+        require(
+            ShareClassId.unwrap(shareClassId[poolId]) == ShareClassId.unwrap(shareClassId_),
+            IShareClassManager.ShareClassNotFound()
```
peculiarity:
```suggestion
            ShareClassNotFound()
```

src/SingleShareClass.sol
```diff
@@ -629,7 +636,10 @@ contract SingleShareClass is Auth, ISingleShareClass {
     ///
     /// @param poolId Identifier of the pool.
     /// @param shareClassId_ Identifier of the share class to be checked.
-    function _ensureShareClassExists(PoolId poolId, bytes16 shareClassId_) private view {
-        require(shareClassId[poolId] == shareClassId_, IShareClassManager.ShareClassNotFound());
+    function _ensureShareClassExists(PoolId poolId, ShareClassId shareClassId_) private view {
+        require(
+            ShareClassId.unwrap(shareClassId[poolId]) == ShareClassId.unwrap(shareClassId_),
```
peculiarity: Nit bit: What about having a function `equals`?
```
shareClassId[poolId].equals(shareClassId_),
```

```
function equals(ShareClassId scId1, ShareClassId scid2) pure returns (bool) {
    return ShareClassId.unwrap(scId1) == ShareClassId.unwrap(scid2);
}

using {isNull, equals} for ShareClassId global;
```

peculiarity (approved):

src/SingleShareClass.sol
wischli: I have tried everything I could think of to get it working but the compiler does not recognize `equals` for the ShareClassId. Can you help me how to properly apply it?

[ac13c5f](https://github.com/centrifuge/pools-internal/pull/78/commits/ac13c5f11915ff07b59527c6a1b83332dc7b912d)

src/SingleShareClass.sol
```diff
@@ -314,7 +314,7 @@ contract SingleShareClass is Auth, ISingleShareClass {
         IERC7726 valuation,
         uint32 endEpochId
     ) public auth returns (uint128 payoutAssetAmount, uint128 payoutPoolAmount) {
-        _ensureShareClassExists(poolId, shareClassId_);
+        require(shareClassId_.equals(shareClassId[poolId]), ShareClassNotFound);
```
peculiarity: `ShareClassNotFound()` :sweat_smile:

PR #80, opened 2025-02-05
Remove PoolRegistry address storage

lemunozm:
We no longer need to store the escrow address in the PoolRegistry, because we do not have escrow contracts and the address can be derived automatically from `(poolId, ShareClassId, EscrowId)` being `EscrowId` either `shareClassEscrow` or `pendingShareClassEscrow`


wischli (approved):
Escrows, get lost!

onnovisser (approved):

PR #81, opened 2025-02-05
Feat: Gateway impl

lemunozm:
WIP

### DIFF
--- a/src/Gateway.sol
+++ b/src/Gateway.sol
@@ -6,6 +6,8 @@ import {AssetId} from "src/types/AssetId.sol";
 import {PoolId} from "src/types/PoolId.sol";
 
 import {IGateway} from "src/interfaces/IGateway.sol";
+import {IMessageHandler} from "src/interfaces/IMessageHandler.sol";
+import {IRouter} from "src/interfaces/IRouter.sol";
 import {IPoolManagerHandler} from "src/interfaces/IPoolManager.sol";
 
 import {CastLib} from "src/libraries/CastLib.sol";
@@ -14,11 +16,7 @@ import {MessageType, MessageLib} from "src/libraries/MessageLib.sol";
 
 import {Auth} from "src/Auth.sol";
 
-interface IRouter {
-    function sendMessage(uint32 chainId, bytes memory message) external;
-}
-
-contract Gateway is Auth, IGateway {
+contract Gateway is Auth, IGateway, IMessageHandler {
     using MessageLib for bytes;
     using BytesLib for bytes;
     using CastLib for string;
@@ -28,12 +26,13 @@ contract Gateway is Auth, IGateway {
     IRouter public router;
     IPoolManagerHandler public handler;
 
-    constructor(IRouter router_, address deployer) Auth(deployer) {
+    constructor(IRouter router_, IPoolManagerHandler handler_, address deployer) Auth(deployer) {
         router = router_;
+        handler = handler_;
     }
 
     function sendNotifyPool(uint32 chainId, PoolId poolId) external auth {
-        router.sendMessage(chainId, abi.encodePacked(MessageType.AddPool, poolId.raw()));
+        router.send(chainId, abi.encodePacked(MessageType.AddPool, poolId.raw()));
     }
 
     function sendNotifyShareClass(
@@ -45,7 +44,7 @@ contract Gateway is Auth, IGateway {
         uint8 decimals,
         bytes32 hook
     ) external auth {
-        router.sendMessage(
+        router.send(
             chainId,
             abi.encodePacked(
                 MessageType.AddTranche,
@@ -64,7 +63,7 @@ contract Gateway is Auth, IGateway {
             ? abi.encodePacked(MessageType.AllowAsset, poolId.raw(), assetId.raw())
             : abi.encodePacked(MessageType.DisallowAsset, poolId.raw(), assetId.raw());
 
-        router.sendMessage(assetId.chainId(), message);
+        router.send(assetId.chainId(), message);
     }
 
     function sendFulfilledDepositRequest(
@@ -75,7 +74,7 @@ contract Gateway is Auth, IGateway {
         uint128 shares,
         uint128 investedAmount
     ) external auth {
-        router.sendMessage(
+        router.send(
             assetId.chainId(),
             abi.encodePacked(
                 MessageType.FulfilledDepositRequest,
@@ -97,7 +96,7 @@ contract Gateway is Auth, IGateway {
         uint128 shares,
         uint128 investedAmount
     ) external auth {
-        router.sendMessage(
+        router.send(
             assetId.chainId(),
             abi.encodePacked(
                 MessageType.FulfilledRedeemRequest,
@@ -118,7 +117,7 @@ contract Gateway is Auth, IGateway {
         bytes32 investor,
         uint128 cancelledAmount
     ) external auth {
-        router.sendMessage(
+        router.send(
             assetId.chainId(),
             abi.encodePacked(
                 MessageType.FulfilledCancelDepositRequest,
@@ -139,7 +138,7 @@ contract Gateway is Auth, IGateway {
         bytes32 investor,
         uint128 cancelledShares
     ) external auth {
-        router.sendMessage(
+        router.send(
             assetId.chainId(),
             abi.encodePacked(
                 MessageType.FulfilledCancelRedeemRequest,
@@ -153,12 +152,12 @@ contract Gateway is Auth, IGateway {
     }
 
     function sendUnlockAssets(AssetId assetId, bytes32 receiver, uint128 assetAmount) external auth {
-        router.sendMessage(
+        router.send(
             assetId.chainId(), abi.encodePacked(MessageType.TransferAssets, assetId.raw(), receiver, assetAmount)
         );
     }
 
-    function handleMessage(bytes calldata message) external auth {
+    function handle(bytes calldata message) external auth {
         MessageType kind = message.messageType();
 
         if (kind == MessageType.RegisterAsset) {
@@ -170,7 +169,7 @@ contract Gateway is Auth, IGateway {
             );
         } else if (kind == MessageType.TransferAssets) {
             handler.handleLockedTokens(
-                AssetId.wrap(message.toUint128(1)), address(bytes20(message.toBytes32(9))), message.toUint128(39)
+                AssetId.wrap(message.toUint128(1)), address(bytes20(message.toBytes32(16))), message.toUint128(49)
             );
         } else if (kind == MessageType.DepositRequest) {
             handler.handleRequestDeposit(
--- a/src/Gateway.sol
+++ b/src/Gateway.sol
@@ -1,0 +1,131 @@
+// SPDX-License-Identifier: BUSL-1.1
+pragma solidity 0.8.28;
+
+import {ShareClassId} from "src/types/ShareClassId.sol";
+import {AssetId} from "src/types/AssetId.sol";
+import {PoolId} from "src/types/PoolId.sol";
+
+import {IGateway} from "src/interfaces/IGateway.sol";
+import {IPoolManagerHandler} from "src/interfaces/IPoolManager.sol";
+
+import {CastLib} from "src/libraries/CastLib.sol";
+
+import {Auth} from "src/Auth.sol";
+
+import {
+    MessageType,
+    SerializationLib,
+    DeserializationLib,
+    RegisterAssetMsg,
+    AddPoolMsg,
+    AddTrancheMsg,
+    AllowAssetMsg,
+    DisallowAssetMsg,
+    TransferAssetsMsg
+} from "src/libraries/MessageLib.sol";
+
+interface IRouter {
+    function sendMessage(uint32 chainId, bytes memory message) external;
+}
+
+contract Gateway is Auth, IGateway {
+    using DeserializationLib for bytes;
+    using CastLib for string;
+    using CastLib for bytes;
+    using CastLib for bytes32;
+
+    IRouter public router;
+    IPoolManagerHandler public handler;
+
+    constructor(IRouter router_, address deployer) Auth(deployer) {
+        router = router_;
+    }
+
+    function sendNotifyPool(uint32 chainId, PoolId poolId) external auth {
+        router.sendMessage(chainId, AddPoolMsg(poolId.raw()).serialize());
+    }
+
+    function sendNotifyShareClass(
+        uint32 chainId,
+        PoolId poolId,
+        ShareClassId scId,
+        string memory name,
+        string memory symbol,
+        uint8 decimals,
+        bytes32 hook
+    ) external auth {
+        router.sendMessage(
+            chainId,
+            AddTrancheMsg(poolId.raw(), scId.toBytes(), bytes(name), symbol.toBytes32(), decimals, hook).serialize()
+        );
+    }
+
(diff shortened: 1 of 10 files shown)


src/libraries/MessageLib.sol
```diff
@@ -0,0 +1,68 @@
...
+}
+
+function serialize(RegisterAssetMsg calldata data) pure returns (bytes memory) {
+    return abi.encodePacked(uint8(Call.RegisterAsset), data.assetId, data.name, data.symbol, data.decimals);
+}
+
+// Others serialize methods
+// ...
+
+using {serialize} for RegisterAssetMsg global;
```
lemunozm: I would like to propose the following structure (which differs from the current CV implementation) for serializing and deserializing messages.

I think PROs from this structure are:
1. All protocol-related stuff in the same file. We do not serialize in one contract and deserialize in another one.
1. Serializing & deserializing are extracted from the business logic, so now it's much easier to test them.
1. Using a struct to define the body of a message forces that when one field is removed or moved around, we obtain a solidity compilation error in both serialize & deserialize methods. Which is super great for avoiding annoying bugs.
1. If we merge CV and CP we can use a single file for all serializations

Usage later in the business logic part:
```solidity
bytes memory message = RegisterAssetMsg(usdc, "USD Coin", "USDC", 6).serialize();

using DeserializationLib for bytes;
RegisterAssetMsg memory value = message.deserializeRegisterAsset();
```

lemunozm: After much testing and playing around with it, I've decided to go with hardcoding the serialization/deserialization in the Gateway without structs, at least for the very beginning PoC.

We still have properties 1. and 2. from above because the Gateway is quite agnostic from the business logic, and everything is in the same file.

We lose property 3 of the above, but anyway, the typing in solidity is very soft.

src/Gateway.sol
```diff
@@ -0,0 +1,209 @@
...
+                scId.raw(),
+                assetId.raw(),
+                investor,
+                cancelledAmount,
+                cancelledAmount
+            )
+        );
+    }
+
+    function sendFulfilledCancelRedeemRequest(
```
peculiarity, line 145: This comment applies to all these custom `send` functions.
You can have a single `send` method which accepts the already encoded message. This is how it is done in CV as well. Gateway is responsible just sending the message to different routers  and not constructing the msg itself. Also as I mentioned, we will need an array of routers.

lemunozm: Yeah, I thought about that, but I do not like the fact that the serialization happens in the `PoolManager` and the deserialization in the `Gateway`. For testing purposes, and organizational, it's easier to have both in this contract. 

PoolManager should not know about the internal protocol encoding; he should say that he wants to send some information, without knowing about the how

src/libraries/BytesLib.sol
```diff
@@ -0,0 +1,174 @@
+// SPDX-License-Identifier: BUSL-1.1
```
peculiarity, line 1: Missing  tests  for the libs

lemunozm: File copied from CV.

src/Gateway.sol
mustermeiszer: Agree with Luis. But maybe I am missing something wrt. to upgrading and extending the gateway functionality later.

src/libraries/BytesLib.sol
lemunozm: What we should test, as a reminder for ourselves, is `CastLib.stringToBytes128`, which is a new utility needed that wasn't in CV

src/interfaces/IGateway.sol
```diff
@@ -5,9 +5,24 @@ import {ShareClassId} from "src/types/ShareClassId.sol";
 import {AssetId} from "src/types/AssetId.sol";
 import {PoolId} from "src/types/PoolId.sol";
 
-// TODO: WIP
+/// @notice Interface for a gateway
```
wischli: Nit: It appears to me that the Gateway is solely used for message dispatch but not reception. 
```suggestion
/// @notice Interface for dispatch-only gateway
```

mustermeiszer (approved):
Looks good! 

QUESTION
* Do I get it right, that destination `IROuter` will store all adapters and logic to route to the right destination? E.g. locally if on same system?

src/interfaces/IGateway.sol
```diff
@@ -5,9 +5,24 @@ import {ShareClassId} from "src/types/ShareClassId.sol";
...
+    /// @notice Updates a contract parameter.
+    /// @param what Name of the parameter to update.
+    /// Accepts a `bytes32` representation of 'router' and 'handler' string values.
+    /// @param data New value given to the `what` parameter
+    function file(bytes32 what, address data) external;
+
+    /// @notice Creates and send the message
     function sendNotifyPool(uint32 chainId, PoolId poolId) external;
+
+    /// @notice Creates and send the message
```
wischli, line 25: Nit: Incomplete natspec but let's do that pre-audit as it's mostly copypasta from other calls.

src/interfaces/IRouter.sol
```diff
@@ -0,0 +1,10 @@
+// SPDX-License-Identifier: BUSL-1.1
+pragma solidity 0.8.28;
+
+/// @notice Generic interface for entities that represents routers
```
wischli: Nit:
```suggestion
/// @notice Generic interface for entities that represent routers
```

src/Gateway.sol
lemunozm: Will probably face that problem when we add the adapters part 👍🏻

lemunozm:
> Do I get it right, that destination IROuter will store all adapters and logic to route to the right destination? E.g. locally if on same system?

`IRouter` is just a placeholder. Actually, it should be all the adapters. For sure, something WIP in this moment that probably will change. (Added a TODO and rename IRouter to IAdapter)

src/interfaces/IRouter.sol
```diff
@@ -0,0 +1,10 @@
...
+interface IRouter {
```
wischli: Nit: AFAICT, this contract should rather be named `IAdapter` in order to follow naming of CV. However, I believe that will be solved by merging into a monorepo in which the `IAdapter` surely is a common contract.

src/Gateway.sol
```diff
@@ -0,0 +1,220 @@
...
+        uint128 investedAmount
+    ) external auth {
+        router.send(
+            assetId.chainId(),
+            abi.encodePacked(
+                MessageType.FulfilledDepositRequest,
+                poolId.raw(),
+                scId.raw(),
+                assetId.raw(),
+                investor,
```
wischli: CR: Incorrect order ([ref](https://centrifuge.hackmd.io/Soy4JiFbRSyovj53KW-a4A?view#FulfilledDepositRequest))
```suggestion
                investor,
                assetId.raw(),
```

src/Gateway.sol
```diff
@@ -0,0 +1,220 @@
...
+        uint128 investedAmount
+    ) external auth {
+        router.send(
+            assetId.chainId(),
+            abi.encodePacked(
+                MessageType.FulfilledRedeemRequest,
+                poolId.raw(),
+                scId.raw(),
+                assetId.raw(),
+                investor,
```
wischli: CR: Incorrect order ([ref](https://centrifuge.hackmd.io/Soy4JiFbRSyovj53KW-a4A?view#FulfilledRedeemRequest))

```suggestion
                investor,
                assetId.raw(),
```

src/Gateway.sol
```diff
@@ -0,0 +1,220 @@
...
+        uint128 cancelledAmount
+    ) external auth {
+        router.send(
+            assetId.chainId(),
+            abi.encodePacked(
+                MessageType.FulfilledCancelDepositRequest,
+                poolId.raw(),
+                scId.raw(),
+                assetId.raw(),
+                investor,
```
wischli: CR: Incorrect order ([ref](https://centrifuge.hackmd.io/Soy4JiFbRSyovj53KW-a4A?view#FulfilledCancelDepositRequest))
```suggestion
                investor,
                assetId.raw(),
```

src/Gateway.sol
```diff
@@ -0,0 +1,220 @@
...
+        uint128 cancelledShares
+    ) external auth {
+        router.send(
+            assetId.chainId(),
+            abi.encodePacked(
+                MessageType.FulfilledCancelRedeemRequest,
+                poolId.raw(),
+                scId.raw(),
+                assetId.raw(),
+                investor,
```
wischli: CR: Incorrect order ([ref](https://centrifuge.hackmd.io/Soy4JiFbRSyovj53KW-a4A?view#FulfilledCancelRedeemRequest))
```suggestion
                investor,
                assetId.raw(),
```

src/Gateway.sol
```diff
@@ -0,0 +1,220 @@
...
+                AssetId.wrap(message.toUint128(25)),
+                message.toBytes32(41),
+                message.toUint128(73)
+            );
+        } else if (kind == MessageType.RedeemRequest) {
+            handler.handleRequestRedeem(
+                PoolId.wrap(message.toUint64(1)),
+                ShareClassId.wrap(message.toBytes16(9)),
+                AssetId.wrap(message.toUint128(25)),
+                message.toBytes32(41),
```
wischli: CR: Same as above Same as above ([ref spec](https://centrifuge.hackmd.io/Soy4JiFbRSyovj53KW-a4A?view#RedeemRequest))
```suggestion
                message.toBytes32(25),
                AssetId.wrap(message.toUint128(57)),
```

src/Gateway.sol
```diff
@@ -0,0 +1,220 @@
...
+        } else if (kind == MessageType.TransferAssets) {
+            handler.handleLockedTokens(
+                AssetId.wrap(message.toUint128(1)), address(bytes20(message.toBytes32(16))), message.toUint128(49)
+            );
+        } else if (kind == MessageType.DepositRequest) {
+            handler.handleRequestDeposit(
+                PoolId.wrap(message.toUint64(1)),
+                ShareClassId.wrap(message.toBytes16(9)),
+                AssetId.wrap(message.toUint128(25)),
+                message.toBytes32(41),
```
wischli: CR: Should be `57` instead because the investor address is received as 32 bytes, i.e. from 25 to 56. I suppose the issue is that you mixed up the order of the 32byte investor address and the uin128 assetId in all messages ([ref spec](https://centrifuge.hackmd.io/Soy4JiFbRSyovj53KW-a4A?view#DepositRequest))
```suggestion
                message.toBytes32(25),
                AssetId.wrap(message.toUint128(57)),
```

src/Gateway.sol
```diff
@@ -0,0 +1,220 @@
...
+                message.toUint128(73)
+            );
+        } else if (kind == MessageType.CancelDepositRequest) {
+            handler.handleCancelDepositRequest(
+                PoolId.wrap(message.toUint64(1)),
+                ShareClassId.wrap(message.toBytes16(9)),
+                AssetId.wrap(message.toUint128(25)),
+                message.toBytes32(41)
```
wischli: CR: Same as above Same as above ([ref spec](https://centrifuge.hackmd.io/Soy4JiFbRSyovj53KW-a4A?view#CancelDepositRequest))
```suggestion
                message.toBytes32(25),
                AssetId.wrap(message.toUint128(57)),
```

src/Gateway.sol
```diff
@@ -0,0 +1,220 @@
...
+            );
+        } else if (kind == MessageType.CancelRedeemRequest) {
+            handler.handleCancelRedeemRequest(
+                PoolId.wrap(message.toUint64(1)),
+                ShareClassId.wrap(message.toBytes16(9)),
+                AssetId.wrap(message.toUint128(25)),
+                message.toBytes32(41)
```
wischli: Same as above ([ref spec](https://centrifuge.hackmd.io/Soy4JiFbRSyovj53KW-a4A?view#CancelRedeemRequest))
```suggestion
                message.toBytes32(25),
                AssetId.wrap(message.toUint128(57)),
```

wischli (commented):
Apart from the mixed up order for `bytes32 investor` and `uint128 assetId` in all messages, LGTM! I can imagine some namings to be refactored as part of merging into a monorepo.

AFAICT, CV uses `Adapter` for our `Router` here and `Router` for our `MessageHandler`.

src/Gateway.sol
```diff
@@ -0,0 +1,220 @@
...
+    function sendNotifyShareClass(
+        uint32 chainId,
+        PoolId poolId,
+        ShareClassId scId,
+        string memory name,
+        string memory symbol,
+        uint8 decimals,
+        bytes32 hook
+    ) external auth {
+        router.send(
```
wischli: Missed to flag the missing `call_type` message id serialization which is necessary for CV deserialization: https://github.com/centrifuge/vaults-internal/blob/ce5a95c3eb8aa7b4069c2919be3cc9037eb48967/src/InvestmentManager.sol#L192

src/Gateway.sol
```diff
@@ -0,0 +1,220 @@
...
+            assetId.chainId(), abi.encodePacked(MessageType.TransferAssets, assetId.raw(), receiver, assetAmount)
+        );
+    }
+
+    function handle(bytes calldata message) external auth {
+        MessageType kind = message.messageType();
+
+        if (kind == MessageType.RegisterAsset) {
+            handler.handleRegisterAsset(
```
wischli, line 173: Missed to flag the missing deserialization of the `call_type` message id which is part of the CV-serialized message: https://github.com/centrifuge/vaults-internal/blob/ce5a95c3eb8aa7b4069c2919be3cc9037eb48967/src/InvestmentManager.sol#L83

wischli (commented):
We are also missing the (de-)serialization of the message id.

src/interfaces/IGateway.sol
lemunozm: Correct for the IGateway 👍🏻

src/interfaces/IRouter.sol
lemunozm: I can rename it by now to avoid confusions

src/Gateway.sol
lemunozm: Such a good catch!

src/interfaces/IGateway.sol
lemunozm: Agree. I am also not sure how verbose we want to be with auto-interpreted fields as `poolId` or `shareClassId`

src/Gateway.sol
lemunozm: Thanks for going deeper on this now, you've saved a lot of time in the future with these comments!

I'm a bit sad that we'll lose all the effort done in CC for cross-chain 
integration tests. But definitely we need them again here

src/Gateway.sol
lemunozm: The call type is encoded under `MessageType.AddTranche`, but it's true that maybe it's not encoded as just one bytes.

src/Gateway.sol
lemunozm: It is just above. Also note that first parameter start a `1`, skipping that value in the message

src/Gateway.sol
wischli: Ah sorry, you are right!

src/Gateway.sol
wischli: Ouch, great failure. Happy to see I was wrong!

wischli (approved):
Thanks for renaming to `Adapter` and fixing the minor issues! LGTM

PR #82, opened 2025-02-06
TransientValuation Support for different prices.

lemunozm:
When using `TransientValuation` in big multicalls, you will require fetching different values for different assets. It isn't easy to know at what moments you should call to `setPrice()` to set the price to the correct assets. Also, making a mistake is not a warning.

This PR allows having multiple prices for different pairs (base, quote) and also warns if a required price is not set.

### DIFF
--- a/src/TransientValuation.sol
+++ b/src/TransientValuation.sol
@@ -14,17 +14,21 @@ import {BaseValuation} from "src/BaseValuation.sol";
 
 contract TransientValuation is BaseValuation, ITransientValuation {
     /// @notice Temporal price set and used to obtain the quote.
-    D18 public /*TODO: transient*/ price;
+    mapping(address base => mapping(address quite => D18)) public /*TODO: transient*/ price;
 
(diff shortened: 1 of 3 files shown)


src/TransientValuation.sol
```diff
@@ -14,17 +14,21 @@ import {BaseValuation} from "src/BaseValuation.sol";
 
 contract TransientValuation is BaseValuation, ITransientValuation {
     /// @notice Temporal price set and used to obtain the quote.
-    D18 public /*TODO: transient*/ price;
+    mapping(address base => mapping(address quite => D18)) public /*TODO: transient*/ price;
```
wischli, line 17: Love the idea of a transient map!

wischli (approved):

src/TransientValuation.sol
lemunozm: Not sure if you suggested something some days ago, but definitelly, it is more robust

mustermeiszer (approved):
Love it!!

PR #85, opened 2025-02-11
Feat: CallEscrow

lemunozm:
# Description

Allow the FM to choose the contract that will be used to call the methods added in `PoolManager.execute()`. So the FM have control about the `msg.sender` used.

### Problem

As an example, a FM will want to perform the following:
```solidity
execute(PoolA)
   poolManager.approve()
   erc20.transfer() // or erc6909 or uniswap.swap ...
   poolManager.issue()
```
   
Nevertheless, the above has a problem. The `msg.sender` of the above calls is  `PoolLocker.multicall`. To be able to transfer, the FM needs to call `erc20.allowance()` for the multicall. This opens a security hole because other FM from other pools will also use the same multicall, or even the multicall can be used standalone. So, anyone can perform the `transfer()` on behalf.

### Solution

Optional call escrows. Similar concept to the escrows used for ERC20. 
- If the FM does not need to make permission and external calls, it can just use the current and default multicall. 
- If they need to identify the sender for external calls, they can configure an `ICallEscrow` contract address attached to the pool to perform allowances from other contracts. The FM owns that address and can trust it as the sender for other contracts.

i.e.
```solidity
// Configure OPTIONALY a call escrow and give transfer permisions
fm -> execute(PoolA)
         poolManager.setAddressFor("call-escrow", CALL_ESCROW_A);
fm -> erc20.allowance(CALL_ESCROW_A);
         
// Use as normal         
fm -> execute(PoolA)
         poolManager.approve()
         erc20.transfer() // or erc6909 or uniswap.swap ...
         poolManager.issue()
```

## Changes

- Revert the removal of `PoolRegisty.addresFor()` to store the multicall escrow address.
- Add `ICallEscrow` that just wrap a call to be the sender of the wrapped call.
- Adapt `PoolLocker` to use optionally the call escrow

### DIFF
--- a/src/CallEscrow.sol
+++ b/src/CallEscrow.sol
@@ -1,0 +1,14 @@
+// SPDX-License-Identifier: MIT
+pragma solidity ^0.8.28;
+
+import {ICallEscrow} from "src/interfaces/ICallEscrow.sol";
+import {Auth} from "src/Auth.sol";
+
+contract CallEscrow is Auth, ICallEscrow {
+    constructor(address deployer) Auth(deployer) {}
+
+    /// @inheritdoc ICallEscrow
+    function call(address target, bytes calldata data) external returns (bool success, bytes memory results) {
+        return target.call(data);
+    }
+}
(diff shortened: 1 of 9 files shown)


lemunozm:
CR. As an addition, we should avoid exposing the `aggregate()` method as public, to avoid somebody from bypassing the pool locker checks

mustermeiszer:
I do not understand how this solves the security issue tbh. Can not anybody just again use the `CallEscrow` from other pools?

lemunozm:
Closed in order to handle this issue at a higher level

PR #86, opened 2025-02-11
Check allow holding asset automatically

lemunozm:
Thread: https://kflabs.slack.com/archives/C07PG2EUR9C/p1739272539175059

### DIFF
--- a/src/PoolManager.sol
+++ b/src/PoolManager.sol
@@ -154,17 +154,6 @@ contract PoolManager is Auth, PoolLocker, IPoolManager, IPoolManagerHandler {
     }
 
     /// @inheritdoc IPoolManagerAdminMethods
-    function allowHoldingAsset(AssetId assetId, bool allow) external poolUnlocked {
-        PoolId poolId = unlockedPoolId();
-
-        if (!allow) {
-            require(!poolRegistry.isInvestorAssetAllowed(poolId, assetId), InvestorAssetStillAllowed());
-        }
-
-        holdings.allowAsset(poolId, assetId, allow);
-    }
-
-    /// @inheritdoc IPoolManagerAdminMethods
     function allowInvestorAsset(AssetId assetId, bool allow) external poolUnlocked {
         PoolId poolId = unlockedPoolId();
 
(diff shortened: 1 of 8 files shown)


src/PoolManager.sol
```diff
@@ -153,17 +153,6 @@ contract PoolManager is Auth, PoolLocker, IPoolManager, IPoolManagerHandler {
         poolRegistry.updateAdmin(unlockedPoolId(), account, allow);
     }
 
-    /// @inheritdoc IPoolManagerAdminMethods
-    function allowHoldingAsset(AssetId assetId, bool allow) external poolUnlocked {
-        PoolId poolId = unlockedPoolId();
-
-        if (!allow) {
-            require(!poolRegistry.isInvestorAssetAllowed(poolId, assetId), InvestorAssetStillAllowed());
-        }
```
lemunozm, line 162: We get rid of this weird logic!

mustermeiszer (approved):

src/PoolManager.sol
```diff
@@ -153,17 +153,6 @@ contract PoolManager is Auth, PoolLocker, IPoolManager, IPoolManagerHandler {
...
-
-        holdings.allowAsset(poolId, assetId, allow);
-    }
-
     /// @inheritdoc IPoolManagerAdminMethods
     function allowInvestorAsset(AssetId assetId, bool allow) external poolUnlocked {
```
wischli, line 157: Based on the mentioned [Slack thread](https://kflabs.slack.com/archives/C07PG2EUR9C/p1739272539175059), you proposed
> inside `allowInvestorAsset()`, as a pre-check to allow the investor asset, check if there is a holding created for a pair `scId` and `assetId` as indicator for allowing holding asset.

Curious why this is missing here.

lemunozm: The check exists in line 161 at `isAssetAllowed()`, that method will return true if there has been created a holding for that asset.

wischli (approved):
Thanks for clearing my question. Ready to be meeeeeeeerged

PR #88, opened 2025-02-13
chore: ensure SingleShareClass rounds towards system

### LINKED ISSUE #54: Ensure proper rounding in `SingleShareClass`
Follow up from #35

The goal of this ticket is to
1. Add rounding variants of D18 methods such as `mulUint256` similar to https://github.com/centrifuge/liquidity-pools/blob/e556c1a7a0ec7f6d700b47841eb586f5f4801406/src/libraries/MathLib.sol#L105 ([ref comment](https://github.com/centrifuge/pools-internal/pull/35#discussion_r1916538358))
2. Ensure rounding in `SingleShareClass` works properly: I.e. when subtracting in `_update{Deposit, Redeem}Request`, we should not over-/underflow as well as ensure that being off by `1` is not possible ([Slack discussion](https://kflabs.slack.com/archives/C07PG2EUR9C/p1736940247983679))
  * Result from issues on Centrifuge Chain where some states cannot be cleared due to rounding issues leaving states with `1` amounts which cannot be cleared organically ([comment](https://github.com/centrifuge/pools-internal/pull/35#discussion_r1916257843))
  * Consider rounding up during issuance and else rounding down ([comment](https://github.com/centrifuge/pools-internal/pull/35#discussion_r1916271291))
  
 This ticket can be split into two separate ones for 1. and 2.

wischli:
Fixes #54

* Adds tests for rounding edge cases:
  * When there is only a single atom to claim (share or asset amount) for multiple investors, then no investor can claim it independent of the pending amount
  * When an even number of investors with similar pending amounts claims an uneven amount, there is a single atom surplus for the system 
  * When an uneven number of investors with similar pending amounts claims an even amount, there is a single atom surplus for the system

* Fixes issue in claiming when claimed amount is 0 but pending epoch amount `pending{Deposits, Redeems}[shareClassId][assetId]` was already decreased in approval such that we need to re-add the approved amount
  * When this happens, we don't emit `Claimed*` event
* Adds lower bound of `> 0` to approval rates
* Refactors SingleShareClass UT to use `1e18` pool denomination instead of `1e4`


src/SingleShareClass.sol
```diff
@@ -377,25 +379,35 @@ contract SingleShareClass is Auth, ISingleShareClass {
...
-                poolId,
-                shareClassId_,
-                epochId_,
-                investor,
-                depositAssetId,
-                approvedAssetAmount,
-                userOrder.pending,
-                claimableShareAmount
-            );
+            if (claimableShareAmount > 0) {
```
wischli: NOTE: Depending on the approval ratio and `navPerShare`, it is possible that the approved user amount is not zero but the claimable amount is zero.

test/unit/SingleShareClass.t.sol
```diff
@@ -929,17 +940,200 @@ contract SingleShareClassTransientTest is SingleShareClassBaseTest {
...
+        uint128 pendingRedeem =
+            claimRedeemSharePayementAmount == 0 ? redeemAmount : redeemAmount - epochAmounts.redeemSharesRevoked;
         _assertRedeemRequestEq(scId, USDC, investor, UserOrder(pendingRedeem, epochId));
         _assertEpochAmountsEq(scId, USDC, 2, epochAmounts);
         _assertEpochAmountsEq(scId, USDC, epochId, EpochAmounts(d18(0), d18(0), 0, 0, 0, 0, 0));
     }
 }
 
+///@dev Contains all deposit tests which deal with rounding edge cases
+contract SingleShareClassRoundingEdgeCasesDeposit is SingleShareClassBaseTest {
```
wischli: @mustermeiszer I'd love your approval on these 3 deposit edge cases to ensure my assumptions are not incorrect. Let me know if you want me to walk you through the tests. See the PR description for a general explanation about the tests.

test/unit/SingleShareClass.t.sol
```diff
@@ -929,17 +940,200 @@ contract SingleShareClassTransientTest is SingleShareClassBaseTest {
...
+
+        assertEq(claimedA, claimedB, "Claimed shares should be equal");
+        assertEq(claimedB, claimedC, "Claimed shares should be equal");
+        assertEq(claimedA + claimedB + claimedC + 1, issuedShares, "System should have 1 share class token surplus");
+        assert(paymentA != paymentB && paymentB != paymentC && paymentC != paymentA);
+    }
+}
+
+///@dev Contains all deposit tests which deal with rounding edge cases
+contract SingleShareClassRoundingEdgeCasesRedeem is SingleShareClassBaseTest {
```
wischli: @mustermeiszer I'd love your approval on these 3 redeem edge cases to ensure my assumptions are not incorrect. Let me know if you want me to walk you through the tests. See the PR description for a general explanation about the tests.

lemunozm:
Just a side note, but related to this PR I guess, we have some castings in this file from `uint128` to `int128` without the safe `MathLib.toInt128`. We should check that due:

- `2^127` will be turned into `0` from `int128` perspective, which can skip some checks. Someone can probably complain about it in the audit.

wischli:
> Just a side note, but related to this PR I guess, we have some castings in this file from `uint128` to `int128` without the safe `MathLib.toInt128`. We should check that due:
> 
> * `2^127` will be turned into `0` from `int128` perspective, which can skip some checks. Someone can probably complain about it in the audit.

Thanks for spotting this! Fixed in [a40b279](https://github.com/centrifuge/pools-internal/pull/88/commits/a40b279c04ea1936ad46df7b57f1d816ac02c12d)

lemunozm:
> Thanks for spotting this! Fixed in [a40b279](https://github.com/centrifuge/pools-internal/pull/88/commits/a40b279c04ea1936ad46df7b57f1d816ac02c12d)

Thanks! I think the `cancel(Redeem/Deposit)Request` also has an `-int128()` that should be protected.

Will perform a proper review about this later!

src/SingleShareClass.sol
lemunozm: In that case, should we also dispatch the `ClaimDeposit` event? (in the `else` part)

src/SingleShareClass.sol
```diff
@@ -432,26 +444,35 @@ contract SingleShareClass is Auth, ISingleShareClass {
...
+                    epochId_,
+                    investor,
+                    payoutAssetId,
+                    approvedShareAmount,
+                    userOrder.pending,
+                    claimableAssetAmount
+                );
+            } else {
+                // Increase pending by approved amount as it did not lead to claimable amount
+                pendingRedeem[shareClassId_][payoutAssetId] += approvedShareAmount;
```
lemunozm: Should we also modify the `userOrder.pending` amount with the approvedShareAmount which can be non zero? I mean, adding:

```solidity
userOrder.pending -= approvedShareAmount;
```

src/SingleShareClass.sol
```diff
@@ -377,25 +379,35 @@ contract SingleShareClass is Auth, ISingleShareClass {
...
+                    investor,
+                    depositAssetId,
+                    approvedAssetAmount,
+                    userOrder.pending,
+                    claimableShareAmount
+                );
+            } else {
+                // Increase pending by approved amount as it did not lead to claimable amount
+                pendingDeposit[shareClassId_][depositAssetId] += approvedAssetAmount;
+            }
```
lemunozm: Same here with the asset amount instead:
```solidity
userOrder.pending -= approvedAssetAmount;
```

lemunozm (commented):
~~Just a concern, otherwise~~ looks fine!

src/SingleShareClass.sol
lemunozm: My fault! pending is only modified if the claiming amount was success

src/SingleShareClass.sol
wischli: IMO that does not make sense as we would dispatch it with 0 amount. But thinking about it, could be preferred. WDYT @mustermeiszer ?

mustermeiszer: I would always dispatch the events. Otherwise indexing thinks something is missing.

test/unit/SingleShareClass.t.sol
```diff
@@ -929,17 +940,200 @@ contract SingleShareClassTransientTest is SingleShareClassBaseTest {
...
+
+        shareClass.requestDeposit(poolId, scId, 1, INVESTOR_A, USDC);
+        shareClass.requestDeposit(poolId, scId, approvedAssetAmount - 1, INVESTOR_B, USDC);
+        _approveAllDepositsAndIssue(issuedShares, navPerShare);
+
+        (uint128 claimedA, uint128 paymentA) = shareClass.claimDeposit(poolId, scId, INVESTOR_A, USDC);
+        (uint128 claimedB, uint128 paymentB) = shareClass.claimDeposit(poolId, scId, INVESTOR_B, USDC);
+
+        assertEq(claimedA, claimedB, "Claimed shares should be equal");
+        assertEq(claimedA + claimedB + 1, issuedShares, "System should have 1 share class token surplus");
```
mustermeiszer: CR

We can not isse more shares. We need to burn or not issue the delta. Is that possible?

test/unit/SingleShareClass.t.sol
```diff
@@ -929,17 +940,200 @@ contract SingleShareClassTransientTest is SingleShareClassBaseTest {
...
+
+        shareClass.requestDeposit(poolId, scId, 49 * approvedAssetAmount / 100, INVESTOR_A, USDC);
+        shareClass.requestDeposit(poolId, scId, 51 * approvedAssetAmount / 100, INVESTOR_B, USDC);
+        _approveAllDepositsAndIssue(issuedShares, navPerShare);
+
+        (uint128 claimedA, uint128 paymentA) = shareClass.claimDeposit(poolId, scId, INVESTOR_A, USDC);
+        (uint128 claimedB, uint128 paymentB) = shareClass.claimDeposit(poolId, scId, INVESTOR_B, USDC);
+
+        assertEq(claimedA, claimedB, "Claimed shares should be equal");
+        assertEq(claimedA + claimedB + 1, issuedShares, "System should have 1 share class token surplus");
```
mustermeiszer: CR

We can not isse more shares. We need to burn or not issue the delta. Is that possible?

test/unit/SingleShareClass.t.sol
```diff
@@ -929,17 +940,200 @@ contract SingleShareClassTransientTest is SingleShareClassBaseTest {
...
+        shareClass.requestDeposit(poolId, scId, 39 * approvedAssetAmount / 100, INVESTOR_C, USDC);
+        _approveAllDepositsAndIssue(issuedShares, navPerShare);
+
+        (uint128 claimedA, uint128 paymentA) = shareClass.claimDeposit(poolId, scId, INVESTOR_A, USDC);
+        (uint128 claimedB, uint128 paymentB) = shareClass.claimDeposit(poolId, scId, INVESTOR_B, USDC);
+        (uint128 claimedC, uint128 paymentC) = shareClass.claimDeposit(poolId, scId, INVESTOR_C, USDC);
+
+        assertEq(claimedA, claimedB, "Claimed shares should be equal");
+        assertEq(claimedB, claimedC, "Claimed shares should be equal");
+        assertEq(claimedA + claimedB + claimedC + 1, issuedShares, "System should have 1 share class token surplus");
```
mustermeiszer: CR

We can not isse more shares. We need to burn or not issue the delta. Is that possible?

mustermeiszer (commented):
Nice to have that covered. 
* Regarding the asset surplus: That stays in the pool, that is fine. 
* Regarding the share surplus: COULD we also just NOT issue this share-atom? If this is hard to fix, ignore my CRs.

### Open Questions
* Does this mean the single share class implementation DEFAULTS to 18 decimals everywhere?
* Would it be possible for a fund to choose a starting price for its shares or one share?

wischli:
> * Regarding the share surplus: COULD we also just NOT issue this share-atom? If this is hard to fix, ignore my CRs.

As discussed on Slack, we are ignoring this ([ref](https://kflabs.slack.com/archives/C07PG2EUR9C/p1739958929680789?thread_ts=1739958924.214889&cid=C07PG2EUR9C))

> * Does this mean the single share class implementation DEFAULTS to 18 decimals everywhere?

Right now, it defaults to the pool currency denomination. Technical debt to decouple that as discussed in Weekly Sync ([ref](https://www.notion.so/Weekly-Standup-1192eac24e1780419e0ac4cdf24850e9?pvs=4#19d2eac24e178169a5a1fa972ddf9725))

> * Would it be possible for a fund to choose a starting price for its shares or one share?

Yes via `navPerShare`

lemunozm (approved):
LGTM!

PR #89, opened 2025-02-18
Feat/ref: Allow investor asset per ShareClassId

lemunozm:
Thread: https://kflabs.slack.com/archives/C07PG2EUR9C/p1739369288420839

### Changes
- No longer need a `Holdings.isAllowHolding` mapping 🎉 
- No longer need a `PoolMananger.notifyInvestorAsset` call 🎉 
- Changes to pass CI due an update in forge / foundry.

NOTE. Despite the refactorizations, nothing changes in the requirements for allowing an asset. The mandatory flow is:
```
CV -> CP: registerAsset()
CP: createHolding() //creating a holding acts as allowing a holding asset
CP: allowInvestorAsset()
```

### DIFF
--- a/src/pools/PoolManager.sol
+++ b/src/pools/PoolManager.sol
@@ -135,13 +135,6 @@ contract PoolManager is Auth, PoolLocker, IPoolManager, IPoolManagerHandler {
     }
 
     /// @inheritdoc IPoolManagerAdminMethods
-    function notifyAllowedAsset(ShareClassId scId, AssetId assetId) external poolUnlocked {
-        PoolId poolId = unlockedPoolId();
-
-        gateway.sendNotifyAllowedAsset(poolId, scId, assetId, poolRegistry.isInvestorAssetAllowed(poolId, assetId));
-    }
-
-    /// @inheritdoc IPoolManagerAdminMethods
     function setPoolMetadata(bytes calldata metadata) external poolUnlocked {
         poolRegistry.setMetadata(unlockedPoolId(), metadata);
     }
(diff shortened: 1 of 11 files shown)


src/misc/libraries/MathLib.sol
```diff
@@ -60,7 +60,7 @@ library MathLib {
     /// @dev    Original credit to Remco Bloemen under MIT license (https://xn--2-umb.com/21/muldiv)
     ///         with further edits by Uniswap Labs also under MIT license.
     // slither-disable-start divide-before-multiply
-    function mulDiv(uint256 x, uint256 y, uint256 denominator) internal pure returns (uint256 result) {
+    function mulDiv(uint256 x, uint256 y, uint256 denominator) public pure returns (uint256 result) {
```
hieronx: This changes the library from embedded to external, which we almost certainly don't want. Why is this required? 

https://eip2535diamonds.substack.com/p/the-difference-between-solidity-libraries

lemunozm: Interesting link!

I did it due [this](https://kflabs.slack.com/archives/C07PG2EUR9C/p1739878179231799). But it seems like there's a config to allow this without modifying the `internal` modifier. Will do!

lemunozm: Updated!

lemunozm: Could you confirm @hieronx  if the same is true for free functions (as the ones found in `D18.sol`)? Should they also be `internal`?

lemunozm: BTW, the issue related to why the `InvestmentContract` in CV was increased in size was related to this. Fixed too.

hieronx: I am actually not sure about free functions, have honestly never used them!

lemunozm: After some searching, for reference, free functions have internal behavior automatically.

wischli: > This changes the library from embedded to external, which we almost certainly don't want. Why is this required?
> 
> https://eip2535diamonds.substack.com/p/the-difference-between-solidity-libraries

This was news to me, thanks for sharing that!

> After some searching, for reference, free functions have internal behavior automatically.

Confirming this with the [official Solidity docs](https://docs.soliditylang.org/en/latest/contracts.html):
> Functions outside of a contract, also called “free functions”, always have implicit internal [visibility](https://docs.soliditylang.org/en/latest/contracts.html#visibility-and-getters). Their code is included in all contracts that call them, similar to internal library functions.

src/pools/PoolManager.sol
```diff
@@ -134,13 +134,6 @@ contract PoolManager is Auth, PoolLocker, IPoolManager, IPoolManagerHandler {
         );
     }
 
-    /// @inheritdoc IPoolManagerAdminMethods
-    function notifyAllowedAsset(ShareClassId scId, AssetId assetId) external poolUnlocked {
-        PoolId poolId = unlockedPoolId();
-
-        gateway.sendNotifyAllowedAsset(poolId, scId, assetId, poolRegistry.isInvestorAssetAllowed(poolId, assetId));
```
mustermeiszer, line 141: @lemunozm if we decide to add a new message to tell the CAL side what is allowed, we could do that here, right?

mustermeiszer (commented):
LGTM

src/pools/PoolManager.sol
lemunozm: Correct. Or CV could somehow send that message too to the CAL side. Like a router that forward the message to two different endpoints. Just thinking out loud

wischli (approved):
LGTM now! Great to have learned more about free and internal vs. external functions as part of this review!

PR #90, opened 2025-02-19
Rename AssetManager to AssetRegistry

lemunozm:
Bonus: `OneToOneValuation` to `IdentityValuation`, given we still refer to it with the old name.


mustermeiszer (approved):
🫶🎉

wischli (approved):
In favor!

PR #94, opened 2025-02-24
feat: Balance Sheet Manager

mustermeiszer:
The `BalanceSheetManager` implements logic for these different actions

Incoming triggers from CP
* withdraw assets on the CV side &rarr; triggers a return message
* deposit assets on the CV side &rarr; triggers a return message
* issue shares on the CV side &rarr; triggers a return message
* revoke shares on the CV side &rarr; triggers a return message

Local triggers from CV
* deposit assets onto the balance sheet of a pool
* withdraw assets from the balance sheet of a pool
* update the accounting on the CP side
* update the holding on the CP side with a value from the CV side
* issue shares from the CV side
* revoke shares from the CV side

## Out of Scope / Follow-Up
* Batching
* Gateway fee payment
* Triggering of CP to CV messages targeted to the BSM - pool router logic


src/common/libraries/MessageLib.sol
```diff
@@ -62,13 +70,16 @@ enum MessageCategory {
...
+    BalanceSheet,
     Other
 }
 
 library MessageLib {
     using MessageLib for bytes;
     using BytesLib for bytes;
+    using JournalEntryLib for bytes;
     using CastLib for *;
+    using JournalEntryLib for JournalEntry[];
```
wischli, line 86: Ultra nit: Keep `using JournalEntryLib` in subsequent lines or use asterix import.

src/common/libraries/MessageLib.sol
```diff
@@ -122,6 +136,9 @@ library MessageLib {
             length += message.toUint16(length - 2); //payloadLength
         } else if (kind == uint8(MessageType.UpdateContract)) {
             length += message.toUint16(length - 2); //payloadLength
+        } else if (kind == uint8(MessageType.UpdateHolding)) {
+            length += message.toUint16(length - 2); // credits length
+            length += message.toUint16(length - 4); //debits length
```
wischli: Why do we subtract 4 for debits but only 2 for credits?

src/common/libraries/MessageLib.sol
```diff
@@ -929,4 +971,135 @@ library MessageLib {
...
+    //---------------------------------------
+    //    UpdateShares
+    //---------------------------------------
+
+    struct UpdateShares {
+        uint64 poolId;
+        bytes16 scId;
+        bytes32 who;
+        uint128 shares;
+        uint256 timestamp;
```
wischli, line 985: Curious why we use `uint64 timestamp` for `UpdateHolding` but here we use `uint256 timestamp`. Can we not make the type consistent?

src/common/types/JournalEntry.sol
```diff
@@ -0,0 +1,74 @@
...
+     * @dev Decodes a big-endian, tight-encoded bytes array back into an array of JournalEntry.
+     *      The array length must be a multiple of 20 bytes.
+     */
+    function toJournalEntries(bytes memory _bytes, uint256 _start, uint16 _length)
+        internal
+        pure
+        returns (JournalEntry[] memory)
+    {
+        require(_bytes.length >= _start + _length, "decodeJournalEntries_outOfBounds");
+        uint256 count = _length / 20;
```
wischli, line 60: Nit: Let's add a check to ensure the target array length is indeed a multiple of 20 bytes
```suggestion
        require(_length % 20 == 0, "decodeJournalEntries_invalidLength");    
        uint256 count = _length / 20;
```

src/vaults/BalanceSheetManager.sol
```diff
@@ -0,0 +1,439 @@
...
+    }
+
+    // --- Incoming ---
+    function adaptNotedWithdraw(
+        uint64 poolId,
+        bytes16 scId,
+        uint128 assetId,
+        address receiver,
+        uint128 amount,
+        bytes32 encoded,
```
lemunozm: How do you know if it's a number of an address?

src/common/types/JournalEntry.sol
```diff
@@ -0,0 +1,74 @@
+// SPDX-License-Identifier: BUSL-1.1
+pragma solidity 0.8.28;
+
+struct JournalEntry {
+    uint128 amount;
+    uint32 accountId;
+}
+
+library JournalEntryLib {
```
wischli, line 17: Code looks right to me but before merging, we definitely need extensive fuzz tests.

src/vaults/BalanceSheetManager.sol
```diff
@@ -0,0 +1,439 @@
...
+    IGateway public gateway;
+    IMessageProcessor public sender;
+    IPoolManager public poolManager;
+
+    mapping(uint64 => mapping(bytes16 => mapping(address => bool))) public permission;
+    mapping(uint64 => mapping(bytes16 => mapping(address => mapping(uint128 => Noted)))) public notedWithdraw;
+    mapping(uint64 => mapping(bytes16 => mapping(address => mapping(uint128 => Noted)))) public notedDeposit;
+
+    constructor(address escrow_) Auth(msg.sender) {
+        escrow = IPerPoolEscrow(escrow_);
```
wischli, line 47: It appears to me that the `BalanceSheetManager` will be deployed once per CV instance. Nevertheless, we initialize the corresponding escrow as **per pool** which seems misleading to me. Curious about your thoughts on this.

src/vaults/BalanceSheetManager.sol
```diff
@@ -0,0 +1,439 @@
...
+        uint128 amount,
+        IERC7726 valuation,
+        Meta calldata m
+    ) external authOrPermission(poolId, scId) {
+        _deposit(
+            poolId, scId, poolManager.assetToId(asset), asset, tokenId, provider, amount, _getPrice(valuation, asset), m
+        );
+    }
+
+    function deposit(
```
wischli: IMO we should keep only a single one of these variants. Given their current structure, it seems like the variant without `uint128 assetId` is superiour due to not requiring a storage read on the `CV.PoolManager`.

However, if you believe both variants to be called quite frequently without requiring knowledge of the `tokenId` when calling with `uint128 assetId` input, then both should continue to exist in order to keep gas cost lower due to not requiring a storage read for retrieving `uint128 assetId <-> (address asset, uint256 tokenId)`.

src/vaults/BalanceSheetManager.sol
```diff
@@ -0,0 +1,439 @@
...
+
+        sender.sendIssueShares(poolId, scId, to, shares, block.timestamp);
+        emit IssueShares(poolId, scId, to, shares);
+    }
+
+    function revoke(uint64 poolId, bytes16 scId, address from, uint128 shares)
+        external
+        authOrPermission(poolId, scId)
+    {
+        address token = poolManager.getTranche(poolId, scId);
```
wischli: Heads up that I proposed renaming this to `tranche` in #98

src/vaults/BalanceSheetManager.sol
```diff
@@ -0,0 +1,439 @@
...
+    // --- Incoming ---
+    function adaptNotedWithdraw(
+        uint64 poolId,
+        bytes16 scId,
+        uint128 assetId,
+        address receiver,
+        uint128 amount,
+        bytes32 encoded,
+        Meta calldata m
+    ) external authOrPermission(poolId, scId) {
```
wischli: I'm curious in which scenario `auth` is not sufficient for the incoming messages. I believe these calls will be called from the `CV.MessageProcessor`?

src/vaults/BalanceSheetManager.sol
```diff
@@ -0,0 +1,439 @@
...
+        Noted storage noted = notedWithdraw[poolId][scId][receiver][assetId];
+
+        noted.amount = amount;
+        noted.encoded = encoded;
+        noted.m = m;
+
+        if (noted.amount == 0) {
+            delete notedWithdraw[poolId][scId][receiver][assetId];
+        } else {
+            notedWithdraw[poolId][scId][receiver][assetId] = noted;
```
wischli: Redundant: We are already updating exactly that storage entry thanks to declaring `noted` as `storage`.

src/vaults/BalanceSheetManager.sol
```diff
@@ -0,0 +1,439 @@
...
+        }
```
wischli: Proposal for saving gas:
```suggestion
        if (amount == 0) {
                delete notedWithdraw[poolId][scId][receiver][assetId];
        } else {
                Noted storage noted = notedWithdraw[poolId][scId][receiver][assetId];
                noted.amount = amount;
                noted.encoded = encoded;
                noted.m = m;
        }
```

src/vaults/BalanceSheetManager.sol
```diff
@@ -0,0 +1,439 @@
...
+
+        noted.amount = amount;
+        noted.encoded = encoded;
+        noted.m = m;
+
+        if (noted.amount == 0) {
+            delete notedDeposit[poolId][scId][provider][assetId];
+        } else {
+            notedDeposit[poolId][scId][provider][assetId] = noted;
+        }
```
wischli: Same as above:
```suggestion
        if (amount == 0) {
                delete notedDeposit[poolId][scId][receiver][assetId];
        } else {
                Noted storage noted = notedDeposit[poolId][scId][receiver][assetId];
                noted.amount = amount;
                noted.encoded = encoded;
                noted.m = m;
        }
```

src/vaults/BalanceSheetManager.sol
```diff
@@ -0,0 +1,439 @@
...
+    function _getPrice(address valuation, address asset) internal view returns (D18) {
+        // TODO: Check cast?
+        return d18(
+            uint128(
+                IERC7726(valuation).getQuote(
+                    1, // TODO: Fix this - e.g. 1 * poolManager.poolDecimals(poolId),
+                    address(0), // TODO: Fix this - e.g. poolManager.poolCurrency(poolId),
+                    asset /* TODO: tokenId compatible */
+                )
+            )
```
wischli: Nit: Use safe cast which requires adding `using MathLib for ...` to the header of your contract
```suggestion
                IERC7726(valuation).getQuote(
                    1, // TODO: Fix this - e.g. 1 * poolManager.poolDecimals(poolId),
                    address(0), // TODO: Fix this - e.g. poolManager.poolCurrency(poolId),
                    asset /* TODO: tokenId compatible */
                ).toUint128();
```

src/vaults/BalanceSheetManager.sol
```diff
@@ -0,0 +1,439 @@
...
+            // noted.m.timestamp = block.timestamp;
+        }
+
+        D18 pricePerUnit;
+        if (noted.isRawPrice()) {
+            pricePerUnit = noted.asRawPrice();
+        } else {
+            // TODO: Check cast?
+            pricePerUnit = _getPrice(noted.asValuation(), asset);
+        }
```
wischli: Nit: Could also use ternary operator
```suggestion
        D18 pricePerUnit = noted.isRawPrice() ? noted.asRawPrice() :  pricePerUnit = _getPrice(noted.asValuation(), asset);
```

wischli (commented):
Overall makes sense to me but I need to do another review round in the next days. I suppose for the messages with array data, we might need some kind of max bound? Apart from that, undecided on the current `bytes32` price approach.

src/common/libraries/MessageLib.sol
lemunozm: Due the position of the sub message size variable in the message

lemunozm (commented):
Will do a proper review tomorrow

src/common/types/JournalEntry.sol
```diff
@@ -0,0 +1,74 @@
...
+    /**
+     * @dev Packs an array of JournalEntry into a tight bytes array of length (entries.length * 20).
+     *      Each entry = 20 bytes:
+     *         - amount (uint128) is stored in 16 bytes (big-endian)
+     *         - accountId (uint32) in 4 bytes (big-endian)
+     */
+    function encodePacked(JournalEntry[] memory entries) internal pure returns (bytes memory) {
```
lemunozm, line 24: NIT. Maybe renaming this as `fromJournalEntries` as opposite to `toJournalEntries` (?)

src/vaults/MessageProcessor.sol
```diff
@@ -189,6 +300,34 @@ contract MessageProcessor is Auth, IMessageProcessor, IMessageHandler {
             } else {
                 revert InvalidMessage(uint8(kind));
             }
+        } else if (cat == MessageCategory.BalanceSheet) {
+            if (kind == MessageType.UpdateHolding) {
+                MessageLib.UpdateHolding memory m = message.deserializeUpdateHolding();
+                Meta memory meta = Meta({timestamp: m.timestamp, debits: m.debits, credits: m.credits});
+                if (m.execute) {
+                    if (m.isIncrease) {
+                        balanceSheetManager.withdraw(
```
lemunozm: When `isIcreased`, then we _reduce_ with withdraw. Just double-check this is the intended behavior.

test/common/unit/libraries/MessageLib.t.sol
```diff
@@ -36,6 +39,7 @@ contract TestMessageLibCategories is Test {
         assert(MessageCategory.Investment == uint8(MessageType.CancelRedeemRequest).category());
         assert(MessageCategory.Investment == uint8(MessageType.FulfilledCancelDepositRequest).category());
         assert(MessageCategory.Investment == uint8(MessageType.FulfilledCancelRedeemRequest).category());
+        assert(MessageCategory.BalanceSheet == uint8(MessageType.UpdateHolding).category());
```
lemunozm, line 40: We should also add `UpdateShares` and `UpdateJournal`

src/vaults/BalanceSheetManager.sol
```diff
@@ -0,0 +1,439 @@
...
+        uint256 tokenId,
+        address provider,
+        uint128 amount,
+        IERC7726 valuation,
+        Meta calldata m
+    ) external authOrPermission(poolId, scId) {
+        _deposit(
+            poolId, scId, poolManager.assetToId(asset), asset, tokenId, provider, amount, _getPrice(valuation, asset), m
+        );
+    }
```
lemunozm: I think here is where you need to call `gateway.setPayableSource(msg.sender)`, the same for the rest of outgoing methods.

src/vaults/BalanceSheetManager.sol
```diff
@@ -0,0 +1,439 @@
...
+
+    function issue(uint64 poolId, bytes16 scId, address to, uint128 shares, bool asAllowance)
+        external
+        authOrPermission(poolId, scId)
+    {
+        address token = poolManager.getTranche(poolId, scId);
+
+        if (asAllowance) {
+            ITranche(token).mint(address(this), shares);
+            IERC20(token).approve(address(to), shares);
```
lemunozm: This could be IERC6909 when that PR gets merged right?

src/vaults/BalanceSheetManager.sol
```diff
@@ -0,0 +1,439 @@
...
+import {ITranche} from "src/vaults/interfaces/token/ITranche.sol";
+
+contract BalanceSheetManager is Auth, IRecoverable, IBalanceSheetManager, IUpdateContract {
+    IPerPoolEscrow public immutable escrow;
+
+    IGateway public gateway;
+    IMessageProcessor public sender;
+    IPoolManager public poolManager;
+
+    mapping(uint64 => mapping(bytes16 => mapping(address => bool))) public permission;
```
lemunozm: NIT. Given we already have PoolId and ShareClassId in common, we could use them here too.

src/vaults/BalanceSheetManager.sol
lemunozm: Agree 👀

src/vaults/BalanceSheetManager.sol
```diff
@@ -0,0 +1,439 @@
...
+        address token = poolManager.getTranche(poolId, scId);
+
+        if (asAllowance) {
+            ITranche(token).mint(address(this), shares);
+            IERC20(token).approve(address(to), shares);
+        } else {
+            ITranche(token).mint(address(to), shares);
+        }
+
+        sender.sendIssueShares(poolId, scId, to, shares, block.timestamp);
```
lemunozm: I think here is where you need to call `gateway.setPayableSource(msg.sender)`, the same for the rest of outgoing methods.
```suggestion
        gateway.setPayableSource(msg.sender)
        sender.sendIssueShares(poolId, scId, to, shares, block.timestamp);
```

lemunozm: I will think if it makes sense a refactor where `send*` methods also receive the payable `source`.

src/vaults/BalanceSheetManager.sol
```diff
@@ -0,0 +1,439 @@
...
+            amount,
+            encoded,
+            m.timestamp,
+            m.debits,
+            m.credits
+        );
+    }
+
+    function executeNotedWithdraw(uint64 poolId, bytes16 scId, address asset, uint256 tokenId, address receiver)
+        external
```
lemunozm, line 215: Add `authOrPermission`

src/vaults/BalanceSheetManager.sol
```diff
@@ -0,0 +1,439 @@
...
+            // TODO: Check cast?
+            pricePerUnit = _getPrice(noted.asValuation(), asset);
+        }
+
+        _withdraw(
+            poolId, scId, assetId, asset, tokenId, receiver, noted.amount, pricePerUnit, noted.allowance(), noted.m
+        );
+    }
+
+    function executeNotedWithdraw(uint64 poolId, bytes16 scId, uint128 assetId, address receiver) external {
```
lemunozm: Add `authOrPermission`

src/vaults/BalanceSheetManager.sol
```diff
@@ -0,0 +1,439 @@
...
+            receiver,
+            noted.amount,
+            pricePerUnit,
+            noted.allowance(),
+            noted.m
+        );
+    }
+
+    function executeNotedDeposit(uint64 poolId, bytes16 scId, address asset, uint256 tokenId, address provider)
+        external
```
lemunozm: Add `authOrPermission`

src/vaults/BalanceSheetManager.sol
```diff
@@ -0,0 +1,439 @@
...
+            pricePerUnit = noted.asRawPrice();
+        } else {
+            // TODO: Check cast?
+            pricePerUnit = _getPrice(noted.asValuation(), asset);
+        }
+
+        _deposit(poolId, scId, assetId, asset, tokenId, provider, noted.amount, pricePerUnit, noted.m);
+    }
+
+    function executeNotedDeposit(uint64 poolId, bytes16 scId, uint128 assetId, address provider) external {
```
lemunozm: Add `authOrPermission`

lemunozm (commented):
Simple than I thought!! Good job here 🚀  

I'm in favor of reducing message number using a boolean 👍🏻 in most cases. If not we would need to duplicate a lot of code, with the consecuent error-prone by the cost of just one byte

src/common/libraries/MessageLib.sol
mustermeiszer: I don't get that.

src/common/libraries/MessageLib.sol
mustermeiszer: Yes. Open todo ^^

src/vaults/BalanceSheetManager.sol
mustermeiszer: The interface is a `PerPoolEscrow` on a single contract. Open for renaming.

src/vaults/BalanceSheetManager.sol
mustermeiszer: Sounds good! I wanted to have a complete interface, but I agree, that `asset + tokenId` is sufficient.

src/vaults/BalanceSheetManager.sol
mustermeiszer: These here will be called by 

* `CV.MessageProcessor` &rarr; CP induced messages
* Contracts that an FM wants to give access to on the CV &rarr; per pool and share class access

src/vaults/BalanceSheetManager.sol
mustermeiszer: ah, wasn't aware of that!

src/vaults/BalanceSheetManager.sol
mustermeiszer: Love it!

src/vaults/BalanceSheetManager.sol
mustermeiszer: Use some signal bit somehwere in the bytes32.

src/common/types/JournalEntry.sol
mustermeiszer: My main take was that it is a packed representation, but no strong opinion. Whatever you like more.

test/common/unit/libraries/MessageLib.t.sol
mustermeiszer: Yes!

src/vaults/BalanceSheetManager.sol
mustermeiszer: Great! Wanted to ask that! What could be a source for payments for the incoming messages? Do you already have an idea on how we could allow funding pool messages on the CV side?

src/common/libraries/MessageLib.sol
lemunozm: `using JournalEntryLib for *` is also valid to enable for any type the library uses.

src/vaults/BalanceSheetManager.sol
mustermeiszer: Yes! Really good point.

src/vaults/BalanceSheetManager.sol
mustermeiszer: Yes

src/vaults/BalanceSheetManager.sol
mustermeiszer: It is intended to be permissionless. So that smart contracts can receive tokens or allowances. Do you see an issue with that?

src/vaults/BalanceSheetManager.sol
lemunozm: Actually just before sending here: https://github.com/centrifuge/protocol-v3/pull/94#discussion_r2000791902 is a better place.

It should be `msg.sender`. There are two ways of funding:

- When adding this action to the `VaultRouter`, the multicall can be paid, so `setPayableSource` takes no effect. (This is the normal usage).
- If this call is performed by another contract interaction that does not support to pay the method, then the Gateway should `receive()` the payment, and later the GasSerive would allow the message to be paid because the msg.sender was in the whitelist.

I think these methods should always be called from the VaultRouter. If that's the case, I think we do not even need setPayableSource

src/vaults/BalanceSheetManager.sol
lemunozm: That would mean that anyone can withdraw funds to the `receiver` they want. Would it not be a way of stealing funds?

src/vaults/MessageProcessor.sol
```diff
@@ -189,6 +301,40 @@ contract MessageProcessor is Auth, IMessageProcessor, IMessageHandler {
             } else {
                 revert InvalidMessage(uint8(kind));
             }
+        } else if (cat == MessageCategory.BalanceSheet) {
+            if (kind == MessageType.UpdateHolding) {
+                MessageLib.UpdateHolding memory m = message.deserializeUpdateHolding();
```
lemunozm: Could not have here a concurrence problem if both CP and CV modify the holdings?

i.e:
Suppose we have `100 `as holding value.

1. CV decreaseHoldings `80` (message still not reach CP) => Remains `20`
1. CP decreaseHoldings `70` => Remains `30`
1. CP receives the message from CV, but `30 - 80` is to much to decrease, so it fails.
1. CV receives the message from CP, but `20 - 70` is to much to decrease, so it fails.

Final state. 
- CV things `80` were decreased, 
- CP things `70` were decreased.

src/common/libraries/MessageLib.sol
wischli, line 86: Or as alternative
```suggestion
    using CastLib for *;
    using JournalEntryLib for bytes;
    using JournalEntryLib for JournalEntry[];
```

src/common/libraries/MessageLib.sol
```diff
@@ -929,4 +976,122 @@ library MessageLib {
...
+    }
+
+    function serialize(UpdateHolding memory t) internal pure returns (bytes memory) {
+        bytes memory debits = t.debits.encodePacked();
+        bytes memory credits = t.credits.encodePacked();
+
+        bytes memory partial1 = abi.encodePacked(MessageType.UpdateHolding, t.poolId, t.scId, t.assetId, t.who);
+        bytes memory partial2 =
+            abi.encodePacked(partial1, t.amount, t.pricePerUnit, t.timestamp, t.isIncrease, t.asAllowance);
+        bytes memory partial3 = abi.encodePacked(partial2, uint16(debits.length), uint16(credits.length));
```
wischli: Why do we need to encode three times? if we can easily manage to go with one encoding, this would decrease gas cost.

src/vaults/MessageProcessor.sol
mustermeiszer: CP will NOT be allowed to decrease holdings. It must ALWAYS go through CV

mustermeiszer: From now on

lemunozm: What it does then the `UpdateHolding` message sent from CP to CV?

src/vaults/BalanceSheetManager.sol
mustermeiszer: No actually. THis is our tranche which is erc20

src/vaults/MessageProcessor.sol
mustermeiszer: I renamed that to 

* `TriggerUpdateHolding` -> CP to CV - deposits or withdraws assets, sends `UpdateHolding` to CP, CP must update state ONLY then

src/pools/MultiShareClass.sol
```diff
@@ -531,6 +532,24 @@ contract MultiShareClass is Auth, IMultiShareClass {
     }
 
 
+    /// @inheritdoc IShareClassManager
+    function increaseShareClassIssuance(PoolId poolId, ShareClassId shareClassId_, uint128 amount) external auth {
+        require(exists(poolId, shareClassId_), ShareClassNotFound());
+
+        metrics[shareClassId_].totalIssuance += amount;
+        // TODO: emit event?
```
wischli: Ref: https://github.com/centrifuge/protocol-v3/pull/173#discussion_r2011548439

test/pools/integration/Cases.t.sol
```diff
@@ -244,4 +248,74 @@ contract TestCases is PoolsDeployer, Test {
...
+
+        require(poolDecimals == 1e18, "Pool decimals should be 1e18");
+
+
+        JournalEntry[] memory debits = new JournalEntry[](0);
+        (JournalEntry[] memory credits, uint256 i) = (new JournalEntry[](1), 0);
+        credits[i++] = JournalEntry(130 * poolDecimals, holdings.accountId(poolId, scId, USDC_C2, uint8(AccountType.GAIN)));
+
+
+        cv.updateHolding(poolId, scId, USDC_C2, 1000, D18.wrap(1e18), true, debits, credits);
```
mustermeiszer: @lemunozm and @onnovisser I need your help here. 

* Onno made the test work with the `decimalDiff` between - [see here](https://github.com/centrifuge/protocol-v3/pull/173/files#diff-08bb4526d2d0f8a2a8da15a2a117bd70b08519fe9eb953351658eff03539a641)
* I think we need to use the actual pool decimals 
* But the current `c.updateHolding` triggers a 
```
[FAIL: panic: arithmetic underflow or overflow (0x11)] testCalUpdateHolding() (gas: 1565580)
```

Am I setting this up wrongly.

src/common/libraries/MessageLib.sol
mustermeiszer: stack overflow

test/pools/integration/Cases.t.sol
onnovisser: I think if you multiply the value with the pool decimals, then you need to multiply the asset amount by the asset decimals.
`cv.updateHolding(poolId, scId, USDC_C2, 1000 * 1e6, D18.wrap(1e18), true, debits, credits);`

mustermeiszer: YES! That sounds about right. Will try. Thanks!

script/VaultsDeployer.s.sol
```diff
@@ -114,10 +123,15 @@ contract VaultsDeployer is CommonDeployer {
...
         investmentManager.file("poolManager", address(poolManager));
         investmentManager.file("gateway", address(gateway));
         investmentManager.file("sender", address(messageProcessor));
+
+        balanceSheetManager.file("poolManager", address(poolManager));
+        balanceSheetManager.file("gateway", address(gateway));
+        balanceSheetManager.file("sender", address(messageProcessor));
     }
 
     function removeVaultsDeployerAccess(address deployer) public {
```
wischli: CR (not blocking merge): Missing removal of deployer access for BSM

src/common/libraries/MessageLib.sol
```diff
@@ -34,7 +37,13 @@ enum MessageType {
     CancelRedeemRequest,
     FulfilledCancelDepositRequest,
     FulfilledCancelRedeemRequest,
-    TriggerRedeemRequest
+    TriggerRedeemRequest,
+    // -- BalanceSheetManager messages 25 - 27
```
wischli, line 41: Nit: Comment is off
```suggestion
    // -- BalanceSheetManager messages 25 - 29
```

src/common/types/JournalEntry.sol
wischli: In order to not block this PR, I added this to the cleanup epic: https://github.com/centrifuge/protocol-v3/issues/162#issuecomment-2735951245

src/common/types/JournalEntry.sol
wischli: Also added to the epic, not blocking

src/misc/types/D18.sol
```diff
@@ -84,14 +84,24 @@ function d18(uint128 num, uint128 den) pure returns (D18) {
     return D18.wrap(MathLib.mulDiv(num, 1e18, den).toUint128());
 }
 
+function eq(D18 a, D18 b) pure returns (bool) {
+    return D18.unwrap(a) == D18.unwrap(b);
+}
+
+function raw(D18 d) pure returns (uint128) {
+    return D18.unwrap(d);
+}
```
wischli, line 93: Nit: Missing tests

src/pools/MultiShareClass.sol
```diff
@@ -531,6 +532,29 @@ contract MultiShareClass is Auth, IMultiShareClass {
...
+        uint128 newIssuance = metrics[shareClassId_].totalIssuance + amount;
+        metrics[shareClassId_].totalIssuance = newIssuance;
+
+        emit IssuedShares(poolId, shareClassId_, epochId[poolId], navPerShare, navPerShare.mulUint128(newIssuance), amount);
+    }
+
+    /// @inheritdoc IShareClassManager
+    function decreaseShareClassIssuance(PoolId poolId, ShareClassId shareClassId_, D18 navPerShare, uint128 amount) external auth {
+        require(exists(poolId, shareClassId_), ShareClassNotFound());
+        require(metrics[shareClassId_].totalIssuance >= amount, "Issuance too low");
```
wischli, line 548: Nit: Use custom error as pointed out in #173

src/pools/MultiShareClass.sol
```diff
@@ -531,6 +532,29 @@ contract MultiShareClass is Auth, IMultiShareClass {
...
+
+        uint128 newIssuance = metrics[shareClassId_].totalIssuance - amount;
+        metrics[shareClassId_].totalIssuance = newIssuance;
+
+        // TODO: Maybe remove the redeemAssets part from the event?
```
wischli, line 553: We could do that, though it will degrade the information for actual share revocation. @mustermeiszer Do you think it's critical to transmit the asset amount of revoked shares during revocation?

src/vaults/BalanceSheetManager.sol
```diff
@@ -0,0 +1,328 @@
...
+    function issue(PoolId poolId, ShareClassId scId, address to, D18 pricePerShare, uint128 shares, bool asAllowance)
+        external
+        authOrPermission(poolId, scId)
+    {
+        _issue(poolId, scId, to, pricePerShare, shares, asAllowance);
+    }
+
+    /// @inheritdoc IBalanceSheetManager
+    function journalEntry(PoolId poolId, ShareClassId scId, Meta calldata m) external authOrPermission(poolId, scId) {
+        // @dev we do not need to ensure the meta here. Could be part of a batch and must not be balanced
```
wischli, line 165: Ultra-Nit: `must not` in english means it absolutely shouldn't whereas you mean, it doesn't necessarily have to be balanced
```suggestion
        // @dev we do not need to ensure the meta here. Could be part of a batch and does not have to be balanced
```

src/vaults/BalanceSheetManager.sol
wischli: Reminder about this comment @mustermeiszer

src/vaults/BalanceSheetManager.sol
```diff
@@ -0,0 +1,328 @@
...
+
+        escrow.deposit(asset, tokenId, poolId.raw(), scId.raw(), amount);
+        sender.sendIncreaseHolding(
+            poolId, scId, assetId, provider, amount, pricePerUnit, block.timestamp, m.debits, m.credits
+        );
+
+        emit Deposit(poolId, scId, asset, tokenId, provider, amount, pricePerUnit, block.timestamp, m.debits, m.credits);
+    }
+
+    function _ensureEntries(uint128 amount, Meta memory m) internal pure {
```
wischli, line 314: Nit: Can we make this name a bit more explicit?
```suggestion
    function _ensureBalancedEntries(uint128 amount, Meta memory m) internal pure {
```

test/pools/integration/Cases.t.sol
```diff
@@ -260,4 +265,97 @@ contract TestCases is PoolsDeployer, Test {
...
+        (JournalEntry[] memory debits, uint256 i) = (new JournalEntry[](3), 0);
+        debits[i++] = JournalEntry(1000, holdings.accountId(poolId, scId, USDC_C2, uint8(AccountType.ASSET)));
+        debits[i++] = JournalEntry(250, extraAccountId);
+        debits[i++] = JournalEntry(130, holdings.accountId(poolId, scId, USDC_C2, uint8(AccountType.EQUITY)));
+
+        (JournalEntry[] memory credits, uint256 j) = (new JournalEntry[](2), 0);
+        credits[j++] = JournalEntry(1250, holdings.accountId(poolId, scId, USDC_C2, uint8(AccountType.EQUITY)));
+        credits[j++] = JournalEntry(130, holdings.accountId(poolId, scId, USDC_C2, uint8(AccountType.LOSS)));
+
+        cv.updateJournal(poolId, scId, debits, credits);
```
wischli, line 289: Shouldn't we do a bunch of assertions after this call? Seems like this test only ensures the call doesn't revert.

wischli (dismissed):
No blockers from my side based on what we discussed in the Sync earlier, only minor nits and questions. Great job @mustermeiszer and @onnovisser :rocket:

src/common/libraries/MessageLib.sol
lemunozm: I think we can remove just categories and avoid these issues.

src/common/MessageProcessor.sol
```diff
@@ -4,23 +4,27 @@ pragma solidity 0.8.28;
...
+    IBalanceSheetManagerGatewayHandler
 } from "src/common/interfaces/IGatewayHandlers.sol";
 import {IVaultMessageSender, IPoolMessageSender} from "src/common/interfaces/IGatewaySenders.sol";
 
-import {ShareClassId} from "src/pools/types/ShareClassId.sol";
-import {AssetId} from "src/pools/types/AssetId.sol";
-import {PoolId} from "src/pools/types/PoolId.sol";
+import {ShareClassId} from "src/common/types/ShareClassId.sol";
+import {AssetId} from "src/common/types/AssetId.sol";
+import {PoolId} from "src/common/types/PoolId.sol";
```
lemunozm, line 27: Thanks for updating my latest type additions

src/common/MessageProcessor.sol
```diff
@@ -261,6 +268,149 @@ contract MessageProcessor is Auth, IMessageProcessor {
...
+        ShareClassId scId,
+        AssetId assetId,
+        address provider,
+        uint128 amount,
+        D18 pricePerUnit,
+        uint256 timestamp,
+        JournalEntry[] calldata debits,
+        JournalEntry[] calldata credits
+    ) external auth {
+        MessageLib.UpdateHolding memory data = MessageLib.UpdateHolding({
```
lemunozm: CR (not blocker): After the latest rebase, you would need to add the bypass here. Should be trivial but I can do it in a follow-up if you want to skip this task

src/common/interfaces/IGatewayHandlers.sol
```diff
@@ -29,6 +32,30 @@ interface IPoolRouterGatewayHandler {
...
+    /// @notice Update a holding by request from CAL.
+    function updateHoldingAmount(
+        PoolId poolId,
+        ShareClassId scId,
+        AssetId assetId,
+        uint128 amount,
+        D18 pricePerUnit,
+        bool isIncrease,
+        JournalEntry[] memory debits,
+        JournalEntry[] memory credits
```
lemunozm, line 48: NIT, I think `calldata` here will avoid a copy (same for `updateJournal()`)

src/common/interfaces/IGatewaySenders.sol
```diff
@@ -102,4 +105,65 @@ interface IVaultMessageSender is ICentrifugeChainId {
...
+        PoolId poolId,
+        ShareClassId shareClassId,
+        AssetId assetId,
+        address receiver,
+        uint128 amount,
+        D18 pricePerUnit,
+        uint256 timestamp,
+        JournalEntry[] calldata debits,
+        JournalEntry[] calldata credits
+    ) external;
```
lemunozm, line 133: NIT. Given that the message already contains an `isIncrease`, maybe we could use the export one method here. Also helps it becomes symmetric with the CV side.

src/common/MessageProcessor.sol
```diff
@@ -261,6 +268,149 @@ contract MessageProcessor is Auth, IMessageProcessor {
...
+            amount: 0,
+            pricePerUnit: pricePerUnit,
+            timestamp: timestamp,
+            isIncrease: false,
+            debits: debits,
+            credits: credits
+        });
+
+        gateway.send(poolId.chainId(), data.serialize());
+    }
```
lemunozm: If debits/credits/amount/who is always zero. Could we split this message into a new one in a future PR? (Not sure if you already had this in mind)

src/common/libraries/MessageLib.sol
```diff
@@ -604,6 +631,29 @@ library MessageLib {
         return abi.encodePacked(UpdateContractType.VaultUpdate, t.vaultOrFactory, t.assetId, t.kind);
     }
 
+    //---------------------------------------
+    //   UpdateContract.Permission (submsg)
+    //---------------------------------------
+
+    struct UpdateContractPermission {
+        address who;
```
lemunozm, line 639: CR: Must be `bytes32` to have multichain compatibility

src/common/libraries/MessageLib.sol
```diff
@@ -870,4 +920,215 @@ library MessageLib {
...
+    //    UpdateShares
+    //---------------------------------------
+
+    struct UpdateShares {
+        uint64 poolId;
+        bytes16 scId;
+        bytes32 who;
+        D18 pricePerShare;
+        uint128 shares;
+        uint256 timestamp;
```
lemunozm, line 985: NIT. Could `uint256` be unnecessarily large for a `timestamp`?

src/common/libraries/MessageLib.sol
```diff
@@ -870,4 +920,215 @@ library MessageLib {
...
+
+    //---------------------------------------
+    //    UpdateShares
+    //---------------------------------------
+
+    struct UpdateShares {
+        uint64 poolId;
+        bytes16 scId;
+        bytes32 who;
+        D18 pricePerShare;
```
lemunozm, line 983: NIT (but desirable). I would avoid custom types in the message struct description. These types seem too opaque for messaging and we can tomorrow modify the type representation, breaking the message specs without noticing. 

Also, for other readers, this file becomes our "specs", so I think we need to focus on clarity

(apply all messages)

src/common/libraries/MessageLib.sol
```diff
@@ -870,4 +920,215 @@ library MessageLib {
...
+
+        return h;
+    }
+
+    function serialize(TriggerUpdateHolding memory t) internal pure returns (bytes memory) {
+        bytes memory debits = t.debits.encodePacked();
+        bytes memory credits = t.credits.encodePacked();
+
+        bytes memory partial1 = abi.encodePacked(MessageType.TriggerUpdateHolding, t.poolId, t.scId, t.assetId, t.who);
+        bytes memory partial2 = abi.encodePacked(partial1, t.amount, t.pricePerUnit, t.isIncrease, t.asAllowance);
```
lemunozm, line 1088: NIT. Can be `partial1` and `partial2` encoded in one go?

src/pools/PoolRouter.sol
```diff
@@ -328,20 +338,39 @@ contract PoolRouter is Auth, Multicall, IPoolRouter, IPoolRouterGatewayHandler {
...
+                    holdings.accountId(unlockedPoolId, scId, assetId, uint8(AccountType.EXPENSE)), uint128(diff)
+                );
+            } else {
+                accounting.addCredit(
+                    holdings.accountId(unlockedPoolId, scId, assetId, uint8(AccountType.GAIN)), uint128(diff)
+                );
+                accounting.addDebit(
+                    holdings.accountId(unlockedPoolId, scId, assetId, uint8(AccountType.ASSET)), uint128(diff)
+                );
+            }
```
lemunozm, line 357: This is related to: https://github.com/centrifuge/protocol-v3/pull/173#discussion_r2011036629, maybe the above helps with what we discussed in sync?

src/vaults/BalanceSheetManager.sol
```diff
@@ -0,0 +1,328 @@
...
+        address token = poolManager.checkedTranche(poolId.raw(), scId.raw());
+
+        if (asAllowance) {
+            ITranche(token).mint(address(this), shares);
+            IERC20(token).approve(address(to), shares);
+        } else {
+            ITranche(token).mint(address(to), shares);
+        }
+
+        sender.sendIssueShares(poolId, scId, to, pricePerShare, shares, block.timestamp);
```
lemunozm, line 234: I think with the subsizing pool feature I'm working on, this can be fixed 🤞🏻

src/vaults/BalanceSheetManager.sol
```diff
@@ -0,0 +1,328 @@
...
+        AssetId assetId,
+        address asset,
+        uint256 tokenId,
+        address provider,
+        uint128 amount,
+        D18 pricePerUnit,
+        Meta memory m
+    ) internal {
+        _ensureEntries(pricePerUnit.mulUint128(amount), m);
+        escrow.pendingDepositIncrease(asset, tokenId, poolId.raw(), scId.raw(), amount);
```
lemunozm, line 298: NIT. A bit of out scope from this PR, but can we rename this to `increasePendingDeposit()` (same for decrease), the way it's written feels like it's a getter that returns the pending amount.

test/common/unit/libraries/MessageLib.t.sol
```diff
@@ -451,4 +458,181 @@ contract TestMessageLibIdentities is Test {
...
+
+    function testTriggerUpdateShares() public pure {
+        MessageLib.TriggerUpdateShares memory a = MessageLib.TriggerUpdateShares({
+            poolId: 1,
+            scId: bytes16("sc"),
+            who: bytes32("alice"),
+            pricePerShare: d18(123456),
+            shares: 100,
+            isIssuance: true,
+            asAllowance: true
```
lemunozm, line 623: Can we set one of these to `false`? If not, the identity test can pass even though the message is serialization doesn't match

test/vaults/unit/BalanceSheetManager.t.sol
```diff
@@ -0,0 +1,448 @@
...
+import {MessageLib} from "src/common/libraries/MessageLib.sol";
+import {PoolId} from "src/common/types/PoolId.sol";
+import {ShareClassId} from "src/common/types/ShareClassId.sol";
+import {AssetId} from "src/common/types/AssetId.sol";
+import {AccountId} from "src/common/types/AccountId.sol";
+
+import {IBalanceSheetManager} from "src/vaults/interfaces/IBalanceSheetManager.sol";
+import {BalanceSheetManager} from "src/vaults/BalanceSheetManager.sol";
+
+contract BalanceSheetManagerTest is BaseTest {
```
lemunozm, line 20: CR. If this is an integration test, could you move it to the `integration` folder?

src/vaults/BalanceSheetManager.sol
mustermeiszer: This shouldnt be here

lemunozm (commented):
Amazing work! Despite the amount of lines changes, it's easy to follow!

I left some CR, but can be don in a follow up PR. I do not want to block

test/pools/integration/Cases.t.sol
lemunozm: I agree

lemunozm:
~~A critical thing missing is that the BSM must be handled through the `VaultRouter`. Just a reminder, not a blocker. That can be added in a follow-up PR~~ See: https://kflabs.slack.com/archives/C07PG2EUR9C/p1742987634108869?thread_ts=1742984764.392399&cid=C07PG2EUR9C

src/vaults/BalanceSheetManager.sol
```diff
@@ -0,0 +1,328 @@
...
+{
+    using MathLib for *;
+
+    IPerPoolEscrow public immutable escrow;
+
+    IGateway public gateway;
+    IVaultMessageSender public sender;
+    IPoolManager public poolManager;
+
+    mapping(PoolId => mapping(ShareClassId => mapping(address => bool))) public permission;
```
hieronx, line 44: I think we can index this just by share class ID, since that already incorporates the pool ID, right?

src/common/libraries/MessageLib.sol
hieronx: yeah I think it can be `uint64`, that is what we have in other mressages

src/common/MessageProcessor.sol
```diff
@@ -330,6 +337,149 @@ contract MessageProcessor is Auth, IMessageProcessor {
...
+
+    /// @inheritdoc IVaultMessageSender
+    function sendIssueShares(
+        PoolId poolId,
+        ShareClassId scId,
+        address receiver,
+        D18 pricePerShare,
+        uint128 shares,
+        uint256 timestamp
+    ) external {
```
hieronx:
```suggestion
    ) external auth {
```

src/common/MessageProcessor.sol
```diff
@@ -330,6 +337,149 @@ contract MessageProcessor is Auth, IMessageProcessor {
...
+
+    /// @inheritdoc IVaultMessageSender
+    function sendRevokeShares(
+        PoolId poolId,
+        ShareClassId scId,
+        address provider,
+        D18 pricePerShare,
+        uint128 shares,
+        uint256 timestamp
+    ) external {
```
hieronx:
```suggestion
    ) external auth {
```

src/common/MessageProcessor.sol
```diff
@@ -330,6 +337,149 @@ contract MessageProcessor is Auth, IMessageProcessor {
...
+        );
+    }
+
+    /// @inheritdoc IVaultMessageSender
+    function sendJournalEntry(
+        PoolId poolId,
+        ShareClassId scId,
+        JournalEntry[] calldata debits,
+        JournalEntry[] calldata credits
+    ) external {
```
hieronx:
```suggestion
    ) external auth {
```

wischli (approved):
Re-approving!

src/vaults/BalanceSheetManager.sol
lemunozm: We decided to make `ShareClassId` unique (by using the `PoolId` internally) but never retrieve the `PoolId` from the `ShareClassId`.

The reason behind this is that we can easily skip using some pool checks if the user mixes PoolId (from PoolA) and ShareClassId (from poolB).

lemunozm: We want to be super safe avoiding cross pool issues

script/VaultsDeployer.s.sol
```diff
@@ -101,24 +109,39 @@ contract VaultsDeployer is CommonDeployer {
...
 
         // Rely on VaultRouter
         gateway.rely(address(vaultRouter));
         poolManager.rely(address(vaultRouter));
     }
 
     function _vaultsFile() public {
+        messageDispatcher.file("poolManager", address(poolManager));
+        messageDispatcher.file("investmentManager", address(investmentManager));
+        messageDispatcher.file("balanceSheetManager", address(investmentManager));
```
wischli, line 129: @onnovisser @lemunozm This seems incorrect because the `investmentManager` doesn't implement `IBalanceSheetManagerGatewayHandler` but the BSM does. Seems like this slipped through or am I missing something? Can fix as part of #109.

lemunozm: Right, that's incorrect!

It should be:
```
messageDispatcher.file("balanceSheetManager", address(balanceSheetManager));
```

Let me fix it in a PR where I'm updating the `messageDispatcher` that seems more related. Good eye @wischli !

lemunozm: Of fix it in your PR if you feel like it fits there! (Will do anyways in mine to avoid missing this)

wischli: I definitely resolved it already but I don't think we will merge the PR before Monday.

src/common/MessageProcessor.sol
lemunozm: Done: https://github.com/centrifuge/protocol-v3/pull/182

src/common/interfaces/IGatewaySenders.sol
lemunozm: Done: https://github.com/centrifuge/protocol-v3/pull/182

src/common/MessageProcessor.sol
lemunozm: Done https://github.com/centrifuge/protocol-v3/pull/182

src/common/interfaces/IGatewayHandlers.sol
lemunozm: Discarded given memory values are used as input

src/common/libraries/MessageLib.sol
lemunozm: Done https://github.com/centrifuge/protocol-v3/pull/182

src/common/libraries/MessageLib.sol
lemunozm: Done https://github.com/centrifuge/protocol-v3/pull/182

src/common/libraries/MessageLib.sol
lemunozm: Done https://github.com/centrifuge/protocol-v3/pull/182

src/common/libraries/MessageLib.sol
lemunozm: Done https://github.com/centrifuge/protocol-v3/pull/182

Also needed to avoid stack too deep issue

src/vaults/BalanceSheetManager.sol
lemunozm: We would need to add a `setPayableSource` here.

test/common/unit/libraries/MessageLib.t.sol
lemunozm: Done https://github.com/centrifuge/protocol-v3/pull/182

test/vaults/unit/BalanceSheetManager.t.sol
lemunozm: Done: https://github.com/centrifuge/protocol-v3/pull/182

src/common/libraries/MessageLib.sol
lemunozm: Done: https://github.com/centrifuge/protocol-v3/pull/182

script/VaultsDeployer.s.sol
lemunozm: Done https://github.com/centrifuge/protocol-v3/pull/182

src/common/libraries/MessageLib.sol
lemunozm: Done https://github.com/centrifuge/protocol-v3/pull/182

PR #96, opened 2025-02-24
feat: Per Pool Escrow

mustermeiszer:
PR adds a per pool-share-class layer on top of the existing escrow contract. 

The additional layer ensures
* funds of pools are separated
* increasing the deposited amount MUST happen after the actual token holdings of the said token have been increased


lemunozm:
Not sure if commented before. Can it be inherited from `Escrow`? I think we should leave the `Escrow` generic as it is

mustermeiszer:
Of course, but Why? We don't need the escrow for anything else, right? 

EDIT: 

Just saw your comment in the old PR. IMO we should not have 2 escrow contracts. The router will always just use the same escrow address as the rest of the system.

lemunozm:
> Of course, but Why? We don't need the escrow for anything else, right?

Well, not now. I think from an organizational perspective and conceptually, the `Escrow` is something super generic, and here we're adding a custom business behavior. Feels like two different layers of logic.

But not a super strong opinion!

src/vaults/Escrow.sol
```diff
@@ -26,4 +34,82 @@ contract Escrow is Auth, IEscrow {
...
+        require(pendingWithdraws[token][tokenId][poolId][scId] >= value, "Escrow/insufficient-pending-withdraws");
+
+        pendingWithdraws[token][tokenId][poolId][scId] -= value;
+    }
+
+    function withdraw(address token, uint256 tokenId, uint64 poolId, uint16 scId, uint256 value) external auth {
+        require(availableBalanceOf(token, tokenId, poolId, scId) > value, "Escrow/insufficient-funds");
+
+        holdings[token][tokenId][poolId][scId] -= value;
+    }
```
lemunozm: CR: I think we should also decrease `pendingWithdraws` here

lemunozm (commented):
Just one CR. It looks good!

src/vaults/Escrow.sol
mustermeiszer: The reason I did not do that was that `increasePendingWithdraws` acts as a blocker against withdrawing funds if they are "locked" for something. 

E.g. 
* approving redemptions &rarr; enforces FM can not take funds out of the escrow anymore if they are locked

If we make `withdraw` symetric with `deposit` we don't have that anymore. For the UI it is also hard to show what is available in these cases. Open for counter arguments of course.

lemunozm: Not fully understand. So once you withdraw, you will still have pending withdraw? That would mean that you can withdraw again, right? Does not breaking the rules?

1. increasePedingWithdraws 100.
2. withdraw 100
3. withdraw 100 (this should not be valid)

mustermeiszer: I will rename: 

* `increasePendingWithdraw` &rarr; `increaseLockedAmount`
* `decreasePendingWithdraw` &rarr; `decreaseLockedAmount`

So that was the usage that I had in mind

lemunozm: I think I'm missing something, in which way `increaseLockedAmount` is "locked" something if I can call `withdraw()` the times I want and be able to withdraw an unlimited amount?

mustermeiszer: You can't. The available balance of blocks that.

lemunozm: Correct. After another look, I now understand why it should be renamed to locked!

src/vaults/Escrow.sol
```diff
@@ -26,4 +34,82 @@ contract Escrow is Auth, IEscrow {
...
+        override
+        auth
+    {
+        require(pendingWithdraws[token][tokenId][poolId][scId] >= value, "Escrow/insufficient-pending-withdraws");
+
+        pendingWithdraws[token][tokenId][poolId][scId] -= value;
+    }
+
+    function withdraw(address token, uint256 tokenId, uint64 poolId, uint16 scId, uint256 value) external auth {
+        require(availableBalanceOf(token, tokenId, poolId, scId) > value, "Escrow/insufficient-funds");
```
lemunozm:
```suggestion
        require(availableBalanceOf(token, tokenId, poolId, scId) >= value, "Escrow/insufficient-funds");
```

If you have 100 not locked and you want to withdraw 100, you should be allowed to do it.

src/vaults/Escrow.sol
```diff
@@ -2,14 +2,22 @@
...
 /// @notice Escrow contract that holds tokens.
 ///         Only wards can approve funds to be taken out.
-contract Escrow is Auth, IEscrow {
+contract Escrow is Auth, IPerPoolEscrow, IEscrow {
+    mapping(address token => mapping(uint256 tokenId => mapping(uint64 poolId => mapping(uint16 scId => uint256))))
+        internal pendingWithdraws;
+    mapping(address token => mapping(uint256 tokenId => mapping(uint64 poolId => mapping(uint16 scId => uint256))))
+        internal pendingDeposits;
+    mapping(address token => mapping(uint256 tokenId => mapping(uint64 poolId => mapping(uint16 scId => uint256))))
+        internal holdings;
```
lemunozm: Question. Should not poolId & scId be the first key of the map, to allow some UX tools to index fastly things per poolId?

Not an expert on indexing anyways.

mustermeiszer: Good point. But will create some events and change ordering.

src/vaults/Escrow.sol
mustermeiszer: Definitely! Good catch

src/vaults/Escrow.sol
mustermeiszer: Done

src/vaults/Escrow.sol
wischli: Reminder this has not been fixed despite being resolved.

src/vaults/interfaces/IEscrow.sol
```diff
@@ -4,11 +4,50 @@ pragma solidity 0.8.28;
...
+    event PendingDeposit(
+        address indexed token, uint256 indexed tokenId, uint64 indexed poolId, uint16 scId, uint256 value
+    );
+    event Deposit(address indexed token, uint256 indexed tokenId, uint64 indexed poolId, uint16 scId, uint256 value);
+    event PendingWithdraw(
+        address indexed token, uint256 indexed tokenId, uint64 indexed poolId, uint16 scId, uint256 value
+    );
+    event Withdraw(address indexed token, uint256 indexed tokenId, uint64 indexed poolId, uint16 scId, uint256 value);
+
+    function pendingDepositIncrease(address token, uint256 tokenId, uint64 poolId, uint16 scId, uint256 value)
```
wischli: And the pile of missing docs grows 💩  I absolutely get we can deal with this later.

test/vaults/unit/Escrow.t.sol
```diff
@@ -4,6 +4,7 @@ pragma solidity 0.8.28;
 import {Escrow} from "src/vaults/Escrow.sol";
 import {IAuth} from "src/misc/interfaces/IAuth.sol";
 import "test/vaults/BaseTest.sol";
+import {ERC6909} from "../../../src/misc/ERC6909.sol";
```
wischli: Nit CR:

```suggestion
import {ERC6909} from "src/misc/ERC6909.sol";
import "test/vaults/BaseTest.sol";
```

src/vaults/Escrow.sol
```diff
@@ -2,14 +2,22 @@
...
 import {IERC20} from "src/misc/interfaces/IERC20.sol";
+import {IERC6909} from "src/misc/interfaces/IERC6909.sol";
 import {SafeTransferLib} from "src/misc/libraries/SafeTransferLib.sol";
 
 /// @title  Escrow
 /// @notice Escrow contract that holds tokens.
 ///         Only wards can approve funds to be taken out.
-contract Escrow is Auth, IEscrow {
+contract Escrow is Auth, IPerPoolEscrow, IEscrow {
+    mapping(uint64 poolId => mapping(uint16 scId => mapping(address token => mapping(uint256 tokenId => uint256))))
```
wischli: CR: Should either be `bytes16 scId` or `ShareClassId scId` which corresponds to uint128.

wischli (commented):
AFAICT, the share class id type for the storage maps is incorrect

src/vaults/Escrow.sol
lemunozm: Such an eye!

src/vaults/interfaces/IEscrow.sol
mustermeiszer: WhaT?

src/vaults/Escrow.sol
mustermeiszer: Will use `bytes16` for now. Don't wanna move the types from pools now.

src/vaults/interfaces/IEscrow.sol
wischli: Just saying that for audits its generally recommended to have as much docs as possible and we are increasing our "docs debt" on a daily basis which is fine.

mustermeiszer: Did a doc clean-up for this one.

mustermeiszer: https://github.com/centrifuge/protocol-v3/pull/96/commits/79c6982bbf6801faf69fa9bcec4cfc2f87e42813

src/vaults/Escrow.sol
mustermeiszer: Fixed https://github.com/centrifuge/protocol-v3/pull/96/commits/79c6982bbf6801faf69fa9bcec4cfc2f87e42813

test/vaults/unit/Escrow.t.sol
```diff
@@ -32,4 +36,193 @@ contract EscrowTest is BaseTest {
...
+
+        escrow.pendingDepositIncrease(address(erc20), TEST_TOKEN_ID, TEST_POOL_ID, TEST_SC_ID, 300);
+        erc20.mint(address(escrow), 300);
+        escrow.deposit(address(erc20), TEST_TOKEN_ID, TEST_POOL_ID, TEST_SC_ID, 100);
+
+        assertEq(escrow.availableBalanceOf(address(erc20), TEST_TOKEN_ID, TEST_POOL_ID, TEST_SC_ID), 0, "100 - 100 = 0");
+
+        escrow.deposit(address(erc20), TEST_TOKEN_ID, TEST_POOL_ID, TEST_SC_ID, 200);
+        assertEq(
+            escrow.availableBalanceOf(address(erc20), TEST_TOKEN_ID, TEST_POOL_ID, TEST_SC_ID), 200, "100 - 100 = 0"
```
wischli: Ultra nit:
```suggestion
            escrow.availableBalanceOf(address(erc20), TEST_TOKEN_ID, TEST_POOL_ID, TEST_SC_ID), 200, "300 - 100 = 200"
```

wischli (approved):
Thanks so much for the docs! The last missing piece for me would be testing the event emissions, which we are not doing right now. But I don't want to block this PR.

lemunozm (approved):
LGTM! I understand a follow up PR will use this new escrow methods

test/vaults/unit/Escrow.t.sol
```diff
@@ -3,44 +3,50 @@ pragma solidity 0.8.28;
...
         escrow.approveMax(address(erc20), spender);
 
+        vm.expectEmit();
+        emit IEscrow.Approve(address(erc20), spender, type(uint256).max);
         escrow.approveMax(address(erc20), spender);
         assertEq(erc20.allowance(address(escrow), spender), type(uint256).max);
     }
 
     function testUnapprove() public {
+        assertEq(address(erc20), address(erc20));
```
wischli: Nit: Can be removed

test/vaults/unit/Escrow.t.sol
```diff
@@ -3,44 +3,50 @@ pragma solidity 0.8.28;
...
+        emit IEscrow.Approve(address(erc20), spender, 0);
         escrow.unapprove(address(erc20), spender);
         assertEq(erc20.allowance(address(escrow), spender), 0);
     }
 
     // -------------------------------------------------
     // Tests for PerPoolEscrow
     // -------------------------------------------------
-    uint256 constant TEST_TOKEN_ID = 0; // We'll treat this as an ERC20 scenario
+    uint256 constant TEST_erc20Addr_ID = 0; // We'll treat this as an ERC20 scenario
```
wischli: Was this rename desired? Seems like not

wischli: FYI, you could also use `defaultTokenId` which is also 0 and exposed via `BaseTest`.

wischli (approved):
Thanks for adding event emission tests!

wischli (approved):
Really appreciate applying all my nits. Finally, I have nothing to complain about anymore, really solid PR now!

Issue #95, opened 2025-02-24
Refactor pools `Multicall.aggregate`


wischli:
1. Upgrade the new transient usage to the new solidity version.
2. Factorize `Multicall` to misc (i.e. unite `CentrifugeRouter.multicall` and `MultiCall.aggregate`). 
3. Merge `PoolLocker` into `PoolManager` inheriting from `Multicall`

For 2., we should disallow external calls as we currently do in `CentrifugeRouter.multicall` in order to protect against reentrency attacks and potentially unchecked arbitrary code execution. Currently, it's hard to guarantee that our authorization checks cannot be bypassed by a malicious actor.

PR #97, opened 2025-02-24
Use new transient keyword

lemunozm:
First part of: https://github.com/centrifuge/protocol-v3/issues/95

Use `transient` keyword everywhere in the project. Avoid previous transient usage.

Note, `transient` is not still supported by the formatter, so I added the files which use transient to the formatter ignore list.

Next in line (follow up PR):
- Factorize `Multicall` to `misc`
- Merge `PoolLocker` into `PoolManager` inheriting from `Multicall`


src/misc/TransientValuation.sol
```diff
@@ -13,7 +13,7 @@ import {BaseValuation} from "src/misc/BaseValuation.sol";
 
 contract TransientValuation is BaseValuation, ITransientValuation {
     /// @notice Temporal price set and used to obtain the quote.
-    mapping(address base => mapping(address quite => D18)) public /*TODO: transient*/ price;
+    mapping(address base => mapping(address quite => D18)) public /*transient*/ price; // FIX
```
lemunozm: Sadly, we can not use a storage mapping 😢, my intention is at least to keep the base and quote as other transients to check the price before using it to avoid wrong use cases.

test/pools/unit/SingleShareClass.t.sol
```diff
@@ -155,22 +165,9 @@ abstract contract SingleShareClassBaseTest is Test {
...
-            // Load the current value of the storage slot
-            bytes32 currentValue = vm.load(address(shareClass), slot);
-
-            // Clear only the first 4 bytes (corresponding to `_epochIncrement`)
-            // and preserve the rest
-            bytes32 clearedValue = currentValue & ~bytes32(uint256(0xFFFFFFFF));
-
-            // Set `_epochIncrement` to 0
-            vm.store(address(shareClass), slot, clearedValue);
+            shareClass.setEpochIncrement(0);
```
lemunozm: @wischli can I also remove the `WITH_TRANSIENT`? What is the purpose of that constant?

src/misc/TransientValuation.sol
hieronx: It should be possible to use assembly transient storage, like we did in the `CentrifugeRouter` contract, and then generate the storage key using the mapping indices `base` and `quite`.

test/pools/unit/SingleShareClass.t.sol
wischli: Yes, `WITH_TRANSIENT` can be removed. It was kept for convenience reasons as an artifact.

test/pools/unit/SingleShareClass.t.sol
```diff
@@ -61,11 +61,21 @@ contract OracleMock is IERC7726 {
     }
 }
 
+contract SingleShareClassExt is SingleShareClass {
+    constructor(IPoolRegistry poolRegistry, address deployer) SingleShareClass(poolRegistry, deployer) {
+        poolRegistry = poolRegistry;
+    }
+
+    function setEpochIncrement(uint32 epochIncrement) public {
```
wischli, line 67: CR: In theory, this call should only exist in the context of tests as a temporary solution until foundry supports `vm.storeTransient` (https://github.com/foundry-rs/foundry/issues/8165). I understand that this does not seem possible. So my proposal would be to use assembly `tstore` for `SingleShareClassBaseTest._resetTransientEpochIncrement` instead, assuming that should work. WDYT? I'm happy to do that and open a PR against this branch.

Right now, this call is not even permissioned to via `Auth` so at least that should be added with a `FIXME` that we need to fix pre-audit.

foundry.toml
```diff
@@ -72,7 +72,16 @@ fail_on_revert = false
...
-  ignore = ["test/*.sol"]
+  ignore = [
+    "test/*.sol",
+    #Check if https://github.com/foundry-rs/foundry/issues/9088 is fixed to remove the following:
+    "src/misc/TransientValuation.sol", #Because transient keyword
+    "src/pools/Accounting.sol", #Because transient keyword
+    "src/pools/PoolLocker.sol", #Because transient keyword
+    "src/pools/SingleShareClass.sol", #Because transient keyword
+    "src/vaults/CentrifugeRouter.sol", #Because transient keyword
+    "src/vaults/gateway/Gateway.sol", #Because transient keyword
```
wischli, line 83: 😢  Really inconvenient. Hope https://github.com/foundry-rs/foundry/issues/9088 will be fixed ASAP..

src/misc/TransientValuation.sol
lemunozm: That makes sense! Will play with it. Thanks

test/pools/unit/SingleShareClass.t.sol
lemunozm: Note that the contract only lives in the test, for testing purposes. So, there is nothing to worry about this call. If even so, you prefer to use the `txStore`, nothing against that. Would appreciate the PR! Whatever is simpler/easier.

foundry.toml
lemunozm: Yes please 🙏🏻

test/pools/unit/SingleShareClass.t.sol
wischli: Ouch, I missed the `.t` in the file name 🙈 Then let's keep as is because the current approach is more readable.

src/misc/TransientValuation.sol
lemunozm: For reference, it seems to be supported soon: https://github.com/ethereum/solidity/issues/15007

lemunozm: Done! Really good idea Jeroen! 🚀

test/pools/unit/SingleShareClass.t.sol
lemunozm: Done

hieronx (approved):
Transient storage usage looks good to me!

PR #98, opened 2025-02-24
feat: ERC-6909 Compatibility

mustermeiszer:
Needed in CAL branch in order to have correct storage mappings of existing assets as a sanity check. When duing stuff on the `BalanceSheetManager`

* Adds support for ERC6909 tokens
* Impacts CV API which relied on `address asset` beforehand as now `uint256 tokenId` is also required to distinguish between ERC20 (zero token id) and ERC6909 (non-zero token id) assets
  * adds token id to vault storage -> affects deploying new vaults
  * recover tokens


src/vaults/PoolManager.sol
```diff
@@ -46,14 +48,19 @@ contract PoolManager is Auth, IPoolManager {
     ITrancheFactory public trancheFactory;
     IGasService public gasService;
 
+    struct AssetIdKey {
+        address asset;
+        uint256 tokenId;
```
wischli: Q: Is `tokenId` the same as `assetId` or something else?

src/vaults/PoolManager.sol
```diff
@@ -46,14 +48,19 @@ contract PoolManager is Auth, IPoolManager {
...
     mapping(uint64 poolId => Pool) internal _pools;
     mapping(address => VaultAsset) internal _vaultToAsset;
     mapping(uint64 poolId => mapping(bytes16 => UndeployedTranche)) internal _undeployedTranches;
 
     /// @inheritdoc IPoolManager
-    mapping(uint128 assetId => address) public idToAsset;
+    mapping(uint128 assetId => AssetIdKey) public idToAsset;
     /// @inheritdoc IPoolManager
-    mapping(address => uint128 assetId) public assetToId;
+    mapping(AssetIdKey assedIdKey => uint128 assetId) public assetToId;
```
wischli: Not a fan of the asymmetry which removes the possibility of retrieving either the asset address from the id or vice versa. IIUC, the asset id now always needs to be known.

src/vaults/PoolManager.sol
mustermeiszer: `tokenId` is referring to a possible id from the ERC6909 standard. 

If `tokenId == 0` &rarr; is ERC20
else &rarr; is ERC6909

src/vaults/PoolManager.sol
mustermeiszer: It is symmetric still. or am I missing something? For all existing vaults you simply choose `tokenId = 0`.

lemunozm (commented):
Everything LGTM!

lemunozm (dismissed):

src/vaults/PoolManager.sol
```diff
@@ -115,14 +123,76 @@ contract PoolManager is Auth, IPoolManager {
...
+                symbol = _symbol;
+            } catch {}
+
+            try IERC6909MetadataExt(asset, tokenId).decimals() returns (uint8 _decimals) {
+                require(_decimals >= MIN_DECIMALS, "PoolManager/too-few-asset-decimals");
+                require(_decimals <= MAX_DECIMALS, "PoolManager/too-many-asset-decimals");
+                decimals = _decimals;
+            } catch {
+                revert("PoolManager/asset-missing-decimals");
+            }
```
lemunozm: Crazy idea, instead of having this pattern everywhere `if tokenId==0 else`... (also in per pool escrow PR) why not let the ERC6909 implementation to call `IERC20(token).methods()` when `tokenId` == 0?

It means, from PoolManager/Escrow, only exists ERC6909 to handle, and ERC6909 will call to ERC20 when tokenId == 0

mustermeiszer: We do not control the implementation of `ERC6909` so we can not do that.

lemunozm: Ok, it's not like in CP 👍🏻

src/vaults/PoolManager.sol
```diff
@@ -115,14 +123,77 @@ contract PoolManager is Auth, IPoolManager {
...
+    // @inheritdoc IPoolManager
+    function registerAsset(address asset, uint256 tokenId, uint32 destinationChain) external returns (uint128 assetId) {
+        string memory name;
+        string memory symbol;
+        uint8 decimals;
+
+        if (tokenId == 0) {
+            try IERC20Metadata(asset).name() returns (string memory _name) {
+                name = string(bytes(_name).sliceZeroPadded(0, 128));
+            } catch {}
```
lemunozm: I'm curious about this try/catch pattern. If there is an error getting the value, should it not revert?

wischli: I suppose we should extend the reversion for missing decimals also to name and symbol, yes.

src/vaults/PoolManager.sol
```diff
@@ -115,14 +123,77 @@ contract PoolManager is Auth, IPoolManager {
...
+            } catch {
+                revert("PoolManager/asset-missing-decimals");
+            }
+        }
+
+        bytes32 key = _formatAssetIdKey(asset, tokenId);
+        assetId = _assetToId[key];
+        if (assetId == 0) {
+            _assetCounter++;
+            assetId = uint128(bytes16(abi.encodePacked(uint32(block.chainid), _assetCounter)));
```
hieronx, line 142: I would add `uint32 private immutable _chainId;` as storage value and in the constructor, and use this rather than `block.chainid`, per this Slack thread: https://kflabs.slack.com/archives/C07PG2EUR9C/p1740659674849429?thread_ts=1740654438.583919&cid=C07PG2EUR9C
```suggestion
            assetId = uint128(bytes16(abi.encodePacked(_chainId, _assetCounter)));
```

lemunozm: What's the reason behind this?

If the deployer chooses a different chainId than the real chainId, then the messages from CP can no longer be redirected to the correct chain in the gateway/adapter, extracting the `chainId` from the AssetId (as we currently do).

hieronx: It would allow setting a custom chain ID if `block.chainid > type(uint32).max`

lemunozm: And how would the adapter on the other side know where to send the message then? i.e.

1. CV in Base uses `1234` as chainId. But the real Base chainId is `8453`
2. CV creates an asset using that chain Id `AssetId(chainId, counter)`, and sends that asset to CP.
3. CP sends a fulfilled message to the chain identified by the AssetId: `1234`.
4. The adapter does not find that chain and the message does not reach Base.

Should we configure the adapters in other chains also to reinterpret some chainIds? 🤔

hieronx: > Should we configure the adapters in other chains also to reinterpret some chainIds? 🤔

Yes when we configure a new chain, we should ensure the CP side and CV side is all in sync.

lemunozm: Ok, we'll need some other mapping in the adapters. But maybe to relax this PR, I would leave this chain id feature to a follow up PR

src/vaults/PoolManager.sol
mustermeiszer: I think that is not a requirement from a standards perspective. @hieronx  did it on purpose IIRC.

src/vaults/PoolManager.sol
mustermeiszer: @hieronx and @lemunozm adding bigger `chainId` would result in breaking the concatenation process for `AssetId(chainId, counter)`, right? 

Not sure, that is easily extendable like envisioned here.

lemunozm: What do you mean by "bigger"?

src/vaults/PoolManager.sol
hieronx: Yeah, technically name, symbol and decimals are all optional per ERC20. We obviously need decimals, but name and symbol can be considered optional for CP I would say.

In practice, I have never really seen a token without name and symbol.

src/vaults/PoolManager.sol
mustermeiszer: Larger than u32

lemunozm: I think what @hieronx means is that if any chain exceeds u32, we can map that `u64` to a non-used `u32`. So there is no increment in size.

mustermeiszer: Ahhh! Perfect!

lemunozm:
As a note, not sure if we've defined how to send this registerAsset to the different chains. Should we specify the chains where to register the asset?

src/vaults/PoolManager.sol
```diff
@@ -115,14 +124,84 @@ contract PoolManager is Auth, IPoolManager {
         emit TransferTrancheTokens(poolId, trancheId, msg.sender, destinationDomain, destinationId, recipient, amount);
     }
 
+    // @inheritdoc IPoolManager
+    function registerAsset(address asset, uint256 tokenId, uint32 /* destinationChain */ )
```
lemunozm: If we want to send this asset to different chains, the counter will be increased each time. I think we do not want that (?)

mustermeiszer: good catch! Will adapt that

mustermeiszer: Actually it won't .

lemunozm: Correct, I saw it wrong! My apologies 🙏🏻

test/vaults/BaseTest.sol
```diff
@@ -158,27 +162,23 @@ contract BaseTest is Deployer, GasSnapshot, Test {
         poolManager.updateTranchePrice(poolId, trancheId, assetId, uint128(10 ** 18), uint64(block.timestamp));
 
         // TODO: Use .update() from poolManager if possible
```
wischli: Reminder to look into this

mustermeiszer: Its hard to retrieve the `vault` address from the `.update()` flow. Don't thing it is worth it. Throught about that today a bit, so wanted to drop that here.

wischli:
> As a note, not sure if we've defined how to send this registerAsset to the different chains. Should we specify the chains where to register the asset?

Maybe I am missing something but we specify the destination chain in the arguments of `registerAsset(..., uint32 destChainid)` which seems sufficient to me?

test/vaults/BaseTest.sol
wischli: Turns out you can get the `vault` via the tranche, so it actually wasn't complicated. Applied here: [8b0d5fa](https://github.com/centrifuge/protocol-v3/pull/98/commits/8b0d5fa731ebb98b2c5e581b6528e21c838507cc)

src/vaults/PoolManager.sol
wischli: Applied here: [16dbcd2](https://github.com/centrifuge/protocol-v3/pull/98/commits/16dbcd2aca5c6bd14ecb7a722c46745836c4625e)

lemunozm:
Yeah, I removed that message time ago, not sure why you were able to see it 😂

src/vaults/PoolManager.sol
```diff
@@ -165,17 +221,18 @@ contract PoolManager is Auth, IPoolManager, IUpdateContract {
     /// @inheritdoc IPoolManager
     function allowAsset(uint64 poolId, uint128 assetId) public auth {
         require(isPoolActive(poolId), "PoolManager/invalid-pool");
-        address asset = idToAsset[assetId];
+        address asset = idToAsset_[assetId].asset;
         require(asset != address(0), "PoolManager/unknown-asset");
 
+        // TODO: Fix for ERC6909 once escrow PR is in
```
wischli: @mustermeiszer Can you elaborate on what needs to be changed here? Happy to do that.

src/vaults/PoolManager.sol
```diff
@@ -42,23 +52,27 @@ contract PoolManager is Auth, IPoolManager, IUpdateContract {
...
     ITrancheFactory public trancheFactory;
 
+    uint32 internal _assetCounter;
+
     mapping(uint64 poolId => Pool) internal _pools;
     mapping(address => VaultAsset) internal _vaultToAsset;
     mapping(address factory => bool) public vaultFactory;
 
+    mapping(uint128 assetId => AssetIdKey) public idToAsset_;
+    mapping(address asset => uint128 assetId) public assetToId;
```
wischli: @mustermeiszer I had to diverge from your symmetrical approach because your `assetToId` required knowledge of the `uint256 tokenId` which could not be obtained from storage if you only knew the `address asset`. So I first introduced a third storage `mapping(address asset => uint256) assetTokenIds` but then realized we would store tokenIds twice which is inefficient. I believe the current solution is the most efficient one but lacking symmetry.

src/vaults/PoolManager.sol
```diff
@@ -112,15 +126,57 @@ contract PoolManager is Auth, IPoolManager, IUpdateContract {
...
+
+        if (tokenId == 0) {
+            IERC20Metadata meta = IERC20Metadata(asset);
+            name = meta.name();
+            symbol = meta.symbol();
+        } else {
+            IERC6909MetadataExt meta = IERC6909MetadataExt(asset);
+            name = meta.name(tokenId);
+            symbol = meta.symbol(tokenId);
+        }
```
wischli, line 137: @mustermeiszer I refactored the `try-catch` blocks because they actually didn't catch anything. Apparently, the new `_safeGetAssetDecimals` is safer and for the name and symbol, we can just just normal getters.

test/misc/mocks/MockERC6909.sol
```diff
@@ -1,9 +1,10 @@
 // SPDX-License-Identifier: BUSL-1.1
 pragma solidity 0.8.28;
 
+import {ERC6909} from "src/misc/ERC6909.sol";
 import {IERC6909MetadataExt} from "src/misc/interfaces/IERC6909.sol";
 
-contract MockERC6909 is IERC6909MetadataExt {
+contract MockERC6909 is IERC6909MetadataExt, ERC6909 {
```
wischli: @lemunozm Curious if you are fine with this change. I first wanted to go the minimal effort path to be able to create ERC6909s with metadata extension for `registerAsset` unit tests.

test/vaults/integration/CentrifugeRouter.t.sol
```diff
@@ -51,7 +51,8 @@ contract CentrifugeRouterTest is BaseTest {
         router.requestDeposit{value: 1 wei}(vault_, amount, self, self, 1 wei);
         centrifugeChain.updateMember(vault.poolId(), vault.trancheId(), self, type(uint64).max);
 
-        uint256 gas = estimateGas() + GAS_BUFFER;
+        // Multiply estimated gas by two due to outgoing registerAsset and requestDeposit messages
+        uint256 gas = 2 * estimateGas() + GAS_BUFFER;
```
wischli: Since `registerAsset` is now part of `deploySimpleVault` which dispatches an outbound CV message, the test gas estimation was off. Curious if you agree with the changes.

wischli (commented):
PR is ready to be reviewed @mustermeiszer and @lemunozm. Fixed all commented out tests of the CV PoolManager to the best of my knowledge.

src/vaults/PoolManager.sol
mustermeiszer: That will not work for ERC-6909 IIUC. 

Where do we require knowledge about the `tokenId` that needs to be read from storage?

src/vaults/PoolManager.sol
lemunozm: Could we leave this feature for when we have better defined how `chainId` is working everywhere?

We must handle chainIds in Gateway, GasService, Adapters, and message recoveries. 

If we open this door now, we lose the invariant of `AssetId.chainId == block.chainId`, which is now really appreciated for having a first MVP working.

IMO this feature must be done once we have CP-CV integration tests and most of the chain issues are already solved

lemunozm: To extend the above, if we do this, I'll use the `Gateway` as single source of truth for this aliased `chainId`, so the `PoolManager` could do `AssetId(gateway.chainId(), counter)` and the adapter can also ask to the gateway to correctly mapping everything. But let's do this once we've everything well defined! 🙏🏻

src/vaults/PoolManager.sol
lemunozm: Should not it be 
```suggestion
    mapping(address asset => mapping(uint256 tokenId => uint128 assetId)) public assetToId;
```
to support ERC6909 tokens?

src/vaults/PoolManager.sol
```diff
@@ -112,15 +124,57 @@ contract PoolManager is Auth, IPoolManager, IUpdateContract {
...
+
+        assetId = assetToId[asset];
+        if (assetId == 0) {
+            _assetCounter++;
+            assetId = uint128(bytes16(abi.encodePacked(uint32(_chainId), _assetCounter)));
+
+            idToAsset_[assetId] = AssetIdKey(asset, tokenId);
+            assetToId[asset] = assetId;
```
lemunozm: and here 
```suggestion
            assetToId[asset][tokenId] = assetId;
```

src/vaults/PoolManager.sol
```diff
@@ -112,15 +124,57 @@ contract PoolManager is Auth, IPoolManager, IUpdateContract {
...
+            IERC20Metadata meta = IERC20Metadata(asset);
+            name = meta.name();
+            symbol = meta.symbol();
+        } else {
+            IERC6909MetadataExt meta = IERC6909MetadataExt(asset);
+            name = meta.name(tokenId);
+            symbol = meta.symbol(tokenId);
+        }
+
+        assetId = assetToId[asset];
```
lemunozm: and here 
```suggestion
        assetId = assetToId[asset][tokenId];
```

src/vaults/PoolManager.sol
lemunozm: IIUC, the whole allowAsset() can be removed due to the updateContract addition.

Also disallow and the messages in `vaults/ProcessMessage.sol`

src/vaults/PoolManager.sol
```diff
@@ -444,7 +478,28 @@ contract PoolManager is Auth, IPoolManager, IUpdateContract {
         return _vaultToAsset[vault].isLinked;
     }
 
+    /// @inheritdoc IPoolManager
     function isAllowedAsset(uint64 poolId, address asset) public view override returns (bool) {
         return allowedAssets[poolId][asset];
     }
+
+    /// @inheritdoc IPoolManager
+    function idToAsset(uint128 assetId) public view returns (address asset) {
```
lemunozm:
```suggestion
    function idToAsset(uint128 assetId) public view returns (address asset, uint256 tokenId) {
```

src/vaults/PoolManager.sol
```diff
@@ -165,17 +219,18 @@ contract PoolManager is Auth, IPoolManager, IUpdateContract {
...
     function allowAsset(uint64 poolId, uint128 assetId) public auth {
         require(isPoolActive(poolId), "PoolManager/invalid-pool");
-        address asset = idToAsset[assetId];
+        address asset = idToAsset_[assetId].asset;
         require(asset != address(0), "PoolManager/unknown-asset");
 
+        // TODO: Fix for ERC6909 once escrow PR is in
         allowedAssets[poolId][asset] = true;
         emit AllowAsset(poolId, asset);
     }
```
mustermeiszer: CR: Fully remove all logic around allow or disallow asset.

src/vaults/PoolManager.sol
```diff
@@ -165,17 +219,18 @@ contract PoolManager is Auth, IPoolManager, IUpdateContract {
...
 
     /// @inheritdoc IPoolManager
     function disallowAsset(uint64 poolId, uint128 assetId) public auth {
         require(isPoolActive(poolId), "PoolManager/invalid-pool");
-        address asset = idToAsset[assetId];
+        address asset = idToAsset_[assetId].asset;
```
mustermeiszer: CR: Fully remove all logic around allow or disallow asset.

test/vaults/integration/CentrifugeRouter.t.sol
lemunozm: I do not understand why this increment the required gas at this point 🤔, this gas value is for the next `requiresDeposit` message, which should be unaffected

test/misc/mocks/MockERC6909.sol
lemunozm: The intention of using mocks instead of directly calling other contracts, is to avoid breaking tests if those contracts change. Given ERC6909 follows a fixed standard this is not as important.

It's much more work to give support to the mock just implementing `IERC6909`? I think only `allowance()` is used under the hood (?).

Question . Could be only the ERC6909Metadata the used methods by the system? If that's the case, maybe we do not even need ERC6909 methods for testing (?)

lemunozm (commented):
Good job refactoring all those tests!!

I have some CRs, and the one about removing the chainId support for PoolManager.

mustermeiszer (commented):
* Need to fully remove all logic around `AllowAsset` and `DisallowAsset`
* Need to fix ERC6909 registration as discussed on slack

src/vaults/PoolManager.sol
wischli: As discussed with @mustermeiszer , I will revert back to his

```solidity
mapping(bytes32 assetIdKey => assetId) assetToId_;
```

test/misc/mocks/MockERC6909.sol
wischli: > I think only `allowance()` is used under the hood (?).

That and `approve()`.

lemunozm: No strong opinion! We can easily modify in the future with a more complete mock and use by now the default implementation.

Question. Could you link me where `approve()` is called? I'm not able to see it.

wischli: > Question. Could you link me where `approve()` is called? I'm not able to see it.

Here as a result of `escrow.approveMax` during asset registration: https://github.com/centrifuge/protocol-v3/blob/b577f49b29fc76fd01772d8dfa46c60a0e57662c/src/vaults/Escrow.sol#L42

test/vaults/integration/CentrifugeRouter.t.sol
wischli: Thanks for being skeptical. Even though this code change fixes the test, it's actually applied at the wrong place: https://github.com/centrifuge/protocol-v3/pull/98/commits/4c7c6d605a4ae914e25ecdea9e1a03d9d22d8f62#diff-80f00688f402c131024046fbf945a68776c3c46831d0910b2ad0ad5a1d1da020R74-R77

src/vaults/PoolManager.sol
wischli: @lemunozm Reverted in [4b7fa5d](https://github.com/centrifuge/protocol-v3/pull/98/commits/4b7fa5d2acc8c6bcf5771cf7a2f6c9ac8df80725)

src/vaults/PoolManager.sol
wischli: Removed in [8eb6e15](https://github.com/centrifuge/protocol-v3/pull/98/commits/8eb6e15fd30179d7df8bfa6b1fc927e9d3721db1)

src/vaults/PoolManager.sol
wischli: Actually went with you proposal @lemunozm because it's more explicit and doesn't really increase storage size as keys are hashed automatically.

test/misc/mocks/MockERC6909.sol
wischli: Decoupled `MockERC6909` from `ERC6909` in [4d03f26](https://github.com/centrifuge/protocol-v3/pull/98/commits/4d03f2694f8a801f34494c8e4b7b1af04f00104b)

src/vaults/PoolManager.sol
wischli: Done in [8eb6e15](https://github.com/centrifuge/protocol-v3/pull/98/commits/8eb6e15fd30179d7df8bfa6b1fc927e9d3721db1)

src/vaults/PoolManager.sol
mustermeiszer: Shouldnt we also male the baust Factory accept an tokenId then? Then we are flexible for the future

wischli: Yes, I think that makes sense. I first wanted to ship the minimal (from my perspective) required changes to get everything working and then wait on your feedback.

src/vaults/PoolManager.sol
lemunozm: 🙏🏻

test/misc/mocks/MockERC6909.sol
lemunozm: Awesome! Finally it was easy to do as long as we do not need to implement mandatorily the IERC6909

lemunozm (dismissed):
Everything solved! Thanks for this!! Finally we'll have the same RegistryAsset message in both sides 🥹

src/vaults/PoolManager.sol
```diff
@@ -47,15 +55,15 @@ contract PoolManager is Auth, IPoolManager, IUpdateContract {
...
 
+    uint32 internal _assetCounter;
+
     mapping(uint64 poolId => Pool) internal _pools;
     mapping(address => VaultAsset) internal _vaultToAsset;
     mapping(address factory => bool) public vaultFactory;
 
+    mapping(uint128 assetId => AssetIdKey) public idToAsset_;
+    mapping(address asset => mapping(uint256 tokenId => uint128 assetId)) public assetToId;
     mapping(uint64 poolId => mapping(address asset => bool)) allowedAssets;
```
mustermeiszer: CR

Remove

src/vaults/PoolManager.sol
```diff
@@ -110,6 +118,46 @@ contract PoolManager is Auth, IPoolManager, IUpdateContract {
         emit TransferTrancheTokens(poolId, trancheId, msg.sender, destinationId, recipient, amount);
     }
 
+    // @inheritdoc IPoolManager
+    function registerAsset(address asset, uint256 tokenId, uint32 destChainId) external returns (uint128 assetId) {
+        string memory name;
+        string memory symbol;
+        uint8 decimals;
+
+        require(asset.code.length > 0, "PoolManager/invalid-asset-contract");
```
mustermeiszer: QUESTION

Why do we need that here? Isn't it sufficient to call the `IERCxxx` interfaces?

src/vaults/PoolManager.sol
```diff
@@ -110,6 +118,46 @@ contract PoolManager is Auth, IPoolManager, IUpdateContract {
...
+
+        decimals = _safeGetAssetDecimals(asset, tokenId);
+        require(decimals >= MIN_DECIMALS, "PoolManager/too-few-asset-decimals");
+        require(decimals <= MAX_DECIMALS, "PoolManager/too-many-asset-decimals");
```
mustermeiszer, line 127: QUESTION

@hieronx why do we have these limitations?

src/vaults/PoolManager.sol
```diff
@@ -201,7 +229,7 @@ contract PoolManager is Auth, IPoolManager, IUpdateContract {
         TrancheDetails storage tranche = _pools[poolId].tranches[trancheId];
         require(tranche.token != address(0), "PoolManager/tranche-does-not-exist");
 
-        address asset = idToAsset[assetId];
+        address asset = idToAsset_[assetId].asset;
```
mustermeiszer: CR

- make ERC6909 compatible
- tranche storage must be `asset`-`tokenId` mapping

src/vaults/PoolManager.sol
```diff
@@ -307,14 +314,15 @@ contract PoolManager is Auth, IPoolManager, IUpdateContract {
         address[] memory vaultWards = new address[](0);
 
         // Deploy vault
+        address asset = idToAsset(assetId);
```
mustermeiszer: CR

Make factory ERC6909 compatible. 

`newVault(..., asset, tokenId, ...)`

src/vaults/PoolManager.sol
```diff
@@ -307,14 +314,15 @@ contract PoolManager is Auth, IPoolManager, IUpdateContract {
...
         address vault =
             IVaultFactory(factory).newVault(poolId, trancheId, asset, tranche.token, address(escrow), vaultWards);
 
-        // Check whether the ERC20 token is a wrapper
+        // Check whether asset is an ERC20 token wrapper
         try IERC20Wrapper(asset).underlying() returns (address) {
-            _vaultToAsset[vault] = VaultAsset(asset, true, false);
+            _vaultToAsset[vault] = VaultAsset(asset, assetId, true, false);
```
mustermeiszer: CR

Make ERC6909 compatible. Add `tokenId` field

src/vaults/PoolManager.sol
```diff
@@ -328,25 +336,25 @@ contract PoolManager is Auth, IPoolManager, IUpdateContract {
...
 
     /// @inheritdoc IPoolManager
-    function linkVault(uint64 poolId, bytes16 trancheId, address asset, address vault) public auth {
+    function linkVault(uint64 poolId, bytes16 trancheId, uint128 assetId, address vault) public auth {
         TrancheDetails storage tranche = _pools[poolId].tranches[trancheId];
         require(tranche.token != address(0), "PoolManager/tranche-does-not-exist");
 
         address manager = IBaseVault(vault).manager();
-        uint128 assetId = assetToId[asset];
+        address asset = idToAsset(assetId);
```
mustermeiszer: CR 

`VaultManager` should take a `tokenId` too

src/vaults/PoolManager.sol
```diff
@@ -388,7 +396,7 @@ contract PoolManager is Auth, IPoolManager, IUpdateContract {
     function getVaultAssetId(address vault) public view override returns (uint128) {
```
mustermeiszer: QUESTION

@lemunozm said we should have naming convention `vaultToAssetId` instead of `getXXX`

src/vaults/PoolManager.sol
```diff
@@ -400,7 +408,28 @@ contract PoolManager is Auth, IPoolManager, IUpdateContract {
         return _vaultToAsset[vault].isLinked;
     }
 
+    /// @inheritdoc IPoolManager
     function isAllowedAsset(uint64 poolId, address asset) public view override returns (bool) {
```
mustermeiszer: CR 

Remove

mustermeiszer (commented):
Logic looks good, but I think as we are tackling the erc6909 compatibility now, we should go the last mile and make it fully compatible. 

See my CRs.

src/vaults/PoolManager.sol
wischli: I suppose we can remove the check such that `_safeGetAssetDecimals` reverts with `PoolManager/asset-missing-decimals` when the checked code doesn't support either ERC20 or ERC6909 `decimals`

src/vaults/PoolManager.sol
wischli: IMO we should store either `uint128 assetId` or `uint256 tokenId` as together with the address, either can be retrieved from `(address, *other)`. IMO, assetId makes more sense here, as the protocol tries to abstract the tokenId away as much as possible. WDYT?

src/vaults/PoolManager.sol
wischli: @hieronx Can we change `TrancheDetails` storage or is that a blocker due to Morpho and/or Maker?

https://github.com/centrifuge/protocol-v3/blob/3b29368bfadc65da0378ff71a1064f5fb38f67f4/src/vaults/interfaces/IPoolManager.sol#L13-L21

hieronx: No that's fine, can be changed! 

One thing we should make sure: the old vault linked to Morpho will also need to be updated with the new investment manager when we migrate.

src/vaults/PoolManager.sol
hieronx: If you have significantly more than 18 decimals, `uint128` that we use for balances etc might not be sufficient.

src/vaults/PoolManager.sol
wischli: Can you elaborate on that? A vault address now corresponds to 
1. `address asset`
2. `uint128 assetId`
3. `uint256 tokenId` - TBD if hop over `CV.PoolManager.idToAsset_` storage or retrieved via `_vaultToAsset`. 

So IMO, we don't need to bake `tokenId` into the manager as well?

src/vaults/PoolManager.sol
mustermeiszer: Ah yes. Of course. I did read that wrongly. Makes total sense.

src/vaults/PoolManager.sol
mustermeiszer: Makes sense. Maybe we can even remove `asset` and make it solely rely on `assetId`?

src/vaults/PoolManager.sol
wischli: Done in [de56985](https://github.com/centrifuge/protocol-v3/pull/98/commits/de5698579e48ba986f2e932568fab8d6ed76122c)

src/vaults/PoolManager.sol
wischli: Good catch! Done in [b4ab26b](https://github.com/centrifuge/protocol-v3/pull/98/commits/b4ab26b11f44612f6909064965ab64657fc6a1af)

src/vaults/PoolManager.sol
wischli: @mustermeiszer Done in [56ca183](https://github.com/centrifuge/protocol-v3/pull/98/commits/56ca183ce7dc77c2b4b1becf60802b9a6e81253a)

src/vaults/PoolManager.sol
wischli: Done in [56ca183](https://github.com/centrifuge/protocol-v3/pull/98/commits/56ca183ce7dc77c2b4b1becf60802b9a6e81253a)

src/vaults/PoolManager.sol
wischli: Done in [89086bf](https://github.com/centrifuge/protocol-v3/pull/98/commits/89086bff1969592bdbc8cd6c3b047eccfb35621e)

src/vaults/PoolManager.sol
wischli: Done in [b4ab26b](https://github.com/centrifuge/protocol-v3/pull/98/commits/b4ab26b11f44612f6909064965ab64657fc6a1af)

wischli:
@mustermeiszer Applied all your CRs, thanks for your eagle eyes! Looking forward to your next review!

lemunozm (dismissed):
LGTM!! 🚀 although would be good to have a Frederik approval too

lemunozm:
Just one minor CR. Could you remove this: https://github.com/centrifuge/protocol-v3/blob/main/src/common/Gateway.sol#L144

wischli:
> Just one minor CR. Could you remove this: https://github.com/centrifuge/protocol-v3/blob/main/src/common/Gateway.sol#L144

Sure, good catch! I am certain I had removed this already but rebasing probably reintroduced it 🥲

src/vaults/interfaces/IPoolManager.sol
```diff
@@ -39,6 +39,8 @@ struct UndeployedTranche {
 }
 
 struct VaultAsset {
```
mustermeiszer: NIT

reanme to `VaultDetails` or something like that.

mustermeiszer: NIT

add `tokenID` for symmetry, but could also retrieve that from `idToAsset_`

src/vaults/PoolManager.sol
```diff
@@ -360,35 +363,36 @@ contract PoolManager is Auth, IPoolManager, IUpdateContract {
...
+        AssetIdKey memory assetIdKey = idToAsset_[assetId];
+        TranchePrice memory value = _pools[poolId].tranches[trancheId].prices[assetIdKey.asset][assetIdKey.tokenId];
         require(value.computedAt > 0, "PoolManager/unknown-price");
         price = value.price;
         computedAt = value.computedAt;
     }
 
     /// @inheritdoc IPoolManager
-    function getVaultAsset(address vault) public view override returns (address, bool) {
+    function vaultToAssetAddress(address vault) public view returns (address, bool) {
```
mustermeiszer: NIT

Does this still make sense? Shouldn't we just expose `vaultToVaultDetails` and return the datastructure? 

Or we expose specific view method like

* `isERC20(address vault) public view returns (address, bool)`

But no strong opinion, we can also close that one now.

mustermeiszer (commented):
Great work! Thanks for handling this scope. That was much more as we anticipated ^^ 

Approving!

src/vaults/ERC7540Vault.sol
```diff
@@ -39,6 +39,7 @@ contract ERC7540Vault is Auth, IERC7540Vault {
 
     /// @inheritdoc IERC7575
     address public immutable asset;
+    uint256 public immutable tokenId;
```
hieronx: Wonder if we should add docs somewhere that we cannot rely on `vault.tokenId()` being present since old vaults won't have it.

wischli: Thanks for seeing this. Will definitely add docs, that it should never be used outside of the `ERC7540Vault` contract.

src/vaults/PoolManager.sol
wischli: I decided on just exposing `function vaultDetails(address vault) view returns (VaultDetails memory)`: [4d77c44](https://github.com/centrifuge/protocol-v3/pull/98/commits/4d77c442ffa2b235029435b671d5c953cf06648d)

wischli: I kept `isLinked` for now but we could also deprecate that. WDYT?

src/vaults/ERC7540Vault.sol
wischli: I think we should change the visibility to `internal` then such that it's impossible to be accessed from external contracts. WDYT?

mustermeiszer: agree!

mustermeiszer (commented):
Approving! Like how it looks, really clean now!

lemunozm (dismissed):
Let's merge this!!

wischli (approved):
The last commit [3978fdf](https://github.com/centrifuge/protocol-v3/pull/98/commits/3978fdfdb4730704be07f22e19e053922a403596) after approvals from @lemunozm and @mustermeiszer just added missing `PoolManager` test coverage and fixed the check for a valid hook interface, which resulted as necessary from improving the coverage.

lemunozm (approved):
LGTM! Thanks for being so perfectionist in this regard!

PR #99, opened 2025-02-25
Ref: PoolManager with new multicall method

### LINKED ISSUE #95: Refactor pools `Multicall.aggregate`

lemunozm:
Fixes https://github.com/centrifuge/protocol-v3/issues/95

### Changes
- `PoolManager.execute()` now works only for methods from the `PoolManager` checking re-entrance issues.
- `PoolLocker` and the previous `Multicall` no longer exist.
- `multicall()` from `CentrifugeRouter` has been extracted to `misc/Multicall`.
- `PoolManager` uses the same `Multicall` as `CentrifugeRouter`.

NOTE: all methods that previously had `poolUnlocked()` + permissionless methods now have `protected` modifier. It means that we can now mix permissionless methods with admin methods in the same multicall, which is an extra cool feature!

To recap modifier purposes:
- `poolUnlocked` modifier: means that the method requires to be called through execute() to have the `pool` open.
- `protected` modifier: means that the method can be called through multicall and checks for reentrancy issues.

`poolUnlocked` always requires to be `protected` but the opposite does not need to be true.

Note that both `execute()` and `multicall()` can be used. The second one is more generic without unlocking the pool.


hieronx (approved):
Looks great to me! 🔥

src/misc/Multicall.sol
```diff
@@ -3,22 +3,40 @@ pragma solidity ^0.8.28;
...
-            // Forward the error happened in target.call().
+abstract contract Multicall is IMulticall {
+    address private transient _initiator;
+
+    modifier protected() {
+        if (_initiator == address(0)) {
+            // Single call re-entrancy lock
+            _initiator = msg.sender;
+            _;
+            _initiator = address(0);
```
lemunozm: I'm getting an "unreachable code" warning in this line 14 (making our CI fail). I'm running out of ideas. Should not be unreachable. Actually, the code follows the same pattern it existed before in CentrifugeRouter.

Can I ask for another couple of eyes here? @hieronx

hieronx: I don't get it either 🤔 It is a warning not an error, maybe it can be ignored?

lemunozm: I'm afraid that avoiding it we're hidden some bug related to it. AFAIK, if it's unreachable code, the bytecode related to it could not be included by the compiler 😖

Will add a test!

src/misc/Multicall.sol
```diff
@@ -3,22 +3,40 @@ pragma solidity ^0.8.28;
...
+                require(length != 0, CallFailed());
+
+                assembly ("memory-safe") {
+                    revert(add(32, returnData), length)
                 }
             }
-            results[i] = result;
         }
+
+        _initiator = address(0);
```
wischli: Q: Why is it necessary to reset to `address(0)` after the multicall? Shouldn't the transaction lifetime be over anyways such that `_initiater` is reset by being transient?

src/misc/interfaces/IMulticall.sol
```diff
@@ -1,17 +1,23 @@
...
+    /// @notice Dispatched when multicall is called inside of one protected method or recursively
+    error AlreadyInitiated();
+
+    /// @notice Allows caller to execute multiple ( batched ) messages calls in one transaction.
+    /// @param data An array of encoded methods of the same contract.
+    /// @dev No reentrant execution is allowed.
     /// If one call fails, it reverts the whole transaction.
-    function aggregate(Call[] calldata calls) external returns (bytes[] memory results);
+    /// In order to provide the correct value for functions that require top up,
+    /// the caller must estimate separate, in advance, how much each of the message call will cost.
```
wischli: Ultra nit:
```suggestion
    /// the caller must estimate separately, in advance, how much each of the message call will cost.
```

wischli (approved):
What an improvement, great work! Maybe we could add your on-point PR description recap somewhere to the docs, i.e. of the modifiers

> * `poolUnlocked` modifier: means that the method requires to be called through execute() to have the `pool` open.
> * `protected` modifier: means that the method can be called through multicall and checks for reentrancy issues.
> 
> `poolUnlocked` always requires to be `protected` but the opposite does not need to be true.

src/misc/Multicall.sol
wischli: AFAICT, the "unreachable code" happens whenever there is a function with `protected` modifier which returns or reverts always such that anything below the `_;` placeholder for the executed function body is indeed unreachable. I verified this by temporarily removing the return statement from `(I)PoolManager.addShareClass` which removed the warning.

lemunozm:
Will add!

src/misc/Multicall.sol
lemunozm: In a high-level you could want to wrap several execute() in another multicall. If we do not reset, we block that behavior

src/misc/Multicall.sol
lemunozm: I've removed it too to try it, and the warning still happens to me (note that warnings are not shown in the contract is not recompiled due to incremental compilation).

But what you said makes sense with the `err()` method used in `Multicall.t.sol`, commenting that method the warning disappears.

I also tested it, and the variable is correctly cleared for a normal flow. So it seems we're safe, and it is just a false-alarm warning

src/misc/Multicall.sol
```diff
@@ -3,22 +3,40 @@ pragma solidity ^0.8.28;
...
+        uint256 totalBytes = data.length;
+        for (uint256 i; i < totalBytes; ++i) {
+            (bool success, bytes memory returnData) = address(this).delegatecall(data[i]);
             if (!success) {
-                assembly {
-                    // Reverting the error originated in the above call.
-                    // First 32 bytes contains the size of the array, rest the error data
-                    revert(add(result, 32), mload(result))
+                uint256 length = returnData.length;
+                require(length != 0, CallFailed());
```
lemunozm: @hieronx Curious here how `CallFailed()` can happen. I'm not able to test this error 🤔

src/misc/Multicall.sol
wischli: > I also tested it, and the variable is correctly cleared for a normal flow. So it seems we're safe, and it is just a false-alarm warning

Did you test a multicall which includes `addShareClass` and at least one subsequent call?

lemunozm: In `Cases.t.sol`, there is one.

src/misc/Multicall.sol
```diff
@@ -1,24 +1,46 @@
...
+            _initiator = address(0);
+        } else {
+            // Multicall re-entrancy lock
+            require(msg.sender == _initiator, UnauthorizedSender());
+            _;
+        }
+    }
+
+    function multicall(bytes[] calldata data) public payable {
+        require(_initiator == address(0), AlreadyInitiated());
```
lemunozm: I think this check is not needed if the own multicall is protected.

Without this, we allow having multicalls of multicalls, which could be desired.

PR showing this: https://github.com/centrifuge/protocol-v3/pull/100

hieronx: Yeah it can be removed. With CV side ,there was really no purpose for nested multicalls, so this was a simplification. I agree with multiple pools and the unlock pattern with `execute`, it makes sense to allow this. I don't see any security issues.

src/misc/Multicall.sol
hieronx: This is for `revert()`.

src/misc/Multicall.sol
lemunozm: Great!
I'll merge the other PR then

PR #101, opened 2025-02-26
Cantina audit fix review

lemunozm:
Port from https://github.com/centrifuge/vaults-internal/pull/96

NOTE: IouCfg related stuff is not added.


lemunozm:
The good thing about bringing the whole history from CV is that it's pretty easy to move PRs between both repos! 🥳

hieronx (approved):
LGTM

PR #102, opened 2025-02-26
feat: ShareClass Metadata

### LINKED ISSUE #84: Add name, symbol and hook to `SingleShareClass.addShareClass`
Currently, when we add a share class in `SingleShareClass.sol`, we ignore the `bytes32 calldata data` which should include 
* The name of the share class token
* The symbol of the share class token
* The hook address

These data points are required for our protocol message `AddTranche` ([ref](https://centrifuge.hackmd.io/Soy4JiFbRSyovj53KW-a4A?view#AddTranche)) which is dispatched as a result of `PoolManager.notifyShareClass`. Therefore, we need to store these data points for each share class id as part of the `SingleShareClass.sol`.

This PR could also include adding a call to update the hardcoded hook address for some share class id in case we want to keep the direction of CP->CV for this based on the current CC->CV flow ([ref](https://kflabs.slack.com/archives/C07PG2EUR9C/p1738764933481839?thread_ts=1738764514.674309&cid=C07PG2EUR9C)).

wischli:
# Fixes #84

* Adds decoding of calldata bytes provided in `addShareClass` to `ShareClassMetadata` which is stored under a new storage key `mapping(scId => ShareClassMetadata) metadata`
* Adds new call to update metadata
* Applies fixes to PoolManager


src/pools/SingleShareClass.sol
```diff
@@ -57,13 +65,20 @@ contract SingleShareClass is Auth, ISingleShareClass {
     using MathLib for D18;
     using MathLib for uint128;
     using MathLib for uint256;
+    using CastLib for bytes;
+    using BytesLib for bytes;
+
+    uint32 constant META_NAME_LENGTH = 128;
+    uint32 constant META_SYMBOL_LENGTH = 128;
```
lemunozm: Symbol name in the messaging has only 32 bytes. I am not sure if we should protect this here, or it is a protocol concern to clamp the name there 🤔

src/pools/SingleShareClass.sol
```diff
@@ -502,6 +515,12 @@ contract SingleShareClass is Auth, ISingleShareClass {
         return (_shareClassNavPerShare[shareClassId_], totalIssuance[shareClassId_]);
     }
 
+    function setMetadata(PoolId poolId, ShareClassId shareClassId_, bytes calldata metadata_) external {
```
lemunozm: NIT. Should it be called `updateMetadata()`? Because the metadata is already set in the creation of the share class?

src/pools/SingleShareClass.sol
```diff
@@ -631,6 +650,21 @@ contract SingleShareClass is Auth, ISingleShareClass {
...
+        require(metadata_.length == META_NAME_LENGTH + META_SYMBOL_LENGTH + META_HOOK_LENGTH, InvalidMetadataSize());
+
+        string memory name = metadata_.slice(0, 128).bytes128ToString();
+        require(bytes(name).length != 0, InvalidMetadataName());
+
+        string memory symbol = metadata_.slice(128, 128).bytes128ToString();
+        require(bytes(symbol).length != 0, InvalidMetadataSymbol());
+
+        bytes32 hook = metadata_.toBytes32(256);
+        require(hook != bytes32(""), InvalidMetadataHook());
```
lemunozm: What if I only want to update the name and symbol but not the hook? Should it be hook "optional"?

test/pools/integration/Cases.t.sol
```diff
@@ -138,6 +146,14 @@ contract TestConfiguration is TestCommon {
...
         // From this point, pool is ready for investing from CV side
     }
+
+    function _encodeScMetadata(string memory name, string memory symbol, bytes32 hook)
+        internal
+        pure
+        returns (bytes memory metadata)
+    {
+        return abi.encodePacked(bytes(name.stringToBytes128()), bytes(symbol.stringToBytes128()), hook);
+    }
```
lemunozm: Nice utility! Maybe we can move it to `SingleShareClass.sol` like `previewShareClassId` to be able to use it in other places (as CP-CV integration tests, some spells/migrations, or other contracts that interact with this)

lemunozm: NOTE for checking later: `stringToBytes128()` is a utility I added to `CastLib` that is not audited. I think if utf8 does not take into account `\0` (which seems not) then we can simplify it

test/pools/integration/Cases.t.sol
```diff
@@ -138,6 +146,14 @@ contract TestConfiguration is TestCommon {
 
         // From this point, pool is ready for investing from CV side
     }
+
+    function _encodeScMetadata(string memory name, string memory symbol, bytes32 hook)
```
wischli: This duplicates code from `SingleShareClass.t.sol` but I wasn't sure if it's worth moving to some library as its tightly couple to the `SingleShareClass`.

test/pools/integration/Cases.t.sol
```diff
@@ -91,7 +95,9 @@ contract TestConfiguration is TestCommon {
 
         (bytes[] memory cs, uint256 c) = (new bytes[](4), 0);
         cs[c++] = abi.encodeWithSelector(poolManager.setPoolMetadata.selector, bytes("Testing pool"));
-        cs[c++] = abi.encodeWithSelector(poolManager.addShareClass.selector, bytes(""));
+        cs[c++] = abi.encodeWithSelector(
+            poolManager.addShareClass.selector, _encodeScMetadata(SC_META_NAME, SC_META_SYMBOL, SC_META_HOOK)
+        );
```
lemunozm: Most Ultra NIT comment ever. If we name them SC_NAME/SC_SYMBOL, maybe it will fit in just one line?

lemunozm (commented):
Thanks for this awesome work!

test/pools/integration/Cases.t.sol
lemunozm: Follows https://github.com/centrifuge/protocol-v3/pull/102#discussion_r1971264058

src/pools/SingleShareClass.sol
wischli: The hook must always be set. IMO, when you want to update not all fields, you should provide the current data. IIUC, you prefer having separate metadata update functions for each field?

src/pools/SingleShareClass.sol
wischli: Good catch! I should have just checked the `AddTranche` message instead of getting the limits from Centrifuge Chain were we allow both strings to contain up to 128 bytes because `orml_asset_registry::AssetMetadata` only supports a single string limit both for the name and symbol: https://github.com/centrifuge/centrifuge-chain-internal/blob/60b59b571834a9d6b69c9acb443fa1082ba7b3b0/libs/types/src/tokens.rs#L30-L35

src/pools/SingleShareClass.sol
lemunozm: I was thinking about let using `""` as value for hook, in case it can be optional

wischli: I would say we either treat empty {name, symbol, hook} as `None` and query the current data or we require the update to set all three. Currently, we enforce the latter. Should I refactor to the former?

lemunozm: I would treat empty as none. At least for the hook (although I do not understand correctly the implications of having an empty hook, so avoid this if I'm wrong).

I think just removing:
```sol
require(hook != bytes32(""), InvalidMetadataHook());
```
is enough.

mustermeiszer (commented):
Why do we need to store the hook?

wischli:
> Why do we need to store the hook?

Currently, the source of truth for the hook is Centrifuge Chain (see `AddTranche` and `UpdateTrancheHook` ([ref](https://centrifuge.hackmd.io/Soy4JiFbRSyovj53KW-a4A?view#UpdateTrancheHook)) messages). I couldn't find notes where we discussed changing that but maybe that was done verbally? 

Right now, it points to the `CV.RestrictionManager` for each tranche. When anyone makes transfers on the tranche tokens, internally CV calls the hook: https://github.com/centrifuge/protocol-v3/blob/c872d3020680b543f1b4d57f3260e1b15a40be1d/src/vaults/token/Tranche.sol#L109

I thought we wanted to store the hook such that `notifyShareClass` can query it from storage and does not require it as an input parameter. But I am happy to remove it from storage and add it as input parameter. What's your preference @mustermeiszer ?

mustermeiszer:
Yes. Lets remove it! UI can look it up from the CV side.

src/pools/SingleShareClass.sol
```diff
@@ -86,17 +104,17 @@ contract SingleShareClass is Auth, ISingleShareClass {
     }
 
     /// @inheritdoc IShareClassManager
-    function addShareClass(PoolId poolId, bytes calldata /* data */ )
-        external
-        auth
-        returns (ShareClassId shareClassId_)
-    {
+    function addShareClass(PoolId poolId, bytes calldata data) external auth returns (ShareClassId shareClassId_) {
```
lemunozm: I was thinking... if all share classes will need a name and a symbol because it's needed from PoolManager to get them (through `metadata()`. Then why do we need to encode them through `data`? Could not we have `addShareClass(poolId, name, symbol, data)` then?

TLDR: `data` should be any extra external property, but `name` and `symbol` are common to all SCM implementations

test/pools/integration/Cases.t.sol
```diff
@@ -128,6 +132,10 @@ contract TestConfiguration is TestCommon {
 
         // From this point, pool is ready for investing from CV side
     }
+
+    function encodeMetadata(string memory name, string memory symbol) internal pure returns (bytes memory metadata) {
+        return abi.encodePacked(bytes(name.stringToBytes128()), symbol.toBytes32());
+    }
```
lemunozm: No longer needed because the one used is the one from SingleShareClass (in case we still need it, see my previous comment)

lemunozm (commented):
Just a couple of thoughts, but everything looks good!

test/pools/integration/Cases.t.sol
lemunozm: I just saw you removed it!

wischli: Yeah sorry, missed that one :D

src/pools/SingleShareClass.sol
wischli: Yeah I agree this will be universally true independent of future SCM implementations. Will apply that!

wischli: [a2d2bf8](https://github.com/centrifuge/protocol-v3/pull/102/commits/a2d2bf866ff479fed63550661bae86c12eac406c)

lemunozm (approved):
Thanks so much for tackling my comments. Awesome PR 🚀

lemunozm:
I do not know how to avoid the snapshot changes... It's super annoying, and we expect to have even more! 

BTW, do you use any optimization when running the test? the gas has decreased a lot

wischli:
> I do not know how to avoid the snapshot changes... It's super annoying, and we expect to have even more!

I will look into that today as well. Maybe I can find something that we haven't considered yet.

> BTW, do you use any optimization when running the test? the gas has decreased a lot

No, just running `forge t`. I can imagine that it's related to my foundry update which I did yesterday in order to fix `forge inspect storage $contract`. I'm now on `forge 1.0.0-stable (e144b82070 2025-02-13T20:02:34.979686000Z)`. Unfortunately, I don't remember which version I used beforehand. It's a bummer that we cannot pin the foundry version in the `foundry.toml`.

PR #103, opened 2025-02-26
feat: remove transfer func

mustermeiszer:
Removes all functionalities around transferring assets to general addresses. This will be replaced by increasing and decreasing holdings on a per pool-share-class ownership basis.


lemunozm (approved):
Thanks for the great clean up! 🧹

PR #108, opened 2025-02-26
feat: Update contract

mustermeiszer:
Introduces a message `UpdateContract` that allows for the CP side to call into any other contract on the CV side that implement `IUpdateContract`. 

This feature is then used in the `CV.PoolManager` to 
* deploy vaults through a chosen, but allowlisted factory
* link and unlink vaults - i.e. allowing a vault to be used for investing into the given pool, implicitly allowing an asset through that vault

### Follow-Up
* Removing everything around `Dis/AllowAsset` message flow will be done as part of https://github.com/centrifuge/protocol-v3/pull/98


mustermeiszer:
Closing in favor of https://github.com/centrifuge/protocol-v3/pull/130

src/vaults/InvestmentManager.sol
```diff
@@ -57,6 +57,28 @@ contract InvestmentManager is Auth, IInvestmentManager {
         SafeTransferLib.safeTransfer(token, to, amount);
     }
 
+    // --- IVaultManager ---
+    function addVault(uint64 poolId, bytes16 trancheId, address vault) public auth override {
+        IERC7540Vault vault_ = IERC7540Vault(vault);
+        address token = vault_.share();
```
hieronx, line 70: Nit: `token` and `asset` are only used once so can just be inlined

src/vaults/InvestmentManager.sol
```diff
@@ -57,6 +57,28 @@ contract InvestmentManager is Auth, IInvestmentManager {
...
+        address asset = vault_.asset();
+
+        IAuth(token).rely(vault);
+        ITranche(token).updateVault(asset, vault);
+        Auth(address(this)).rely(vault);
+
+    }
+
+    function rmVault(uint64 poolId, bytes16 trancheId, address vault) public auth override {
```
hieronx: nit:
```suggestion
    function removeVault(uint64 poolId, bytes16 trancheId, address vault) public auth override {
```

src/vaults/PoolManager.sol
```diff
@@ -407,17 +416,36 @@ contract PoolManager is Auth, IPoolManager {
...
+        return this.getVaults(poolId, trancheId, idToAsset[assetId]);
     }
 
     /// @inheritdoc IPoolManager
-    function getVault(uint64 poolId, bytes16 trancheId, address asset) public view returns (address) {
-        address vault = ITranche(_pools[poolId].tranches[trancheId].token).vault(asset);
-        require(vault != address(0), "PoolManager/unknown-vault");
-        return vault;
+    /// @dev This function MUST NOT be called onchain
+    function getVaults(uint64 poolId, bytes16 trancheId, address asset) public view returns (address[]) {
```
hieronx: I think this should be removed, UIs can implement this (and our SDK can)

src/common/libraries/MessageLib.sol
```diff
@@ -466,6 +467,31 @@ library MessageLib {
...
+        bytes payload;
+    }
+
+    function deserializeUpdateContract(bytes memory data) internal pure returns (UpdateContract memory) {
+        require(messageType(data) == MessageType.UpdateContract, UnknownMessageType());
+        return UpdateContract({
+            poolId: data.toUint64(1),
+            scId: data.toBytes16(9),
+            target: data.toBytes32(25),
+            payload: data.slice(57, data.length - 57)
```
lemunozm: Note. Once you rebase, this would not work with batches, because data.length is the whole set of contiguous messages.

Take a look how `UpdateRestriction` works after the rebase. You would need to encode the length of the payload.

src/vaults/PoolManager.sol
```diff
@@ -289,92 +313,77 @@ contract PoolManager is Auth, IPoolManager {
...
-        require(canTrancheBeDeployed(poolId, trancheId), "PoolManager/tranche-not-added");
-
-        address[] memory trancheWards = new address[](2);
-        trancheWards[0] = investmentManager;
-        trancheWards[1] = address(this);
+    // --- IUpdateContract implementation ---
+    /// @inheritdoc IUpdateContract
+    /// @notice The pool manager either deploys the vault if a factory address is provided or it simply links/unlinks the vault
+    function update(uint64 poolId, bytes16 trancheId, bytes memory payload) public auth {
+        (factory, assetId, isLinked, vault) = abi.decode(payload, (address, uint128, bool, address));
```
lemunozm: Question. Why send a payload if later you assume it has some static structure defined? If different payloads can exists, should we provide a submessage type?

lemunozm (commented):
A couple of comments about messaging. I'll go deeper into the logic later

src/vaults/PoolManager.sol
mustermeiszer: update will be a generic way to call into contracts on the CV side. don't want to limit the structure in any form to some data

lemunozm: IIUC, the "submessage type" of the payload would be the `target` parameter right? That's why you assume in this implementation that payload has that structure

mustermeiszer: We can also decode at the `function handle()` level if you prefer that.

lemunozm: I think it's good where it's because the structure is defined by the implementor of `target` 👍🏻

src/common/libraries/MessageLib.sol
```diff
@@ -466,6 +467,31 @@ library MessageLib {
...
+    //---------------------------------------
+    //    UpdateContract
+    //---------------------------------------
+
+    struct UpdateContract {
+        uint64 poolId;
+        bytes16 scId;
+        bytes32 target;
+        bytes payload;
+    }
```
lemunozm, line 617: I'm curious how this message would work with any non-evm implementation of CV (i.e. future solana implementation). Could the caller from CP know the kind of destination chain to encode correctly the call?

mustermeiszer: Not sure tbh. I just choose target 32bytes as we have that everywhere else too. Can also reduce to 20

src/vaults/InvestmentManager.sol
mustermeiszer: removed asset, kept token.

src/common/libraries/MessageLib.sol
mustermeiszer: Done

src/common/libraries/MessageLib.sol
lemunozm: I'll keep 32 as identifier of the program where to run the payload in the other generic chain. I think it would work, as long as the FM knows the CV chain when he sets the target & payload.

src/vaults/factories/ERC7540VaultFactory.sol
```diff
@@ -2,43 +2,39 @@
...
         ERC7540Vault vault = new ERC7540Vault(poolId, trancheId, asset, tranche, root, investmentManager);
 
         vault.rely(root);
+        vault.rely(investmentManager);
         uint256 wardsCount = wards_.length;
         for (uint256 i; i < wardsCount; i++) {
             vault.rely(wards_[i]);
         }
 
-        Auth(investmentManager).rely(address(vault));
```
lemunozm, line 35: I think you need to keep this line to be able to let the vault to call investmentManager

src/vaults/InvestmentManager.sol
```diff
@@ -35,6 +34,8 @@ contract InvestmentManager is Auth, IInvestmentManager {
     IGateway public gateway;
     IPoolManager public poolManager;
 
+    mapping(uint64 poolId => mapping(bytes16 trancheId => mapping(uint128 assetId => address vault))) public _vault;
```
lemunozm:
```suggestion
    mapping(uint64 poolId => mapping(bytes16 trancheId => mapping(uint128 assetId => address vault))) public vault;
```

It's a public name

src/vaults/InvestmentManager.sol
```diff
@@ -651,4 +694,14 @@ contract InvestmentManager is Auth, IInvestmentManager {
         ITranche share = ITranche(IERC7540Vault(vault).share());
         return share.checkTransferRestriction(from, to, value);
     }
+
+    /// @inheritdoc IInvestmentManager
+    function getVault(uint64 poolId, bytes16 trancheId, uint128 assetId) public view returns (address) {
+        return _vault[poolId][trancheId][assetId];
+    }
```
lemunozm: We do not need this implementation. It's automatically exposed as `vault`.

src/vaults/InvestmentManager.sol
```diff
@@ -651,4 +694,14 @@ contract InvestmentManager is Auth, IInvestmentManager {
...
+
+    /// @inheritdoc IInvestmentManager
+    function getVault(uint64 poolId, bytes16 trancheId, address asset) public view returns (address) {
+        return _vault[poolId][trancheId][poolManager.assetToId(asset)];
+    }
```
lemunozm: To be aligned with the code style, this should be `vaultByAddress` or some name without  "get" prefix

src/vaults/PoolManager.sol
```diff
@@ -43,38 +44,51 @@ contract PoolManager is Auth, IPoolManager {
...
-        vaultFactory = ERC7540VaultFactory(vaultFactory_);
         trancheFactory = ITrancheFactory(trancheFactory_);
+
+        for (uint256 i = 0; i < vaultFactories.length; i++) {
+            address factory = vaultFactories[i];
+            vaultFactory[factory] = true;
+        }
+        // Need to be ward in itself in order to call into `IUpdateContract(target).update(..) where target is
+        // `address(this)`
+        rely(address(this));
```
lemunozm: If it helps, you can also do a delegate call if `target == address(this)` to avoid these issues:

i.e:
```solidity
if target == address(this) {
    target.delegateCall(IUpdateContract.update, ...);
} else {
    IUpdateContract(target).update(..)
}
```

And therefore avoid calling the contract as an external if it's the same (which means avoid calling rely/deny to itself in several points in the codebase).

Note that this works as long as the IUpdateContract has the same reliance as PoolManager, which, in this case, is the Gateway.

src/vaults/PoolManager.sol
```diff
@@ -43,38 +44,51 @@ contract PoolManager is Auth, IPoolManager {
...
     }
 
+    function file(bytes32 what, address factory, bool status) external auth {
+        if (what == "vaultFactory") {
+            vaultFactory[factory] = status;
+        } else {
+            revert("PoolManager/file-unrecognized-param");
+        }
+        emit File(what, factory, status);
+    }
```
lemunozm, line 85: Question. I also saw this in the gateway. Curious, why is there overuse of `file()` naming? Could not be called `setVaultFactory()` which is more expressive? cc @hieronx

src/vaults/PoolManager.sol
```diff
@@ -291,91 +312,93 @@ contract PoolManager is Auth, IPoolManager {
...
+    /// @notice The pool manager either deploys the vault if a factory address is provided or it simply links/unlinks
+    /// the vault
+    function update(uint64 poolId, bytes16 trancheId, bytes memory payload) public auth {
+        if (payload.length != 128) {
+            revert("PoolManager/invalid-update-payload-length");
         }
 
-        _pools[poolId].tranches[trancheId].token = token;
+        (address factory, uint128 assetId, bool isLinked_, address vault) =
+            abi.decode(payload, (address, uint128, bool, address));
```
lemunozm: I still think we need some kind of index or "submesage id" here to avoid failing with a generic decode error in case the message is sent to a non-expected `IContractUpdate`.

If the payload is generic, it means that we need to be able to choose between different payload structures, but we can't if we can not identify it.

src/vaults/PoolManager.sol
```diff
@@ -291,91 +312,93 @@ contract PoolManager is Auth, IPoolManager {
...
-        );
-
-        if (undeployedTranche.hook != address(0)) {
-            ITranche(token).file("hook", undeployedTranche.hook);
+    // --- IUpdateContract implementation ---
+    /// @inheritdoc IUpdateContract
+    /// @notice The pool manager either deploys the vault if a factory address is provided or it simply links/unlinks
+    /// the vault
+    function update(uint64 poolId, bytes16 trancheId, bytes memory payload) public auth {
+        if (payload.length != 128) {
```
lemunozm: Curious about this limitation, why it exists?

src/vaults/PoolManager.sol
```diff
@@ -291,91 +312,93 @@ contract PoolManager is Auth, IPoolManager {
...
         }
 
-        // Link vault to tranche token
-        IAuth(tranche.token).rely(vault);
-        ITranche(tranche.token).updateVault(asset, vault);
+        address manager = IBaseVault(vault).manager();
+        // NOTE - Reverting the three actions below is not easy. We SHOULD do that if we phase-out a manager
+        IAuth(tranche.token).rely(manager);
+        escrow.approveMax(tranche.token, manager);
+        escrow.approveMax(asset, manager);
```
lemunozm, line 379: Question. What's the difference here between `token` and `asset`?

src/vaults/PoolManager.sol
lemunozm: Given we also need to encode it in CP, could we add it to `MessageLib`? Similar to how the submessages of `UpdateRestriction` are added.

Note we also need to perform encoding/decoding later in integration tests. So at least we have 4 points in the code where to set this same structure.

src/vaults/PoolManager.sol
```diff
@@ -407,29 +425,30 @@ contract PoolManager is Auth, IPoolManager {
...
     /// @inheritdoc IPoolManager
-    function isAllowedAsset(uint64 poolId, address asset) public view returns (bool) {
-        return _pools[poolId].allowedAssets[asset];
+    function isLinked(uint64, /* poolId */ bytes16, /* trancheId */ address, /* asset */ address vault)
+        public
+        view
+        returns (bool)
+    {
+        // TODO: Check whether to check against asset of vault in storage??
+        return _vaultToAsset[vault].isLinked;
```
lemunozm, line 456: Reminder here to close it before merge.

I think it makes sense as it is (?)

src/vaults/PoolManager.sol
```diff
@@ -407,29 +425,30 @@ contract PoolManager is Auth, IPoolManager {
...
+    }
+
+    function isAllowedAsset(uint64 poolId, address asset) public view override returns (bool) {
+        return allowedAssets[poolId][asset];
```
lemunozm, line 460: This is already autogenerated as `allowedAssets()` so it can be removed.

lemunozm (commented):
Despite the amount of comments, I think the main logic is correct! Good job here 🚀

src/vaults/PoolManager.sol
mustermeiszer: You mean a submessage per receiving contract, that only the receiving contract of the call itself is aware about? Feels like just more overhead.

src/vaults/PoolManager.sol
mustermeiszer: asset -> e.g. USDC
token -> e.g. LTF

src/vaults/PoolManager.sol
mustermeiszer: We will never know all submessages as this is also intended to work for new contracts or contracts that an FM wants to use. 

Not sure it really helps IMO.

src/vaults/PoolManager.sol
mustermeiszer: Old structure. Don't wanna touch that here.

src/vaults/PoolManager.sol
hieronx: 1) an old convention we've had for a long time, adopted from Maker, 2) it is quite nice for monitoring: you just need to listen to `File` events on the contracts to know if any of the key parameters or dependencies have changed.

hieronx: It is only intended to be used for parameter changes that are not updated during regular usage of the protocol.

See style guide for details

src/vaults/PoolManager.sol
lemunozm: Ah ok, right. Just remember I previously understood your point. The contract itself acts as discriminator of the payload 👍🏻

src/vaults/PoolManager.sol
lemunozm: Got it!. It would make sense then, call the method with the proper name, and emit the `File` event inside of it?

src/vaults/PoolManager.sol
lemunozm: As far as `PoolManager` is concerned, right now, we know the message structure, because the deserialization happens inside of it. 

If tomorrow there are other contracts added to the repo we just need to update the `MessageLib`. The point is that if we now modify some of these message, it's very dificult to update all parts correctly in the codebase.

src/vaults/PoolManager.sol
lemunozm: Old structure? Not sure if I follow. Isn't it a new method added?

src/vaults/PoolManager.sol
mustermeiszer: I will keep the current structure. We can change that rather in a follow up for all files IMO.

src/vaults/PoolManager.sol
mustermeiszer: nice!

src/vaults/PoolManager.sol
mustermeiszer: No. It's an old one. Will be removed in the ERC6909 pr. So don't wanna conflict there.

src/vaults/PoolManager.sol
lemunozm: Sure! Just asking for curiosity. Nothing to apply in this PR

src/vaults/PoolManager.sol
lemunozm: Ah ok 👍🏻

src/vaults/PoolManager.sol
mustermeiszer: Makes total sense. Adapted to submessage.

src/vaults/PoolManager.sol
mustermeiszer: Removed by using a submessage.

src/vaults/PoolManager.sol
mustermeiszer: Done in https://github.com/centrifuge/protocol-v3/pull/108/commits/d3253905856fe80adb37067544dd6dad13d56fd2

src/vaults/PoolManager.sol
mustermeiszer: Done in https://github.com/centrifuge/protocol-v3/pull/108/commits/3b9c7b0941b9365c64f23db3f387a1ad9d149b3f

src/vaults/InvestmentManager.sol
mustermeiszer: Done

src/vaults/InvestmentManager.sol
mustermeiszer: Done

src/vaults/PoolManager.sol
```diff
@@ -266,7 +266,7 @@ contract PoolManager is Auth, IPoolManager, IUpdateContract {
     /// @inheritdoc IPoolManager
     function updateContract(uint64 poolId, bytes16 trancheId, address target, bytes memory update_) public auth {
         if (target == address(this)) {
-            (bool success, bytes memory data) = address(this).delegatecall(
+            (bool success, ) = address(this).delegatecall(
```
lemunozm: It's interesting to return the error from data in case it fails. You can check how `Multicall.sol` does it.

src/misc/Auth.sol
```diff
@@ -22,13 +22,13 @@ abstract contract Auth is IAuth {
...
     /// @inheritdoc IAuth
-    function rely(address user) external auth {
+    function rely(address user) public auth {
         wards[user] = 1;
         emit Rely(user);
     }
 
     /// @inheritdoc IAuth
-    function deny(address user) external auth {
+    function deny(address user) public auth {
```
lemunozm, line 31: I think these can be `external` again.

src/vaults/PoolManager.sol
mustermeiszer: Done

src/misc/Auth.sol
mustermeiszer: No, the investment manager is still using them from `this`

lemunozm: Could you link me to that line?

mustermeiszer: * https://github.com/centrifuge/protocol-v3/blob/ec10fb2a5fe42765fd24e1ccd826ea0c5970a1ba/src/vaults/InvestmentManager.sol#L79
* https://github.com/centrifuge/protocol-v3/blob/ec10fb2a5fe42765fd24e1ccd826ea0c5970a1ba/src/vaults/InvestmentManager.sol#L98

lemunozm: Ok, I see, later the vault needs to call manager 👍🏻.

lemunozm (approved):
LGTM! Great work 🚀

src/vaults/PoolManager.sol
wischli: @mustermeiszer Can I remove the TODO comment in #98?

PR #109, opened 2025-02-26
feat: Sync Invest ERC-7540

mustermeiszer:
### New contracts
* sync invest manager: `SyncRequests`
* `SyncDepositVaultFactory`
* `BaseVault`, `AsyncRedeemVault`, `BaseSyncDepositVault`, `SyncDepositVault`
  *  TODO: Add `AsyncDepositVault`

### Refactors

sync-deposit-refactor_v6

[Plantuml Editor](fLTDRziu4BtpLmnQe07QAkj3ls1uZJRfRciiKXWDsalR0osDBI4IgIMffCQY_xr3od98YRQ5R2wnEM_Ul1aEX_HRRRWoJLMoCoXnRS18KCKcDv0L2bDJIC6ChEtoIXeZAs_zR8SactEzj-YRGjHSyGhAGg3PrWXIcLnw1Y4pr5Z3ouctgBYuiOjl9eoLk46HEMO6C5lDPwaifPfk5AAOpvwaEQOtyzaVRfN-h-Omctt2k_aSu4-W3v3AgfO2HUY_AXiaTwwKl4jS0Y791Td1MJNnL5pm3IfGM7ARciwBMWTyTQFgqZBUAHHf7f8jdV6UhgZeQqMI75cFbT9jasPUITAIQmt_KSWBdjvib6n4Tca3TCa-HhvU8xS0gJ9KeN4oUVLgCc6_6EtMGXXKQvwY1awLlMfJqwmcOle4YuMuHMsil9tDnIqwDrXSS8r3q2vO27sFjTJ5SSndp12hev2ll2ax4h84ts4Apu1dsNEgOoLl4MwjCOxI6iqYiwrw6EGi_nHK_q32ExqLwPXq7714JZlSlmtLqVQIa0QxqoNNFJifgnNU5hBHvJQoHI7rzKYfrtw43a5Dicy55wRh4DS6RIEUq7vCUaHvotbAp0Tky8vlFrBJqZZPz9joGEpZlJ1Ng3MPbqhI2T1oO7XVw9eR6YkAWhUM2rvoaU9rZhXNtLgMKfRTobwewwYsMdobZUAfsOLuP4-ECp_oDgGFyXhe9DOIUhizm7lRzNNNyR6zShPme_xwVFdwzOj9vtKM0f_1kYa9nUtY_Z3TRwIlPo-7401TbVwcBAGgi2rK-zTNssAsFNkVVuVoiWRuPJkdFtsU9jyUjtFwHn8z95C8fxcTqom6739vtHxZ2EV1q_Fpyu6RpcLJPkvk-2uAGGKsZK8Mmlp0IGAd8zF0l-9rJGT6mnDO11UFPWT6UPZUg2AD8GgB4ESvMkZO31vLxo7B62snS25Lx_9u64b_MXva6P7K_oCPcywXTXmZuiIS3XMD86JxcHfrtK_HkFS1vzXOZHCCmWn5FDXxd8AerWVmUBL8oDi-a7yayqEllJs5UqXt3-HXGW7LiPFNyhZxfBjHeva5X8TcGKlcStaxuH4SwAyW9_Tkz9Y6YkxvulEuJpUMxQYwG1CRt9bKfwIZw3OzLKXdJGzYXYwHRkaSMV_n1DRXv96UGgF9U_30UeYt7vgylGSO0NFd0b_gZ2WkfN1F9UPZagIfa-56xHo_YzRrYlzSgYB5TnjaB82s9Oey27sIGki6CxhGkIBXCN1BCop8rBxEYlKMBfVqr9EDWO2BkErFjiQ9p22NPKONEsDlwTSY_NZ_3G00)

_NOTE: I will provide a diagram of the refactored interfaces and contracts on Wednesday. For now this description must suffice._

- ❗Split `IInvestmentManager` into lots of fragments: `IBaseInvestmentManager`, `IDepositManager`, `IRedeemManager`, `I{Async, Sync}DepositManager` and `I{Async, Sync}RedeemManager`
  - `IBaseInvestmentManager` ⊂ `{IDepositManager, IRedeemManager}` ⊂ `{IAsyncDepositManager, IAsyncRedeemManager}`
  - `IDepositManager` ⊂ `ISyncDepositManager`
- ❗ Rename `InvestmentManager` to `AsyncRequests`
- ❗Add `BaseVault` and `AsyncRedeemVault` based on previous `ERC7540Vault`
  - Rename `ERC7540Vault` to `AsyncVault`
  - `BaseVault` ⊂ `AsyncRedeemVault` ⊂ `ERC7540Vault`
  - `BaseVault` ⊂ [`AsyncRedeemVault`, `BaseSyncDepositVault`] ⊂ `SyncDepositVault`
  - `BaseVault` should be used for ANY vault, independent of (a)synchronicity
- @mustermeiszer : "I’ve also moved the price conversion logic from the investment manager into a new library, so this can be re-used by the instant deposit manager"

### Open questions for reviewers
1. ✅  @hieronx: Can we assume all vaults to be `ERC7575`? Previously, that was the case. I tried to decouple `BaseVault` from it but it lead to a lot of problems. --> Yes
2. ✅  Contract inheritance of `BaseVault` ⊂ `AsyncRedeemVault` ⊂ `{ERC7540Vault, SyncDepositAsyncRedeemVault}` okay? --> [Yes](https://kflabs.slack.com/archives/C07PG2EUR9C/p1742981107592769?thread_ts=1742491274.772479&cid=C07PG2EUR9C)
3. ✅  Adapt similar inheritance for investment managers? I.e. create contracts for sync deposit, async deposit, async redeem --> Doesn't make sense because we will only have fully sync and async investment managers

### What is next, for fully working MVP:
- Decide on and update vault deployment flow (permissionless, and require UpdateContract to enable?)
- ~~Update InvestmentManager to support multiple vaults per pool id + tranche id + asset~~

### Later
- Add rate limiting
- Separate price retrieval in instant manager into a new oracle contact, so we can customize this later without swapping the instant manager (e.g. adding linear interpolation for certain use cases)


src/vaults/interfaces/IERC7540.sol
```diff
@@ -143,7 +143,7 @@ interface IERC7540Redeem is IERC7540Operator {
 interface IERC7540CancelDeposit {
     event CancelDepositRequest(address indexed controller, uint256 indexed requestId, address sender);
     event CancelDepositClaim(
-        address indexed controller, address indexed receiver, uint256 indexed requestId, address sender, uint256 assets
+        address indexed receiver, address indexed controller, uint256 indexed requestId, address sender, uint256 assets
```
wischli: Was this change intended or an undesired merge conflict @hieronx ? AFAICT, we are diverging from the standard here.

src/vaults/InstantManager.sol
```diff
@@ -0,0 +1,115 @@
...
+        // TODO: implement rate limit
+        return type(uint256).max;
+    }
+
+    /// @inheritdoc IInstantManager
+    function previewDeposit(address vault, address, /* sender */ uint256 assets) public view returns (uint256 shares) {
+        IBaseVault vault_ = IBaseVault(vault);
+        (uint128 latestPrice, uint64 computedAt) =
+            poolManager.getTranchePrice(vault_.poolId(), vault_.trancheId(), vault_.asset());
+        require(block.timestamp - computedAt <= maxPriceAge[vault], PriceTooOld());
```
mustermeiszer: COMMENT

`maxPriceAge` should be "settable" by the CP side. I would propose the contract implements `IUpdateContract` which would be the easiest. 

Ther will also be in the long additional settings like a `maxDeposits` which sets the maximum of deposits per timewindow. But that is out of scope for this PR. Just do let you know that we will use the `IUpdateContract` here in detail.

mustermeiszer: as an edition it would be gread if we can set an valuation source through the Update method. different ones for invest/redeem if both are synchronous. 

Then we can check 
* max age
* forward both computed at and price and amount to the valuation method

just thinking out loud.

wischli: @mustermeiszer I am unsure what should be scope and what is out of scope for this PR: AFAICT, this PR should add a new `UpdateContract` type (and probably two messages?) for setting the `maxDeposit` and `maxMint` per investor in CV.  

In a subsequent PR we will then
> Separate price retrieval in instant manager into a new oracle contact, so we can customize this later without swapping the instant manager (e.g. adding linear interpolation for certain use cases)

Do you agree? Am I missing something?

src/vaults/interfaces/IERC7540.sol
hieronx: That is indeed a mistake, not sure how that happened.

src/vaults/InstantManager.sol
hieronx: Right now we should just support setting the `maxPriceAge` value per vault through `UpdateContract`.

Then in subsequent PRs, we should:
- Separate price retrieval like you mention
- Add rate limit, but this wouldn't be per investor, rather per vault per day (e.g. max 100 assets deposited per day across all users for 1 vault).

mustermeiszer:
> his PR should add a new UpdateContract type (and probably two messages?) for setting the maxDeposit and maxMint per investor in CV.

For this PR the `update` logic should allow to 
* set `maxPriceAge`
* set `maxDeposit` &rarr; globally
* set `maxMint` &rarr; globally

hieronx:
> set maxDeposit → globally
> set maxMint → globally

I don't think we should set these parameters directly. I responded here too: https://github.com/centrifuge/protocol-v3/pull/109/files/b4e08e246d2034973b4867995a4041811bc558cb#r2016102858 But basically we should add a rate limit, but this wouldn't be per investor, rather per vault per day (e.g. max 100 assets deposited per day across all users for 1 vault).

And I would still make this a next step in a future PR.

src/vaults/interfaces/IERC7540.sol
lemunozm: Good catch!

wischli:
> > set maxDeposit → globally
> > set maxMint → globally
> 
> I don't think we should set these parameters directly. I responded here too: https://github.com/centrifuge/protocol-v3/pull/109/files/b4e08e246d2034973b4867995a4041811bc558cb#r2016102858 But basically we should add a rate limit, but this wouldn't be per investor, rather per vault per day (e.g. max 100 assets deposited per day across all users for 1 vault).
> 
> And I would still make this a next step in a future PR.

I didn't include this in this PR. Given the time constraints, I also propose to do it in a follow-up.

src/vaults/BaseVaults.sol
```diff
@@ -0,0 +1,404 @@
...
+
+enum VaultKind {
+    /// @dev Refers to ERC7540Vault
+    Async,
+    /// @dev not yet supported
+    Sync,
+    /// @dev Refers to SyncDepositAsyncRedeemVault
+    SyncDepositAsyncRedeem,
+    /// @dev Not yet supported
+    AsyncDepositSyncRedeem
```
hieronx: Fwiw I cannot think of a use case for this today, think it's likely we'll never build this.

hieronx:
> Can we assume all vaults to be ERC7575? Previously, that was the case. I tried to decouple BaseVault from it but it lead to a lot of problems.

Yeah that should be fine

src/vaults/SyncInvestmentManager.sol
```diff
@@ -0,0 +1,210 @@
...
+
+        IAuth(token).deny(vaultAddr);
+        ITranche(token).updateVault(vault_.asset(), address(0));
+        deny(vaultAddr);
+    }
+
+    // --- IDepositManager ---
+    /// @inheritdoc IDepositManager
+    function mint(address vaultAddr, uint256 shares, address receiver, address owner)
+        external
```
hieronx: `mint` and `deposit` should be `auth`, only callable by vaults.

src/vaults/SyncDepositAsyncRedeemVault.sol
```diff
@@ -0,0 +1,41 @@
...
+import "src/vaults/interfaces/IERC7540.sol";
+import "src/vaults/interfaces/IERC7575.sol";
+
+/// @title  SyncDepositAsyncRedeemVault
+/// @notice Partially (a)synchronous Tokenized Vault implementation with synchronous deposits and asynchronous
+/// redemptions following ERC-7540.
+///
+/// @dev    Each vault issues shares of Centrifuge tranches as restricted ERC-20 or ERC-6909 tokens
+///         against asset deposits based on the current share price.
+contract SyncDepositAsyncRedeemVault is AsyncRedeemVault, SyncDepositVault {
```
hieronx: nit:
```suggestion
contract SyncDepositAsyncRedeemVault is SyncDepositVault, AsyncRedeemVault {
```

mustermeiszer:
> "I’ve also moved the price conversion logic from the investment manager into a new library, so this can be re-used by the instant deposit manager"

BSM side will love this!!

src/common/libraries/MessageLib.sol
```diff
@@ -654,6 +655,29 @@ library MessageLib {
         return abi.encodePacked(UpdateContractType.Permission, t.who, t.allowed);
     }
 
+    //---------------------------------------
+    //   UpdateContract.MaxPriceAge (submsg)
+    //---------------------------------------
+
+    struct UpdateContractMaxPriceAge {
+        bytes32 vault;
```
mustermeiszer, line 663: QUESTION

Could we also make that on an `assetId` base ? I guess the vault represents that atm, right?

src/vaults/BaseVaults.sol
```diff
@@ -0,0 +1,404 @@
...
+    }
+
+    /// @notice Ensures msg.sender can operate on behalf of controller.
+    function _validateController(address controller) internal view {
+        require(controller == msg.sender || isOperator[controller][msg.sender], "ERC7540Vault/invalid-controller");
+    }
+}
+
+abstract contract AsyncRedeemVault is BaseVault, IAsyncRedeemVault {
+    IAsyncRedeemManager public asyncManager;
```
mustermeiszer: CLARIFICATION

I am not sure, but IIRC, the `poolManager.linkVault(..)` logic gives permission to the manager retrieved vie `vault.manager()`. Isn't it a problem for the logic, if we have multiple managers per vault?

mustermeiszer: Also, a vault using multiple managers needs to be added to the `IM.addVault()` logic, to multiple, right?

mustermeiszer: I think we can simplify that, but just hardcoding `AsyncInvemstneManager` and `SyncInvesmentManager` to the `PoolManager`

src/vaults/ERC7540Vault.sol
mustermeiszer, line 1: QUESTIONS

Why don't we have `AsyncInvestVault` in the `BasVaults.sol` and make this one here simply inherent from the three relevant impls - `BaseVault`, `AsyncRedeemVault`, and `AsyncInvestVault`?

src/vaults/PoolManager.sol
```diff
@@ -354,9 +356,17 @@ contract PoolManager is Auth, IPoolManager, IUpdateContract, IPoolManagerGateway
...
-        IVaultManager(manager).addVault(poolId, trancheId, vault, assetIdKey.asset, assetId);
+
+        IBaseInvestmentManager manager = IBaseVault(vault).manager();
+        IVaultManager(address(manager)).addVault(poolId, trancheId, vault, assetIdKey.asset, assetId);
+
+        // For sync deposit & async redeem vault, also add vault to async manager (base manager is sync one)
+        (bool isSyncDepositAsyncRedeemVault, address asyncRedeemManager) = isPartiallySyncVault(vault, manager);
+        if (isSyncDepositAsyncRedeemVault) {
+            IVaultManager(asyncRedeemManager).addVault(poolId, trancheId, vault, assetIdKey.asset, assetId);
+        }
```
mustermeiszer: NIT

No blocker, but potential follow-up. What about we hardcode the different investment managers in the `PoolManager` again. I removed those, but know it feels like it is creating a level of abstraction that is more confusing than helping. 

Something like 

* `ISyncInvestmentManager public syncInvestmentManager`
* `IAsyncInvestmentManager public asyncInvestmentManager` 

and then somehow have a clearer adding functionality? This is confusing me a bit and I worry we lose knowledge about this in the long run.

What about we do a default `addVault()` for all vaults to both managers and then we do the `isPartiallySyncVault` in the managers itself and only add them if it is relevant for them?

src/vaults/SyncInvestmentManager.sol
```diff
@@ -0,0 +1,210 @@
...
+
+    /// @dev Issues shares to the receiver and instruct the Balance Sheet Manager to react on the issuance and the
+    /// updated holding value
+    function _issueShares(address vaultAddr, address receiver, uint128 shares) internal {
+        SyncDepositAsyncRedeemVault vault = SyncDepositAsyncRedeemVault(vaultAddr);
+
+        uint64 poolId_ = vault.poolId();
+        bytes16 scId_ = vault.trancheId();
+        VaultDetails memory vaultDetails = poolManager.vaultDetails(vaultAddr);
+        D18 pricePerShare = d18(_pricePerShare(vaultAddr, poolId_, scId_, vaultDetails.assetId));
```
mustermeiszer: NIT

why wrap here and not in `_pricePerShare()`. Wrapping into `D18` like that just feels like it should be done by the provider.

src/vaults/SyncInvestmentManager.sol
```diff
@@ -0,0 +1,210 @@
...
+    function addVault(uint64 poolId, bytes16 trancheId, address vaultAddr, address asset_, uint128 assetId)
+        public
+        override
+        auth
+    {
+        SyncDepositAsyncRedeemVault vault_ = SyncDepositAsyncRedeemVault(vaultAddr);
+        address token = vault_.share();
+
+        require(vault_.asset() == asset_, "SyncInvestmentManager/asset-mismatch");
+        require(vault[poolId][trancheId][assetId] == address(0), "SyncInvestmentManager/vault-already-exists");
```
mustermeiszer: NIT

Sync manager can theoretically handle multiple vaults per asset.

mustermeiszer: CR 

The storage lies in the `BaseInvestmentManager`, which is a problem as there cannot be a fully-async and fully-sync vault in parallel, which must be possible.

src/vaults/SyncInvestmentManager.sol
```diff
@@ -0,0 +1,210 @@
...
+        returns (uint256 assets)
+    {
+        assets = previewMint(vaultAddr, owner, shares);
+
+        _issueShares(vaultAddr, receiver, shares.toUint128());
+    }
+
+    /// @inheritdoc IDepositManager
+    function deposit(address vaultAddr, uint256 assets, address receiver, address owner)
+        external
```
mustermeiszer: CR

`auth`

src/vaults/SyncInvestmentManager.sol
```diff
@@ -0,0 +1,210 @@
...
+        D18 pricePerShare = d18(_pricePerShare(vaultAddr, poolId_, scId_, vaultDetails.assetId));
+
+        PoolId poolId = PoolId.wrap(poolId_);
+        ShareClassId scId = ShareClassId.wrap(scId_);
+
+        // Mint shares for receiver notify CP about issued shares
+        balanceSheetManager.issue(poolId, scId, receiver, pricePerShare, shares.toUint128(), false);
+
+        // Notify CP about updated holding value
+        balanceSheetManager.updateValue(poolId, scId, vaultDetails.asset, vaultDetails.tokenId, pricePerShare);
```
mustermeiszer: CR

The CP side needs to be informed about the amount of holdings that have been provided. Needs to call 

`balanceSheetManager.deposit(poolId, scId, vaultDetails.asset, vaultDetails.tokenId, PROVIDER?,  assets,  PRICE_OF_ASSET_IN_POOL_CURRENCY)`

* `PROVIDER?` &rarr; probably the escrow, as the vaults are sending to it.
* `PRICE_OF_ASSET_IN_POOL_CURRENCY` &rarr; that is really hard to solve. I would opt for `d18(1,1)` for now.

mustermeiszer: I think I will implement the solution via providing a price of share class in pool currency, then we can always get the asset price by the div between share class pool currency price and asset price.

src/vaults/SyncInvestmentManager.sol
```diff
@@ -0,0 +1,210 @@
...
+        returns (uint128 latestPrice)
+    {
+        (uint128 price, uint64 computedAt) = poolManager.tranchePrice(poolId, trancheId, assetId);
+        require(block.timestamp - computedAt <= maxPriceAge[vaultAddr], PriceTooOld());
+        return price;
+    }
+
+    /// @dev Issues shares to the receiver and instruct the Balance Sheet Manager to react on the issuance and the
+    /// updated holding value
+    function _issueShares(address vaultAddr, address receiver, uint128 shares) internal {
```
mustermeiszer: CR

Missing `IPerPoolEscrow.increasePendingDeposit()` & `IPerPoolEscrow.decreaseDeposit(..)` for said asset so that it actually becomes available.

src/vaults/AsyncInvestmentManager.sol
mustermeiszer, line 1: CR 

COULD also be a follow-up PR. But we are missing all logic around handling `IPerPoolEscrow`. This is not an exploit at the moment, just means, that in the current logic issuers do not have access to their funds.

mustermeiszer (commented):
Love the structure of the vaults and managers. Very easy to understand like that. Got a few blocking CRs. 

I would propose we close all CRs - excluding the `IPerPoolEscrow` logic and do this in a follow-up PR. As the missing `IPerPoolEscrow` is not an exploit in the current state, it just means, that issuers won't have access to their funds.

src/common/libraries/MessageLib.sol
wischli: > Could we also make that on an `assetId` base?

Can you clarify if you want to support it with either `assetId` or `vault` or just with `assetId`? Definitely not a problem as we can map an asset Id easily to its vault address with the PM.

src/vaults/BaseVaults.sol
wischli: > I am not sure, but IIRC, the `poolManager.linkVault(..)` logic gives permission to the manager retrieved vie `vault.manager()`. Isn't it a problem for the logic, if we have multiple managers per vault?

The current code executes `{add, remove}Vault` as well as `_approveManager` in case the target vault has a secondary (async) manager. See `isPartiallySyncVault` executions in the the `PoolManager.sol` contract.

wischli: > I think we can simplify that, but just hardcoding `AsyncInvemstneManager` and `SyncInvesmentManager` to the `PoolManager`

WDYT about that @hieronx ?

src/vaults/ERC7540Vault.sol
wischli: > Why don't we have `AsyncInvestVault` in the `BasVaults.sol` and make this one here simply inherent from the three relevant impls - `BaseVault`, `AsyncRedeemVault`, and `AsyncInvestVault`?

Great question: This is an open TODO which I didn't prioritize to get this PR into a reviewable state.

src/vaults/SyncInvestmentManager.sol
wischli: We are using the `uint128` type of it for `previewMint` and `previewDeposit` and the `D18` one only here.

src/vaults/BaseVaults.sol
hieronx: Yeah good idea, that simplifies it quite a bit!

src/vaults/SyncInvestmentManager.sol
wischli: Sorry, I mistakenly thought the _1 async vault per triplet_ also extended to the sync vaults. Will fix this in a follow up PR if that's fine with you.

src/vaults/SyncInvestmentManager.sol
wischli: Would it be fine for you if I provide this in a follow up PR?

src/vaults/BaseVaults.sol
wischli: Will remove then, thanks!

src/vaults/PoolManager.sol
```diff
@@ -440,6 +458,41 @@ contract PoolManager is Auth, IPoolManager, IUpdateContract, IPoolManagerGateway
...
+        returns (bool isPartial, address otherManager)
+    {
+        if (
+            manager.supportsInterface(type(ISyncInvestmentManager).interfaceId)
+                && IERC165(vault).supportsInterface(type(IAsyncRedeemVault).interfaceId)
+        ) {
+            isPartial = true;
+            otherManager = address(IAsyncRedeemVault(vault).asyncManager());
+        }
+    }
```
mustermeiszer: CLARIFICATION 

We have [this](https://github.com/centrifuge/vaults-internal/blob/ce5a95c3eb8aa7b4069c2919be3cc9037eb48967/src/ERC7540Vault.sol#L278-L284) `function supportsInterface(...)` in our legacy vaults. 

IIUC, @wischli, we have no compatibility issues with out current logic, right? Just want to ensure, we are not creating incompatability.

src/vaults/SyncInvestmentManager.sol
mustermeiszer: ahh, okay, makes sense!

src/vaults/SyncInvestmentManager.sol
mustermeiszer: Fine for me!

src/vaults/SyncInvestmentManager.sol
mustermeiszer: Fine for me!

src/common/libraries/MessageLib.sol
mustermeiszer: Na, all good! Lets keep it per vault.

src/vaults/PoolManager.sol
wischli: Thanks for checking! Our legacy vaults will just return `(false, address(0))` here because their corresponding manager doesn't implement `ISyncInvestmentManager` directly (but implicitly) and the vault doesn't implement `IAsyncRedeemVault` directly (but implicitly). So the `supportsInterface` checks will just be false which is fine because those vaults are not partially sync.

src/vaults/BaseVaults.sol
wischli: For me, this was the best solution which also didn't require reintroducing investment manager storage to the pool manager: [bb98751](https://github.com/centrifuge/protocol-v3/pull/109/commits/bb987510d8aa050c18d05b5d3cfacf38d2781532)

src/vaults/PoolManager.sol
wischli: See https://github.com/centrifuge/protocol-v3/pull/109#discussion_r2020729287

src/vaults/SyncInvestmentManager.sol
wischli: @mustermeiszer I am unsure how to construct the journal entry meta for the `BalanceSheetManager.deposit` call which requires it: AFAICT, for `SyncManager.deposit`, the `credits` must reflect the deposited assets by the corresponding account. In case I am right, then for `SyncManager.mint` both credits and debits should be empty, right?

wischli: Also, what would be the corresponding `AccountId` in case of a deposit? I have no clue how to generate the `uint24 prefix` correctly with the given information. I feel like the `Meta` struct should actually be generated as part of `BalanceSheetManager.deposit`.

foundry.toml
```diff
@@ -65,8 +65,8 @@ max_test_rejects = 65_536 # default value
   ]
 
 [profile.smt.model_checker.contracts]
-  "src/vaults/ERC7540Vault.sol" = ["ERC7540Vault"]
-  "src/vaults/InvestmentManager.sol" = ["InvestmentManager"]
+  "src/vaults/AsyncVault.sol" = ["AsyncVault"]
+  "src/vaults/AsyncManager.sol" = ["InvestmentManager"]
```
lemunozm:
```suggestion
  "src/vaults/AsyncManager.sol" = ["AsyncManager"]
```

script/VaultsDeployer.s.sol
```diff
@@ -146,14 +169,16 @@ contract VaultsDeployer is CommonDeployer {
...
+        IAuth(syncDepositVaultFactory).deny(msg.sender);
+        IAuth(trancheFactory).deny(msg.sender);
+        IAuth(restrictionManager).deny(msg.sender);
+        IAuth(restrictedRedemptions).deny(msg.sender);
+        asyncManager.deny(msg.sender);
+        syncManager.deny(msg.sender);
+        poolManager.deny(msg.sender);
+        escrow.deny(msg.sender);
+        routerEscrow.deny(msg.sender);
+        vaultRouter.deny(msg.sender);
```
lemunozm: This comes from some git conflict, I guess. It should be `deployer` now.

src/vaults/BaseInvestmentManager.sol
```diff
@@ -0,0 +1,80 @@
...
+
+abstract contract BaseInvestmentManager is Auth, IBaseInvestmentManager {
+    using MathLib for uint256;
+
+    address public immutable root;
+    address public immutable escrow;
+
+    IPoolManager public poolManager;
+
+    constructor(address root_, address escrow_) Auth(msg.sender) {
```
lemunozm, line 26: @hieronx given the current new understanding of how the deployer works. We can use `msg.sender` here (as it was in the original InvestmentManager) in all our contracts and avoid the `deployer`.

That should work, and we have the current `InvestmentManager` as proof of that.

src/vaults/BaseInvestmentManager.sol
```diff
@@ -0,0 +1,80 @@
...
+        root = root_;
+        escrow = escrow_;
+    }
+
+    /// @inheritdoc IRecoverable
+    function recoverTokens(address token, uint256 tokenId, address to, uint256 amount) external auth {
```
lemunozm, line 32: We should add a ticket to support a recoverTokens method with the previous interface here. Or do you want to add in this PR? Should just be:

```solidity
    /// @inheritdoc IRecoverable
    function recoverTokens(address token, address to, uint256 amount) external auth {
        SafeTransferLib.safeTransfer(token, to, amount);
    }
```

src/vaults/AsyncManager.sol
```diff
@@ -593,100 +501,116 @@ contract InvestmentManager is Auth, IInvestmentManager, IInvestmentManagerGatewa
...
+        view
+        override(BaseInvestmentManager, IERC165)
+        returns (bool)
+    {
+        return super.supportsInterface(interfaceId) || interfaceId == type(IVaultManager).interfaceId
+            || interfaceId == type(IDepositManager).interfaceId || interfaceId == type(IAsyncDepositManager).interfaceId
+            || interfaceId == type(IRedeemManager).interfaceId || interfaceId == type(IAsyncRedeemManager).interfaceId
+            || interfaceId == type(IAsyncManager).interfaceId
+            || interfaceId == type(IInvestmentManagerGatewayHandler).interfaceId
+            || interfaceId == type(IAsyncRedeemManager).interfaceId || interfaceId == type(IMessageHandler).interfaceId;
```
lemunozm: `IAsyncRedeemManager` is repeated in the list. 

To avoid weird formatting, I would recommend here to use: `// forgefmt: disable-next-item` and use all in the same line (ref MessageLib.MESSAGE_LENGTHS)

src/vaults/AsyncVault.sol
```diff
@@ -0,0 +1,165 @@
...
+    }
+
+    /// @dev Strongly-typed accessor to the generic async redeem manager
+    function asyncManager() public view returns (IAsyncManager) {
+        return IAsyncManager(address(IAsyncRedeemVault(address(this)).asyncRedeemManager()));
+    }
+
+    /// @dev Preview functions for ERC-7540 vaults revert
+    function previewDeposit(uint256) external pure returns (uint256) {
+        revert();
```
lemunozm, line 149: NIT. Could we add an "unimplemented" message or so?

src/vaults/SyncInvestmentManager.sol
mustermeiszer: You can ignore the `Meta`. THe CP side will use the default accounts.

mustermeiszer: `Meta` is just optional.

src/vaults/BaseVaults.sol
```diff
@@ -299,12 +161,10 @@ contract ERC7540Vault is Auth, IERC7540Vault {
...
     /// @inheritdoc IERC165
-    function supportsInterface(bytes4 interfaceId) external pure override returns (bool) {
-        return interfaceId == type(IERC7540Deposit).interfaceId || interfaceId == type(IERC7540Redeem).interfaceId
-            || interfaceId == type(IERC7540Operator).interfaceId || interfaceId == type(IERC7540CancelDeposit).interfaceId
-            || interfaceId == type(IERC7540CancelRedeem).interfaceId || interfaceId == type(IERC7575).interfaceId
-            || interfaceId == type(IERC7741).interfaceId || interfaceId == type(IERC7714).interfaceId
-            || interfaceId == type(IERC165).interfaceId;
+    function supportsInterface(bytes4 interfaceId) public pure virtual override returns (bool) {
+        return interfaceId == type(IERC7540Operator).interfaceId || interfaceId == type(IERC7741).interfaceId
+            || interfaceId == type(IERC7714).interfaceId || interfaceId == type(IERC7575).interfaceId
```
lemunozm, line 166: `IERC7714` duplicated

src/vaults/BaseVaults.sol
```diff
@@ -408,51 +337,52 @@ contract ERC7540Vault is Auth, IERC7540Vault {
...
         revert();
     }
+}
 
-    // --- Event emitters ---
-    /// @inheritdoc IERC7540Vault
-    function onRedeemRequest(address controller, address owner, uint256 shares) external auth {
-        emit RedeemRequest(controller, owner, REQUEST_ID, msg.sender, shares);
-    }
+abstract contract AbstractSyncDepositVault is BaseVault {
```
lemunozm: NIT. Seems like maybe `BaseSyncDepositVault` is the correct name following the same style as others contracts (?)

src/vaults/AsyncManager.sol
```diff
@@ -14,79 +16,62 @@ import {MessageType, MessageLib} from "src/common/libraries/MessageLib.sol";
...
+import {IAsyncVault} from "src/vaults/interfaces/IERC7540.sol";
+import {PriceConversionLib} from "src/vaults/libraries/PriceConversionLib.sol";
+import {BaseInvestmentManager} from "src/vaults/BaseInvestmentManager.sol";
 
 /// @title  Investment Manager
 /// @notice This is the main contract vaults interact with for
 ///         both incoming and outgoing investment transactions.
-contract InvestmentManager is Auth, IInvestmentManager, IInvestmentManagerGatewayHandler {
-    using MessageLib for *;
+contract AsyncManager is BaseInvestmentManager, IAsyncManager {
```
lemunozm: Why not call it like in the diagram `AsyncInvestmentManager` and `SyncInvestmentManager` ? Feel more readable to know what it does.

src/vaults/PoolManager.sol
```diff
@@ -354,9 +356,11 @@ contract PoolManager is Auth, IPoolManager, IUpdateContract, IPoolManagerGateway
         TrancheDetails storage tranche_ = pools[poolId].tranches[trancheId];
         require(tranche_.token != address(0), "PoolManager/tranche-does-not-exist");
 
-        address manager = IBaseVault(vault).manager();
         AssetIdKey memory assetIdKey = _idToAsset[assetId];
-        IVaultManager(manager).addVault(poolId, trancheId, vault, assetIdKey.asset, assetId);
+
+        IBaseInvestmentManager manager = IBaseVault(vault).manager();
+        IVaultManager(address(manager)).addVault(poolId, trancheId, vault, assetIdKey.asset, assetId);
```
lemunozm, line 362: If you expect `IBaseInvestmentManager` to always be convertible to `IVaultManager`, why not add this restriction as an inheritance in the interfaces?

I think `IAsyncManager` should implement `IBaseInvestmentManager`

src/vaults/SyncManager.sol
```diff
@@ -0,0 +1,240 @@
...
+        view
+        override(BaseInvestmentManager, IERC165)
+        returns (bool)
+    {
+        return super.supportsInterface(interfaceId) || interfaceId == type(IVaultManager).interfaceId
+            || interfaceId == type(IDepositManager).interfaceId || interfaceId == type(ISyncDepositManager).interfaceId
+            || interfaceId == type(ISyncManager).interfaceId;
+    }
+
+    /// @dev Retrieve the latest price for the share class token
```
hieronx: nit for followup PR: Should be moved to `Internal methods` section

src/vaults/PoolManager.sol
```diff
@@ -440,6 +446,26 @@ contract PoolManager is Auth, IPoolManager, IUpdateContract, IPoolManagerGateway
         require(assetId != 0, "PoolManager/unknown-asset");
     }
 
+    /// @dev Sets up permissions for the base vault manager and potentially a secondar manager (in case of partially
```
hieronx: nit for followup PR: 
```suggestion
    /// @dev Sets up permissions for the base vault manager and potentially a secondary manager (in case of partially
```

src/vaults/interfaces/investments/IAsyncRedeemManager.sol
```diff
@@ -0,0 +1,65 @@
+// SPDX-License-Identifier: BUSL-1.1
+pragma solidity 0.8.28;
+
+import {IRedeemManager} from "src/vaults/interfaces/investments/IRedeemManager.sol";
+import {IVaultManager} from "src/vaults/interfaces/IVaultManager.sol";
+
+/// @dev Vault requests and deposit/redeem bookkeeping per user
+struct AsyncRedeemState {
```
hieronx: I believe this struct is unused now?

hieronx (dismissed):
🚀

src/vaults/factories/SyncDepositVaultFactory.sol
```diff
@@ -0,0 +1,47 @@
...
+        address tranche,
+        address, /* escrow */
+        address[] calldata wards_
+    ) public auth returns (address) {
+        SyncDepositVault vault =
+            new SyncDepositVault(poolId, trancheId, asset, tokenId, tranche, root, syncManager, asyncManager);
+
+        vault.rely(root);
+        vault.rely(syncManager);
+        vault.rely(asyncManager);
```
lemunozm: Is relying on `asyncManager` needed here? This is the sync vault

src/vaults/interfaces/IPoolManager.sol
```diff
@@ -3,6 +3,8 @@ pragma solidity 0.8.28;
 
 import {IRecoverable} from "src/common/interfaces/IRoot.sol";
 
+import {IBaseInvestmentManager} from "src/vaults/interfaces/investments/IBaseInvestmentManager.sol";
```
lemunozm: NIT. I think this can be removed

lemunozm (commented):
Such a great PR! I can not imagine how you deep you've need to go to accomplish this!

I left some comments/questions/nits/minor CRs but nothing severe! Amazing job!

Just a concern, I understand we can handle vaults independently of if they are current ERC7540Vaults or new ones, right? We should ensure retrocompatibility with the current living vaults.

BTW, regarding the diagram, in case you do not know about it, the UML standard already defines your purple lines as white-filled arrows towards the parent.

script/VaultsDeployer.s.sol
hieronx: Good catch

src/vaults/AsyncVault.sol
hieronx: I would keep it as is, it's been that way with the current vaults, no reason to change it.

src/vaults/AsyncManager.sol
hieronx: My subjective opinion is that `AsyncManager` vs `SyncManager` is more easily separable because they are shorter.

src/vaults/BaseInvestmentManager.sol
hieronx: Agreed, for any contract that is not deterministically deployed, that will be fine!

src/vaults/AsyncManager.sol
lemunozm: But can not confuse a new reader that reads: PoolManager, AsyncManager, SyncManager, VaultManager, BalanceSheetManager (?)

It is not clear that those managers are focused on investment and have nothing to do with PoolManager or BalanceSheetManager

src/vaults/AsyncVault.sol
lemunozm: Ok, my only reason was debugging-friendly, but not a strong opinion!

src/vaults/AsyncManager.sol
hieronx: We could rename to `AsyncRequests` and `SyncRequests` to solve that. Similar to `Holdings`

I feel the long names are quite unusable for the expert reader of the code (us + auditors), who already knows what the contracts do and that they are related to investments, and needs to clearly separate between the async vs sync one.

src/vaults/BaseInvestmentManager.sol
lemunozm: I created an issue.

Could you add an example of some contracts that forcibly need to receive a `deployer`instead of just use `msg.sender`?

hieronx: Anything we plan to deploy deterministically. In vaults, `tokenFactory`, `escrow`, `routerEscrow`, `root`, `restrictionManager`, `restrictedRedemptions`.

hieronx: Actually, once we implement https://github.com/centrifuge/protocol-v3/issues/132, we should maybe just deploy everything deterministically. No strong reason not to.

lemunozm: Then, IIUC, we should not use `msg.sender` inside the contructor and better pass the `deployer`, right?

foundry.toml
wischli: Eagle eye! Find & replace has failed

src/vaults/BaseInvestmentManager.sol
hieronx: Yeah exactly

src/vaults/BaseInvestmentManager.sol
wischli: Separate PR preferred as this needs to be applied to other contracts as well

src/vaults/AsyncManager.sol
wischli: Then I will rename to `AsyncRequests` and `SyncRequests` in this PR to keep follow up PRs less noisy.

src/vaults/PoolManager.sol
wischli: > I think `IAsyncManager` should implement `IBaseInvestmentManager`

That's the case, see above diagram - it comes from `IBaseInvestmentManager ⊂ I{Deposit, Redeem}Manager ⊂ IAsyncRedeemManager`.

wischli: > If you expect `IBaseInvestmentManager` to always be convertible to `IVaultManager`, why not add this restriction as an inheritance in the interfaces?

Unfortunately, in practice we don't have a bijection, i.e. `IBaseInvestmentManager <-- IVaultManager` only holds true because only `I{Sync, Async}Manager` impl `IVaultManager`. However, we cannot restrict `BaseVault.manager` to either of that. However, we know that due to the abstract contract, the eventual vault contracts use either the sync or async manager.

src/vaults/BaseVaults.sol
wischli: Actually not, its ERC77**41** and ERC77**14** 😄

mustermeiszer (commented):
One thing to note: We do not expose the `IERC4626` compatibility anywhere. IMO we should to that? Right, @hieronx

src/vaults/factories/SyncDepositVaultFactory.sol
wischli: It's the sync deposit async redeem vault. Therefore, the `asyncManager` needs to be relied because each `SyncDepositAyncRedeemVault` is actually added twice, once for each the investment manager

src/vaults/AsyncVault.sol
```diff
@@ -0,0 +1,165 @@
...
+    /// @inheritdoc IERC7575
+    function maxDeposit(address controller) public view returns (uint256 maxAssets) {
+        maxAssets = asyncManager().maxDeposit(address(this), controller);
+    }
+
+    /// @inheritdoc IERC7540Deposit
+    function deposit(uint256 assets, address receiver, address controller) public returns (uint256 shares) {
+        _validateController(controller);
+        shares = asyncManager().deposit(address(this), assets, receiver, controller);
+        emit Deposit(receiver, controller, assets, shares);
```
mustermeiszer: CR 

Ordering is different compared to the one on main atm - Shouldnt it be `contoller`, `receiver`?

src/vaults/AsyncVault.sol
```diff
@@ -0,0 +1,165 @@
...
+    /// @inheritdoc IERC7575
+    function maxMint(address controller) public view returns (uint256 maxShares) {
+        maxShares = asyncManager().maxMint(address(this), controller);
+    }
+
+    /// @inheritdoc IERC7540Deposit
+    function mint(uint256 shares, address receiver, address controller) public returns (uint256 assets) {
+        _validateController(controller);
+        assets = asyncManager().mint(address(this), shares, receiver, controller);
+        emit Deposit(receiver, controller, assets, shares);
```
mustermeiszer: CR

Same here, see above comment

wischli:
> One thing to note: We do not expose the `IERC4626` compatibility anywhere. IMO we should to that? Right, @hieronx

Good catch. Should be added to `IBaseVault` as another implemented interface as the base vault indeed already implements it.

wischli:
> One thing to note: We do not expose the `IERC4626` compatibility anywhere. IMO we should to that? Right, @hieronx

Actually, this causes a lot of conflicts since `IERC7575` seems to be the superior one here which also implements all calls of `IERC4626` and BaseVault implements `IERC7575`.

src/vaults/AsyncVault.sol
wischli: IMO, you are right. This was a leftover from the unwanted change in the interface: https://github.com/centrifuge/protocol-v3/pull/109#discussion_r1978921259

Good catch!

hieronx: Yeah true it should be `controller`, then `receiver`, in the async impl

hieronx:
> Actually, this causes a lot of conflicts since IERC7575 seems to be the superior one here which also implements all calls of IERC4626 and BaseVault implements IERC7575.

Yeah ERC7575 interface is sufficient.

README.md
```diff
@@ -28,8 +28,8 @@ Investors can invest in multiple tranches for each RWA pool. Each of these tranc
...
 - [**Tranche Token**](https://github.com/centrifuge/protocol-v3/blob/main/src/vaults/token/Tranche.sol): An [ERC-20](https://ethereum.org/en/developers/docs/standards/tokens/erc-20/) token for the tranche, linked to a [`RestrictionManager`](https://github.com/centrifuge/protocol-v3/blob/main/src/vaults/token/RestrictionManager.sol) that manages transfer restrictions. Prices for tranche tokens are computed on Centrifuge.
 
-The deployment of these tranches and the management of investments is controlled by the underlying InvestmentManager, PoolManager, Gateway and Adapters.
-- [**Investment Manager**](https://github.com/centrifuge/protocol-v3/blob/main/src/vaults/InvestmentManager.sol): The core business logic contract that handles pool creation, tranche deployment, managing investments and sending tokens to the [`Escrow`](https://github.com/centrifuge/protocol-v3/blob/main/src/vaults/Escrow.sol), and more.
+The deployment of these tranches and the management of investments is controlled by the underlying AsyncManager, PoolManager, Gateway and Adapters.
```
hieronx:
```suggestion
The deployment of these tranches and the management of investments is controlled by the underlying AsyncRequests, PoolManager, Gateway and Adapters.
```

README.md
```diff
@@ -28,8 +28,8 @@ Investors can invest in multiple tranches for each RWA pool. Each of these tranc
...
+- [**Investment Manager**](https://github.com/centrifuge/protocol-v3/blob/main/src/vaults/AsyncManager.sol): The core business logic contract that handles pool creation, tranche deployment, managing investments and sending tokens to the [`Escrow`](https://github.com/centrifuge/protocol-v3/blob/main/src/vaults/Escrow.sol), and more.
```
hieronx:
```suggestion
- [**Async Requests**](https://github.com/centrifuge/protocol-v3/blob/main/src/vaults/AsyncRequests.sol): The core business logic contract that handles pool creation, tranche deployment, managing investments and sending tokens to the [`Escrow`](https://github.com/centrifuge/protocol-v3/blob/main/src/vaults/Escrow.sol), and more.
```

src/vaults/PoolManager.sol
wischli: As far as the PoolManager is concerned, the `IBaseVault(vaultAddr).manager()` should return `IVaultManager` and not `IBaseInvestmentManager` but due to the above, the base vault cannot do that. Since we don't know the type of vault at this point, we also shouldn't check any higher level vault interface.

src/vaults/factories/SyncDepositVaultFactory.sol
wischli: You were right! Removed in f7e40561d9f3d2b9a16b8879be0bd643f61632cf. I think this was an artifact of an earlier implementation

src/vaults/BaseVaults.sol
lemunozm: OMG! That's correct 😆

src/vaults/PoolManager.sol
lemunozm: 🤔, I don't fully understand. AFAIK, All `IVaultManager` implementations end up implementing `BaseInvestmentManager`. It means that we can force that interface to require `IBaseInvestmentManager`, isn't?

What I'm missing?

wischli: No, you are right. We could change that but IMO it just adds more obfuscation without providing "type safety" to the cast `IVaultManager(address(IBaseInvestManager(mg)))` because when `IVaultManager` impls `IBaseInvestmentManager`, it does not add a guarantee that every base investment manager is indeed a vault manager.

hieronx (dismissed):

lemunozm (approved):
Let's merge this!

PR #112, opened 2025-02-27
Remove domain from TransferTrancheTokens

lemunozm:
thread: https://kflabs.slack.com/archives/C07PG2EUR9C/p1740654438583919


lemunozm:
NOTE: the high increment in gas is because the latest PR from William decreased them a lot. Maybe different versions of foundry/machine (?)

hieronx (approved):

PR #117, opened 2025-02-27
Ref: common message lib for CV and CP

### LINKED ISSUE #104: Fix `handleLockedTokens` indices
https://github.com/Recon-Fuzz/centrifuge-review/issues/16#issuecomment-2684591593

lemunozm:
Fixes https://github.com/centrifuge/protocol-v3/issues/104

Moves all serialization content from CP and CV to one place `MessageLib.sol` and redirect everything to it. There is no longer explicit calls to `abi.encodePacked`, not even in tests.

NOTE: Adapted some messages with the latest changes from CP development (please double-check):
- `hook` as `bytes32`
- addresses in `ScheduleUpgrade`, `CancelUpgrade` and `RecoverTokens` as `bytes32`
- `*MessageRecovery::adapter` as bytes32
- `shareClassId` param to `AllowAsset` & `DisallowAsset` messages
- `FulfillCancel*` message no longer needs two amounts.

Modifying Batch messages will go in a follow up PR


test/pools/integration/Cases.t.sol
```diff
@@ -94,22 +94,20 @@ contract TestConfiguration is TestCommon {
...
+        MessageLib.NotifyPool memory m0 = MessageLib.deserializeNotifyPool(cv.lastMessages(0));
+        assertEq(m0.poolId, poolId.raw());
+
+        MessageLib.NotifyShareClass memory m1 = MessageLib.deserializeNotifyShareClass(cv.lastMessages(1));
+        assertEq(m1.poolId, poolId.raw());
+        assertEq(m1.scId, scId.raw());
+        assertEq(m1.name, SC_NAME);
+        assertEq(m1.symbol, SC_SYMBOL.toBytes32());
+        assertEq(m1.decimals, 18);
+        assertEq(m1.hook, SC_HOOK);
```
lemunozm, line 107: I think the new way of checking the message content in tests is a big win ^^.

If we have some mismatch now, we do not have a horrible list of bytes, we know which parameter of the message is wrong.

src/pools/Gateway.sol
```diff
@@ -79,20 +78,19 @@ contract Gateway is Auth, IGateway, IMessageHandler {
...
-                investedAmount
-            )
+            MessageLib.FulfilledDepositRequest({
+                poolId: poolId.raw(),
+                scId: scId.raw(),
+                investor: investor,
+                assetId: assetId.raw(),
+                shareAmount: shareAmount,
+                assetAmount: assetAmount
+            }).serialize()
```
lemunozm, line 92: I know I'm being a bit more verbose, but I would prefer to initialize the structs by fields as shown to ensure I'm not flipping values. i.e. `shareAmount` and `assetAmount`

wischli: I definitely prefer the improved readability!

test/pools/integration/Cases.t.sol
wischli: That's awesome!

src/vaults/gateway/Gateway.sol
```diff
@@ -239,9 +241,9 @@ contract Gateway is Auth, IGateway, IRecoverable {
...
         } else if (id == 8) {
             manager = address(gasService);
-        } else if (id >= 9 && id <= 19) {
+        } else if (id >= 9 && id <= 18) {
             manager = poolManager;
-        } else if (id >= 20 && id <= 28) {
+        } else if (id >= 19 && id <= 27) {
             manager = investmentManager;
         } else {
             // Dynamic path for other managers, to be able to easily
```
lemunozm, line 249: This code in the gateway feels very scary. I think it's super easy to add/remove a message, and have unknown failures without any clue. I propose introducing a concept of "categories" that allows us to have tests to ensure the above conditions. See this commit: [b4db33b](https://github.com/centrifuge/protocol-v3/pull/117/commits/b4db33b3d4b84dcf6c5c843724875dad1493dd93)

The gateway then will if/elseif/else against the categories

src/misc/libraries/CastLib.sol
```diff
@@ -32,18 +33,6 @@ library CastLib {
         return string(bytesArray);
     }
 
-    function stringToBytes128(string memory str) internal pure returns (bytes memory) {
```
lemunozm, line 35: NOTE. This was introduced by me in a previous PR, but it's no longer needed.

src/vaults/PoolManager.sol
```diff
@@ -89,8 +92,11 @@ contract PoolManager is Auth, IPoolManager {
         ITranche tranche = ITranche(getTranche(poolId, trancheId));
         require(address(tranche) != address(0), "PoolManager/unknown-token");
         tranche.burn(msg.sender, amount);
+
+        // TODO: use destinationId to determine the destination chainId where to send the tranches
```
lemunozm: This TODO means that the adapter/gateway requires a chainId parameter, to know where to send the shares. I propose to do this in a follow-up PR

src/vaults/gateway/Gateway.sol
lemunozm: Tell me if you agree with this @hieronx, and I will proceed to adapt that check in the gateway.

src/vaults/PoolManager.sol
mustermeiszer: Sounds good! I would love to have all calls to gateway already be based on `chainId` as parameter and then gatway ignores it for now.

src/vaults/gateway/Gateway.sol
mustermeiszer: No strong opinion from my side. But makes it easier to test IMO.

mustermeiszer (commented):
Love the struct based approach! Much better type insurance. Not approving as I want to wait for @hieronx feedback regarding the "type" check. As long as we can extend the messages with dynamic handlers I like the new approach.

src/common/libraries/MessageLib.sol
```diff
@@ -2,30 +2,734 @@
...
+    //    FulfilledDepositRequest
+    //---------------------------------------
+
+    struct FulfilledDepositRequest {
+        uint64 poolId;
+        bytes16 scId;
+        bytes32 investor;
+        uint128 assetId;
+        uint128 shareAmount;
+        uint128 assetAmount;
```
wischli: Ultra nit: Could we reorder to follow the (de-)serialization order?
```suggestion
        uint128 assetAmount;
        uint128 shareAmount;
```

src/common/libraries/MessageLib.sol
```diff
@@ -2,30 +2,734 @@
...
+    //    FulfilledRedeemRequest
+    //---------------------------------------
+
+    struct FulfilledRedeemRequest {
+        uint64 poolId;
+        bytes16 scId;
+        bytes32 investor;
+        uint128 assetId;
+        uint128 shareAmount;
+        uint128 assetAmount;
```
wischli: Ultra nit: Could we reorder to follow the (de-)serialization order?

```suggestion
        uint128 assetAmount;
        uint128 shareAmount;
```

src/common/libraries/MessageLib.sol
```diff
@@ -2,30 +2,734 @@
...
+    //---------------------------------------
+    //    FulfilledCancelDepositRequest
+    //---------------------------------------
+
+    struct FulfilledCancelDepositRequest {
+        uint64 poolId;
+        bytes16 scId;
+        bytes32 investor;
+        uint128 assetId;
+        uint128 cancelledAmount;
```
wischli, line 652: Just wanted to raise that we hereby drop one of the two amount fields of the v2 message. This makes sense to me, which is why I flagged it in January ([ref Slack thread](https://kflabs.slack.com/archives/C07PG2EUR9C/p1736933248761209))

src/pools/SingleShareClass.sol
```diff
@@ -60,10 +60,6 @@ function previewShareClassId(PoolId poolId) pure returns (ShareClassId) {
     return ShareClassId.wrap(bytes16(uint128(PoolId.unwrap(poolId))));
 }
 
-function encodeMetadata(string memory name, string memory symbol) pure returns (bytes memory metadata) {
-    return abi.encodePacked(bytes(CastLib.stringToBytes128(name)), CastLib.toBytes32(symbol));
-}
-
```
wischli, line 66: Thanks for cleaning up my leftover!

src/vaults/PoolManager.sol
```diff
@@ -100,48 +106,37 @@ contract PoolManager is Auth, IPoolManager {
     // --- Incoming message handling ---
     /// @inheritdoc IPoolManager
     function handle(bytes calldata message) external auth {
-        MessagesLib.Call call = MessagesLib.messageType(message);
+        MessageType kind = MessageLib.messageType(message);
 
-        if (call == MessagesLib.Call.AddAsset) {
+        if (kind == MessageType.RegisterAsset) {
+            // TODO: This must be removed
```
wischli, line 112: Yes, I will do this as part of https://github.com/centrifuge/protocol-v3/pull/98

src/vaults/gateway/Gateway.sol
wischli: > No strong opinion from my side. But makes it easier to test IMO.

+1 From my side

test/vaults/integration/Gateway.t.sol
```diff
@@ -73,17 +77,17 @@ contract GatewayTest is BaseTest {
         uint64 poolId = 999;
         uint128 assetId = defaultAssetId + 1;
         MockERC20 newAsset = deployMockERC20("newAsset", "NEW", 18);
-        bytes memory _addPool = abi.encodePacked(uint8(MessagesLib.Call.AddPool), poolId);
-
-        bytes memory _addAsset = abi.encodePacked(uint8(MessagesLib.Call.AddAsset), assetId, address(newAsset));
-        bytes memory _allowAsset = abi.encodePacked(uint8(MessagesLib.Call.AllowAsset), poolId, assetId);
+        bytes memory _addPool = MessageLib.NotifyPool(poolId).serialize();
+        // TODO: remove next line when registerAsset feature is supported
```
wischli, line 81: I'm on that!

wischli (commented):
What an incredible improvement! I personally love the `category` checks. The idempotency checks are also extremely helpful. Similar to @mustermeiszer, I am only soft-approving in order to let @hieronx have the final word.

src/vaults/gateway/Gateway.sol
hieronx: Makes sense to me! Actually wonder if we shouldn't include these categories in the message encoding such that it becomes
```
[ category id ] [ message id ] [ message content ]
```

instead of 
```
[ message id ] [ message content ]
```

Would add only `uint8` to every message, but would make it even cleaner and easier to test + would make it easier to add new messages within categories.

lemunozm: Interesting!

Indeed, we could use the first 3 bytes (8 categories) to specify the category, and the remaining 5 bytes (32 messages per category) to define the message, and save that extra byte.

src/common/libraries/MessageLib.sol
lemunozm: Sure! Actually the ser/de was already reordered. It was only the struct. Done!

src/vaults/PoolManager.sol
lemunozm: Done here [b20f984](https://github.com/centrifuge/protocol-v3/pull/117/commits/b20f98433a481015d3bdee88ce319cecbb396054)

src/vaults/gateway/Gateway.sol
lemunozm: I need to think a bit more about its implications, do you mind @hieronx if I do this in a follow-up PR, once the batch refactor is added?

mustermeiszer (approved):
Soft approval from my side. So no blockers.

wischli (approved):

src/vaults/PoolManager.sol
hieronx: Comment can be removed then, right?

src/vaults/gateway/Gateway.sol
hieronx: Sure!

hieronx (approved):

PR #120, opened 2025-03-03
Batch simplification based on known message sizes

lemunozm:
This simplification removes the Batch message, and assumes every sent/received message is itself a set of messages.

Properties:
- Easier serialization
- Easier deserialization
- Autobatching without user interaction or internal message modifications 🔝 

### Changes:
- Messages have a length associated
- UpdateRestriction submessages handled by MessageLib to know the size of their variants
- Batch processing in the gateway with the new method.


src/common/libraries/MessageLib.sol
```diff
@@ -66,24 +112,42 @@ library MessageLib {
         return message.toUint8(0);
     }
 
+    function messageLength(bytes memory message) internal pure returns (uint16 length) {
+        uint8 kind = message.toUint8(0);
+        require(kind <= uint8(type(MessageType).max), UnknownMessageType());
+
+        length = uint16(uint8(bytes32(MESSAGE_LENGTHS)[31 - kind]));
```
lemunozm, line 111: Instead of having a large list of if/else if with each message type, we can encode their lengths in a bytes32 and read from it in `O(0)` from a constant storage. Also the bytecode is reduced.

In case a message exceeds 256 bytes per message (none by now), we can expand the encoding to `uint16`

test/common/unit/libraries/MessageLib.t.sol
```diff
@@ -49,15 +48,19 @@ contract TestMessageLibIdentities is Test {
...
-    function testInitializeMessageRecovery() public pure {
+    function testInitiateMessageRecovery() public pure {
         MessageLib.InitiateMessageRecovery memory a =
             MessageLib.InitiateMessageRecovery({hash: bytes32("hash"), adapter: bytes32("adapter")});
         MessageLib.InitiateMessageRecovery memory b = MessageLib.deserializeInitiateMessageRecovery(a.serialize());
 
         assertEq(a.hash, b.hash);
         assertEq(a.adapter, b.adapter);
+
+        assertEq(a.serialize().messageLength(), a.serialize().length);
```
lemunozm, line 63: Lengths are ensured that they always reflect the correct message size

lemunozm:
Support for dynamic UpdateRestriction messages: [080a497](https://github.com/centrifuge/protocol-v3/pull/120/commits/080a49796cd8d442f8fa520e24501accf9222434)

src/common/libraries/MessageLib.sol
```diff
@@ -58,6 +65,37 @@ library MessageLib {
 
     error UnknownMessageType();
 
+    /// @dev Encode all message lengths in this constant to avoid a large list of if/elseif checks
+    /// and reduce generated bytecode
+    // forgefmt: disable-next-item
+    uint256 constant MESSAGE_LENGTHS =
+        (33 << uint8(MessageType.MessageProof) * 8) +
+        (65 << uint8(MessageType.InitiateMessageRecovery) * 8) +
+        (65 << uint8(MessageType.DisputeMessageRecovery) * 8) +
```
lemunozm, line 74: Sadly, solidity does not support constant arrays. But we can encode all the message lengths of our protocol into a `uint256`. If some day we overpass 32 messages, we can just add another `MESSAGE_LENGTHS_2` constant

src/vaults/gateway/Gateway.sol
```diff
@@ -212,47 +199,34 @@ contract Gateway is Auth, IGateway, IRecoverable {
...
+                manager = address(root);
+            } else if (cat == MessageCategory.Gas) {
+                manager = address(gasService);
+            } else if (cat == MessageCategory.Pool) {
+                manager = poolManager;
+            } else if (cat == MessageCategory.Investment) {
+                manager = investmentManager;
+            } else {
+                revert("Gateway/unexpected-category");
+            }
```
lemunozm, line 216: This is the only usage of categories. I think it could be a bit overkill to encode an extra byte or handling bits in the first message byte just for this single part of the code. cc @hieronx.

Also, from CP perspective the usage of categories is not needed, so maybe put more work on this is not necessary. Maybe not now (not sure in the future)

hieronx: I disagree. The main value is that we can later add a new message for example in the gas category, without needing to migrate all CP + CV instances simultaneously. Since if we don't, and all the indices after this category are bumped by 1, then any messages in flight during the upgrade will fail.

If we don't do this, we realistically can only add messages in a new category at the end.

lemunozm: Can we not assume that a category has elements that are not contiguous?

i.e:
```
if (code >= 4 && code <= 6 || code == 29) {
    MessageCategory.Root;
}
```

lemunozm: The downsides I see with both current approaches:
- Using an extra byte: we use a byte that can be derived from the message type, so we are coding information in a message that is not really needed.
- Using `b11100000` as category and `b00011111` as messageType masks. Makes it more difficult to encode the `MESSAGE_LENGTHS` array.

If using the above `if (code >= 4 && code <= 6 || code == 29)`, works for now, I would prefer that. I do not expect crazy chaos in these indexes. If it happens, probably it's because we need a ProtocolV4, and we can reorganize the indexes again, because an entire deployment is required.

mustermeiszer (commented):
Looks nice! I am indifferent regarding the message categories - as long as we can extend later down the road without a V4 upgrade.

src/vaults/gateway/Gateway.sol
lemunozm: Will leave as it is right now: https://kflabs.slack.com/archives/C07PG2EUR9C/p1741080699503619

Until we see more properly how everything works together

src/vaults/token/RestrictionManager.sol
```diff
@@ -87,13 +90,21 @@ contract RestrictionManager is Auth, IRestrictionManager, IHook {
...
     // --- Incoming message handling ---
     /// @inheritdoc IHook
-    function updateRestriction(address token, bytes memory update) external auth {
-        RestrictionUpdate updateId = RestrictionUpdate(update.toUint8(0));
-
-        if (updateId == RestrictionUpdate.UpdateMember) updateMember(token, update.toAddress(1), update.toUint64(33));
-        else if (updateId == RestrictionUpdate.Freeze) freeze(token, update.toAddress(1));
-        else if (updateId == RestrictionUpdate.Unfreeze) unfreeze(token, update.toAddress(1));
-        else revert("RestrictionManager/invalid-update");
+    function updateRestriction(address token, bytes memory payload) external auth {
```
hieronx, line 93: One important note: we are not going to be migrating existing tranche tokens, including their hooks. Thus, the update restriction messages need to stay compatible with the current implementation. Or we do need to migrate the hooks, but that would be nontrivial.

hieronx (approved):

src/common/libraries/MessageLib.sol
```diff
@@ -66,24 +104,39 @@ library MessageLib {
...
     }
 
+    function messageLength(bytes memory message) internal pure returns (uint16 length) {
+        uint8 kind = message.toUint8(0);
+        require(kind <= uint8(type(MessageType).max), UnknownMessageType());
+
+        length = uint16(uint8(bytes32(MESSAGE_LENGTHS)[31 - kind]));
+
+        if (kind == uint8(MessageType.UpdateRestriction)) {
+            length += message.toUint16(length - 2); //payloadLength
```
wischli, line 114: It doesn't feel correct to me to increment the length of `UpdateRestriction` again by amount of bytes minus two of the serialized message. Shouldn't it be `=` instead of `+=`?

src/common/libraries/MessageLib.sol
wischli: Just a clarification question: Are you casting to `uint16` solely for future extensions such that we need to update less contracts then?

wischli (commented):
Two quick questions before I resume

src/common/libraries/MessageLib.sol
lemunozm: The UpdateRestriction length only contains the length of the "header" of those messages. So in that line we're reading the length of the submessages that is encoded in the own message (or header) of UpdateRestriction.

The total real lenght is the header lenght + submessage lenght

src/vaults/token/RestrictionManager.sol
lemunozm: Mmmm... then we can not add the submessage length to those messages, and the indexes should be compatible 🤔. Will think about, maybe we can add some unique condition for those somehow.

src/common/libraries/MessageLib.sol
lemunozm: Exactly! By now all message length keep in uint8 but could not be the case in the future, we support until `uint16` in the API

PR #129, opened 2025-03-04
Allow AM to set salt per share class

### LINKED ISSUE #121: Allow AM to set salt per share class
Ref: https://kflabs.slack.com/archives/C07PG2EUR9C/p1741004752979359

hieronx:
Fixes https://github.com/centrifuge/protocol-v3/issues/121


src/pools/SingleShareClass.sol
```diff
@@ -74,6 +75,7 @@ contract SingleShareClass is Auth, ISingleShareClass {
     /// Storage
     uint32 internal transient _epochIncrement;
     IPoolRegistry public poolRegistry;
+    mapping(bytes32 salt => bool) public salts;
```
lemunozm, line 78: I think it should it be per pool. At least we should block other FM to take the salt for them and not allowing to use it for the intended FM. i.e:
- I deploy using `salt = 5`
- Malicious FM uses `salt = 5`
- I can not update from CP using `salt = 5`

src/pools/SingleShareClass.sol
```diff
@@ -660,14 +662,19 @@ contract SingleShareClass is Auth, ISingleShareClass {
...
 
         uint256 sLen = bytes(symbol).length;
         require(sLen > 0 && sLen <= 32, InvalidMetadataSymbol());
 
-        metadata[shareClassId_] = ShareClassMetadata(name, symbol);
+        // Either the salt has not changed, or the salt was never used before by any share class token
+        require(salt == metadata[shareClassId_].salt || !salts[salt], AlreadyUsedSalt());
+        salts[salt] = true;
+
+        metadata[shareClassId_] = ShareClassMetadata(name, symbol, salt);
```
lemunozm, line 677: Should we protect against using `salt == 0`? If not, you can not reuse `salt = 0` (as null) for other share classes because it will be `true` with the first null entry. Although not sure if nullability is a feature here.

src/pools/SingleShareClass.sol
hieronx: It needs to be global since otherwise there will be collisions of tranche token deployments on the vaults side

hieronx: In your example, the second bullet point would fail because `salt = 5` is already used

src/pools/SingleShareClass.sol
hieronx: Yeah agreed it makes sense to prevent this, will add a require statement

wischli (approved):
Thanks for tackling this so quickly!

src/pools/SingleShareClass.sol
lemunozm: Correct! I did not see the whole flow in the correct order.

PR #130, opened 2025-03-04
Fix pools ITs, add PoolRouter and remove allow/disallow

lemunozm:
### Change list
- Fix current integration tests from pools.
- Remove allow/disallow asset
- Minor accounting check
- Minor shareClassManager check
- Foundry CI to nightly
- PoolRouter


lemunozm:
@mustermeiszer Would you add a PR on top of this one, right? Why the change in the name of the PR? I'll assume it's a typo

src/pools/SingleShareClass.sol
```diff
@@ -342,6 +342,8 @@ contract SingleShareClass is Auth, ISingleShareClass {
                 continue;
             }
 
+            require(epochAmounts_.redeemSharesRevoked <= totalIssuance_, RevokeMoreThanIssued());
```
lemunozm, line 347: @wischli This is a NIT as output from CP-IT. And I can remove it if you think it's overkill. 

Isolated, from CP perspective, it can process a redeemRequest of a higher amount than the deposited amount. It means that when you revoke, you will revoke more than what was issued. This is not an issue itself, because CV protect against that, tracking correct amounts. But it's easy from a CP integration test perspective to see this error as an underflow, which is super unhelpful.

test/pools/integration/Cases.t.sol
```diff
@@ -174,7 +172,7 @@ contract TestInvestments is TestConfiguration {
         assertEq(m0.scId, scId.raw());
         assertEq(m0.investor, INVESTOR);
         assertEq(m0.assetId, USDC_C2.raw());
-        assertEq(m0.assetAmount, PERCENT_80.mulUint128(INVESTOR_AMOUNT));
+        assertEq(m0.assetAmount, PERCENT_20.mulUint128(PERCENT_20.mulUint128(INVESTOR_AMOUNT)));
```
wischli: Nit: Logically, the correct amount would be 

```solidity
        assertEq(
            m0.assetAmount,
            NAV_PER_SHARE.mulUint128(
                uint128(valuation.getQuote(PERCENT_20.mulUint128(SHARE_AMOUNT), USD.addr(), USDC_C2.addr()))
            ),
        );
```

Because the user redeems `SHARE_AMOUNT` which is approved with 20% and multiplied with `NAV_PER_SHARE`. Raising this because if any of the amounts are changed, this will break without providing information on why the current calculation was chosen.

src/pools/SingleShareClass.sol
wischli: Fine for me to keep.

lemunozm:
I've needed to do a little refactor to make the transient `unlockedPoolId` to live in `PoolManager` instead of `PoolRooter`: [f4a7ebe](https://github.com/centrifuge/protocol-v3/pull/130/commits/f4a7ebe505b2eac502535ae8206dd1ada0e4df16)

That way the `PoolManager` is conscious about when things are locked and it can, for example, send the batch there. Also, the own contract invariants fulfill in the own contract.

NOTE. We need to have both `unlockPool` and `auth` modifiers in the `PoolManager`. `unlockPool` is not enough to avoid reentrancy issues.

src/pools/SingleShareClass.sol
wischli: Nit: We are not covering that line with a test. I can do it as part of #135 if you want to

src/pools/PoolManager.sol
```diff
@@ -86,59 +74,64 @@ contract PoolManager is Auth, Multicall, IPoolManager, IPoolManagerHandler {
...
+        returns (PoolId poolId)
+    {
         // TODO: add fees
-        return poolRegistry.registerPool(msg.sender, currency, shareClassManager);
+        return poolRegistry.registerPool(admin, currency, shareClassManager);
     }
 
     /// @inheritdoc IPoolManager
-    function claimDeposit(PoolId poolId, ShareClassId scId, AssetId assetId, bytes32 investor) external protected {
+    function claimDeposit(PoolId poolId, ShareClassId scId, AssetId assetId, bytes32 investor) external {
```
wischli, line 106: Just to be sure: Let's assume the `msg.sender` performs a multicall with claims for different pools on different CVs. Do we ensure that the corresponding `Fulfilled*` messages will not be put into the same batch?

src/pools/PoolManager.sol
```diff
@@ -148,38 +141,38 @@ contract PoolManager is Auth, Multicall, IPoolManager, IPoolManagerHandler {
...
-    function allowPoolAdmin(address account, bool allow) external poolUnlocked protected {
+    /// @inheritdoc IPoolManager
+    function allowPoolAdmin(address account, bool allow) external auth poolUnlocked {
         poolRegistry.updateAdmin(unlockedPoolId, account, allow);
     }
 
-    /// @inheritdoc IPoolManagerAdminMethods
-    function allowInvestorAsset(ShareClassId scId, AssetId assetId, bool allow) external poolUnlocked protected {
+    /// @inheritdoc IPoolManager
+    function allowAsset(ShareClassId scId, AssetId assetId, bool /*allow*/ ) external view auth poolUnlocked {
```
wischli, line 155: Just to be sure: The `view` is temporary until the `TODO` is resolved right?

src/pools/PoolRegistry.sol
```diff
@@ -6,6 +6,7 @@ import {MathLib} from "src/misc/libraries/MathLib.sol";
 
 import {PoolId, newPoolId} from "src/pools/types/PoolId.sol";
 import {AssetId} from "src/pools/types/AssetId.sol";
+import {ShareClassId} from "src/pools/types/ShareClassId.sol";
```
wischli: Artifact? Unused in the contract

src/pools/PoolRouter.sol
```diff
@@ -0,0 +1,164 @@
...
+        poolManager.revokeShares(scId, payoutAssetId, navPerShare, valuation);
+    }
+
+    /// @inheritdoc IPoolRouter
+    function createHolding(ShareClassId scId, AssetId assetId, IERC7726 valuation, uint24 prefix) external protected {
+        poolManager.createHolding(scId, assetId, valuation, prefix);
+    }
+
+    /// @inheritdoc IPoolRouter
+    function increaseHolding(ShareClassId scId, AssetId assetId, IERC7726 valuation, uint128 amount) public protected {
```
wischli: Quick question: Why are `createHolding`, `increaseHolding` and `createAccount` public and the rest external?

src/pools/interfaces/IPoolRegistry.sol
```diff
@@ -3,21 +3,20 @@ pragma solidity 0.8.28;
 
 import {PoolId} from "src/pools/types/PoolId.sol";
 import {AssetId} from "src/pools/types/AssetId.sol";
+import {ShareClassId} from "src/pools/types/ShareClassId.sol";
```
wischli: Artifact? Seems redundant

src/pools/interfaces/IPoolRouter.sol
```diff
@@ -0,0 +1,87 @@
...
+import {AccountId} from "src/pools/types/AccountId.sol";
+import {PoolId} from "src/pools/types/PoolId.sol";
+import {IShareClassManager} from "src/pools/interfaces/IShareClassManager.sol";
+
+/// @notice Entry point to the system
+interface IPoolRouter {
+    /// @notice Main method to unlock the pool and call the rest of the admin methods
+    function execute(PoolId poolId, bytes[] calldata data) external payable;
+
+    /// @notice See counterpart in PoolManager contract
```
wischli, line 18: Comment: I wish we could just reference `IPoolManager` but that only works if `IPoolRouter` implemented it 😞

test/pools/integration/Cases.t.sol
wischli: @lemunozm Do you want to stick with the current solution or apply my suggestion? If sticking with yours, I'd love to have a comment added for future reference

wischli (commented):
A few questions from my side

test/pools/integration/Cases.t.sol
lemunozm: My rationale was that I wanted to avoid using the `valuation` as part of the checker of the result, due to the fact that if I have some error on it, it would not be reflected in the check. That's why I fixed to the static values.

Not sure if I follow your comment above. From my understanding:
- The first 20% is due when I `approveDeposit(20)`.
- The second 20% is due when I `approveRedeem(20)` from the resulting above, so `0.2 * 0.2 * initial_amount`

In which way the `NAV_PER_SHARE` affects here? Should not that be cancelled in the process due both conversions in different dirrections?

src/pools/PoolManager.sol
lemunozm: Yes, that gateway will dispatch that (incomming PR)

src/pools/PoolManager.sol
lemunozm: Yes. Forced by the warning

src/pools/PoolRouter.sol
lemunozm: Because they were used internally in PoolManager. They must be external here. Good eye!

src/pools/interfaces/IPoolRouter.sol
lemunozm: Me too! Not sure if there is some way to link docs as in Rustdoc

test/pools/integration/Cases.t.sol
lemunozm: Note. Will be more tests with more complex casuistics where navs and valuation changes in the process. Probably, there we would need a more detailed check taking into account the NAV used.

src/pools/SingleShareClass.sol
lemunozm: Correct! Thanks! 🙌🏻

test/pools/integration/Cases.t.sol
wischli: The approved deposit is irrelevant here because if `cv.requestRedeem(poolId, scId, USDC_C2, INVESTOR, SHARE_AMOUNT)` succeeds, it implies the user has that many shares. So the claimed fulfilment amount `m0.assetAmount` is the approved amount `20% * SHARE_AMOUNT` multiplied with the `NAV_PER_SHARE = d18(2, 1)` in denomination of `USDC_C2` (due to using `IdentityValuation`). 

Therefore, your solution works by coincidence:
* Your amount: `INVESTOR_AMOUNT * 0.2 * 0.2 = 100e6 * 0.2 * 0.2 = 4e6`
* Correct amount: `poolToUsdc(SHARE_AMOUNT* 2 * 0.2) = 10e18 * 2 * 0.2 * 1e6 / 1e18 = 4e18 * 1e6 / 1e18 = 4e6`

I know this is just a test and I don't want to block this PR. I just want to make our lives easier in the future, when we create similar tests. One might think, that the claimed redeem amount is dependent on the deposit process when in fact it is not as long as the user has sufficient shares

However, if you don't want to use the valuation nor hardcode the redenomination here, then I don't have a better proposal other than adding a comment explaining my above thought process.

lemunozm: Ok, seems my understanding was wrong! Thanks for the long explanation. So it worked just by coincidence 😲. Added your solution above ✅

wischli (approved):
Thanks for applying my comment! IMO, ready to be merged to fix main!

PR #131, opened 2025-03-05
Add PoolRouter as a wrapper of PoolManager

lemunozm:
Problem (described [here](https://kflabs.slack.com/archives/C07PG2EUR9C/p1741162462573139)) is that `pools/PoolManager` exceeds the required size for a contract.

- ❌ Optimizations do not reduce enough
- ❌ Extracting the gateway method leaves the contract in the limit, not leaving any extra space for a minimal addition.
- ✅ Adding a `PoolRouter` on top of `PoolManager`, moving all checks and multicall part to a new router seems to be the best option. For a max of 24.5K:
  - PoolRouter size: 11.4K
  - PoolManager size: 15.7K

  Leaving a big space for new additions.
  
I also think this separation makes sense from a logic perspective:
- PoolManager: compose the logic of the system.
- PoolLocker: the single entry point that enables pool interaction by multicall doing proper reentrancy checks.


src/pools/PoolRouter.sol
```diff
@@ -0,0 +1,207 @@
...
+import {IHoldings} from "src/pools/interfaces/IHoldings.sol";
+import {IPoolManager, IPoolManagerHandler, EscrowId, AccountType} from "src/pools/interfaces/IPoolManager.sol";
+import {IPoolRouter} from "src/pools/interfaces/IPoolRouter.sol";
+
+contract PoolRouter is Multicall, IPoolRouter {
+    /// @dev Represents the unlocked pool Id in the multicall
+    PoolId public transient unlockedPoolId;
+
+    IPoolManager public poolManager;
+    IPoolRegistry public poolRegistry;
```
hieronx:
```suggestion
    IPoolManager public immutable poolManager;
    IPoolRegistry public immutable poolRegistry;
```

wischli (commented):
LGTM except for the missing PoolRouter tests for all public calls. But I suppose these will be done as part of CP-CV integration tests?

lemunozm:
I added tests for all methods to check the `PoolLocker` checks, that is, the main addition `PoolRouter` adds over `PoolManager`.

The whole complexity should be tested by `Cases.t.sol`

src/pools/PoolRouter.sol
lemunozm: Correct! Done

src/pools/PoolRouter.sol
```diff
@@ -0,0 +1,207 @@
...
+
+        multicall(data);
+
+        poolManager.lockAccounting();
+        unlockedPoolId = PoolId.wrap(0);
+    }
+
+    /// @inheritdoc IPoolRouter
+    function createPool(AssetId currency, IShareClassManager shareClassManager) external returns (PoolId poolId) {
+        return poolManager.createPool(msg.sender, currency, shareClassManager);
```
hieronx, line 59: I think we want to allow the user to specify the admin. They might be using a hot wallet to create a pool but want to use another cold wallet as the initial admin

hieronx (approved):
One non-blocking comment, feel free to address in a later PR if you agree

src/pools/PoolRouter.sol
lemunozm: They can then `updateAdmin()` on its cold wallet and disable the hot one. But if you think it's anyways interesting I can add it

wischli:
> I added tests for all methods to check the `PoolLocker` checks, that is, the main addition `PoolRouter` adds over `PoolManager`.
> 
> The whole complexity should be tested by `Cases.t.sol`

I understand. However, the coverage flags basically all calls from the `PoolRouter` to the `PoolManager` as uncovered

src/pools/PoolRouter.sol
hieronx: Seems inefficient but true that could work!

lemunozm: I can still do it, if, after knowing this, you think it's beneficial

lemunozm:
Yeah... not sure how to test that correctly. Probably the ITs should cover automatically those when we have more in the future.

PR #134, opened 2025-03-05
feat: Add fuzz tests for types

### LINKED ISSUE #107: Add fuzz testing to `/types`
For all `/types/*.sol`, add fuzz unit tests. Currently, this includes

* AccountId
* AssetId
* ShareClassId
* PoolId

If easily possible, we should also include types to the coverage report

## Ref

> Add fuzz to `Types` so we [auditors] know they don't have gotchas (check my [chainId comment](https://github.com/Recon-Fuzz/centrifuge-review/pull/11#discussion_r1969910912) and decide if ack or fix)
>
> Fuzz + FV | These are perfect for testing because the testing is easy and gives a ton of certainties
IMO you should get these done now so you know you're building on solid code
The one piece that should cause issues is chainID which seems smaller than the programmed int64 in geth and reth

https://github.com/Recon-Fuzz/centrifuge-review/pull/11

### Chain Id

There are networks not listed as testnets which exceed `type(uint32).max`, e.g. Molereum Network with chain id 6022140761023. Scroll down here: https://chainid.network/

onnovisser:
Add fuzz tests for pools types. Not sure if they might be made more elaborate somehow.

Closes: #107


test/pools/unit/ShareClassId.t.sol
```diff
@@ -0,0 +1,18 @@
...
+
+import "forge-std/Test.sol";
+
+import {ShareClassId} from "src/pools/types/ShareClassId.sol";
+
+contract ShareClassIdTest is Test {
+    function testShareClassId(bytes16 id) public pure {
+        vm.assume(id > 0);
+        ShareClassId shareClassId = ShareClassId.wrap(id);
+        ShareClassId shareClassId2 = ShareClassId.wrap(id);
```
lemunozm: Why shareClassId2?

test/pools/unit/PoolId.t.sol
```diff
@@ -0,0 +1,16 @@
...
+
+import "forge-std/Test.sol";
+
+import {PoolId, newPoolId} from "src/pools/types/PoolId.sol";
+
+contract PoolIdTest is Test {
+    function testPoolId(uint32 id) public view {
+        PoolId poolId = newPoolId(id);
+
+        assertEq(poolId.isNull(), false);
```
lemunozm: SHould we `assume` id != 0 for some runs?

lemunozm (commented):
One CR. Could we move these tests to a `test/pools/unit/types` folder?

test/pools/unit/ShareClassId.t.sol
onnovisser: just to test the == operator

lemunozm: Ah, nice!

lemunozm (approved):
Thanks Onno!

wischli (approved):
The broken test in CI should have been fixed by #135, just need to rebase. Thanks a lot!

PR #135, opened 2025-03-05
feat: SCM absolute approval amounts instead of ratios

### LINKED ISSUE #128: CP: SCM approvals with absolute amounts instead of ratios
For both approval calls, switch to requiring absolute approval amounts from the FM instead of percentages to reduce the impact of the approval-request race condition ([ref](https://github.com/Recon-Fuzz/centrifuge-review/issues/26))

Ref: https://kflabs.slack.com/archives/C07PG2EUR9C/p1741013141374949

wischli:
Fixes #128

- [x] Fix redeem tests
- [ ] Maybe: remove refund if claimable amount is zero


lemunozm:
A minor CR if you can do it in this PR (not mandatory, I could do it later). Could you move `metadata()` from `ISingleShareClass` to `IShareClassManager`? Currently the `PoolManager` is assuming the configured SCM is the single one to get the metadata when it should be generic.

wischli:
> A minor CR if you can do it in this PR (not mandatory, I could do it later). Could you move `metadata()` from `ISingleShareClass` to `IShareClassManager`? Currently the `PoolManager` is assuming the configured SCM is the single one to get the metadata when it should be generic.

I might do this in a quick follow-up PR where we also remove the refund to pending amounts ([ref](https://github.com/Recon-Fuzz/centrifuge-review/issues/28)).

wischli:
> A minor CR if you can do it in this PR (not mandatory, I could do it later). Could you move `metadata()` from `ISingleShareClass` to `IShareClassManager`? Currently the `PoolManager` is assuming the configured SCM is the single one to get the metadata when it should be generic.

If we go that path, then `metadata` should also return an arbitrary `bytes memory additional` to prepare for future implementations with additional metadata. Do you agree?

src/misc/libraries/MathLib.sol
```diff
@@ -177,12 +177,32 @@ library MathLib {
...
+
+    /// @notice Returns the largest of two numbers.
+    function maxUint128(uint128 a, uint128 b) internal pure returns (uint128) {
+        return a > b ? a : b;
+    }
+
     /// @notice Returns the largest of two numbers.
     function max(uint256 a, uint256 b) internal pure returns (uint256) {
         return a > b ? a : b;
     }
```
lemunozm: Should not this be enough to support all uint types with just this method?

Do all these variants come from some Alex security issue?

test/pools/integration/Cases.t.sol
```diff
@@ -130,7 +130,9 @@ contract TestInvestments is TestConfiguration {
         IERC7726 valuation = holdings.valuation(poolId, scId, USDC_C2);
 
         (bytes[] memory cs, uint256 c) = (new bytes[](2), 0);
-        cs[c++] = abi.encodeWithSelector(poolRouter.approveDeposits.selector, scId, USDC_C2, PERCENT_20, valuation);
+        cs[c++] = abi.encodeWithSelector(
+            poolRouter.approveDeposits.selector, scId, USDC_C2, PERCENT_20.mulUint128(INVESTOR_AMOUNT), valuation
+        );
```
lemunozm, line 136: Maybe now it makes sense to define a constant as `APPROVED_INVESTOR_AMOUNT = 20 * 1e6`, I mean we no longer need the PERCENT_20 constant used before as param.

src/pools/SingleShareClass.sol
```diff
@@ -187,10 +186,10 @@ contract SingleShareClass is Auth, ISingleShareClass {
...
         );
 
-        // Reduce pending
-        approvedAssetAmount = approvalRatio.mulUint128(pendingDeposit[shareClassId_][paymentAssetId]);
-        pendingDeposit[shareClassId_][paymentAssetId] -= approvedAssetAmount;
-        uint128 pendingDepositPostUpdate = pendingDeposit[shareClassId_][paymentAssetId];
+        // Limit in case approved > pending due to race condition of FM approval and async incoming requests
+        uint128 _pendingDeposit = pendingDeposit[shareClassId_][paymentAssetId];
+        approvedAssetAmount = maxApproval.minUint128(_pendingDeposit);
+        require(approvedAssetAmount > 0, ZeroApprovalAmount());
```
lemunozm: What if the race condition comes because the user has canceled? then: `approvedAssetAmount = maxApproval.minUint128(0) = 0`. Do you think we should worry about that?

lemunozm (commented):
Clean and awesome work! I could feel the pain of refactor those tests.

src/pools/SingleShareClass.sol
mustermeiszer: Isn't that already covered or is this new state?

```
        require(approvedAssetAmount > 0, ZeroApprovalAmount());
```

mustermeiszer (approved):
LGTM!

src/pools/SingleShareClass.sol
wischli: Agreeing with @mustermeiszer here. If there is nothing to approve, we will just error out and inform the FM.

src/misc/libraries/MathLib.sol
wischli: You are right. Fixed by 80645193125311d928bfd8dcecd228f3d50dca5f. I think my mistake was that I had to cast the return value of `max` or `min` to `uint128`. However, I instead created max and min variants with same name (i.e. `max` and `min`)  exclusively for `uint128` which didn't work due to name collision. 

My brain wasn't fully functional throughout this PR but I can assure you that I took a lot of care for the rounding accuracy. Seeing that I didn't have to adjust rounding edge cases, I am certain we have the same accuracy as beforehand.

lemunozm: Cool! Thanks for the clean up!

src/pools/SingleShareClass.sol
lemunozm: Correct!

lemunozm (approved):

PR #136, opened 2025-03-06
Feat: Batch addition for CP

lemunozm:
This PR adds full support for creating/reading batches in CP. CV will be in a follow-up PR.

It also set the new architecture to be used by CP and CV in order to process messages and optimize local chain communication. The message flow is the following:

```
                          Send a message ----------------->

pools/PoolManager <-> pools/MessageProcessor <-> common/Gateway <-> common/Adapters

                          <-------------- Receive a message
```

- `PoolManager` will send messages through a `MessageProcessor`. A small contract, easily migrated, that knows how `PoolManager` works and which message it sends and receives. A modification in a message only requires migrating this contract. Vaults will have a homologous for the vaults messages (follow up PR).

- `Gateway` will not be stored in common (factorizing vaults will go in a follow up PR). It contains the common gateway logic, and the logic for batching. Basically, will be the same as `vaults/Gateway` with batching logic and generic enough to work with CP and CV. Will be two deployed instances of this contract, for CP and CV.

- All contracts able to send ongoing messages will implement `IMessageSender`, and all contracts able to handle income message will implement `IMessageHandler`

### DIFF
--- a/script/pools/Deployer.s.sol
+++ b/script/pools/Deployer.s.sol
@@ -6,31 +6,29 @@ import "forge-std/Script.sol";
 import {TransientValuation} from "src/misc/TransientValuation.sol";
 import {IdentityValuation} from "src/misc/IdentityValuation.sol";
 
+import {Gateway} from "src/common/Gateway.sol";
+
 import {AssetId, newAssetId} from "src/pools/types/AssetId.sol";
-import {IGateway} from "src/pools/interfaces/IGateway.sol";
 import {IAdapter} from "src/pools/interfaces/IAdapter.sol";
 import {PoolRegistry} from "src/pools/PoolRegistry.sol";
 import {SingleShareClass} from "src/pools/SingleShareClass.sol";
 import {Holdings} from "src/pools/Holdings.sol";
 import {AssetRegistry} from "src/pools/AssetRegistry.sol";
 import {Accounting} from "src/pools/Accounting.sol";
-import {Gateway} from "src/pools/Gateway.sol";
+import {MessageProcessor} from "src/pools/MessageProcessor.sol";
 import {PoolManager, IPoolManager} from "src/pools/PoolManager.sol";
 import {PoolRouter} from "src/pools/PoolRouter.sol";
 
 contract Deployer is Script {
-    /// @dev Identifies an address that requires to be overwritten by a `file()` method before ending the deployment.
-    /// Just a placesholder and visual indicator.
-    address constant ADDRESS_TO_FILE = address(123);
-
     // Core contracts
     PoolRegistry public poolRegistry;
     AssetRegistry public assetRegistry;
     Accounting public accounting;
     Holdings public holdings;
     SingleShareClass public singleShareClass;
-    PoolManager public poolManager;
     Gateway public gateway;
+    PoolManager public poolManager;
+    MessageProcessor public messageProcessor;
     PoolRouter public poolRouter;
 
     // Utilities
@@ -45,11 +43,10 @@ contract Deployer is Script {
         assetRegistry = new AssetRegistry(address(this));
         accounting = new Accounting(address(this));
         holdings = new Holdings(poolRegistry, address(this));
-
         singleShareClass = new SingleShareClass(poolRegistry, address(this));
-        poolManager =
-            new PoolManager(poolRegistry, assetRegistry, accounting, holdings, IGateway(ADDRESS_TO_FILE), address(this));
-        gateway = new Gateway(IAdapter(address(0 /* TODO */ )), poolManager, address(this));
+        gateway = new Gateway(address(this));
+        poolManager = new PoolManager(poolRegistry, assetRegistry, accounting, holdings, gateway, address(this));
+        messageProcessor = new MessageProcessor(gateway, poolManager, address(this));
         poolRouter = new PoolRouter(poolManager);
 
         transientValuation = new TransientValuation(assetRegistry, address(this));
@@ -61,7 +58,8 @@ contract Deployer is Script {
     }
 
     function _file() public {
-        poolManager.file("gateway", address(gateway));
+        poolManager.file("sender", address(messageProcessor));
+        gateway.file("handle", address(messageProcessor));
     }
 
     function _rely() private {
@@ -70,9 +68,12 @@ contract Deployer is Script {
         holdings.rely(address(poolManager));
         accounting.rely(address(poolManager));
         singleShareClass.rely(address(poolManager));
-        poolManager.rely(address(gateway));
-        poolManager.rely(address(poolRouter));
         gateway.rely(address(poolManager));
+        gateway.rely(address(messageProcessor));
+        poolManager.rely(address(messageProcessor));
+        poolManager.rely(address(poolRouter));
+        messageProcessor.rely(address(poolManager));
+        messageProcessor.rely(address(gateway));
     }
 
     function _initialConfig() private {
@@ -85,8 +86,9 @@ contract Deployer is Script {
         accounting.deny(address(this));
         holdings.deny(address(this));
         singleShareClass.deny(address(this));
-        poolManager.deny(address(this));
         gateway.deny(address(this));
+        messageProcessor.deny(address(this));
+        poolManager.deny(address(this));
 
         transientValuation.deny(address(this));
         identityValuation.deny(address(this));
--- a/script/pools/Deployer.s.sol
+++ b/script/pools/Deployer.s.sol
@@ -57,7 +57,7 @@ contract Deployer is Script {
         _initialConfig();
     }
 
-    function _file() public {
+    function _file() private {
         poolManager.file("sender", address(messageProcessor));
         gateway.file("handle", address(messageProcessor));
     }
--- a/src/common/Gateway.sol
+++ b/src/common/Gateway.sol
@@ -1,0 +1,100 @@
+// SPDX-License-Identifier: BUSL-1.1
+pragma solidity 0.8.28;
+
+import {CastLib} from "src/misc/libraries/CastLib.sol";
+import {BytesLib} from "src/misc/libraries/BytesLib.sol";
+import {Auth} from "src/misc/Auth.sol";
+
+import {MessageType, MessageLib} from "src/common/libraries/MessageLib.sol";
+import {IMessageHandler} from "src/common/interfaces/IMessageHandler.sol";
+import {IMessageSender} from "src/common/interfaces/IMessageSender.sol";
+import {IAdapter} from "src/common/interfaces/IAdapter.sol";
+import {IGateway} from "src/common/interfaces/IGateway.sol";
+
+contract Gateway is Auth, IGateway {
+    using MessageLib for *;
+    using BytesLib for bytes;
+    using CastLib for *;
+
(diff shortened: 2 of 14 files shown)


src/common/Gateway.sol
```diff
@@ -0,0 +1,100 @@
...
+            delete batch[chainId];
+            delete chainIds[i];
+        }
+
+        isBatching = false;
+    }
+
+    /// @inheritdoc IMessageHandler
+    function handle(bytes memory message) external auth {
+        // TODO: add some gateway stuff
```
lemunozm, line 65: Move all vaults/Gateway logic for receiving here (follow up PR)

src/common/Gateway.sol
```diff
@@ -0,0 +1,100 @@
...
+            } else {
+                batch[chainId_] = bytes.concat(previousMessage, message);
+            }
+        } else {
+            _send(chainId_, message);
+        }
+    }
+
+    function _send(uint32 chainId, bytes memory message) private {
+        // TODO: some gateway stuff
```
lemunozm, line 97: Move all vaults/Gateway logic for sending here (follow up PR)

src/pools/MessageProcessor.sol
```diff
@@ -6,42 +6,40 @@ import {BytesLib} from "src/misc/libraries/BytesLib.sol";
...
     function sendNotifyPool(uint32 chainId, PoolId poolId) external auth {
-        // TODO: call directly to CV.poolManager if same chain (apply in all send*() methods)
-        _send(chainId, MessageLib.NotifyPool({poolId: poolId.raw()}).serialize());
+        // In case we want to optimize for the same network:
+        //if chainId == uint32(block.chainId) {
+        //    cv.poolManager.notifyPool(poolId);
+        //}
+        //else {
+        sender.send(chainId, MessageLib.NotifyPool({poolId: poolId.raw()}).serialize());
+        //}
```
lemunozm, line 39: FYI, here is an example of how local CP-CV will work

src/common/Gateway.sol
```diff
@@ -0,0 +1,100 @@
...
+    function file(bytes32 what, address data) external auth {
+        if (what == "adapter") adapter = IAdapter(data);
+        else if (what == "handle") handler = IMessageHandler(data);
+        else revert FileUnrecognizedWhat();
+
+        emit File(what, data);
+    }
+
+    /// @inheritdoc IGateway
+    function setPayableSource(address source_) external {
```
hieronx: This should not be permissionless I think? Can just claim you are someone else then.

lemunozm: Correct! I missed that. I'll do proper Gateway tests once I unify everything, in the follow-up PR to catch these things

lemunozm: done

src/common/Gateway.sol
mustermeiszer: What do you mean by that?

mustermeiszer (commented):
IIUC the tasks of the new contracts are 

* `Gateway` &rarr; batching, adapter aggregation (proof and real messages)
* `MessageProcessor` &rarr; distinguish between local and non-local, expose full message interface as function calls to prevent serialization on the core level
* `Adapter` &rarrl bridge logic

Is the reason to have the `IMessageProcessor` in between the core and `Gateway` that the `Gateway` has storage wrt. adapters and we want to prevent migrating that?

src/common/Gateway.sol
lemunozm: We need to add the gateway code about adapters and handling estimated gas here.

lemunozm:
> IIUC the tasks of the new contracts are 

* `Gateway` &rarr; batching, adapter aggregation (proof and real messages)
* `MessageProcessor` &rarr; distinguish between local and non-local, expose full message interface as function calls to prevent serialization on the core level
* `Adapter` &rarrl bridge logic

Perfectly defined 👌🏻 

> Is the reason to have the `IMessageProcessor` in between the core and `Gateway` that the `Gateway` has storage wrt. adapters and we want to prevent migrating that?

Yes, that's the main reason. A second reason is that extracting that part from the Gateway, would allow to reuse the same gateway contract code for both CP and CV. Because the part that is different for each side is just the `MessageProcessor`. Working on this in a follow-up PRs once I extract Root/GasService/Guardian/Adapters...

script/pools/Deployer.s.sol
```diff
@@ -61,7 +58,8 @@ contract Deployer is Script {
     }
 
     function _file() public {
```
wischli: Nit: Either this function should be private or not be prefixed with `_`

src/common/Gateway.sol
```diff
@@ -0,0 +1,100 @@
...
+
+    IAdapter public adapter; // TODO: support multiple adapters
+    IMessageHandler public handler;
+
+    mapping(uint32 chainId => bytes) public /*transient*/ batch;
+    uint32[] public /*transient*/ chainIds;
+    bool public /*transient*/ isBatching;
+
+    /// @notice The payer of the transaction.
+    address public /*transient*/ payableSource;
```
wischli, line 27: Why are we omitting `transient` here? Just in order to be able to format the file while it's still WIP?

wischli: Nit: Most ideally, storages should be ordered by increasing line length.

src/common/Gateway.sol
```diff
@@ -0,0 +1,100 @@
...
+
+    /// @inheritdoc IGateway
+    function endBatch() external auth {
+        require(isBatching, NoBatched());
+
+        for (uint256 i; i < chainIds.length; i++) {
+            uint32 chainId = chainIds[i];
+            _send(chainId, batch[chainId]);
+            delete batch[chainId];
+            delete chainIds[i];
```
wischli, line 57: Once we enable `transient`, these two lines won't be necessary anymore right?

src/common/interfaces/IAdapter.sol
```diff
@@ -0,0 +1,13 @@
+// SPDX-License-Identifier: BUSL-1.1
+pragma solidity 0.8.28;
+
+interface IAdapter {
+    /// @notice Send a payload to Centrifuge Chain
```
wischli: Nit:
```suggestion
    /// @notice Send a payload to the destination chain
```

src/common/interfaces/IMessageHandler.sol
```diff
@@ -1,7 +1,7 @@
 // SPDX-License-Identifier: BUSL-1.1
 pragma solidity 0.8.28;
 
-/// @notice Generic interface for entities that receives messages
+/// @notice Generic interface for entities that handle incoming messages
```
wischli, line 4: My nit take: If this interface is purely meant to be used for incoming message handling, then I'd prefer the name to be more explicit than the current generic `handle`, i.e. `IMessageReceiver` (also nice reference to our architecture on Centrifuge Chain). Especially since we also have `IMessageSender`.

src/pools/PoolManager.sol
```diff
@@ -64,7 +67,8 @@ contract PoolManager is Auth, IPoolManager, IPoolManagerHandler {
 
     /// @inheritdoc IPoolManager
     function file(bytes32 what, address data) external auth {
-        if (what == "gateway") gateway = IGateway(data);
+        if (what == "sender") sender = IMessageProcessor(data);
```
wischli, line 70: Nit: I think it's misleading to have `IMessageSender` interface but set this to `MessageProcessor`. Maybe we can file it to `"processor"`?

wischli (commented):
A bunch of nits from my side, mainly regarding naming. Feel free to ignore all of it :) Great work!

src/common/Gateway.sol
lemunozm: > Why are we omitting transient here? Just in order to be able to format the file while it's still WIP?

The main reason is that `transient` is not supported for mappings and arrays, so we need to do it in assembly. It requires some extra work I do not have time and we should do only once we know everything needs to be as it is.

Also, support transient for mappings and arrays is a milestone for next solidity version. So maybe we can avoid that work if we deploy once it has been released.

lemunozm: > Nit: Most ideally, storages should be ordered by increasing line length.

Curious why this. I feel more readable structuring it by meaning (as long as the layout is optimized, which is not true right now)

src/common/Gateway.sol
lemunozm: We can batch several times in the same transaction. So we need to be sure we remove the content of previous messages if they are sent. i.e:

- multicall:_
  - execute(poolA)
    - init batch
    - enqueued
    - enqueued
    - end batch (message sent)
 - execute(poolB)
   - init batch (should start from an empty state)
   - ...

src/common/Gateway.sol
wischli: My comment is based on @hieronx's subjective preference ([ref](https://kflabs.slack.com/archives/C07PG2EUR9C/p1736948829340439?thread_ts=1736948718.324019&cid=C07PG2EUR9C)). I suppose in the end it's always up to PR owner.

src/common/interfaces/IMessageHandler.sol
lemunozm: I agree with you, I also prefer "receive". This comes from CV where it uses `handle()` for receiving message everywhere. Curious about the @hieronx rationale.

If we decide to go with `IMessageReceiver`, I'll do this rename once I have everything related to this refactor merged, to avoid conflicts

src/common/Gateway.sol
wischli: Thanks for the explanation!

src/pools/PoolManager.sol
lemunozm: I also thought about that. But finally I determined `sender` because the usage of the `MessageProcessor` from `PoolManager` perspective is only for sending. The `MessageProcessor` contract has logic for both sides, sending and receiving. But `sender` only uses the sending part.

Actually, this makes me think that `IMessageProcessor` should not inherit from `IMessageHandler`, only `MessageProcessor` should inherit from it.

Will leave this as it is right now, if you do not mind, but I definitely think about these names once I have both MessageProcessors in CV and CP. Thanks for being NIT here.

script/pools/Deployer.s.sol
lemunozm: Correct. It should be private.

src/common/interfaces/IAdapter.sol
lemunozm: Good doc catch!

src/common/Gateway.sol
lemunozm: Ok, let's decide once we know if we need assembly for those or not

src/common/interfaces/IMessageHandler.sol
hieronx: No strong rationale on handle vs receive, feel free to change it!

wischli (approved):
Thanks for the explanations and follow-up thoughts about my nits!

src/common/interfaces/IMessageHandler.sol
lemunozm: It results that I can not use the name `receive()` because it is a solidity specific method. That could be why `handle()` was used in the beginning here.

PR #137, opened 2025-03-06
Move contracts to common and adapt vaults to interact with multiple chains

lemunozm:
Please, review this first: https://github.com/centrifuge/protocol-v3/pull/136

### Changes
- Adapt `MessageHandler` and `Adapter` traits to work with `chainId` (and all contracts using them)
- Move `Root `to common
- Move `GasService` to common
- Move `AxelarAdapter` to common

In next PRs I'm gonna:
- Adapt impl of `GasService` to work per ChainId
- Adapt impl of `AxelarAdapter` to no longer work with CentrifugeChain and work with any Chain.
- Merge `vaults/Gateway` into the `common/Gateway` providing a `MessageProcessor.sol` for vaults


src/common/GasService.sol
```diff
@@ -81,7 +71,7 @@ contract GasService is IGasService, Auth {
 
     /// --- Estimations ---
     /// @inheritdoc IGasService
-    function estimate(bytes calldata payload) public view returns (uint256) {
+    function estimate(uint32, /*chainId*/ bytes calldata payload) public view returns (uint256) {
```
lemunozm, line 74: Should the result take into account the message length?

src/vaults/gateway/Gateway.sol
```diff
@@ -270,48 +285,46 @@ contract Gateway is Auth, IGateway, IRecoverable {
...
 
-        uint256 messageCost = gasService.estimate(message);
-        uint256 proofCost = gasService.estimate(proof);
+        uint256 messageCost = gasService.estimate(chainId, message);
+        uint256 proofCost = gasService.estimate(chainId, proof);
 
         if (fuel != 0) {
             uint256 tank = fuel;
-            for (uint256 i; i < numAdapters; i++) {
+            for (uint256 i; i < adapters.length; i++) {
```
lemunozm, line 300: I've needed to do this to avoid the deep stack issue.

mustermeiszer (approved):

src/common/GasService.sol
lemunozm: @hieronx curious about your input here. Why does it not take into account the message length?

test/vaults/integration/Admin.t.sol
```diff
@@ -315,7 +312,7 @@ contract AdminTest is BaseTest {
 
     function _send(MockAdapter adapter, bytes memory message) internal {
         vm.prank(address(adapter));
-        gateway.handle(message);
+        gateway.handle(1, message);
```
wischli: Nit: Maybe we want to hardcode this chain id to a constant as done in other tests?

test/vaults/integration/LocalAdapter.sol
```diff
@@ -74,14 +71,15 @@ contract LocalAdapter is Auth, IAdapter {
         string calldata destinationContractAddress,
         bytes calldata payload
     ) public {
-        gateway.handle(payload);
+        // TODO: get chainId
+        gateway.handle(1, payload);
```
wischli, line 75: Note: In `vaults/AxelarAdapter.sol`, we are hardcoding to chain id `0`. I understand the `chainId` ineffective for `handle` right now. Just wanted to point out the asymmetry.

wischli (approved):
Despite the large amount of changed files, this PR is straight forward. Nothing to complain from my side!

test/vaults/integration/Admin.t.sol
lemunozm: I hardcoded to 1 because for the purpose of these tests, the chain is irrelevant

test/vaults/integration/LocalAdapter.sol
lemunozm: Yes, this is still a TODO. It needs to do it properly, but I need to place something by now.

test/vaults/integration/Admin.t.sol
lemunozm: Added as constant anyway

src/common/GasService.sol
hieronx: This is only the base cost of the transaction. So far for that, we used a single overestimated value.

This is later passed through to the adapter: https://github.com/centrifuge/protocol-v3/blob/main/src/vaults/gateway/Gateway.sol#L292

In `adapter.estimate`, likewise we could use the message length, but have so far only used an overestimated value.

It is important we can eventually implement this to optimize the initial costs, but I don't think this is important today.

lemunozm: Gotcha thanks!

PR #139, opened 2025-03-08
Update vaults invariant test suite

hieronx:


lemunozm:
Is this the port of https://github.com/centrifuge/vaults-internal/pull/65, right?

test/CryticAggregatorForkedTester.sol
```diff
@@ -0,0 +1,14 @@
+// SPDX-License-Identifier: GPL-2.0
```
lemunozm, line 1: NOTE. All new files should be moved under `src/vaults/fuzzing/recon-*`

hieronx: > NOTE. All new files should be moved under src/vaults/fuzzing/recon-*

I don't think this works for Recon's tooling unfortunately.

> Is this the port of https://github.com/centrifuge/vaults-internal/pull/65, right?

Yes, but nican0r is already using another branch. See Recon Discord for context.

lemunozm: > I don't think this works for Recon's tooling unfortunately.

I got running them in the `fuzzing` folder.

hieronx: @lemunozm not from getrecon.xyz, but rather directly medusa/echidna, I assume?

But @nican0r could this work? I seem to remember the Recon tooling had some requirement to be in `test/recon-*`, but maybe that has changed now.

lemunozm:
```
echidna . --contract CryticTester --config test/vaults/fuzzing/recon-core/echidna.yaml --test-mode assertion
```

See ref: https://github.com/centrifuge/vaults-internal/pull/97/files

hieronx: Yes but afaik running the suite in getrecon.xyz didn't work. That is what I was referring to

hieronx: They have additional tooling on top of Echidna/Medusa to automate runs etc.

PR #141, opened 2025-03-10
chore: SCM revert pending refund if zero claimable

wischli:
Fixes https://github.com/Recon-Fuzz/centrifuge-review/issues/28

Context: Decided with @hieronx to drop this code path in order to increase safety of our code at very little cost. It seems to complicate invariant tests by orders of magnitude since it can have some unknown implications


src/pools/SingleShareClass.sol
```diff
@@ -413,13 +413,15 @@ contract SingleShareClass is Auth, ISingleShareClass {
...
 
+            // NOTE: If claimableShareAmount == 0, then we still reduced pendingDeposits by userOrder.pending despite not reducing the latter due to zero claimable. 
+            // This extreme edge case has two implications:
+            //  1. The sum of pending user orders <= pendingDeposits (instead of equality)
+            //  2. The sum of claimable user amounts <= amount of minted share class tokens corresponding to the approved deposit asset amount (instead of equality). 
+            //     I.e., it is possible for an epoch to have an excess of a share class token atom which cannot be claimed by anyone.
             if (claimableShareAmount > 0) {
                 userOrder.pending -= approvedAssetAmount;
                 payoutShareAmount += claimableShareAmount;
                 paymentAssetAmount += approvedAssetAmount;
```
wischli: I'm undecided if we might want to revert the `if` branch here as well and always reduce the pending user amount even if nothing could be claimed. Then, the invariance of "the sum of pending user orders equals the total pending amount" instead of leq would be equality again. However, it means that it can happen a user's pending amount is reduced without receiving anything for it. WDYT @mustermeiszer ?

wischli: As an update: The second implication 
> 2. The sum of claimable user amounts <= amount of minted share class tokens corresponding to the approved deposit asset amount (instead of equality).

Would still be true independent of whether we remove this if-branch.

lemunozm (dismissed):
Towards to simplify the testing invariants!

mustermeiszer (dismissed):
I think that is a fair change to simplify invariants. Lets see if we see problems in the tests when getting to these edge cases, so far seems good.

lemunozm (approved):

PR #142, opened 2025-03-10
feat: multiple decoupled share classes for same pool

### LINKED ISSUE #123: CP: Support multiple share classes per pool + custom SCM
https://kflabs.slack.com/archives/C07PG2EUR9C/p1740650649706689
 
* Just allow multiple share classes which co-exist without coupling or Waterfall
* Rename `SingleShareClass.sol` to `ShareClassManager.sol`
* Add PoC `CustomSingleShareClass` which solely takes the NAV from the holdings and uses that to calculate the price
  * Most ideally provide two PoCs: 
    * The one calls `ShareClassManager` internally
    * The other one has no knowledge about `ShareClassManager` and is wired via `PoolManager`

* Rename `SingleShareClass.sol` to something like `RequestManager.sol`

wischli:
Fixes #123 

* Renames `(I)SingleShareClass` to `(I)MultiShareClass`
* Packs `totalIssuance` and `navPerShare` storages into a single one called `ShareClassMetrics`


src/pools/SingleShareClass.sol
```diff
@@ -76,11 +71,13 @@ contract SingleShareClass is Auth, ISingleShareClass {
     uint32 internal transient _epochIncrement;
     IPoolRegistry public poolRegistry;
     mapping(bytes32 salt => bool) public salts;
-    mapping(PoolId poolId => uint32 epochId_) public epochId;
-    mapping(PoolId poolId => ShareClassId) public shareClassId;
+    mapping(PoolId poolId => uint32) public epochId;
+    mapping(PoolId poolId => uint32) public shareClassCount;
```
hieronx, line 85: This should likely be a struct, would save storage reads/writes if accessed in the same transaction.

hieronx: This might actually be true for other storages as well, basically grouping everything that has the same key set.

src/pools/SingleShareClass.sol
```diff
@@ -76,11 +71,13 @@ contract SingleShareClass is Auth, ISingleShareClass {
...
     mapping(ShareClassId scId => uint128) public totalIssuance;
     mapping(ShareClassId scId => ShareClassMetadata) public metadata;
     mapping(ShareClassId scId => D18 navPerShare) private _shareClassNavPerShare;
+    mapping(PoolId poolId => mapping(ShareClassId => uint32)) public scIdToIndex;
```
hieronx: Why are these mappings needed? They appear unused.  Do we need to be able to iterate over the share classes somewhere?

wischli: Yes, good point. It's redundant. My reasoning behind this doesn't make sense to me anymore.

src/pools/SingleShareClass.sol
wischli: Apart from adding a share class, these two storages are never read/written mutually. Does it still make sense to unite them?

The shareClassCount is read only when previewing the next share class id and written when adding a share class.

The epochId is initialized when adding a share class and then read/written during each approval, issuance, revocation and claiming.

mustermeiszer (dismissed):
Nice!

src/pools/SingleShareClass.sol
hieronx: For these two I think you're right that it doesn't matter.

But for example it looks like merging these
```solidity
mapping(ShareClassId scId => uint128) public totalIssuance;
mapping(ShareClassId scId => D18 navPerShare) private _shareClassNavPerShare;
```

would help as both are accessed simultaneously

src/pools/SingleShareClass.sol
wischli: Ah, now I remember. It's needed for share class id existence checks. Though for that it suffices to map `poolId -> scId -> bool` instead of mapping to an index which was my initial draft.

src/pools/SingleShareClass.sol
wischli: I called this `ShareClassMetrics`: https://github.com/centrifuge/protocol-v3/pull/142/commits/9d9a9247a808ed8b08e68dc48a952fcbf082a0e2

src/pools/SingleShareClass.sol
wischli: Referring to: https://github.com/centrifuge/protocol-v3/pull/142/files#diff-a50f8c8773aaea424253e8ec9fb6c32974d5836b5a4decca8c626d8f413273ebR584

hieronx: Makes sense!

src/pools/SingleShareClass.sol
```diff
@@ -76,11 +78,12 @@ contract SingleShareClass is Auth, ISingleShareClass {
...
     mapping(bytes32 salt => bool) public salts;
-    mapping(PoolId poolId => uint32 epochId_) public epochId;
-    mapping(PoolId poolId => ShareClassId) public shareClassId;
-    mapping(ShareClassId scId => uint128) public totalIssuance;
+    mapping(PoolId poolId => uint32) public epochId;
+    mapping(PoolId poolId => uint32) public shareClassCount;
+    mapping(ShareClassId scId => ShareClassMetrics) public metrics;
     mapping(ShareClassId scId => ShareClassMetadata) public metadata;
-    mapping(ShareClassId scId => D18 navPerShare) private _shareClassNavPerShare;
+    mapping(PoolId poolId => mapping(ShareClassId => bool)) public shareClassIds;
```
lemunozm, line 88: I think this storage would not need a `PoolId`, right? The pool is encoded in the ShareClassId as the others storages.

lemunozm (approved):
LGTM!

src/pools/SingleShareClass.sol
wischli: This particular one actually needs one because it ensures some share class id belongs to a particular pool

hieronx (dismissed):
LGTM!

lemunozm (dismissed):
LGTM! Thanks for the naming refactor!

docs/pools/architecture.puml
```diff
@@ -59,8 +59,8 @@ interface IShareClassManager {
     + updateMetadata(poolId, shareClassId, name, symbol, data)
 }
 
-class SingleShareClass<wards = PoolManager> <<(C, lightskyblue)>>{ }
-SingleShareClass -down|> IShareClassManager
+class MultiShareClass<wards = PoolManager> <<(C, lightskyblue)>>{ }
+MultiShareClass -down|> IShareClassManager
```
lemunozm, line 63: Thanks for updating this! I should also update some changes in the `PoolManager/Gateway`

src/pools/MultiShareClass.sol
```diff
@@ -103,16 +109,20 @@ contract SingleShareClass is Auth, ISingleShareClass {
...
+        shareClassId_ = previewNextShareClassId(poolId);
+        
+        uint32 index = ++shareClassCount[poolId];
+        indexToScId[poolId][index] = shareClassId_;
+        shareClassIds[poolId][shareClassId_] = true;
+
+        // Initialize epoch with 1 iff first class was added
+        if (index == 1) {
+            epochId[poolId] = 1;
+        }
```
lemunozm, line 121: Question. Why is this needed only for the one SC? If it's a general initial configuration, could be initialized in the constructor?

src/pools/MultiShareClass.sol
```diff
@@ -51,17 +51,22 @@ struct EpochPointers {
...
 }
 
-/// Utility method to determine the ShareClassId for a PoolId
-function previewShareClassId(PoolId poolId) pure returns (ShareClassId) {
-    return ShareClassId.wrap(bytes16(uint128(PoolId.unwrap(poolId))));
+struct ShareClassMetrics {
+    /// @dev Total number of shares
+    uint128 totalIssuance;
+    /// @dev The latest net asset value per share class token
+    D18 navPerShare;
```
lemunozm, line 66: Question. IIUC, this is stored only for UX? Because it's never read.

lemunozm (approved):
A couple of questions, but LGTM!!

src/pools/MultiShareClass.sol
wischli: Its needed for every pool, not every SC. We can't do it in the constructor because the constructor doesn't have knowledge of all the (not-yet-existing) pool ids. So IMO, it is best to regard `addShareClass` as SC constructor context of a pool.

src/pools/MultiShareClass.sol
wischli: If you are referring to the `navPerShare` here, then the answer is that right now, we never use the stored `navPerShare` but in the future probably will. I.e. for automatic NAV calculation the flow would be retrieving the current NAV from holdings/the custom conctract, then updating the low level `navPerShare` in the metrics and then working with that.

src/pools/MultiShareClass.sol
lemunozm: Correct! It's per pool 👍🏻

src/pools/MultiShareClass.sol
lemunozm: Ok, I also think it's interesting for the UX to show the performance of the shareClass and showing the current asset -> share transformation

PR #143, opened 2025-03-10
Recon Invariants

nican0r:
This PR adds a scaffolding and properties defined in the engagement with Recon. 

It also renames the `recon-core` -> `recon-vaults` for consistency with the CV, CP convention being used.

Todo:
- [x] scaffolding
- [x] full coverage
- [x] global properties
- [x] inline properties

Latest runs confirming full coverage (7/3):
- [CP](https://getrecon.xyz/shares/36cd15c9-2064-4a45-b0eb-cd0c1e3c7609)
- [CV](https://getrecon.xyz/shares/9fc4069d-e4d5-4ed7-9db0-a2e509314ad3)


lemunozm:
Hi @nican0r thanks for your work!

Could I ask you if you can fix the dependencies to ease reviews in the PR? Right now, the dependencies files are added to the main repo.

nican0r:
> Hi @nican0r thanks for your work!
> 
> Could I ask you if you can fix the dependencies to ease reviews in the PR? Right now, the dependencies files are added to the main repo.

Hi @lemunozm, sorry about that, should've checked building locally before pushing, I've fixed the dependency issues and modified the suite to work with the latest changes to `main`.

remappings.txt
```diff
@@ -1,3 +1,5 @@
 forge-std/=lib/forge-std/src/
 forge-gas-snapshot/=lib/forge-gas-snapshot/src/
+@recon/=lib/setup-helpers/src/
 @chimera/=lib/chimera/src/
+@recon/=lib/setup-helpers/src/
```
hieronx:
```suggestion
```

hieronx: It's duplicated now

test/pools/fuzzing/recon-pools/README.md
```diff
@@ -0,0 +1,18 @@
...
+
+This allows greater path exploration because the fuzzer would be unlikely to call the functions with the necessary bytes passed directly.  
+
+### Differences From Actual Implementation
+The `adapter` set on the `Gateway` is set to the address of a `MockAdapter` in the `Setup` contract. This is a simplification to allow the necessary logic but it shouldn't forward calls made to the `execute` function because these should be made directly via the `PoolRouterTargets` contract.
+
+The current setup uses one instance of the 
+`SingleShareClass` if the relationship between pools and Pools is one-to-many, more will need to be added.
+
+The `Gateway` contract is mocked since we're not actually using it to call the `PoolManager` functions and the extra transferring of gas adds unnecessary complexity to the tester.
```
lemunozm: Not sure if I understand correctly this last line, but the `Gateway` actually calls to `PoolManager` when we receive messages. It calls through [this interface](https://github.com/centrifuge/protocol-v3/blob/main/src/pools/interfaces/IPoolManager.sol#L173) that `PoolManager` implements.

lemunozm (commented):
I did a review regarding things I know and I can verify, and by now, everything looks good to me!

test/pools/fuzzing/recon-pools/targets/AdminTargets.sol
```diff
@@ -0,0 +1,211 @@
...
+    function poolRouter_execute_clamped(PoolId poolId) public payable updateGhosts asAdmin {
+        // TODO: clamp poolId here to one of the created pools
+        poolRouter.execute{value: msg.value}(poolId, queuedCalls);
+
+        queuedCalls = new bytes[](0);
+    }
+
+    // === MultiShareClass === //
+
+    function singleShareClass_claimDepositUntilEpoch(PoolId poolId, ShareClassId shareClassId_, AssetId depositAssetId, uint32 endEpochId) public updateGhosts asAdmin {
```
wischli: Nit: Still uses `singleShareClass` namespace. We can fix this in a follow-up one though, definitely not a blocker. Just wanted to make you aware for future work.

recon.json
```diff
@@ -0,0 +1,632 @@
...
+      },
+      {
+        "signature": "updateHoldingValuation(bytes16,uint128,address)",
+        "actor": "actor",
+        "mode": "normal"
+      }
+    ],
+    "separated": true
+  },
+  "out/SingleShareClass.sol/SingleShareClass.json": {
```
wischli: I suppose this file is auto-generated?

```suggestion
  "out/MultiShareClass.sol/MultiShareClass.json": {
```

test/pools/fuzzing/recon-pools/README.md
```diff
@@ -0,0 +1,18 @@
...
+### Target Functions
+Because the `PoolManager` functions can only be called via the `PoolRouter`, which calls them via the `execute` function, the target functions in `PoolRouterTargets` use the abi encoding of the `PoolRouter` functions to queue calls in an array. They can then be executed in a single transaction in the `execute_clamped` function which makes all the queued calls and clears the queue.
+
+This allows greater path exploration because the fuzzer would be unlikely to call the functions with the necessary bytes passed directly.  
+
+### Differences From Actual Implementation
+The `adapter` set on the `Gateway` is set to the address of a `MockAdapter` in the `Setup` contract. This is a simplification to allow the necessary logic but it shouldn't forward calls made to the `execute` function because these should be made directly via the `PoolRouterTargets` contract.
+
+The current setup uses one instance of the 
+`SingleShareClass` if the relationship between pools and Pools is one-to-many, more will need to be added.
```
wischli: Nit
```suggestion
`MultiShareClass` if the relationship between pools and Pools is one-to-many, more will need to be added.
```

test/pools/fuzzing/recon-pools/TargetFunctions.sol
```diff
@@ -0,0 +1,418 @@
...
+        );
+
+        // claim deposit as actor
+        bytes32 investor = Helpers.addressToBytes32(_getActor());
+        poolManager_claimDeposit(poolId, scId, isoCode, investor);
+
+        return (poolId, scId);
+    }
+
+    function shortcut_deposit_and_cancel(
```
wischli: Note: Another possible flow would be if the user just requested a deposit and immediately follows up with the cancel. However, I believe it's sufficient to regard your scenario as the only one as its more realistic/common.

test/pools/fuzzing/recon-pools/TargetFunctions.sol
```diff
@@ -0,0 +1,418 @@
...
+        uint32 isoCode,
+        uint128 maxApproval,
+        D18 navPerShare,
+        bool isIdentityValuation
+    ) public {
+        shortcut_redeem(poolId, scId, shareAmount, isoCode, maxApproval, navPerShare, isIdentityValuation);
+        
+        // claim redemption as actor
+        bytes32 investor = Helpers.addressToBytes32(_getActor());
+        poolManager_claimRedeem(poolId, scId, isoCode, investor);
```
wischli: Nit:
```suggestion
        // claim redemption as actor
        shortcut_claim_redemption(poolId, scId, isoCode);
```

wischli: Nit:
```suggestion
        // claim redemption as actor
        shortcut_claim_redemption(poolId, scId, isoCode);
```

test/pools/fuzzing/recon-pools/TargetFunctions.sol
```diff
@@ -0,0 +1,418 @@
...
+        
+        // approve and revoke shares as the pool admin
+        // revokes the shares that were issued in the deposit
+        shortcut_approve_and_revoke_shares(
+            poolId, scId, isoCode, _getMultiShareClassMetrics(scId), navPerShare, isIdentityValuation
+        );
+        
+
+        // claim redemption as actor
+        shortcut_claim_redemption(poolId, scId, isoCode);
```
wischli: Q: Why not use the shortcut? Because of `stack too deep`?
```suggestion
        shortcut_redeem_and_claim(poolId, scId, shareAmount, isoCode, _getMultiShareClassMetrics(scId), navPerShare, isIdentityValuation);
```

test/pools/fuzzing/recon-pools/Properties.sol
```diff
@@ -0,0 +1,54 @@
...
+
+                // loop through all actors
+                for (uint256 k = 0; k < _actors.length; k++) {
+                    address actor = _actors[k];
+                    (uint128 pendingUserRedemption,) = multiShareClass.redeemRequest(scId, assetId, Helpers.addressToBytes32(actor));
+                    // check if the actor has a redeem request
+                    gte(
+                        multiShareClass.pendingRedeem(scId, assetId), 
+                        pendingUserRedemption, 
+                        "cancelled redemption is greater than requested redemption"
```
wischli: Nit: Comment is off
```suggestion
                        "user redemption is greater than total redemption"
```

wischli (commented):
Thanks for your great work to get this running! A few nit comments from my side, not a blocker for sure.

test/pools/fuzzing/recon-pools/targets/AdminTargets.sol
nican0r: Good catch, will fix this

test/pools/fuzzing/recon-pools/TargetFunctions.sol
nican0r: Good point, when I tried adding this in though I'm getting a revert with `ClaimDepositRequired` due to this line: 
image

So it seems like user needs to claim before cancelling their deposit request, is that correct?

test/pools/fuzzing/recon-pools/TargetFunctions.sol
nican0r: Yes exactly, replacing with the shortcut gives a stack too deep error so I used the current implementation as a workaround

test/pools/fuzzing/recon-pools/README.md
nican0r: Sorry could have been clearer, what I meant to say is that our fuzzing suite assumes the `Gateway` behaves correctly as there isn't much interesting logic there to test as it just mostly acts to forward messages received from the vaults and send messages to the vault (from my understanding). 

So I've setup the test suite so that calls that would normally be made by the `Gateway` to the `PoolManager` are made directly by the fuzzer instead so the fuzzer doesn't have to work around message handling in the `Gateway` which is assumed to be correct. The functions that are exposed to the fuzzer to do this are [here](https://github.com/centrifuge/protocol-v3/blob/442cff7f4a4048b228024740c671a020d4222c10/test/pools/fuzzing/recon-pools/targets/AdminTargets.sol#L132-L175).

I then mocked the `Gateway` [here](https://github.com/centrifuge/protocol-v3/blob/feat/recon-invariants/test/pools/fuzzing/recon-pools/mocks/MockGateway.sol) that's used in the setup and called by the other contracts to send messages back to the vaults side since the more recent changes to the `Gateway` regarding forwarding gas would require a decent amount of changes to the existing test suite that wouldn't actually add any more interesting state exploration because this is just forwarding messages out to the vaults side. 

Will add this more detailed description to the readme, pls lmk if there's anything still unclear though @lemunozm

test/pools/fuzzing/recon-pools/TargetFunctions.sol
wischli: > So it seems like user needs to claim before cancelling their deposit request, is that correct?

Only if in between executing the `requestDeposit` and `cancelDepositRequest`, the FM at least approved deposits via `approveDeposits`. Then, `userOrder.lastUpdate <= latestApproval`, such that the user needs to claim (after the FM also issued new shares). The reason is that after an approval happens, some share of the users pending amount was approved which they must not be able to mutate until they claimed such that they have 0 pending approved.

The unit test `testCancelDepositRequest` proves that a user can immediately cancel after requesting a deposit.

test/pools/fuzzing/recon-pools/README.md
lemunozm: Makes perfect sense!! Thanks for the long explanation 🙌🏻

test/pools/fuzzing/recon-pools/TargetFunctions.sol
nican0r: Ah, I see now. I think this is a case worth exploring so added a shortcut for it [here](https://github.com/centrifuge/protocol-v3/blob/3fb3837b2251c6d2062ba89a99d09d2ca411f286/test/pools/fuzzing/recon-pools/TargetFunctions.sol#L140-L162)

test/pools/fuzzing/recon-pools/Properties.sol
```diff
@@ -0,0 +1,402 @@
...
+    using MathLib for uint128;
+    using MathLib for uint256;
+
+
+    /// === Canaries === ///
+
+    /// === Global Properties === ///
+
+    function property_unlockedPoolId_transient_reset() public {
+        eq(_after.ghostUnlockedPoolId.raw(), 0, "unlockedPoolId not reset");
```
hieronx: These are neat!

hieronx (dismissed):

hieronx (dismissed):

hieronx:
Closing in favour of https://github.com/centrifuge/protocol-v3/pull/468

PR #144, opened 2025-03-11
Ref: Use same Gateway contract in CP and CV

lemunozm:
### Overal architecture

image

### Main chain list
- `Guardian` to common
- Create a `MessageProcessor` for CV messages.
- `Gateway` used by CP and CV.

### Pending work for future PRs
- Solve the payment issue for batches: https://github.com/centrifuge/protocol-v3/pull/149
- Make it GasAdapter and Adapter to work correctly multichain: https://github.com/centrifuge/protocol-v3/pull/153
- Recovery messages with multichain support.


src/common/Gateway.sol
```diff
@@ -1,70 +1,151 @@
 // SPDX-License-Identifier: BUSL-1.1
```
lemunozm, line 1: Apologies to the git diff of this file. This file basically is the vaults/Gateway with batch support. Because I already had some stuff in common/Gateway the diff is like if everything is new, but it is not.

src/common/Guardian.sol
```diff
@@ -1,34 +1,29 @@
...
-    function isOwner(address signer) external view returns (bool);
-}
+import {IRoot} from "src/common/interfaces/IRoot.sol";
+import {IGuardian, ISafe} from "src/common/interfaces/IGuardian.sol";
+import {IGateway} from "src/common/interfaces/IGateway.sol";
+import {IAdapter} from "src/common/interfaces/IAdapter.sol";
 
 contract Guardian is IGuardian {
-    Root public immutable root;
+    IRoot public immutable root;
```
lemunozm, line 10: I think this should be `IRoot` to avoid uploading the `Root` bytecode along with this contract. double-check @hieronx

lemunozm (commented):
I hope from this PR I can submit more small PRs with less changes cross-repo 🙏🏻

src/common/Guardian.sol
hieronx: Good point!

script/pools/Deployer.s.sol
```diff
@@ -19,13 +21,19 @@ import {PoolManager, IPoolManager} from "src/pools/PoolManager.sol";
 import {PoolRouter} from "src/pools/PoolRouter.sol";
 
 contract Deployer is Script {
-    // Core contracts
+    uint256 internal constant delay = 48 hours;
```
mustermeiszer, line 24: Is this taken from the CV side?

src/vaults/InvestmentManager.sol
```diff
@@ -202,15 +194,9 @@ contract InvestmentManager is Auth, IInvestmentManager {
...
-            MessageLib.CancelDepositRequest({
-                poolId: vault_.poolId(),
-                scId: vault_.trancheId(),
-                investor: controller.toBytes32(),
-                assetId: poolManager.assetToId(vault_.asset())
-            }).serialize(),
-            source
+        gateway.setPayableSource(source);
+        sender.sendCancelDepositRequest(
+            vault_.poolId(), vault_.trancheId(), controller.toBytes32(), poolManager.assetToId(vault_.asset())
```
mustermeiszer, line 199: Are we always assuming the message handler retrieves the `chainId` from the `PoolId` ?

Not all messages are structured like that, right? Or maybe just `TransferShares`.

src/vaults/MessageProcessor.sol
```diff
@@ -0,0 +1,196 @@
...
+    function sendCancelRedeemRequest(uint64 poolId, bytes16 scId, bytes32 investor, uint128 assetId) external auth {
+        gateway.send(
+            uint32(poolId >> 32),
+            MessageLib.CancelRedeemRequest({poolId: poolId, scId: scId, investor: investor, assetId: assetId}).serialize(
+            )
+        );
+    }
+
+    /// @inheritdoc IMessageHandler
+    function handle(uint32, /* chainId */ bytes memory message) external auth {
```
mustermeiszer, line 99: Why do we have the `ChainId` here, but we never need it? Is this from the sending chain? Doe we emit an event that tells us that we are processing a message from said chain somewhere? 

If not, I would like to add that! That was one of the best things for debugging/checking execution on receiving chains on the Centrifuge Chain.

src/vaults/interfaces/IMessageProcessor.sol
mustermeiszer, line 1: We expect that interface to grow with messages, right?

mustermeiszer (dismissed):
Nothing to complain. Thanks! Like the refactor. 

NIT: Would really like to increase the event emitted when processing messages, so that checking in the explorer, what happened , triggered from where gets easier to understand. 

But nothing blocking.

script/pools/Deployer.s.sol
wischli: Yes

src/vaults/MessageProcessor.sol
lemunozm: The chainId is something the adapter can offer, but we're not using it, right now in any of our messages.

This is why I proposed this: https://kflabs.slack.com/archives/C07PG2EUR9C/p1741256686002279 to save bytes without affect to any PoolId/AssetId stored.

Agree with the event! I think it could even be in the `gateway.handle()` to also consider recovery/proofs messages and be shared between CP and CV.

src/vaults/InvestmentManager.sol
lemunozm: No, we're not assuming that generically. This method is message content-related, and we are assuming the chainId is in the poolId, but just for it. But other methods as `CP.sendNotifyPool()` does not assume that and have a chainId param.

src/vaults/interfaces/IMessageProcessor.sol
lemunozm: Correct 👍🏻

src/vaults/MessageProcessor.sol
lemunozm: Realizing now that we already have an `ExecuteMessage` event in the Gateway emitted each time a message will be handled. Added a `chainId` parameter there! [6cd2d18](https://github.com/centrifuge/protocol-v3/pull/144/commits/6cd2d1885e4fd279c740eee698b9da5834c7112b)

wischli (commented):
LGTM! Getting cleaner day by day :broom:

wischli (approved):

mustermeiszer (approved):
Re-Approving

PR #146, opened 2025-03-12
Ref/Simplify delegate call

lemunozm:
From https://github.com/centrifuge/protocol-v3/pull/108#discussion_r1986758530

I initially proposed the delegateCall to avoid reliance. But we can also perform just an internal call that. Sorry for my previous bad/complex recommendation @mustermeiszer

### DIFF
--- a/src/vaults/PoolManager.sol
+++ b/src/vaults/PoolManager.sol
@@ -264,18 +264,7 @@ contract PoolManager is Auth, IPoolManager, IUpdateContract {
     /// @inheritdoc IPoolManager
     function updateContract(uint64 poolId, bytes16 trancheId, address target, bytes memory update_) public auth {
         if (target == address(this)) {
-            (bool success, bytes memory returnData) = address(this).delegatecall(
-                abi.encodeWithSelector(IUpdateContract.update.selector, poolId, trancheId, update_)
-            );
-
-            if (!success) {
-                uint256 length = returnData.length;
-                require(length != 0, UpdateContractFailed());
-
-                assembly ("memory-safe") {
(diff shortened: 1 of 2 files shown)


src/vaults/PoolManager.sol
```diff
@@ -264,18 +264,7 @@ contract PoolManager is Auth, IPoolManager, IUpdateContract {
     /// @inheritdoc IPoolManager
     function updateContract(uint64 poolId, bytes16 trancheId, address target, bytes memory update_) public auth {
         if (target == address(this)) {
-            (bool success, bytes memory returnData) = address(this).delegatecall(
-                abi.encodeWithSelector(IUpdateContract.update.selector, poolId, trancheId, update_)
-            );
-
-            if (!success) {
-                uint256 length = returnData.length;
-                require(length != 0, UpdateContractFailed());
```
wischli, line 273: Nit: This was the only codepath to revert with `UpdateContractFailed` error. So it's redundant now.

wischli (commented):
Loving the simplification!

src/vaults/PoolManager.sol
```diff
@@ -264,18 +264,7 @@ contract PoolManager is Auth, IPoolManager, IUpdateContract {
...
-
-                assembly ("memory-safe") {
-                    revert(add(32, returnData), length)
-                }
-            }
+            update(poolId, trancheId, update_);
```
mustermeiszer, line 267: Nice!

mustermeiszer (dismissed):

mustermeiszer (approved):
Nice clean up!

PR #147, opened 2025-03-12
ci: re-enable CV coverage

wischli:
Ref: https://kflabs.slack.com/archives/C07PG2EUR9C/p1741766889378979

### DIFF
--- a/foundry.toml
+++ b/foundry.toml
@@ -38,6 +38,18 @@ depth          = 1000
 call_override  = false
 fail_on_revert = false
 
+[profile.ci-coverage]
+verbosity = 2
+
+[profile.ci-coverage.coverage]
+ignore = [
+  "test/vaults/fuzzing/**/*.sol",
+]
+
+[profile.ci-coverage.fuzz]
+runs = 512 # 2x of default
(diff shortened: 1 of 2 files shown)


foundry.toml
```diff
@@ -38,6 +38,18 @@ depth          = 1000
...
+verbosity = 2
+
+[profile.ci-coverage.coverage]
+ignore = [
+  "test/vaults/fuzzing/**/*.sol",
+]
+
+[profile.ci-coverage.fuzz]
+runs = 512 # 2x of default
+max_test_rejects = 65_536 # default value
```
wischli, line 51: AFAICT, either of these were causing the timeouts, most likely due to the Github runner running out of memory. However, we did not apply any profile to the coverage runner and it worked out of the box for the default values, which doesn't make much sense. Anyways, I lowered these compared to our `ci` profile in which we have 1000 runs and 1M max rejects (for `vm.assume`).

lemunozm: Thanks to much to dig into it!!

lemunozm (approved):

PR #149, opened 2025-03-14
Feat: Full batch support with payment

lemunozm:
## First, an overview of the problem

We need two support two kind of message payment `topUp()` and `setPayableSource(source)`
- The first one allows you to pay to the gateway in the same transaction. 
  - Used by users that can choose to pay when calling the transaction.
- The second one allows you to set the `source` as the originator of the call. If that `source` is in the whitelist of the `GasService`, the send process can get funds from the Gateway. Funds should have been paid previously by calling `Gateway.receive()`.
  - Used by other contract that calls to our system and they cannot pay at that point.

On top of this, we need to support optional batching, it means:
- If no batching, the user must pay in the method when the message is sent.
- If batching, the user must pay at the end of the batching. For only one message.

And on top of this, we have solidity restrictions. We can not pay individually for each encoded method in the multicall. We can only pay in the multicall.

## Solution

Only `PoolRouter` and `VaultRouter` (previous `CentrifugeRouter`) can perform batches. These contracts will pay through `topUp()` internally to set the fuel for the multicall (or individual calls).

When processing methods, the payment is performed only if you're not in a multicall. If you are in a multicall, only one message is sent, and you will pay for that at the end of the multicall.

## Change list
- Individual `VaultsRouter` methods no longer require a gas parameter.
- `CentrifugeRouter` renamed to `VaultsRouter`.
- Batch payment at the end of the batch composition.


test/pools/integration/Cases.t.sol
```diff
@@ -185,4 +196,56 @@ contract TestCases is Deployer, Test {
...
+    }
+
+    /// Test the following:
+    /// - multicall()
+    ///   - execute(poolA)
+    ///      - notifyPool()
+    ///   - execute(poolB)
+    ///      - notifyPool()
+    ///
+    /// will pay only for one message. The batch sent is [NotifyPool, NotifyPool].
```
lemunozm, line 233: I'm happy I get a solution where you can compose any number of multicall and only one message will be sent (per chain/adapter)

src/common/Gateway.sol
```diff
@@ -329,9 +334,14 @@ contract Gateway is Auth, IGateway, IRecoverable {
...
     /// @inheritdoc IGateway
     function topUp() external payable {
-        require(payers[msg.sender], "Gateway/only-payers-can-top-up");
-        require(msg.value != 0, "Gateway/cannot-topup-with-nothing");
-        fuel = msg.value;
+        // We only require the top up if:
+        // - We're not in a multicall.
+        // - Or we're in a multicall, but at least one message is required to be sent
+        if (!isBatching || sendWasBuffered) {
+            require(payers[msg.sender], "Gateway/only-payers-can-top-up");
```
lemunozm: I'm entirely sure `payers` are no longer needed. Basically because, `topUp()` can only be called in the Routers, which need to decide how to pay depending if they're batching or not. Both routers already need the `Gateway` to be `Auth` to perform batches.

Moreover, I'm pretty sure `topUp()` can be permissionless as long as we add this condition:
```sol
require(msg.value > fuel)
```
so no other user can try to call `topUp()` in the same transaction to reduce the fuel. This would simplify a bit some stuff.

cc @hieronx

src/common/Gateway.sol
```diff
@@ -329,9 +334,14 @@ contract Gateway is Auth, IGateway, IRecoverable {
...
+            require(msg.value != 0, "Gateway/cannot-topup-with-nothing");
+            fuel = msg.value;
+        }
     }
 
     function setPayableSource(address source) external {
```
lemunozm, line 343: I feel like `topUp()` and `setPayableSource()` have a lot of collision between them and make things very difficult to code, test, and understand them. 

Curious why not the following can not work to pay instead of using `source`:

```
- genericMulticall
  - topUp()
  - ExternalContract -> Vaults -> Gateway.send() // paid with fuel without using VaultsRouter
```
or just:
```
- ExternalContract -> Vaults -> Gateway.send() //already with fuel
- Pay later directly to the adapter. // Or in the same multicall.
```

src/pools/PoolRouter.sol
```diff
@@ -19,9 +21,29 @@ import {IPoolRouter} from "src/pools/interfaces/IPoolRouter.sol";
 
 contract PoolRouter is Multicall, IPoolRouter {
```
hieronx, line 22:
```suggestion
contract PoolsRouter is Multicall, IPoolRouter {
```

src/pools/PoolRouter.sol
```diff
@@ -19,9 +21,29 @@ import {IPoolRouter} from "src/pools/interfaces/IPoolRouter.sol";
...
+    function multicall(bytes[] calldata data) public payable override {
+        bool wasBatching = gateway.isBatching();
+        if (!wasBatching) {
+            gateway.startBatch();
+        }
+
+        super.multicall(data);
+
+        if (!wasBatching) {
+            gateway.setPayableSource(msg.sender);
```
hieronx: Why do we need `setPayableSource` here? Since you are always topping up when calling the pools router.

src/common/Gateway.sol
hieronx: Yeah I agree with the new solution, it makes sense to remove `payers` and rely on `auth`.

hieronx: In terms of making it permissionless, it's hard for me to know right now if there are edge cases that might make this not work. Since we anyways need `auth`, I would suggest keeping `topUp` permissioned. Just restricts the flows more without any downsides.

src/common/Gateway.sol
hieronx: For one, you cannot pay later easily. This requires knowing the specific interface of the specific adapter/bridge we are using. And it generally requires knowing transaction data like the index in the block that you can only know offchain

hieronx: As for the former, the intended users of the `source` is smart contract integrations. Those are by definition not using some specific wrapper like `genericMulticall`. They only know the vault address and that it complies with ERC7540 and thus they can use that interface.

src/common/Gateway.sol
lemunozm: Agree. My only point in favor of permisionless is if it could enable us to remove the `source`, and use the `topUp()` for all cases. i.e. the first case in the below message.

src/pools/PoolRouter.sol
```diff
@@ -19,9 +21,29 @@ import {IPoolRouter} from "src/pools/interfaces/IPoolRouter.sol";
...
+    constructor(IPoolManager poolManager_, IGateway gateway_) {
         poolManager = poolManager_;
+        gateway = gateway_;
+    }
+
+    // --- Administration ---
+    /// @inheritdoc IMulticall
+    /// @notice performs a multicall but all messages sent in the process will be batched
+    function multicall(bytes[] calldata data) public payable override {
+        bool wasBatching = gateway.isBatching();
```
mustermeiszer, line 35: Doesn't that mean others can potentially `startBatching` without the us wanting to batch? 

IMO we should restrict batching to transient storage.

src/pools/PoolRouter.sol
```diff
@@ -34,48 +56,65 @@ contract PoolRouter is Multicall, IPoolRouter {
...
+        external
+        payable
+        returns (PoolId poolId)
+    {
         return poolManager.createPool(msg.sender, currency, shareClassManager);
     }
 
     /// @inheritdoc IPoolRouter
-    function claimDeposit(PoolId poolId, ShareClassId scId, AssetId assetId, bytes32 investor) external protected {
+    function claimDeposit(PoolId poolId, ShareClassId scId, AssetId assetId, bytes32 investor)
```
mustermeiszer, line 67: What would happen IF those are in a multicall?

src/pools/PoolRouter.sol
lemunozm: Yes, probably the whole `setPayableSource` is not needed for pools.

Ideally, following the vaults schema, if you pay with `topUp()` zero, you should then check the `source` to pay with the gateway funds. However, this does not work now because it would require to `topUp()` with zero which is not supported.

Not sure if we would need this path.

mustermeiszer (commented):
Great work! 

Few questions, but also not the best to decide as the knowledge about the payment logic is pretty blurry form my side.

src/pools/PoolRouter.sol
lemunozm: Makes total sense!

src/pools/PoolRouter.sol
lemunozm: Yes, it's already a transient storage.

The above is as it is because you can do nested multicalls, so you need to check if you're the top multicall in order to be the one that finalizes the batch

src/pools/PoolRouter.sol
lemunozm: They would work as any other pool admin method. The payment will be done at the end of the multicall and the message enqueued

lemunozm: Only if they are not called inside of a multicall, then you process the payment just for them

src/pools/PoolRouter.sol
hieronx: I actually think we don't need `topUp` on the pools side at all, since there are no smart contract integrations based on ERC7540 or another standard. Every user would go through the `PoolsRouter` and every user can be forced to `topUp`

mustermeiszer (dismissed):
I really like the new design!

src/pools/PoolRouter.sol
lemunozm: > I actually think we don't need `topUp`

Do you mean `setPayableSource`, right?

If not, I'm not following.

hieronx: Yeah my bad I meant `setPayableSource` there.

src/pools/PoolRouter.sol
lemunozm: After talked, I've modified `VaultsRouter` to `VaultRouter`

src/pools/PoolRouter.sol
lemunozm: Removed [6af7620](https://github.com/centrifuge/protocol-v3/pull/149/commits/6af7620f2d1bab14104013ddd0662d7c3775bb2e)

src/common/Gateway.sol
lemunozm: Removed: [b61ae4a](https://github.com/centrifuge/protocol-v3/pull/149/commits/b61ae4ab0f7ac8c1675a437ceac2a6a96219d61f)

test/pools/integration/Cases.t.sol
```diff
@@ -185,4 +196,56 @@ contract TestCases is Deployer, Test {
         );
         assertEq(m0.shareAmount, APPROVED_SHARE_AMOUNT);
     }
+
+    /// forge-config: default.isolate = true
+    function testExecuteNoSendNoPay() public {
```
wischli, line 199: We are not ensuring, that we are not sending here, are we? I know it's difficult to test that an even has not been emitted.

wischli (commented):
LGTM! Not approving, because I am also fuzzy on the gas payment side.

script/vaults/Deployer.sol
```diff
@@ -126,6 +126,9 @@ contract Deployer is Script {
         investmentManager.rely(address(messageProcessor));
         root.rely(address(messageProcessor));
         gasService.rely(address(messageProcessor));
+
+        // Rely on VaultRouter
```
hieronx, line 128: This is opposite (relying vault router on gateway), but also the comment seems unnecessary
```suggestion
```

src/common/GasService.sol
```diff
@@ -84,7 +84,7 @@ contract GasService is IGasService, Auth {
     }
 
     /// @inheritdoc IGasService
-    function shouldRefuel(address, bytes calldata) public pure returns (bool success) {
-        success = true;
+    function shouldRefuel(address source, bytes calldata) public pure returns (bool success) {
+        return source != address(0);
```
hieronx, line 46: Can you explain this change?

src/common/Gateway.sol
```diff
@@ -33,6 +33,9 @@ contract Gateway is Auth, IGateway, IRecoverable {
 
     uint256 public transient fuel;
 
+    /// @dev Says if a send() was performed, but it's still pending to finalize the batch
+    bool public transient sendWasBuffered;
```
hieronx: This name confused me a bit. Wdyt about `batchIsPending` or `pendingBatch`?

src/common/interfaces/IRoot.sol
```diff
@@ -52,7 +52,7 @@ interface IRoot {
     /// @notice Endorses the `user`
     /// @dev    Endorsed users are trusted contracts in the system. They are allowed to bypass
     ///         token restrictions (e.g. the Escrow can automatically receive tranche tokens by being endorsed), and
-    ///         can automatically set operators in ERC-7540 vaults (e.g. the CentrifugeRouter) is always an operator.
+    ///         can automatically set operators in ERC-7540 vaults (e.g. the VaultsRouter) is always an operator.
```
hieronx:
```suggestion
    ///         can automatically set operators in ERC-7540 vaults (e.g. the VaultRouter) is always an operator.
```

src/common/GasService.sol
lemunozm: It's just a "null check" I added this as protection. 

Rationale. If, in some part of the code, we forget to call `setPayableSource()` before performing an action that sends a message, then that message will use funds from the gateway to be sent when there is no `source` set.

lemunozm: Before, we were forced to pass the `source` in the action, but now it's not. So it's easy to forget it.

test/pools/integration/Cases.t.sol
lemunozm: We can check the `MockVaults` to ensure there are no messages 👍🏻 . Will do!

script/vaults/Deployer.sol
lemunozm: What I want is to enable `router` to call to the `gateway`, so the `gateway` must rely on `router`, right?

test/pools/integration/Cases.t.sol
lemunozm: Done

script/vaults/Deployer.sol
hieronx: > the gateway must rely on router

I guess we just have different definitions of the directionality of "relying".  The way I use/interpret it, calling `a.rely(B)` means relying B on A.

src/common/GasService.sol
hieronx: Got it, makes sense!

script/vaults/Deployer.sol
lemunozm: Mmm... but I'm using the exact same pattern/comments as above in the same function. Are they all wrong?

hieronx: Ha, I see now! Yes in my mind they are "wrong", but I guess whoever wrote this deployment script had your interpretation. Let me see if I can find a correct answer for the definition here ;)  But either way not a blocker for this PR then obviously.

hieronx (approved):

PR #150, opened 2025-03-14
JournalId and TODOs from Accounting contract

onnovisser:
Adds the TransactionId contract and fixes some other leftover TODOs from the Accounting contract work.

- Add events for creating accounts and setting metadata
- Fix the natspec comments


.forge-snapshots/Tranche_transferFrom.snap
```diff
@@ -1 +1 @@
-84198
\ No newline at end of file
+29958
```
lemunozm: NOTE. Have you upgrade your local foundry to the latest nightly? use `foundryup` for that if not

src/pools/TransactionId.sol
```diff
@@ -0,0 +1,23 @@
...
+    mapping(PoolId => uint256) internal _poolCounter;
+
+    /// @inheritdoc ITransactionId
+    bytes32 public transient transactionId;
+
+    constructor(address deployer) Auth(deployer) {}
+
+    /// @inheritdoc ITransactionId
+    function generateTransactionId(PoolId poolId) external auth returns (bytes32) {
+        transactionId = bytes32(++_poolCounter[poolId]);
```
lemunozm: Curious, 
- Do we want to add some timestamp here?
- Do we want to generate a hash of this?

Just wondering about those, I'm not encouraging to do it.

Probably @mustermeiszer can give a product view here

src/pools/interfaces/IAccounting.sol
```diff
@@ -9,6 +9,9 @@ interface IAccounting {
     event Credit(PoolId indexed poolId, bytes32 indexed transactionId, AccountId indexed account, uint128 value);
     event Debit(PoolId indexed poolId, bytes32 indexed transactionId, AccountId indexed account, uint128 value);
 
+    event AccountCreated(PoolId indexed poolId, AccountId indexed account, bool isDebitNormal);
+    event AccountMetadataSet(PoolId indexed poolId, AccountId indexed account, bytes metadata);
```
lemunozm: NIT. Could you add docs here?

lemunozm (commented):
The implementation is just perfect 🥇. LGTM!

Just want to double check one question below, otherwise ready to be merge

onnovisser:
@mustermeiszer I'm reading the CAL specs and am now wondering if this contract is actually still needed if the CAL side sets the JournalId?

lemunozm (dismissed):

mustermeiszer:
> @mustermeiszer I'm reading the CAL specs and am now wondering if this contract is actually still needed if the CAL side sets the JournalId?

I think it can work like that. Not 100% sure though. I haven't specced that fully atm. But roughly it would be  

* Allow to "set" new journal id in transient storage as part of `CP.PoolRouter` flow
* All events use that id in their events - IF non is set, use 0
* Add  a Message `SetJournalId` in the message lib
* Processing the message sets the transient storage

### Why
We need to wire together multiple operations/method calls into different contracts together, so that the indexing logic can make sense of it. 
* We need that from the CP side
* We need that from the CV side 

### Open Questions
* Just use a user provided `JournalId`

lemunozm:
> All events use that id in their events - IF non is set, use 0

This means we need to pass everywhere that information, to all contracts. Could not we just dispatch the event `JournalId()` and the indexer offer then next events already indexed by that `JournalId()`?. I would want if that's is something possible to the indexer to leave the codebase cleaner 🤔 

> Processing the message sets the transient storage

If we set the transient storage, then, any new call by the FM will no longer use that journal Id.

mustermeiszer:
- indexing can not do that efficiently &rarr; needs to be part of the events. I would say rather query from contract than pass as parameter maybe 

- yes, it is intended to work only for a multi call or a batch

src/pools/interfaces/IAccounting.sol
```diff
@@ -34,24 +37,38 @@ interface IAccounting {
...
+    /// @param account The account to set metadata for
+    /// @param metadata The metadata to set
     function setAccountMetadata(PoolId poolId, AccountId account, bytes calldata metadata) external;
 
-    /// @notice Returns the value of an account, returns a negative value for positive balances of credt-normal
-    /// accounts.
+    /// @notice Returns the value of an account
+    /// @param poolId The pool the account belongs to
+    /// @param account The account to get the value of
+    /// @return The value of the account. Will be a negative value for positive balances of credt-normal accounts
```
wischli: Nit: Typo
```suggestion
    /// @return The value of the account. Will be a negative value for positive balances of credit-normal accounts
```

test/pools/unit/TransactionId.t.sol
```diff
@@ -0,0 +1,36 @@
+// SPDX-License-Identifier: BUSL-1.1
+pragma solidity 0.8.28;
+
+import "forge-std/Test.sol";
+
+import {IAuth} from "src/misc/interfaces/IAuth.sol";
+import {PoolId} from "src/pools/types/PoolId.sol";
+import {AccountId, newAccountId} from "src/pools/types/AccountId.sol";
```
wischli: Nit: Not needed

wischli (commented):
LGTM but let's discuss in the sync, how we proceed with this and the unspecced `JournalId` for CAL

src/pools/PoolRegistry.sol
```diff
@@ -79,6 +83,12 @@ contract PoolRegistry is Auth, IPoolRegistry {
         emit UpdatedCurrency(poolId, currency_);
     }
 
+    /// @inheritdoc IPoolRegistry
+    function generateJournalId(PoolId poolId) external auth returns (bytes32) {
```
wischli: CR: Should be a simple `uint256` counter per pool.

src/pools/Accounting.sol
```diff
@@ -12,8 +12,8 @@ contract Accounting is Auth, IAccounting {
 
     uint128 public transient debited;
     uint128 public transient credited;
-    bytes32 private transient _transactionId;
-    PoolId private transient _currentPoolId;
+    bytes32 internal transient _journalId;
```
wischli: IIUC, this storage is necessary because `Accounting` doesn't know the `PoolRegistry` such that it can just access that transient journal id, right?

wischli (commented):
IIRC, @mustermeiszer desired a simple counter journal id in the Standup.

src/pools/Accounting.sol
onnovisser: yes, exactly

src/pools/PoolRegistry.sol
onnovisser: IIRC, we decided to combine with the pool id, because it had to be unique among pools

mustermeiszer: Yes!

mustermeiszer: But I think `uint256` is easier to sort than `bytes32`

mustermeiszer (commented):
We could, with the new indexing idea, just emit the events before `lock` and after `unlock` right? That would work as long as we always lock and unlock.

src/pools/PoolRegistry.sol
```diff
@@ -14,6 +14,10 @@ contract PoolRegistry is Auth, IPoolRegistry {
 
     uint32 public latestId;
 
+    mapping(PoolId => uint64) internal _poolJournalIdCounter;
+    /// @inheritdoc IPoolRegistry
+    bytes32 public transient journalId;
```
lemunozm: We'll need to get back to this in order to process batches correctly. Probably we'll need a transient journalIdPerPool. And we need to evaluate if we should lock/unlock for those messages.

Future PR for sure

onnovisser: yes was wondering the same

onnovisser:
* Moved to Accounting
* Turned id into `uint256`
* Emit journal events at unlock/lock

mustermeiszer (dismissed):
LGTM!

src/pools/PoolRouter.sol
```diff
@@ -93,7 +93,7 @@ contract PoolRouter is Auth, Multicall, IPoolRouter, IPoolRouterHandler {
         require(unlockedPoolId.isNull(), IPoolRouter.PoolAlreadyUnlocked());
         require(poolRegistry.isAdmin(poolId, msg.sender), IPoolRouter.NotAuthorizedAdmin());
 
-        accounting.unlock(poolId, poolRegistry.generateJournalId(poolId));
+        accounting.unlock(poolId, accounting.generateJournalId(poolId));
```
lemunozm, line 96: Question, is `generateJournalId()` only called inside `unlock()`, or do we expect to also call this by a CV message?

lemunozm (dismissed):
Awesome work!

Sorry for the back and forth during this PR

src/pools/Accounting.sol
```diff
@@ -12,8 +12,11 @@ contract Accounting is Auth, IAccounting {
 
     uint128 public transient debited;
     uint128 public transient credited;
-    bytes32 private transient _transactionId;
-    PoolId private transient _currentPoolId;
+    PoolId internal transient _currentPoolId;
+
+    mapping(PoolId => uint64) internal _poolJournalIdCounter;
+    /// @inheritdoc IAccounting
+    uint256 public transient journalId;
```
wischli: Nit: Could we reorder this follow our soft "rule" to have storage ordered by line length?

```suggestion
    /// @inheritdoc IAccounting
    uint256 public transient journalId;
    mapping(PoolId => uint64) internal _poolJournalIdCounter;
```

src/pools/Accounting.sol
```diff
@@ -27,7 +30,7 @@ contract Accounting is Auth, IAccounting {
         acc.totalDebit += value;
         debited += value;
         acc.lastUpdated = uint64(block.timestamp);
-        emit Debit(_currentPoolId, _transactionId, account, value);
+        emit Debit(_currentPoolId, journalId, account, value);
```
wischli: @mustermeiszer Seems redundant to me to still include the journal id in the `Debit` and `Credit` events when we have `{Start, End}JournalId`, do you agree?

wischli (dismissed):
Thanks for adding docs and not losing patience in the back and forth!

src/pools/Accounting.sol
lemunozm: An addition here. We should divide transient from normal storage, because their layout is in a different space IIUC. I would propose start always with `transient` followed by the normal storages.

lemunozm: @wischli not remember if already asked (I think I did but forgot the answer), from where comes the style requirement about order by line length?

src/pools/PoolRouter.sol
onnovisser: I think it has to be called somewhere from a CV message as well, otherwise they can't update Accounting

src/pools/Accounting.sol
onnovisser: Yes, I wasn't sure about that

onnovisser: Removing seems fine to me

src/pools/Accounting.sol
wischli: It's just an aesthetic subjective opinion from @hieronx raised here: https://kflabs.slack.com/archives/C07PG2EUR9C/p1736948829340439?thread_ts=1736948718.324019&cid=C07PG2EUR9C

So far, we have followed this pretty well which is the only reason I brought it up. Here, I also like that the public storage is ordered higher than the internal one.

src/pools/Accounting.sol
mustermeiszer: Agree to remove

mustermeiszer (approved):

PR #153, opened 2025-03-14
Add Wormhole adapter + refactor gas service

### LINKED ISSUE #152: Add Wormhole adapter
Docs: https://wormhole.com/docs/build/contract-integrations/wormhole-relayers/

hieronx:
Wormhole’s gas logic includes onchain estimations, that use the gas limit as the input, and automatically calculate the ETH gas cost from this. This is similar to what we implemented manually in the `GasService` contract. After doing a bit of digging, I found out that Axelar now also supports a similar method, which they hadn’t supported before. Thus, I managed to simplify the gas service quite significantly, and modify the adapters accordingly.

This also means we could remove the `UpdateGasPrice` method and `Gas` category.

Closes https://github.com/centrifuge/protocol-v3/issues/152


src/common/AxelarAdapter.sol
```diff
@@ -88,10 +53,9 @@ contract AxelarAdapter is Auth, IAxelarAdapter {
     }
 
     /// @inheritdoc IAdapter
-    /// @dev Currently the payload (message) is not taken into consideration during cost estimation
-    ///      A predefined `axelarCost` value is used.
-    function estimate(uint32, /*chainId*/ bytes calldata, uint256 baseCost) public view returns (uint256) {
-        return baseCost + axelarCost;
+    function estimate(uint32, /*chainId*/ bytes calldata payload, uint256 gasLimit) public view returns (uint256) {
+        return
+            axelarGasService.estimateGasFee(CENTRIFUGE_ID, CENTRIFUGE_AXELAR_EXECUTABLE, payload, gasLimit, bytes(""));
```
hieronx: Note: still needs to be updated for multiple chain IDs, in a future PR.

src/common/AxelarAdapter.sol
```diff
@@ -1,34 +1,10 @@
 // SPDX-License-Identifier: BUSL-1.1
 pragma solidity 0.8.28;
 
-import {IAxelarAdapter, IAdapter} from "src/common/interfaces/IAxelarAdapter.sol";
+import {IAxelarAdapter, IAdapter, IAxelarGateway, IAxelarGasService} from "src/common/interfaces/IAxelarAdapter.sol";
 import {Auth} from "src/misc/Auth.sol";
 import {IMessageHandler} from "src/common/interfaces/IMessageHandler.sol";
 
-// From https://github.com/axelarnetwork/axelar-cgp-solidity/blob/main/contracts/interfaces/IAxelarGateway.sol
```
hieronx, line 8: No reason for these to not be in the interface file

script/vaults/Axelar.s.sol
```diff
@@ -6,7 +6,7 @@ import {ISafe} from "src/common/interfaces/IGuardian.sol";
 
 import {Deployer} from "script/vaults/Deployer.sol";
 
-// Script to deploy Liquidity Pools with an Axelar Adapter.
+// Script to deploy Vaults with an Axelar Adapter.
```
hieronx, line 9: Unrelated cleanups

script/vaults/Wormhole.s.sol
```diff
@@ -0,0 +1,28 @@
+// SPDX-License-Identifier: BUSL-1.1
+pragma solidity 0.8.28;
+
+import {WormholeAdapter} from "src/common/WormholeAdapter.sol";
+import {ISafe} from "src/common/interfaces/IGuardian.sol";
+
+import {Deployer} from "script/vaults/Deployer.sol";
+
+// Script to deploy Vaults with an Wormhole Adapter.
+contract WormholeScript is Deployer {
```
hieronx, line 10: We should just build a single deploy script that supports deploying multiple adapters. I can do that in a future PR.

src/common/AxelarAdapter.sol
lemunozm: Agree!

src/common/GasService.sol
```diff
@@ -2,85 +2,43 @@
...
-
-    /// @inheritdoc IGasService
-    function updateTokenPrice(uint256 value) external auth {
-        tokenPrice = value;
-        emit UpdateTokenPrice(value);
-    }
-
     /// --- Estimations ---
     /// @inheritdoc IGasService
     function estimate(uint32, /*chainId*/ bytes calldata payload) public view returns (uint256) {
```
lemunozm, line 35: Right now we should iterate at least the payload and sum up all messages as proof or as a normal message. I can do this in a follow-up PR

src/common/Gateway.sol
```diff
@@ -297,13 +297,13 @@ contract Gateway is Auth, IGateway, IRecoverable {
...
 
-                uint256 consumed = currentAdapter.estimate(chainId, payload, isPrimaryAdapter ? messageCost : proofCost);
+                uint256 consumed = currentAdapter.estimate(chainId, payload, isPrimaryAdapter ? messageGasLimit : proofGasLimit);
 
                 require(consumed <= tank, "Gateway/not-enough-gas-funds");
                 tank -= consumed;
 
                 currentAdapter.pay{value: consumed}(chainId, payload, address(this));
-                currentAdapter.send(chainId, payload);
+                currentAdapter.send(chainId, payload, isPrimaryAdapter ? messageGasLimit : proofGasLimit);
```
lemunozm: Should not be `consumed` instead to take into account the `tokenPrice` from the `GasService`?
```suggestion
                currentAdapter.send(chainId, payload, consumed);
```

src/common/interfaces/IWormholeAdapter.sol
```diff
@@ -0,0 +1,97 @@
...
+    // --- Incoming ---
+    /// @notice Execute a message
+    /// @dev    Relies on Wormhole to ensure messages cannot be executed more than once.
+    function receiveWormholeMessages(
+        bytes memory payload,
+        bytes[] memory, /* additionalVaas */
+        bytes32 sourceAddress,
+        uint16 sourceChain,
+        bytes32 /* deliveryHash */
+    ) external;
```
lemunozm: NIT, could we copy: https://github.com/wormhole-foundation/wormhole/blob/main/relayer/ethereum/contracts/interfaces/relayer/IWormholeReceiver.sol into this file and make `IWormholeAdapter` inherit from the above?

Just to track any possible change in the interface better and the fact that this method is actually handled by another project

src/misc/libraries/CastLib.sol
```diff
@@ -6,6 +6,15 @@ import {BytesLib} from "src/misc/libraries/BytesLib.sol";
 
 /// @title  CastLib
 library CastLib {
+    function toAddress(bytes32 addr) internal pure returns (address) {
+        require(bytes12(addr) == 0, "First 12 bytes should be zero");
+        return address(uint160(uint256(addr)));
+    }
```
lemunozm: NIT. Could we express that the address should be left padded also in the name? i.e: `toAddressLeftPadded()` or similar. 

I say this because our messaging is right padded IIRC, to avoid using this when it's not the case.

src/common/AxelarAdapter.sol
```diff
@@ -57,12 +65,4 @@ contract AxelarAdapter is Auth, IAxelarAdapter {
...
             axelarGasService.estimateGasFee(CENTRIFUGE_ID, CENTRIFUGE_AXELAR_EXECUTABLE, payload, gasLimit, bytes(""));
     }
-
-    /// @inheritdoc IAdapter
-    function pay(uint32, /*chainId*/ bytes calldata payload, address refund) public payable {
-        //TODO (chainName, chainExecutable) = chainConfig[chainId];
-        axelarGasService.payNativeGasForContractCall{value: msg.value}(
-            address(this), CENTRIFUGE_ID, CENTRIFUGE_AXELAR_EXECUTABLE, payload, refund
-        );
-    }
```
lemunozm: 🚀

lemunozm (commented):
LGTM! Good to see those simplifications 🚀. Thanks for help us supporting this!!

Just one detail to verify.

src/misc/libraries/CastLib.sol
hieronx: Sure

src/common/interfaces/IWormholeAdapter.sol
hieronx: Good idea!

src/common/Gateway.sol
hieronx: Already fixed!

lemunozm: I do not see it 👀

hieronx: Ah I misread this comment. But no, this is not the consumed/eth gas balance, this is the actual gas limit. That is what Wormhole expects.

src/common/AxelarAdapter.sol
```diff
@@ -1,68 +1,94 @@
...
     {
-        //TODO (chainName, chainExecutable) = chainConfig[chainId];
+        require(msg.sender == address(gateway), NotGateway());
+        AxelarDestination memory destination = destinations[centrifugeId];
+        require(bytes(destination.axelarId).length != 0, UnknownChainId());
 
+        string memory destinationAddress = destination.addr.toString();
         axelarGasService.payNativeGasForContractCall{value: msg.value}(
-            address(this), CENTRIFUGE_ID, CENTRIFUGE_AXELAR_EXECUTABLE, payload, refund
+            address(this), destination.axelarId, destination.addr.toString(), payload, refund
```
lemunozm: NIT. You can reuse here `destinationAddress` instead of `destination.addr.toString()`.

src/common/interfaces/IAxelarAdapter.sol
```diff
@@ -37,21 +37,53 @@ interface IAxelarGasService {
...
 }
+
+struct AxelarSource {
+    uint32 centrifugeId;
+    address addr;
+}
+
+struct AxelarDestination {
+    string axelarId;
+    address addr;
```
lemunozm, line 78: Question. Should we store in `AxelarDestination` the `addr` as `address`? Or should it be a `string` if Axelar expects a string? Maybe some chains expect here other thing different than `address` and we can only encode the correct one if it's a `string`.

I think the same applies to `AxelarSource.addr`

lemunozm (commented):
A couple of minor things!

src/common/interfaces/IAxelarAdapter.sol
hieronx: I thought about this too. The current adapter is anyways specific to EVM since `payNativeGasForContractCall` is EVM specific afaik. The Wormhole adapter likewise is also EVM specific. Changing these to both be non-EVM specific will add quite a bit of complexity that I don't think is currently worth it, especially because swapping adapters is quite easy.

So I would keep them as EVM specific, and thus also store addresses in `address` type. This has the benefit of being easier to `file` and read from state.

lemunozm: Ok! Thanks for your digested thought!

lemunozm (approved):
Let's merge it!!

PR #155, opened 2025-03-17
Fix transient valuation issue by check reentrancy

### LINKED ISSUE #145: Security: Avoid reentrancy issues in TransientValuation
A third party contract can call to `TransientValuation` to modify a price freely. We need to protect that.

- Should `setPrice() be `auth` and called by the `PoolManager`?
- Is there any option that allows not involving the PoolManager?
- Does it make sense to have `setPrices(listOfQuotes)` that could be only set once per transaction?

lemunozm:
Fixes: https://github.com/centrifuge/protocol-v3/issues/145


onnovisser (approved):
LGTM 💯

PR #157, opened 2025-03-17
Simplify PoolRouter calls

lemunozm:
If we add a special multicall implementation where we redirect to the `PoolManager`, we can save tons of lines.

As far as the PoolRouter method are just direct calls to PoolManager, this works.


src/pools/PoolRouter.sol
```diff
@@ -37,7 +37,23 @@ contract PoolRouter is Multicall, IPoolRouter {
...
         }
 
-        super.multicall(data);
+        if (poolManager.unlockedPoolId().isNull()) {
+            super.multicall(data);
+        } else {
+            // Calling from execute will use poolManager as target
+            uint256 totalBytes = data.length;
+            for (uint256 i; i < totalBytes; ++i) {
+                (bool success, bytes memory returnData) = address(poolManager).call(data[i]);
```
lemunozm, line 46: The difference is in this line, `target` is `poolManager` if we're inside `execute()`, and we use `call()` instead of `delegateCall()`

test/pools/integration/Cases.t.sol
```diff
@@ -101,11 +101,11 @@ contract TestCases is Deployer, Test {
...
-        cs[c++] = abi.encodeWithSelector(poolRouter.setPoolMetadata.selector, bytes("Testing pool"));
-        cs[c++] = abi.encodeWithSelector(poolRouter.addShareClass.selector, SC_NAME, SC_SYMBOL, SC_SALT, bytes(""));
-        cs[c++] = abi.encodeWithSelector(poolRouter.notifyPool.selector, CHAIN_CV);
-        cs[c++] = abi.encodeWithSelector(poolRouter.notifyShareClass.selector, CHAIN_CV, scId, SC_HOOK);
-        cs[c++] = abi.encodeWithSelector(poolRouter.createHolding.selector, scId, USDC_C2, identityValuation, 0x01);
+        cs[c++] = abi.encodeWithSelector(poolManager.setPoolMetadata.selector, bytes("Testing pool"));
+        cs[c++] = abi.encodeWithSelector(poolManager.addShareClass.selector, SC_NAME, SC_SYMBOL, SC_SALT, bytes(""));
+        cs[c++] = abi.encodeWithSelector(poolManager.notifyPool.selector, CHAIN_CV);
+        cs[c++] = abi.encodeWithSelector(poolManager.notifyShareClass.selector, CHAIN_CV, scId, SC_HOOK);
+        cs[c++] = abi.encodeWithSelector(poolManager.createHolding.selector, scId, USDC_C2, identityValuation, 0x01);
```
lemunozm, line 108: Later, only for the methods used in execute, we should specify the `poolManager` selector

lemunozm:
Would really appreciate your solidity feedback here @hieronx

PR #158, opened 2025-03-17
Use internal functions in pool manager

hieronx:


src/pools/PoolRouter.sol
```diff
@@ -14,21 +14,69 @@ import {ShareClassId} from "src/pools/types/ShareClassId.sol";
...
+    IMessageProcessor public sender;
     IGateway public gateway;
 
-    constructor(IPoolManager poolManager_, IGateway gateway_) {
-        poolManager = poolManager_;
+    /// @dev A requirement for methods that needs to be called through `execute()`
+    modifier poolUnlocked() {
+        require(!unlockedPoolId.isNull(), IPoolRouter.PoolLocked());
+        _;
+    }
```
lemunozm: Would it make sense to remove this to avoid using it as a modifier? and just move the content to `_protectedAndUnlocked()`

src/pools/PoolRouter.sol
```diff
@@ -47,170 +95,368 @@ contract PoolRouter is Multicall, IPoolRouter {
...
+
+    //----------------------------------------------------------------------------------------------
+    // Gateway owner methods
+    //----------------------------------------------------------------------------------------------
+
+    /// @inheritdoc IPoolRouterHandler
+    function registerAsset(AssetId assetId, string calldata name, string calldata symbol, uint8 decimals)
+        external payable
+    {
+        _protected();
```
lemunozm: It's important that the FM can not choose the gateway intended methods as candidates for the multicall, or they can break some invariants.

This should be `auth` and then it does not need to be `protected`.

test/pools/integration/Cases.t.sol
```diff
@@ -96,7 +95,7 @@ contract TestCases is Deployer, Test {
         assertEq(decimals, 6);
 
         vm.prank(FM);
-        poolId = poolRouter.createPool(USD, multiShareClass);
+        poolId = poolRouter.createPool(address(this), USD, multiShareClass);
```
lemunozm:
```suggestion
        poolId = poolRouter.createPool(FM, USD, multiShareClass);
```

There are more occurrences

src/pools/PoolRouter.sol
hieronx: Yes actually already di!

src/pools/PoolRouter.sol
```diff
@@ -47,170 +90,359 @@ contract PoolRouter is Multicall, IPoolRouter {
...
+        accounting.addCredit(account, amount);
     }
 
+    //----------------------------------------------------------------------------------------------
+    // Gateway owner methods
+    //----------------------------------------------------------------------------------------------
+
+    /// @inheritdoc IPoolRouterHandler
+    function registerAsset(AssetId assetId, string calldata name, string calldata symbol, uint8 decimals)
+        external payable auth
```
hieronx: I guess these don't need to be `payable` actually, if they are called by gateway?

lemunozm: Correct!

lemunozm (approved):

wischli (approved):
Finally just a single `PoolManager` 🎉  Thanks so much both of you!

PR #159, opened 2025-03-18
Fix main branch

lemunozm:


wischli (dismissed):
Thanks for the quick fix!

test/vaults/unit/InvestmentManager.t.sol
```diff
@@ -37,7 +32,6 @@ contract InvestmentManagerTest is BaseTest {
         // permissions set correctly
         assertEq(investmentManager.wards(address(root)), 1);
         assertEq(investmentManager.wards(address(gateway)), 1);
-        assertEq(investmentManager.wards(nonWard), 0);
```
lemunozm, line 40: Unless I'm missing something, we are testing here that a random `address` from Universe is not already set. However, we can not ensure that one non-tested address is already set when it shouldn't.

hieronx: This was definitely intentional. The goal was that we tested only root and gateway were ward, and nothing else.  Why can this not work?

lemunozm: But `nonWard` is a set of random addresses from the set of all possible addresses. It could be an address that should not be a ward but is not chosen by the fuzzer.

lemunozm: It does not work, because `investmentManager` now requires `poolManager` and `messageProcessor` as wards. But this test is very non-deterministic. I'm not sure if it really checks and ensures what we want.

hieronx: It is deterministic, if you keep the `vm.assume` to make exceptions for the addresses that actually should be wards. So you just need to add
```
vm.assume(
    nonWard != address(poolManager) && nonWard != address(messageProcessor)
);
```

lemunozm: Agree that that's the most accurate solution here (that I'm gonna apply). But this test does not ensure deterministically that if tomorrow, we add a new ward to `investmentManager`, that ward will be chosen by the fuzzer as `nonWard`. 

This is what is happening right now, and this is why it passes initially the Github CI (maybe because the random seed used by the fuzzed differs)

hieronx: Well yeah, fuzzers generate random inputs and indeed are not deterministic from that perspective. But given sufficient fuzz runs, the above test helps to rule out some issues. Especially if we do a big run with >1M fuzz runs before the audit.

You could likely use https://github.com/a16z/halmos to deterministically prove this fwiw, but let's keep this as an experiment for later ;)

wischli (approved):

test/vaults/unit/InvestmentManager.t.sol
lemunozm: Do you know if fuzzer uses some heuristic to the choose random addresses used in inputs? Like already used addresses or so? Because if not, having the correct nonWard is like knowing a private key by brute force.

lemunozm: For reference, it seems it does something like that. When you add `vm.label()` or use some address in `setUp`, those addresses are candidates for the fuzzer. Then it makes sense to do it in this way!

hieronx: I have never looked into the details but it definitely is not fully random, there is some dictionary input.

PR #161, opened 2025-03-18
feat: preview share class id

wischli:
Adds missing preview function for generating `ShareClassId` based on a pool id and `uint8 index`. This can be used by Apps to generate all share class ids via

```solidity
uint8 maxIndex = shareClassCount[poolId];
ShareClassId[] memory shareClasses;

for (uint8 i = 1; i <= maxIndex; i++) {
   shareClasses.push(previewShareClassId(poolId, i));
}
```


src/pools/interfaces/IShareClassManager.sol
```diff
@@ -285,4 +285,11 @@ interface IShareClassManager {
     /// @param poolId Identifier of the pool
     /// @return scId Identifier of the next share class
     function previewNextShareClassId(PoolId poolId) external view returns (ShareClassId scId);
+
+    /// @notice Determines the share class id for the given pool and index.
+    ///
+    /// @param poolId Identifier of the pool
+    /// @param index The pool-internal index of the share class id
+    /// @return scId Identifier of the underlying share class
+    function previewShareClassId(PoolId poolId, uint8 index) external pure returns (ShareClassId scId);
```
lemunozm: If it's pure, it could then be a ShareClassId constructor? `createShareClassId()` or so?

lemunozm: Or it does depend on how the MultiShareClass works that could differ from other implementations?

wischli: No, we could definitely do that. I don't believe the share class id generation logic is necessary to be changeable for future share class impls. But I think it's more intuitive for users do this via the `MultiShareClass` interface.

WDYT about adding a `ShareClassId` constructor and still keeping `previewShareClassId` which internally just calls the constructor?

lemunozm: It feels like the best middle solution!!

lemunozm (dismissed):
Thanks for also adding the test!

src/pools/MultiShareClass.sol
```diff
@@ -82,11 +82,10 @@ contract MultiShareClass is Auth, IMultiShareClass {
     IPoolRegistry public poolRegistry;
     mapping(bytes32 salt => bool) public salts;
     mapping(PoolId poolId => uint32) public epochId;
-    mapping(PoolId poolId => uint32) public shareClassCount;
```
hieronx: Any specific reason we are lowering from `uint32` to `uint8`? Not that I think we are likely to need more than 255 share classes in a pool, but could keep the flexibility just in case ;-)

wischli: It was mainly based on your proposal ([Slack comment](https://kflabs.slack.com/archives/C07PG2EUR9C/p1741942129431949?thread_ts=1741864752.645229&cid=C07PG2EUR9C)) that used `uint8` over `uint32`. Happy to stick with `uint32`!

hieronx: Ha, I didn't intentionally put `uint8` in that code example, so let's stick to `uint32` then!

onnovisser:
In order for Apps to have a consistent interface across ShareClassManagers, would it make sense to add `function shareClassCount(PoolId poolId) external view returns (uint32);` (and perhaps other properties) to the IShareClassManager interface?

lemunozm:
I think Apps should point directly to the contract (?), the interface is just for other contract interaction

onnovisser:
I mean doing that mostly to ensure that other ShareClassManager implementations use the same variable names

onnovisser:
If a potential TranchedShareClassManager uses `shareClassCounter` instead of `shareClassCount` for the variable name, then that would be quite painful for Apps

wischli:
> If a potential TranchedShareClassManager uses `shareClassCounter` instead of `shareClassCount` for the variable name, then that would be quite painful for Apps

Done in [e8d13e0](https://github.com/centrifuge/protocol-v3/pull/161/commits/e8d13e0e9e6ce1581ad4a562e0a8f39413e7475f)

src/pools/MultiShareClass.sol
wischli: Done in [24198fb](https://github.com/centrifuge/protocol-v3/pull/161/commits/24198fb698bbd90fefc79c9571e60cfddb106012)

lemunozm (approved):

Issue #162, opened 2025-03-18
[EPIC] Clean-ups before V3 release


hieronx:
- [x] Rename `TrancheToken` to `CentrifugeToken`
- [x] Rename `RestrictedRedemptions` to `FreelyTransferable` and `RestrictionManager` to `RestrictedTransfers`
- [x] Rename trancheId to `scId` / `shareClassId` in CV
  - Except in the vault contract
- [x] Update any remaining references of liquidity pools to vaults
- [x] Update README
- [x] Standardize section header style
- [x] Ensure all contracts follow the style guide
- [x] https://github.com/centrifuge/protocol-v3/issues/127
- [ ] https://github.com/centrifuge/protocol-v3/issues/126
- [x] https://github.com/centrifuge/protocol-v3/issues/156
- [x] https://github.com/centrifuge/protocol-v3/issues/115
- [x] Move `IMultiShareClassManager` events to `IShareClassManager` for indexing

lemunozm:
- [x] Use common/types (`PoolId, ShareCLassId, AssetId`) in Vaults
- [x] (Important) https://github.com/centrifuge/protocol-v3/issues/165
- [x] Upgrade https://github.com/marktoda/forge-gas-snapshot dependency which is archived.
- [x] Check all TODOs in the code
- [x] Correct recovery system through the guardian: https://kflabs.slack.com/archives/C07PG2EUR9C/p1742289360020569?thread_ts=1742288924.901019&cid=C07PG2EUR9C 
- [x] Add a receiver for `TransferShares` message in `vaults/PoolManager`.
- [x] There are some missing messages that are still not supported in CP.
- [x] String errors as typed errors in vaults contracts
- [ ] Update DOCS and all diagrams
  - During audit
- [x] Fix Gas estimation, which now only evaluate the first message in the batch
- [x] Evaluate if we should remove Message Categories
- [x] `CancelDepositRequest` message should only need one cancelled amount.
- [x] Add "Rely on Root" section for `PoolsDeployer`
- [x] Avoid fuzzing PoolId in vault tests and use a POOL_A constant to user everywhere
- [x] Check the order of method parameters are accorded to a basic schema.
  -i.e. `PoolManager.transferShares` has `(poolId, scId, chainId)` when all methods follows a schema like `(chainId, poolId, scId)`
- [x] Add recoverTokens to PoolRouter
- [x] Go through all Manuel comments
- [x] Add pools & common contracts to smt.model_checker ?
[...]

PR #163, opened 2025-03-18
Feat: Add updateContract methods to CP

lemunozm:
### Changelist
- Add generic `updateContract()`
- Add `deployVault()` as a way to create a vault deploying it.
- Add `addVault()` as a way to link a vault.
- Add `removeVault()` as a way to unlink a vault.
- Modify `address` in the message to `bytes32` to be able to support other non-EVM chains.


src/pools/PoolRouter.sol
```diff
@@ -256,6 +249,68 @@ contract PoolRouter is Auth, Multicall, IPoolRouter, IPoolRouterHandler {
...
+            MessageLib.UpdateContractVaultUpdate memory m = MessageLib.deserializeUpdateContractVaultUpdate(payload);
+            if (m.isLinked) {
+                deployVault(scId, AssetId.wrap(m.assetId), target, m.factory, m.vault);
+            }
+            else {
+                removeVault(scId, AssetId.wrap(m.assetId), target, m.vault);
+            }
+        }
+        else {
+            sender.sendUpdateContract(chainId, unlockedPoolId, scId, target, payload);
```
lemunozm: Concern 1. This means that an FM can send any kind of `updateContract()` without any check on the CP side. We could break some invariants with some future messages unknown upfront

src/pools/PoolRouter.sol
```diff
@@ -256,6 +249,68 @@ contract PoolRouter is Auth, Multicall, IPoolRouter, IPoolRouterHandler {
...
+                factory: factory,
+                assetId: assetId.raw(),
+                isLinked: true,
+                vault: vault
+            }).serialize()
+        );
+    }
+
+    /// @inheritdoc IPoolRouter
+    function removeVault(ShareClassId scId, AssetId assetId, bytes32 target, bytes32 vault)
```
lemunozm: Concern 2. We remove a vault without knowing if the vault exists or is linked in the CV side. This message can fail. 

Should we add some storage in CP to handle this? Like `hasVault`? If yes, where to add it?

lemunozm (commented):
A couple of concerns @mustermeiszer

src/pools/PoolRouter.sol
lemunozm: Concern 3. We can send an `updateContract` message to a chain that differs from `assetId.chainId()`, but we are not able to check this.

src/pools/PoolRouter.sol
lemunozm: > where to add it?

I would propose having some generic storage in PoolRegistry for this things. `mapping(PoolId => bytes32 key => uint256 value)` and in this case, generate the key as `hash(scId, assetId)` and storing a bool.

src/pools/PoolRouter.sol
mustermeiszer: > Concern 1. This means that an FM can send any kind of updateContract() without any check on the CP side. We could break some invariants with some future messages unknown upfront

Given that they can only break their own pool it is fine.

mustermeiszer: > Concern 3. We can send an updateContract message to a chain that differs from assetId.chainId(), but we are not able to check this.

Given that they can only break their own pool it is fine.

src/pools/PoolRouter.sol
mustermeiszer: > Should we add some storage in CP to handle this? Like hasVault? If yes, where to add it?

No, again it is their own pool.

mustermeiszer: @hieronx if you think this is critical. I don't think so.

src/pools/PoolRouter.sol
wischli: I am curious why we even `sendUpdateContract` in the outer `else` branch here. Right now, this could only be `Invalid`. My take is that we should extend this in the future once we add more update messages.

lemunozm: The thing is that this "payload" is a submessage. And the submessage could be unknown or exist after deploying this contract.

If that's not true, we can just do normal messages, and we would not need this "payload".

src/pools/PoolRouter.sol
lemunozm: Maybe I'm wrong, but I thought actions performed by messages on the receiver side must not fail, and if they fail for an unknow issue, they should be recoverable (?). 

If the message fails because an invariant has not fulfilled first, then can never be recoverable

mustermeiszer: Not sure I am following here. What do you mean by recoverable?

lemunozm: Dispatching again from the Adapter. Similar to what we did in Centrifuge Chain when we processed failed messages. If what I say no longer applies here, then let's go with the current state. 

Although not sure how handy it is failing asynchronously as an action 🤔. 

i.e. The UI will show to the FM, that the vault has correctly removed by him in CP side, but it's not because the fund manager was wrong with the vault address. CV fails without warning and the FM thinks the state has correctly changed when not, and the vault is still used.

mustermeiszer: > Dispatching again from the Adapter. Similar to what we did in Centrifuge Chain when we processed failed messages. If what I say no longer applies here, then let's go with the current state.

We definitely need that. But those messages would just be stuck then. I am fine with that.

mustermeiszer: > i.e. The UI will show to the FM, that the vault has correctly removed by him in CP side, but it's not because the fund manager was wrong with the vault address. CV fails without warning and the FM thinks the state has correctly changed when not, and the vault is still used.

UI should definitely check the CV state to ensure a vault has been unlinked. THere is and should not be any storage on the CP side to check that.

lemunozm: > UI should definitely check the CV state to ensure a vault has been unlinked

Ok, we should then draw something like "pending to confirmation" in the UI.

Ok, if all concerned are assumed, then if you agree with the PR, it should be ready

src/common/libraries/MessageLib.sol
```diff
@@ -628,10 +628,10 @@ library MessageLib {
...
 
         return UpdateContractVaultUpdate({
-            factory: data.toAddress(1),
-            assetId: data.toUint128(21),
-            isLinked: data.toBool(37),
-            vault: data.toAddress(38)
+            factory: data.toBytes32(1),
+            assetId: data.toUint128(33),
+            isLinked: data.toBool(49),
+            vault: data.toBytes32(50)
```
lemunozm: Optional idea. I think the message can be simplified to:

- `uint128 assetId`
- `bytes32 vaultOrFactory`
- `bytes1 action` enum with 3 states (DEPLOY|LINK|UNLINK)

if DEPLOY `vaultOrFactory` will be `factory`
if LINK|UNLINK `vaultOfFactory` will be `vault`

Saving 32 bytes in the process and simplifying some casuistics processing the message

src/pools/PoolRouter.sol
hieronx: I agree with Frederik, in this particular case it seems the UI should ensure this, and if the message fails asynchronously that is acceptable.

src/vaults/PoolManager.sol
```diff
@@ -275,15 +275,16 @@ contract PoolManager is Auth, IPoolManager, IUpdateContract {
     function update(uint64 poolId, bytes16 trancheId, bytes memory payload) public auth {
         MessageLib.UpdateContractVaultUpdate memory m = MessageLib.deserializeUpdateContractVaultUpdate(payload);
 
-        address vault = m.vault;
-        if (m.factory != address(0) && vault == address(0)) {
-            require(vaultFactory[m.factory], "PoolManager/invalid-vault-factory");
-            vault = deployVault(poolId, trancheId, idToAsset[m.assetId], m.factory);
```
lemunozm: BTW, I've removed this `require ` because it also exists in `deployVault()`

src/common/libraries/MessageLib.sol
wischli: After writing tests for this message, I am in favor of your proposal @lemunozm because right now there are possible permutations which don't make sense, for instance configuring `linkVault: false` (which leads to removing a vault) despite wanting to deploy a new vault.

mustermeiszer: But in order to deploy and link you need a lot more bytes...

mustermeiszer: deploy + unlink will fail as the `investmentManager` will not recognize the vault to be unlinked.

lemunozm: > But in order to deploy and link you need a lot more bytes...

Could you extend? For deploying and linking, only the factory address is needed because the factory already gives you the vault address in CV side.

mustermeiszer: You need 2 messages instead of one.

mustermeiszer: deploy and link can be combined right now. With the enum it can't

lemunozm: Looking at the code, each time you deploy, you also link, you can not deploy and not link 🤔

When I say above DEPLOY, I mean DEPLOY_AND_LINK.

mustermeiszer: As long as we can DEPLOY_AND_LINK all good

lemunozm: Ok, will do it anyway in a follow-up PR to not add more things here.

This, once I rebase, should be ready then

src/pools/PoolRouter.sol
```diff
@@ -257,6 +250,91 @@ contract PoolRouter is Auth, Multicall, IPoolRouter, IPoolRouterGatewayHandler {
         decreaseHolding(scId, payoutAssetId, valuation, payoutAssetAmount);
     }
 
+    /// @inheritdoc IPoolRouter
+    function updateContract(uint16 chainId, ShareClassId scId, bytes32 target, bytes calldata payload) external payable {
+        _protectedAndUnlocked();
+
+        if (payload.updateContractType() == UpdateContractType.VaultUpdate) {
```
mustermeiszer: NIT

I think that can be removed. Why do the double check here? If somebody uses the generic part, let them, right?

mustermeiszer (dismissed):
Just a nit, but nothing blocking.

src/pools/PoolRouter.sol
lemunozm: My thought was that we want to allow checking if holdings already exist when we're making a Vault update. If not, we would allow deposit requests without having the holdings created, which before was a strong precondition.

Has this precondition changed?

src/common/libraries/MessageLib.sol
lemunozm: Finally, did in this PR: [f102bbe](https://github.com/centrifuge/protocol-v3/pull/163/commits/f102bbee3a4e6b09b2045b986a16eba52961c385)

src/pools/PoolRouter.sol
mustermeiszer: I see your point, but I think breaking their own pool is fine for now.

src/pools/PoolRouter.sol
```diff
@@ -257,6 +250,72 @@ contract PoolRouter is Auth, Multicall, IPoolRouter, IPoolRouterGatewayHandler {
...
+            unlockedPoolId,
+            scId,
+            target,
+            MessageLib.UpdateContractVaultUpdate({
+                vaultOrFactory: vault,
+                assetId: assetId.raw(),
+                kind: uint8(VaultUpdateKind.Unlink)
+            }).serialize()
+        );
+    }
```
lemunozm: After simplify the message, we can merge these 3 methods into one if we add a  `VaultUpdateKind` parameter. I can do it if you agree @mustermeiszer. Less code duplication for sure

src/vaults/PoolManager.sol
```diff
@@ -281,20 +281,25 @@ contract PoolManager is Auth, IPoolManager, IUpdateContract, IPoolManagerGateway
...
+            // Needed as safeguard against non-validated vaults
+            // I.e. we only accept vaults that have been deployed by the pool manager
+            require(_vaultDetails[vault].asset != address(0), "PoolManager/unknown-vault");
+
+            if (m.kind == uint8(VaultUpdateKind.Link)) {
+                linkVault(poolId, trancheId, m.assetId, vault);
+            } else if (m.kind == uint8(VaultUpdateKind.Unlink)) {
+                unlinkVault(poolId, trancheId, m.assetId, vault);
+            } else {
+                revert("PoolManager/malformed-vault-update-msg");
```
wischli, line 301: Nit: Missing coverage. But we can do this as part of #162 in which we should aim for ~99% test coverage

wischli (approved):
In love with the simplicity! Great work and thanks for your patience 🚀

src/vaults/PoolManager.sol
lemunozm: Correct! Let's tackle it in the EPIC then. Thanks for being so precise!

PR #166, opened 2025-03-19
Add CFG token

hieronx:


hieronx:
Moved to https://github.com/centrifuge/cfg-token

PR #168, opened 2025-03-19
Write deployment output in script

hieronx:
Generates JSON file with output from the deployment. Stores it in `deployments/latest`, which is added to `.gitignore` so not every testnet deployment bloats the git repo. Instead, a deployer should move the file into `deployments/testnet` or `deployments/mainnet` (which we likely will automate later in CI).


script/CommonDeployer.s.sol
```diff
@@ -86,4 +98,58 @@ contract CommonDeployer is Script {
         gateway.deny(deployer);
         messageProcessor.deny(deployer);
     }
+
+    function register(string memory name, address target) public {
+        deploymentOutput =
+            string(abi.encodePacked(deploymentOutput, '  "', name, '": "0x', _toAsciiString(target), '",\n'));
```
hieronx: Technically on the last registered line, the `,` comma should be removed. I couldn't come up with an easy way to accomplish this, so for now just leaving it as is. The deployer can easily fix this when moving the file.

script/VaultsDeployer.s.sol
```diff
@@ -46,11 +46,24 @@ contract VaultsDeployer is CommonDeployer {
         poolManager = new PoolManager(address(escrow), trancheFactory, vaultFactories);
         router = new VaultRouter(address(routerEscrow), address(gateway), address(poolManager));
 
+        _vaultRegister();
```
lemunozm:
```suggestion
        _vaultsRegister();
```
Just to be aligned with the rest of the names

script/adapters/Localhost.s.sol
```diff
@@ -13,12 +13,11 @@ contract LocalhostAdapterScript is FullDeployer {
         vm.startBroadcast();
 
         deployFull(ISafe(vm.envAddress("ADMIN")), msg.sender);
+        saveDeploymentOutput();
```
lemunozm, line 16: I think, IIUC all calls to `deployFull` should also be followed by `saveDeploymentOutput`. Could then move `saveDeploymentOutput` as part of `deployFull`?

script/CommonDeployer.s.sol
```diff
@@ -86,4 +98,58 @@ contract CommonDeployer is Script {
         gateway.deny(deployer);
         messageProcessor.deny(deployer);
     }
+
+    function register(string memory name, address target) public {
```
lemunozm, line 105: NIT. WDYT about moving `register/saveDeploymentOutput` and their utils as part of a parent Script that `CommonDeployer` could inherit from?

Just to split concerns, but not strong opinion

script/CommonDeployer.s.sol
lemunozm: I saw previously other solutions that start with "," only if there is already some content in the output. Something like:

```solidity
deploymentOutput = (deploymentOutput.length == 0)
   ? string(abi.encodePacked(deploymentOutput, '  "', name, '": "0x', _toAsciiString(target)))
   : string(abi.encodePacked('",\n', deploymentOutput, '  "', name, '": "0x', _toAsciiString(target)));
```

deployments/testnet/sepolia.json
```diff
@@ -0,0 +1,25 @@
+{
```
lemunozm, line 1: Question. Does this file be added to the `.gitignore` too? Why `latest` is added but `testnet` no?

.gitignore
```diff
@@ -7,6 +7,7 @@ broadcast/
 .env.*
 lcov.info
 .vscode
+deployments/latest
```
wischli, line 10: Same question as Luis: Don't we want the opposite? I.e. anything but the latest deployments, including testnets.

script/CommonDeployer.s.sol
```diff
@@ -86,4 +98,58 @@ contract CommonDeployer is Script {
...
+        }
+        return string(s);
+    }
+
+    function _char(bytes1 b) internal pure returns (bytes1 c) {
+        if (uint8(b) < 10) return bytes1(uint8(b) + 0x30);
+        else return bytes1(uint8(b) + 0x57);
+    }
+
+    function _uint2str(uint256 _i) internal pure returns (string memory _uintAsString) {
```
wischli: Ultra nit: Use `To` instead of `2`
```suggestion
    function _uintToStr(uint256 _i) internal pure returns (string memory _uintAsString) {
```

Or just `_toString()`

lemunozm: +1 for `_toString()` Would it make sense to add this to a library? Maybe CastLib.sol`?

script/adapters/Localhost.s.sol
hieronx: But then if you do a separate deployment using `deployPools` or `deployVaults`, you would want it to be run separately right?

script/CommonDeployer.s.sol
hieronx: Fine with me! Wdyt if we merge it and you make this change, in the way you think it should be, in your branch?

deployments/testnet/sepolia.json
hieronx: @lemunozm @wischli We want live deployments for mainnet and testnet to be in the repo. https://github.com/centrifuge/liquidity-pools/tree/main/deployments like it is here. 

Why `latest` but not `testnet`, see PR description!

script/CommonDeployer.s.sol
hieronx: I would consider the scripts similar to test code, we don't want this to be in `src` and subject to every full audit. So I wouldn't put it into a library in `src/misc/libraries`. I guess it could be in a library in `script/`. Personally I think it's fine as is but happy if you want to move it!

script/CommonDeployer.s.sol
lemunozm: Fine for me too 👍🏻

script/CommonDeployer.s.sol
hieronx: Great idea!

script/adapters/Localhost.s.sol
lemunozm: Do we want to have a separate deployment in testnets? AFAIU, the deployment in testnets/production should be at least a `FullDeployment`. Only for non-testnet testing we need those pools/vaults deployments, but for those tests, we do not need the json.

deployments/testnet/sepolia.json
lemunozm: Sorry! I fully missed the PR description 🤦🏻. Now I understand!

script/CommonDeployer.s.sol
lemunozm: I understand! Then it's fine as it is by now until we need to reuse it 👍🏻

script/adapters/Localhost.s.sol
hieronx: The plan so far was to deploy CP only on 1 chain, and CV on 4 chains, for the launch. We can potentially just do a full deployment and only enable pool registration on 1 chain though. Let's discuss more on Slack.

wischli (dismissed):
Solid as a rock! Apologies for skipping over the PR description too quickly. Makes sense to me!

script/adapters/Localhost.s.sol
lemunozm: Interesting, I didn't know that was the idea. I think the deployments still work as they are, except we need a better way to combine adapters + kinds of deployments

lemunozm (dismissed):
Let's merge this! 

WARNING: Please modify the branch to `main` first, given the base PR has been already merged

wischli (approved):

PR #169, opened 2025-03-20
feat: support registerAsset in VaultRouter

wischli:
Adds `registerAsset` call to `VaultRouter`


hieronx:
Good catch!

Actually just thinking, wonder if we should just make `transferTrancheTokens` and `registerAsset` `auth`, and only allow these to be called through the `VaultRouter`. This would prevent users calling these without paying gas. And since these are non standardized interfaces, it seems there is no downside to this.

wischli:
> Good catch!
> 
> Actually just thinking, wonder if we should just make `transferTrancheTokens` and `registerAsset` `auth`, and only allow these to be called through the `VaultRouter`. This would prevent users calling these without paying gas. And since these are non standardized interfaces, it seems there is no downside to this.

Agree. Will add to this PR

lemunozm:
I think we can also remove the `setPayableSource()` call inside `PoolManager.registerAsset`, right? Because this message will always be called through the `VaultsRouter`.  The same for `transferTrancheTokens`. 

This would mean that we can also remove the `gateway` dependency from the `PoolManager`

hieronx:
> I think we can also remove the setPayableSource() call inside PoolManager.registerAsset, right

Yeah true!

script/VaultsDeployer.s.sol
```diff
@@ -106,6 +106,7 @@ contract VaultsDeployer is CommonDeployer {
 
         // Rely on VaultRouter
         gateway.rely(address(vaultRouter));
+        poolManager.rely(address(vaultRouter));
```
wischli, line 109: Just want to make sure this represents a desired missed authorization. Otherwise, `vaultRouter` cannot call authorized PM calls.

script/VaultsDeployer.s.sol
```diff
@@ -95,7 +95,6 @@ contract VaultsDeployer is CommonDeployer {
 
         // Rely on vaultGateway
         investmentManager.rely(address(gateway));
-        poolManager.rely(address(gateway));
```
lemunozm: Is the other way around. PoolManager still need to be called by the gateway

lemunozm (approved):
Thanks for doing this and the speed on doing it!

PR #171, opened 2025-03-20
Use internal centrifuge chain id

lemunozm:
## Description

- No longer use `block.chainId` to identify things that can be sent cross-chain
- Use `MessageProcessor` as the single source of truth for storing the configured `centrifugeChainId`.

Probably later, the adapters should also filter based on that `centrifugeChainId` to discard messages that are not for them.


mustermeiszer (commented):
I am struggling to figure out how Axelar will work like this. 

The axelar adapter needs a way to 
* map the incoming axelar chain id to our centrifuge chain id, right?
* map outgoing from centrifuge chain id to axelar chain id 

or am I missing something?

src/common/MessageProcessor.sol
```diff
@@ -48,7 +48,16 @@ contract MessageProcessor is Auth, IMessageProcessor {
     IPoolManagerGatewayHandler public poolManager;
     IInvestmentManagerGatewayHandler public investmentManager;
 
-    constructor(IMessageSender gateway_, IRoot root_, IGasService gasService_, address deployer) Auth(deployer) {
+    uint16 public centrifugeChainId;
+
+    constructor(
+        uint16 centridugeChainId_,
```
wischli: Typo
```suggestion
        uint16 centrifugeChainId_,
```

lemunozm: Such an eagle eye!

lemunozm:
>I am struggling to figure out how Axelar will work like this. 
>
>The axelar adapter needs a way to 
>* map the incoming axelar chain id to our centrifuge chain id, right?
>* map outgoing from centrifuge chain id to axelar chain id 
>
>or am I missing something?

It's exactly as you say. The adapters are already prepared to be configured with those maps.

lemunozm:
Although I would adapt them to use `uint16`

src/common/interfaces/IWormholeAdapter.sol
```diff
@@ -122,13 +122,13 @@ struct WormholeDestination {
 }
 
 interface IWormholeAdapter is IAdapter, IWormholeReceiver {
-    event File(bytes32 indexed what, uint16 wormholeId, uint32 centrifugeId, address source);
-    event File(bytes32 indexed what, uint32 centrifugeId, uint16 wormholeId, address destination);
+    event FileSource(bytes32 indexed what, uint16 wormholeId, uint16 centrifugeChainId, address source);
+    event FileDestination(bytes32 indexed what, uint16 centrifugeChainId, uint16 wormholeId, address destination);
```
lemunozm, line 126: @hieronx I've needed to modify the event because the parameters are now identical, and we want to distinguish if we're configuring the `source` or the `destination`.

I think this is not an issue as long as we filter the event by prefix name "File", WDYT? Also, filtering by prefix would allow for a better naming for some file events that are difficult to interpret.

mustermeiszer (approved):
LGTM! 
We should keep a list of `EVM chain id` to `CentrifugeChainId` somewhere, wdyt? Any ideas how?

lemunozm:
I think that should be something offchain like [axelar](https://axelarscan.io/resources/chains) or [wormhole](https://wormhole.com/docs/build/reference/chain-ids/) does

PR #173, opened 2025-03-21
feat: BSM UpdateHolding on CP-side

onnovisser:
draft PR for some early feedback whether this is going in the right direction

### DIFF
--- a/src/common/MessageProcessor.sol
+++ b/src/common/MessageProcessor.sol
@@ -519,8 +519,7 @@ contract MessageProcessor is Auth, IMessageProcessor {
                 } else {
                     balanceSheetManager.triggerRevokeShares(m.poolId, m.scId, address(bytes20(m.who)), m.shares);
                 }
-            } 
-            if (kind == MessageType.UpdateHolding) {
+            } else if (kind == MessageType.UpdateHolding) {
                 MessageLib.UpdateHolding memory m = message.deserializeUpdateHolding();
 
                 poolRouter.updateHoldingAmount(
@@ -533,6 +532,13 @@ contract MessageProcessor is Auth, IMessageProcessor {
                     m.debits,
                     m.credits
                 );
+            } else if (kind == MessageType.UpdateJournal) {
+                MessageLib.UpdateJournal memory m = message.deserializeUpdateJournal();
+                poolRouter.updateJournal(
+                    PoolId.wrap(m.poolId), ShareClassId.wrap(m.scId), m.debits, m.credits
+                );
+            } else if (kind == MessageType.UpdateShares) {
+                MessageLib.UpdateShares memory m = message.deserializeUpdateShares();
             } else {
                 revert InvalidMessage(uint8(kind));
             }
--- a/src/common/MessageProcessor.sol
+++ b/src/common/MessageProcessor.sol
@@ -519,6 +519,20 @@ contract MessageProcessor is Auth, IMessageProcessor {
                 } else {
                     balanceSheetManager.triggerRevokeShares(m.poolId, m.scId, address(bytes20(m.who)), m.shares);
                 }
+            } 
+            if (kind == MessageType.UpdateHolding) {
+                MessageLib.UpdateHolding memory m = message.deserializeUpdateHolding();
+
+                poolRouter.updateHoldingAmount(
+                    PoolId.wrap(m.poolId),
+                    ShareClassId.wrap(m.scId),
+                    AssetId.wrap(m.assetId),
+                    m.amount,
+                    m.pricePerUnit,
+                    m.isIncrease,
+                    m.debits,
+                    m.credits
+                );
             } else {
                 revert InvalidMessage(uint8(kind));
             }
--- a/src/common/MessageProcessor.sol
+++ b/src/common/MessageProcessor.sol
@@ -537,8 +537,17 @@ contract MessageProcessor is Auth, IMessageProcessor {
                 poolRouter.updateJournal(
                     PoolId.wrap(m.poolId), m.debits, m.credits
                 );
-            // } else if (kind == MessageType.UpdateShares) {
-            //     MessageLib.UpdateShares memory m = message.deserializeUpdateShares();
+            } else if (kind == MessageType.UpdateShares) {
+                MessageLib.UpdateShares memory m = message.deserializeUpdateShares();
+                if (m.isIssuance) {
+                    poolRouter.increaseShareIssuance(
+                        PoolId.wrap(m.poolId), ShareClassId.wrap(m.scId), m.shares
+                    );
+                } else {
+                    poolRouter.decreaseShareIssuance(
+                        PoolId.wrap(m.poolId), ShareClassId.wrap(m.scId), m.shares
+                    );
+                }
             } else {
                 revert InvalidMessage(uint8(kind));
             }
--- a/src/common/MessageProcessor.sol
+++ b/src/common/MessageProcessor.sol
@@ -535,10 +535,10 @@ contract MessageProcessor is Auth, IMessageProcessor {
             } else if (kind == MessageType.UpdateJournal) {
                 MessageLib.UpdateJournal memory m = message.deserializeUpdateJournal();
                 poolRouter.updateJournal(
-                    PoolId.wrap(m.poolId), ShareClassId.wrap(m.scId), m.debits, m.credits
+                    PoolId.wrap(m.poolId), m.debits, m.credits
                 );
-            } else if (kind == MessageType.UpdateShares) {
-                MessageLib.UpdateShares memory m = message.deserializeUpdateShares();
+            // } else if (kind == MessageType.UpdateShares) {
+            //     MessageLib.UpdateShares memory m = message.deserializeUpdateShares();
             } else {
                 revert InvalidMessage(uint8(kind));
             }
--- a/src/pools/MultiShareClass.sol
+++ b/src/pools/MultiShareClass.sol
@@ -533,6 +533,24 @@ contract MultiShareClass is Auth, IMultiShareClass {
 
 
     /// @inheritdoc IShareClassManager
+    function increaseShareClassIssuance(PoolId poolId, ShareClassId shareClassId_, uint128 amount) external auth {
+        require(exists(poolId, shareClassId_), ShareClassNotFound());
+
+        metrics[shareClassId_].totalIssuance += amount;
+        // TODO: emit event?
+    }
+
+    /// @inheritdoc IShareClassManager
+    function decreaseShareClassIssuance(PoolId poolId, ShareClassId shareClassId_, uint128 amount) external auth {
+        require(exists(poolId, shareClassId_), ShareClassNotFound());
+        require(metrics[shareClassId_].totalIssuance >= amount, "Issuance too low");
+
+        metrics[shareClassId_].totalIssuance -= amount;
+        // TODO: emit event?
+    }
+
+
+    /// @inheritdoc IShareClassManager
     function updateShareClassNav(PoolId poolId, ShareClassId shareClassId_) external view auth returns (uint128, D18) {
         require(exists(poolId, shareClassId_), ShareClassNotFound());
         revert("unsupported");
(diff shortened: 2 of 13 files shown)


src/pools/PoolRouter.sol
```diff
@@ -427,6 +436,51 @@ contract PoolRouter is Auth, Multicall, IPoolRouter, IPoolRouterGatewayHandler {
...
+    function updateHoldingAmount(PoolId poolId, ShareClassId scId, AssetId assetId, uint128 amount, D18 pricePerUnit, bool isIncrease, JournalEntry[] memory debits, JournalEntry[] memory credits)
+        external auth
+    {
+        address poolCurrency = poolRegistry.currency(poolId).addr();
+        transientValuation.setPrice(assetId.addr(), poolCurrency, pricePerUnit);
+        uint128 valueChange = transientValuation.getQuote(amount, assetId.addr(), poolCurrency).toUint128();
+
+        (uint128 debited, uint128 credited) = updateJournal(poolId, scId, debits, credits);
+        uint128 debitValueLeft = valueChange - debited;
+        uint128 creditValueLeft = valueChange - credited;
```
mustermeiszer, line 450: Really like the structure! Great work!! 

NIT

We do not ensure that total debited-credited is actually smaller equal to valueChange. We should check that here and provide a good error. 

OR actually, we should check that at the CV side. ^^

mustermeiszer (commented):
Love the simplicity. Looks good already. Will add checks for debit and credits on the CV side.

lemunozm (commented):
Awesome work!! Super easy to review 🚀

wischli (commented):
Agreeing with the rest: Great simplistic changes! Looking forward to the remainder.

test/pools/integration/Cases.t.sol
```diff
@@ -244,4 +246,69 @@ contract TestCases is PoolsDeployer, Test {
...
+        (JournalEntry[] memory debits, uint256 i) = (new JournalEntry[](3), 0);
+        debits[i++] = JournalEntry(D18.wrap(1000), holdings.accountId(poolId, scId, USDC_C2, uint8(AccountType.ASSET)));
+        debits[i++] = JournalEntry(D18.wrap(250), extraAccountId);
+        debits[i++] = JournalEntry(D18.wrap(130), holdings.accountId(poolId, scId, USDC_C2, uint8(AccountType.EQUITY)));
+
+        (JournalEntry[] memory credits, uint256 j) = (new JournalEntry[](2), 0);
+        credits[j++] = JournalEntry(D18.wrap(1250), holdings.accountId(poolId, scId, USDC_C2, uint8(AccountType.EQUITY)));
+        credits[j++] = JournalEntry(D18.wrap(130), holdings.accountId(poolId, scId, USDC_C2, uint8(AccountType.LOSS)));
+
+        vm.prank(address(poolRouter));
```
lemunozm: Because we're here in IT. I would avoid faking the contract addresses. This test case should work as it would work in a real environment without faking any contract address. I mean, the entire flow should be described and called from real actors (FM, CV or ANY)

In this case, I think you need to be able to call `updateJournal()` from a message sent from the `MockVault.` If it doesn't work because the accounting is not locked, then we need to fix the `poolRouter`

test/pools/integration/Cases.t.sol
```diff
@@ -244,4 +246,69 @@ contract TestCases is PoolsDeployer, Test {
...
+        accounting.unlock(poolId, "TODO");
+
+        vm.prank(address(messageProcessor)); 
+        poolRouter.updateJournal(poolId, debits, credits);
+
+        vm.prank(address(poolRouter));
+        accounting.lock();
+    }
+
+    function testCalUpdateHolding() public {
```
lemunozm, line 272: The same here

src/pools/PoolRouter.sol
```diff
@@ -427,6 +436,51 @@ contract PoolRouter is Auth, Multicall, IPoolRouter, IPoolRouterGatewayHandler {
...
+        (uint128 debited, uint128 credited) = updateJournal(poolId, debits, credits);
+        uint128 debitValueLeft = valueChange - debited;
+        uint128 creditValueLeft = valueChange - credited;
+
+        _updateHoldingWithPartialDebitsAndCredits(poolId, scId, assetId, amount, isIncrease, debitValueLeft, creditValueLeft);
+    }
+
+    /// @inheritdoc IPoolRouterGatewayHandler
+    function updateJournal(PoolId /*poolId*/, JournalEntry[] memory debits, JournalEntry[] memory credits) public auth returns (uint128 debited, uint128 credited) {
+        // TODO: Open accouning for the pool somewhere?
```
mustermeiszer: It should already be open by the gateway, right?

src/common/MessageProcessor.sol
```diff
@@ -519,6 +519,26 @@ contract MessageProcessor is Auth, IMessageProcessor {
...
+                    m.debits,
+                    m.credits
+                );
+            } else if (kind == MessageType.UpdateJournal) {
+                MessageLib.UpdateJournal memory m = message.deserializeUpdateJournal();
+                poolRouter.updateJournal(
+                    PoolId.wrap(m.poolId), m.debits, m.credits
+                );
+            // } else if (kind == MessageType.UpdateShares) {
+            //     MessageLib.UpdateShares memory m = message.deserializeUpdateShares();
```
mustermeiszer: Do we wanna do that one in a follow-up?

mustermeiszer (approved):
Unless we wanna do the remaining todo and the `UpdateShares` here too, we can merge and do a follow up. 

@onnovisser please wait with merging, than I can bring my local changes up first.

test/pools/integration/Cases.t.sol
onnovisser: yes, definitely. I just think there's a couple pieces missing. I think we need the per pool journal id transient storage. Then `updateHoldingAmount` and `updateJournal` can call `accounting.unlock` and don't need to pass the journal id. Then we also need the SetJournalId message to set a journal id in the first place, I suppose. After that I can add a function to the MockVault to test the whole flow

src/common/MessageProcessor.sol
onnovisser: yes, was planning to pick up after this

test/pools/integration/Cases.t.sol
onnovisser: Or for now I can already unlock accounting in those functions and just pass a temp id

onnovisser: @lemunozm pushed

src/pools/PoolRouter.sol
lemunozm: I think only MessageProcessor can open this. But if the accounting can be locked at the end of this method, I would prefer to let this to the own PoolRouter.

lemunozm: Not remember, do we expect messages that leave the accounting unlocked? as `addDebit()` through a message?

test/pools/integration/Cases.t.sol
```diff
@@ -244,4 +246,54 @@ contract TestCases is PoolsDeployer, Test {
 
         poolRouter.multicall{value: GAS}(cs);
     }
+
+    function testCalUpdateJournal() public {
+        (PoolId poolId, ShareClassId scId) = testPoolCreation();
+
+        AccountId extraAccountId = newAccountId(123, uint8(AccountType.ASSET));
+
+        vm.prank(address(poolRouter));
```
lemunozm: Could we avoid `poolRouter` as an actor also here? I think we can just create the account as `poolRouter.createAccount()` from FM

src/pools/PoolRouter.sol
```diff
@@ -427,6 +436,58 @@ contract PoolRouter is Auth, Multicall, IPoolRouter, IPoolRouterGatewayHandler {
...
+        if (isIncrease) {
+            holdings.increase(poolId, scId, assetId, transientValuation, amount);
+            accounting.addDebit(holdings.accountId(poolId, scId, assetId, uint8(debitAccountType)), debitValue);
+            accounting.addCredit(holdings.accountId(poolId, scId, assetId, uint8(creditAccountType)), creditValue);
+        } else {
+            holdings.decrease(poolId, scId, assetId, transientValuation, amount);
+            accounting.addDebit(holdings.accountId(poolId, scId, assetId, uint8(creditAccountType)), debitValue);
+            accounting.addCredit(holdings.accountId(poolId, scId, assetId, uint8(debitAccountType)), creditValue);
+        }
+    }
```
lemunozm: NIT. Could you move them to the bottom of the file to follow the code style?

src/pools/MultiShareClass.sol
```diff
@@ -532,6 +532,24 @@ contract MultiShareClass is Auth, IMultiShareClass {
     }
 
 
+    /// @inheritdoc IShareClassManager
+    function increaseShareClassIssuance(PoolId poolId, ShareClassId shareClassId_, uint128 amount) external auth {
+        require(exists(poolId, shareClassId_), ShareClassNotFound());
+
+        metrics[shareClassId_].totalIssuance += amount;
+        // TODO: emit event?
```
mustermeiszer, line 540: Yes. Definitly!

mustermeiszer (approved):
@wischli would love your eyes on the issuance thing, but we can do that in the BSM main pr, wdyt?

@onnovisser I think it is ready to be merged. Will take care of events on my side.

src/pools/PoolRouter.sol
```diff
@@ -259,25 +264,29 @@ contract PoolRouter is Auth, Multicall, IPoolRouter, IPoolRouterGatewayHandler {
...
         require(assetRegistry.isRegistered(assetId), IAssetRegistry.AssetNotFound());
 
-        AccountId[] memory accounts = new AccountId[](4);
+        AccountId[] memory accounts = new AccountId[](6);
         accounts[0] = newAccountId(prefix, uint8(AccountType.ASSET));
         accounts[1] = newAccountId(prefix, uint8(AccountType.EQUITY));
         accounts[2] = newAccountId(prefix, uint8(AccountType.LOSS));
         accounts[3] = newAccountId(prefix, uint8(AccountType.GAIN));
+        accounts[4] = newAccountId(prefix, uint8(AccountType.EXPENSE));
+        accounts[5] = newAccountId(prefix, uint8(AccountType.LIABILITY));
```
lemunozm, line 280: Am I right to assume that if it is a liability, it only needs EXPENSE and LIABILITY, but it does not use LOSS and GAIN and vice-versa? cc @mustermeiszer @onnovisser

If I am right, maybe could be interesting to add two methods: `createHoldings` & `createLiabilities` or something similar, where each one initializes only the accounts they need (?)

onnovisser: Not sure, but I suppose you might still use GAIN and LOSS on liabilities

onnovisser:
Merged it for now. can pick up other feedback on the main CAL branch

src/pools/MultiShareClass.sol
wischli: @mustermeiszer I'm unsure if we want to use the existing `IssuedShares` event with `navPerShare = 0` or introduce another event. Of course, I would normally prefer introducing a new event. However, then we have 4 instead of 2 events which signal a change in the total issuance, e.g. issued/revoked as well as increased/decreased.

```solidity
    event IssuedShares(
        PoolId indexed poolId,
        ShareClassId indexed scId,
        uint32 indexed epoch,
        D18 navPerShare,
        uint128 nav,
        uint128 issuedShareAmount
    );
 ```

File: src/pools/MultiShareClass.sol [hunk 10]
```diff
@@ -532,6 +532,24 @@ contract MultiShareClass is Auth, IMultiShareClass {
...
+    }
+
+    /// @inheritdoc IShareClassManager
+    function decreaseShareClassIssuance(PoolId poolId, ShareClassId shareClassId_, uint128 amount) external auth {
+        require(exists(poolId, shareClassId_), ShareClassNotFound());
+        require(metrics[shareClassId_].totalIssuance >= amount, "Issuance too low");
```
wischli, line 546: @onnovisser CR: Revert with a custom error, i.e. `RevokeMoreThanIssued` or add a new `IssuanceUnderflow`

src/pools/MultiShareClass.sol
hieronx: I think we will need to track the `nav per share` that they were issued at. This will be needed for the indexing. @mustermeiszer then I think we can index also sync investments on the CP side, together with async investments.

We could do this by adding `navPerShare` to the CV -> CP messages, and then emit here. What do you think?

PR #176, opened 2025-03-24
Feat: local message bypass & end to end testing infrastructure

lemunozm:
This PR adds two main things:
- Local message bypass in `MessageProcessor`. It means, for messages sent to the same chain, it bypasses serialization/deserialization and Gateway
- Set up an infrastructure for end-to-end tests where each test is tested automatically against the same chain and a different chain.

Other minor things:
- `LocalAdapter` is an adapter used only for integration tests that sends its message to another adapter.
- `CommonDeployer` factorized into `JsonRegistry`

NOTE
For CV integration tests, it's important to **never use as chainId for poolId/assetId the same chainId used for sending a message**. If using the same, `MessageProcessor` will bypass those messages and it will crash, given for CV-integration deployment has not Pools configured.


test/integration/EndToEnd.t.sol
```diff
@@ -0,0 +1,93 @@
...
+        PoolId poolId = cp.poolRouter().createPool(FM, deployA.USD(), deployA.multiShareClass());
+
+        (bytes[] memory c, uint256 i) = (new bytes[](1), 0);
+        c[i++] = abi.encodeWithSelector(PoolRouter.notifyPool.selector, cvChainId);
+        assertEq(i, c.length);
+
+        cp.poolRouter().execute{value: GAS}(poolId, c);
+
+        assert(cv.poolManager().pools(poolId.raw()) != 0);
+    }
```
lemunozm, line 92: Example of CP-CV integration test.

`_getDeploys()` retrievers the instances of cp/cv for the same/different chains based on the fuzzed value `sameChain`. So, the same test is performed automatically in both environments.

NOTE. This test differs from others because the test contract is not inherent from the `Deployer`. We can not do this because we want several deployed instances. Instead, `cv` and `cp` contracts should be used to access each environment.

hieronx (commented):
Not approving since I only did cursory review, but everything looks good to me!

script/CommonDeployer.s.sol
```diff
@@ -37,12 +36,12 @@ contract CommonDeployer is Script {
...
     }
 
-    function deployCommon(uint16 centrifugeChainId, ISafe adminSafe_, address deployer) public {
+    function deployCommon(uint16 chainId, ISafe adminSafe_) public {
         if (address(root) != address(0)) {
             return; // Already deployed. Make this method idempotent.
         }
 
-        root = new Root(DELAY, deployer);
+        root = new Root(DELAY, address(this));
```
wischli, line 44: Context for other reviewers: https://kflabs.slack.com/archives/C07PG2EUR9C/p1742466111863379

test/vaults/BaseTest.sol
```diff
@@ -54,22 +54,24 @@ contract BaseTest is VaultsDeployer, GasSnapshot, Test {
...
     uint256 public erc20TokenId = 0;
-    uint128 public defaultAssetId = newAssetId(defaultChainId, 1).raw();
+    uint128 public defaultAssetId = newAssetId(THIS_CHAIN_ID, 1).raw();
     uint128 public defaultPrice = 1 * 10 ** 18;
     uint8 public defaultDecimals = 8;
 
     function setUp() public virtual {
-        vm.chainId(defaultChainId);
+        // We should not use the block ChainID
+        vm.chainId(23);
```
wischli: Nit: Maybe we should define this also as a constant, i.e. `BLOCK_CHAIN_ID`?

test/integration/adapters/LocalAdapter.sol
```diff
@@ -0,0 +1,45 @@
...
+        // Local messages must be bypassed
+        assertEq(sourceChainId, chainId, "Local messages must be bypassed");
+
+        gateway.handle(chainId, message);
+    }
+
+    /// @inheritdoc IAdapter
+    function send(uint16 destinationChainId, bytes calldata payload, uint256, address) external payable {
+        // Local messages must be bypassed
+        assertNotEq(destinationChainId, sourceChainId, "Local messages must by bypassed");
```
wischli, line 37: I am clearly missing something here: Doesn't `LocalAdapter` imply that the source and destination are equal such that this should be equality instead of non-equality? Please explain to me, thanks!

wischli (dismissed):
Congratulations on our first end-to-end test thanks to all your incredible efforts :tada:

test/vaults/BaseTest.sol
lemunozm: Agree! 👍🏻 it's still used in `ERC20`

test/integration/adapters/LocalAdapter.sol
lemunozm: Probably, I can give a better name (open to suggestions!). The `LocalAdapter` is only used in `EndToEnd.t.sol` (will not be used testnets, there we'll use `LocalhostAdapter`) to connect to another adapter.

end-to-end tests emulate test cases between different chains. If an action goes to the same chain, then the `MessageProcessor` will bypass the adapter, only if it goes to different chain, it will reach the adapter.

lemunozm: Well, after thinking it well. It's a `LocalAdapter` because it connects two Full deployment instances in the same Ethereum chain but containing different centrifuge chain Ids.

mustermeiszer: Why would they contain different centrifuge chain Ids? That sounds problematic.

lemunozm: Because they "are" actually in different chains. We're emulating a multichain deployment

mustermeiszer: Okay. Sounds good!

wischli (approved):
Reapproving. Thanks a lot for the explanation regarding the `LocalAdapter`!

PR #179, opened 2025-03-26
Accounting: Transient storage per pool

onnovisser:
Refactor Accounting to have transient storage per pool, so that journalIds can be re-used. Allows a batch from the gateway to have interleaved messages from different pools.


src/pools/Accounting.sol
```diff
@@ -6,61 +6,94 @@ import {Auth} from "src/misc/Auth.sol";
...
-        emit StartJournalId(poolId, journalId_);
+    function unlock(PoolId poolId) external auth {
+        require(!poolId.unlocked(), AccountingAlreadyUnlocked());
+        poolId.setDebited(0);
+        poolId.setCredited(0);
+        poolId.setUnlocked(true);
+
+        if (poolId.journalId() == 0) {
+            poolId.setJournalId(_generateJournalId(poolId));
+        }
```
lemunozm: CR. Is it `unlock` overwriten by the behavior of `journalId`? It means, if we set the a journalId, can not consider the pool unlocked?

I think so:
- If the multicall is done by the FM in CP, then we check in `PoolRouter` that the current `unlockPool` is not set avoiding reentrancies with different pools.
- If the multicall is done by a batch in CV, then we do not increment the JournalId if it's already open.

test/pools/unit/Accounting.t.sol
```diff
@@ -40,21 +39,28 @@ contract AccountingTest is Test {
         accounting.createAccount(POOL_A, GAIN_ACCOUNT, false);
     }
 
+    function beforeTestSetup(bytes4 testSelector) public pure returns (bytes[] memory beforeTestCalldata) {
```
lemunozm, line 42: This is foundry specific, right? Didn't know about it!

src/pools/Accounting.sol
```diff
@@ -89,9 +122,7 @@ contract Accounting is Auth, IAccounting {
         }
     }
 
-    /// @inheritdoc IAccounting
-    function generateJournalId(PoolId poolId) external auth returns (uint256) {
-        journalId = uint256((uint128(poolId.raw()) << 64) | ++_poolJournalIdCounter[poolId]);
-        return journalId;
+    function _generateJournalId(PoolId poolId) internal returns (uint256) {
+        return uint256((uint128(poolId.raw()) << 64) | ++_poolJournalIdCounter[poolId]);
     }
```
lemunozm, line 109: Maybe it deserve to move this generation to the `PoolRouter`. If tomorrow we want to add more info to this number, we can do it easily if the generation happens in the `PoolRouter` without migrating the `Accounting`.

I would propose the following in the PoolRouter.unlock():
```sol
accounting.unlock(poolId, _generateJournalId(poolId, accounting.JournalId));
```

If this is not interesting because we consider this number will never change, I'm fine with the current solution.

cc @mustermeiszer

lemunozm (commented):
I read this and I felt I was reading the code from a Solidity expert of several years. Really amazed with your skills @onnovisser !

test/pools/unit/Accounting.t.sol
onnovisser: Yes! I found out about it today

src/pools/Accounting.sol
onnovisser: The goal is for the journalId to remain set across multiple unlock/locks. If in a batch we have an multiple UpdateHoldings, each call to `updateHoldingAmount` will unlock and lock the accounting. The second call will pick up on the journalId set during the first call.

lemunozm: Ok, I miss the fact that we're locking/unlocking each time a method from the gateway is called.

Then, why do we need a transient per pool? We only need it for the journalId, right?

src/pools/Accounting.sol
lemunozm: Another option is to move all transient stuff to a new contract that can be used by  `Accounting` and can be easily migrated. That way we can update the transient logic stuff if some requirement changes without migrating the `Accounting`

src/pools/Accounting.sol
onnovisser: yes you're right actually. if the individual functions do unlock/lock then we only the the transient journal id per pool and keep the id of the currently unlocked pool.

onnovisser: simplified it again, closer to what it was before

lemunozm (dismissed):
Love how simple the solution to support batches finally is! Thanks @onnovisser !

src/pools/Accounting.sol
mustermeiszer: mmh. Not sure tbh. Gut feeling is that we don't need anything else than a counter. But then again, we will NEVER be able to migrate than one. Tend to keep it here, but will leave this up to you.

lemunozm: My concern is that if in the future we need to enable/disable unlocking the pool while reading different messages in the same batch, we would need to update this logic.

I'm a bit more leaned into split that logic to an `AccountingLock` contract or similar where to handle how to unlock and how to handle the JournalId. But we can review this once everything is complete, before going to production.

mustermeiszer: Fine for me

lemunozm (approved):
LGTM!

PR #181, opened 2025-03-27
Feat: subsidizing pools

lemunozm:
This PR allows subsidizing a pool, for later let the Gateway to pay the adapters for messages related to that pool. NOTE. This only affects non-batching payments. If an action that sends a message had called the `setPauyableSource(PoolA)`, then that message can be paid with the funds attached to that pool. 

To attach funds to a pool to pay messages in the Gateway any user can call the payable `subsidizePool(poolId)`.

NOTE. At this moment, only the `InvestmentManager` makes use of the `setPayableSource()` method.


lemunozm:
Despite the simplicity of this PR, fixing the failing tests took me several hours. The payment methods are getting quite complex. We have at least now 3 ways of payment:

- In the same call (using `topUp()` under the hood). Support optional batching. Only from Vault/Pool Routers. 
- By subsidizing a pool (using `setPayableSource()` under the hood). Only used if `InvestmentManager` is called directly without VaultRouter. It does not support batching.
  - (new incoming feature) Will also be used by other contracts when processing a message implies sending back a new message.
- By not paying anything, which will delay the payment directly yo the adapter after sending the message. Blocked when the user uses Vault/Pool Routers, because we expect some minimal payment. We can only reach this point if there are not enough funds when the pool is subsided and the `InvestmentManager` is called directly.

Related thread with some brainstorming solutions to simplify this: https://kflabs.slack.com/archives/C07PG2EUR9C/p1743004666788769?thread_ts=1743004113.580929&cid=C07PG2EUR9C

src/common/Gateway.sol
```diff
@@ -51,6 +57,9 @@ contract Gateway is Auth, IGateway, IRecoverable {
     mapping(bytes32 messageHash => Message) internal _messages;
     mapping(IAdapter adapter => Adapter) internal _activeAdapters;
 
+    /// @notice Amount of ETH received per pool for paying messages
+    mapping(PoolId => uint256) public subsidy;
```
lemunozm, line 61: NOTE. Right now, `recoverTokens()` could break the invariant of each subsidize pool, i.e.:

- subsidizePool{100}(PoolA)
- recoverTokens(60)

But the system still thinks PoolA can use 100.

NOTE. `recoverTokens()` will be only called by governance, but even if it's called with `recoverTokens(1)`, it already breaks the invariant of `sum(subsidedPools) == address(gateway).balance`. So it only makes sense if we're migrating the contract or so, where the contract will no longer be used anymore.

If not, I think we would need a new solution where the PoolAdmin can "unsubsidize a pool" and recover their tokens.

hieronx: > So it only makes sense if we're migrating the contract or so, where the contract will no longer be used anymore.

The other use case it makes sense, which is the main use case for `recoverTokens`: someone accidentally transfers ETH directly to the Gateway contract, at which point it becomes `sum(subsidedPools) + accidentalTokens == address(gateway).balance`. That is really what this method is for.

lemunozm: Note that now we no longer have a `receive()` method, so I'm not sure if sending ETH accidentally to the `Gateway` is something possible (at least it does not differ from any other contract)

hieronx: You can still send tokens there. `receive` is only a handler for triggering code on receiving ETH

hieronx: Either ERC20 tokens or native ETH

hieronx: > (at least it does not differ from any other contract)

Not different no, and that is why we have `recoverTokens` in all user facing contracts

lemunozm: Ah ok. Now, I fully understand whys/usage. Thanks for the explanation!

src/common/Gateway.sol
```diff
@@ -51,6 +57,9 @@ contract Gateway is Auth, IGateway, IRecoverable {
     mapping(bytes32 messageHash => Message) internal _messages;
     mapping(IAdapter adapter => Adapter) internal _activeAdapters;
 
+    /// @notice Amount of ETH received per pool for paying messages
```
```suggestion
    /// @notice Amount of native tokens received per pool for paying messages
```

E.g. is called CELO on Celo chain

hieronx (dismissed):

src/common/Gateway.sol
lemunozm: Correct!

hieronx (approved):

PR #183, opened 2025-03-28
Refactor `Gateway` to segment per chain ID, extend `Guardian`

hieronx:


test/vaults/BaseTest.sol
```diff
@@ -61,7 +61,7 @@ contract BaseTest is VaultsDeployer, GasSnapshot, Test {
     uint16 public constant OTHER_CHAIN_ID = 1;
     uint16 public constant THIS_CHAIN_ID = OTHER_CHAIN_ID + 100;
     uint32 public constant BLOCK_CHAIN_ID = 23;
-    uint64 immutable POOL_A = newPoolId(OTHER_CHAIN_ID, 1).raw();
+    uint32 immutable POOL_A = 1;
```
lemunozm: I forced this to avoid having local message issues. If PoolId is different from the usage one, the `MessageDispatcher` will never call pools code.

Right now it works because `POOL_A` has chain 0, different that `THIS_CHAIN_ID`. I would like, if possible, to leave this explicit.

script/adapters/Wormhole.s.sol
```diff
@@ -18,7 +18,7 @@ contract WormholeDeployer is FullDeployer {
         deployFull(centrifugeChainId, ISafe(vm.envAddress("ADMIN")), msg.sender);
 
         WormholeAdapter adapter = new WormholeAdapter(gateway, relayer, localChainId, msg.sender);
-        wire(adapter, msg.sender);
+        wire(centrifugeChainId, adapter, msg.sender);
```
lemunozm: CR. I think `centrifugeChainId` should be passed here.

`centrifugeChainId` is the chain where the adapter leaves. That adapter will never be used due to local bypassed messages. I think we need to add another envvar with the chain where this adapter is intended to be connected to.

script/adapters/Localhost.s.sol
```diff
@@ -24,7 +24,7 @@ contract LocalhostDeployer is FullDeployer {
         saveDeploymentOutput();
 
         LocalhostAdapter adapter = new LocalhostAdapter(gateway, msg.sender);
-        wire(adapter, msg.sender);
+        wire(centrifugeChainId, adapter, msg.sender);
```
lemunozm: Given that we have now bypassed messages, I think we can remove this adapter that will never be used.

lemunozm (dismissed):
Super clean and easier than I thought!

Just a CR, everything else looks good!

test/vaults/BaseTest.sol
hieronx: Actually right now, the CV tests are not working, because of incorrect chain IDs in the pools. I was not entirely sure how you had intended to set this up

script/adapters/Localhost.s.sol
hieronx: True! Let's do in followup PR?

script/adapters/Wormhole.s.sol
hieronx: I think this comes back to this Slack discussion: https://kflabs.slack.com/archives/C07PG2EUR9C/p1742816139172279?thread_ts=1742814500.568169&cid=C07PG2EUR9C

Let's remove the `wire` for now, and add a ticket to fix this properly in a followup PR?

lemunozm: Ok, I agree with the ticket 👍🏻

script/adapters/Localhost.s.sol
lemunozm: Yeah, could you add a ticket or add an entry to the EPIC?

test/vaults/BaseTest.sol
```diff
@@ -97,7 +97,7 @@ contract BaseTest is VaultsDeployer, GasSnapshot, Test {
         testAdapters.push(adapter3);
 
         // wire contracts
-        wire(adapter1, address(this));
+        wire(THIS_CHAIN_ID, adapter1, address(this));
```
lemunozm: It should be `OTHER_CHAIN_ID`

test/integration/EndToEnd.t.sol
```diff
@@ -51,7 +51,7 @@ contract TestEndToEnd is Test {
         deploy.deployFull(chainId, safeAdmin, address(deploy));
 
         adapter = new LocalAdapter(chainId, deploy.gateway(), address(deploy));
-        deploy.wire(adapter, address(deploy));
+        deploy.wire(chainId, adapter, address(deploy));
```
lemunozm: This is the "other" `chainId` used in the other deployment

test/vaults/BaseTest.sol
lemunozm: Added a couple of suggestions to fix the tests. I think the problem is the `wire()`

src/vaults/VaultRouter.sol
```diff
@@ -34,12 +34,14 @@ contract VaultRouter is Auth, Multicall, IVaultRouter {
 
     IEscrow public immutable escrow;
     IGateway public immutable gateway;
+    uint16 public immutable localChainId;
```
lemunozm: I think you can get rid of this if you do:

`poolManager.sender().centrifugeId()`

lemunozm (commented):
Thanks for the annoying refactor!

src/vaults/VaultRouter.sol
hieronx: Would be quite expensive, would rather put it into the contract

lemunozm: As long as we ensure the value is the same, we're fine 👍🏻

Curious is `VaultRouter.estimate()` expected to use by other contracts or just by the UX?

hieronx: Mainly by the Centrifuge App, but could be used by smart contracts of third party protocols.

script/adapters/Localhost.s.sol
lemunozm: Finally removed

script/adapters/Wormhole.s.sol
lemunozm: I just added the remoteCentrifugeId, to at least leave the Deployer working, although we would need to iterate over it to create a good deployer for adapters

src/common/Gateway.sol
```diff
@@ -230,37 +230,45 @@ contract Gateway is Auth, IGateway, IRecoverable {
 
         if (kind == MessageType.InitiateMessageRecovery) {
             MessageLib.InitiateMessageRecovery memory m = message.deserializeInitiateMessageRecovery();
-            IAdapter adapter = IAdapter(address(bytes20(m.adapter)));
-            require(_activeAdapters[adapter].id != 0, "Gateway/invalid-adapter");
-            recoveries[adapter][m.hash] = block.timestamp + RECOVERY_CHALLENGE_PERIOD;
-            emit InitiateMessageRecovery(m.hash, adapter);
+            return _initiateMessageRecovery(m.domainId, IAdapter(address(bytes20(m.adapter))), m.hash);
```
wischli: Nit: `_initiateMessageRecovery` doesn't return anything
```suggestion
           _initiateMessageRecovery(m.domainId, IAdapter(address(bytes20(m.adapter))), m.hash);
```

src/common/Guardian.sol
```diff
@@ -24,6 +29,8 @@ contract Guardian is IGuardian {
         _;
     }
 
+    // TODO: add file method for messageDispatcher
```
wischli: Reminder for this TODO

src/common/interfaces/IGateway.sol
```diff
@@ -103,22 +112,22 @@ interface IGateway is IMessageHandler, IMessageSender {
     ///         The quorum is taken from the first adapter.
     ///         Current quorum is the amount of all adapters.
     /// return  Needed amount
-    function quorum() external view returns (uint8);
+    function quorum(uint16 chainId) external view returns (uint8);
```
wischli, line 116: Nit: This and below functions are missing `chainId` docs

src/common/MessageDispatcher.sol
```diff
@@ -20,19 +20,22 @@ import {
...
+interface IMessageDispatcher is IRootMessageSender, IVaultMessageSender, IPoolMessageSender {
     /// @notice Emitted when a call to `file()` was performed.
     event File(bytes32 indexed what, address addr);
 
     /// @notice Dispatched when the `what` parameter of `file()` is not supported by the implementation.
     error FileUnrecognizedParam();
 
+    /// @notice Dispatched when a message is attempted to be executed on the same chain, when that is not allowed.
+    error LocalExecutionNotAllowed();
+
```
hieronx:
```suggestion
```

hieronx (commented):
Great job fixing all the tests!

test/vaults/BaseTest.sol
```diff
@@ -165,74 +164,59 @@ contract BaseTest is VaultsDeployer, GasSnapshot, Test {
...
+        // Trigger new vault deployment via UpdateContract
+        poolManager.update(
+            POOL_A.raw(),
+            trancheId,
+            MessageLib.UpdateContractVaultUpdate({
+                vaultOrFactory: bytes32(bytes20(vaultFactory)),
+                assetId: assetId,
+                kind: uint8(VaultUpdateKind.DeployAndLink)
+            }).serialize()
+        );
```
wischli: Nit: This adds up to running into `stack-too-deep` more quickly. But let's keep as is for now.

test/vaults/fuzzing/recon-aggregator/targets/BiasedTargetFunctions.sol
```diff
@@ -78,19 +78,20 @@ abstract contract BiasedTargetFunctions is BaseTargetFunctions, Properties {
         MockAdapter(address(adapters[calledRouterId])).execute(
             MessageLib.InitiateMessageRecovery({
                 hash: keccak256(message),
-                adapter: bytes32(bytes20(address(adapters[recoverRouterId])))
+                adapter: bytes32(bytes20(address(adapters[recoverRouterId]))),
+                domainId: 0
```
wischli, line 82: Curious why we use `domainId: 0` here instead of some constant from `BaseTest`.

wischli (commented):
Thanks for the minor fixes independent of this PR! LGTM with a few nits. Let's not merge this before #192 though if possible due follow our priority list.

src/common/MessageDispatcher.sol
lemunozm: Good catch!

src/common/Gateway.sol
lemunozm: Dispute either, good catch!

test/vaults/BaseTest.sol
lemunozm: Has had some issue here?

test/vaults/fuzzing/recon-aggregator/targets/BiasedTargetFunctions.sol
lemunozm: It is not inherent in the base test. And from this perspective the `domainId` is unused

src/common/interfaces/IGateway.sol
lemunozm: Added

src/common/Guardian.sol
lemunozm: Added

test/vaults/BaseTest.sol
wischli: No, just saying that each test which deploys a vault, is impacted by that such that the test might have to be refactored using more variables to avoid it. Just wanted to mention that, no need to change!

wischli (approved):
Incredible job to get this working 🚀 !

PR #185, opened 2025-03-30
Extend test data, rename tranche to share classes in vaults

hieronx:

PR #184, opened 2025-03-28
Feat: Update prices of Share Class

mustermeiszer:

PR #192, opened 2025-04-01
Rename `tranche` to `share classes` in vaults & Extend Test Data

wischli:
Supersedes #185 which also includes changes from #184

* Adds missing renames of all _tranche_ occurrences 
* Renames `TrancheTokenFactory` to `TrancheFactory`
* Renames `Tranche` to `ShareToken`
* Renames `TrancheToken` to `CentrifugeToken`
* Renames `shareClass` variables to `scId` (mainly in `MultiShareClass`)
* Renames `RestrictedRedemptions` to `FreelyTransferable` and `RestrictionManager` to `RestrictedTransfers` ([ref](https://kflabs.slack.com/archives/C07PG2EUR9C/p1743509902052659?thread_ts=1743419171.784569&cid=C07PG2EUR9C))

Question for reviewers: In CV, we are 99% referring to the previous _tranche token_ as just _token_. Do you think this is misleading due to also having `tokenId` which is linked to assets instead of `token`? Wondering if we should use `shareToken` instead. 

### Additional public call/event renames extending #185 

PoolManager
```diff
-     emit TransferTrancheTokens(..);
+     emit TransferShares(..);
```


test/vaults/integration/AsyncRequests.sol
```diff
@@ -25,7 +25,8 @@ contract AsyncRequestsTest is BaseTest {
     function testDeployment(address nonWard) public {
         vm.assume(
             nonWard != address(root) && nonWard != address(gateway) && nonWard != address(poolManager)
-                && nonWard != address(messageDispatcher) && nonWard != address(messageProcessor) && nonWard != address(this)
+                && nonWard != address(messageDispatcher) && nonWard != address(messageProcessor)
+                && nonWard != address(syncRequests) && nonWard != address(this)
```
wischli, line 29: Note: Leftover from #109 which seemingly wasn't caught by the fuzzer around the time we merged the PR.

README.md
```diff
@@ -24,13 +24,13 @@ Centrifuge Vaults enable seamless deployment of Centrifuge pools on any EVM-comp
...
-Investors can invest in multiple tranches for each RWA pool. Each of these tranches is a separate deployment of an [ERC-7540](https://eips.ethereum.org/EIPS/eip-7540) Vault and a Tranche Token.
-- [**ERC7540Vault**](https://github.com/centrifuge/protocol-v3/blob/main/src/vaults/ERC7540Vault.sol): An [ERC-7540](https://eips.ethereum.org/EIPS/eip-7540) (extension of [ERC-4626](https://ethereum.org/en/developers/docs/standards/tokens/erc-4626/)) compatible contract that enables investors to deposit and withdraw stablecoins to invest in tranches of pools.
-- [**Tranche Token**](https://github.com/centrifuge/protocol-v3/blob/main/src/vaults/token/Tranche.sol): An [ERC-20](https://ethereum.org/en/developers/docs/standards/tokens/erc-20/) token for the tranche, linked to a [`RestrictionManager`](https://github.com/centrifuge/protocol-v3/blob/main/src/vaults/token/RestrictionManager.sol) that manages transfer restrictions. Prices for tranche tokens are computed on Centrifuge.
+Investors can invest in multiple share classes for each RWA pool. Each of these share classes is a separate deployment of an [ERC-7540](https://eips.ethereum.org/EIPS/eip-7540) Vault and a Share Class Class Token.
```
hieronx:
```suggestion
Investors can invest in multiple share classes for each RWA pool. Each of these share classes is a separate deployment of an [ERC-7540](https://eips.ethereum.org/EIPS/eip-7540) Vault and a Share Token.
```

deployments/verify.sh
```diff
@@ -37,20 +37,20 @@ if ! cast call $contract_address 'share()(address)' --rpc-url $RPC_URL &> /dev/n
     exit 1
 fi
 poolId=$(cast call $contract_address 'poolId()(uint64)' --rpc-url $RPC_URL | awk '{print $1}')
-trancheId=$(cast call $contract_address 'trancheId()(bytes16)' --rpc-url $RPC_URL | cut -c 1-34)
+shareClassId=$(cast call $contract_address 'shareClassId()(bytes16)' --rpc-url $RPC_URL | cut -c 1-34)
```
hieronx: This is calling into the vault so still `trancheId()`

script/VaultsDeployer.s.sol
```diff
@@ -34,16 +34,16 @@ contract VaultsDeployer is CommonDeployer {
...
     function deployVaults(uint16 chainId, ISafe adminSafe_, address deployer) public {
         deployCommon(chainId, adminSafe_, deployer);
 
         escrow = new Escrow{salt: SALT}(deployer);
         routerEscrow = new Escrow{salt: keccak256(abi.encodePacked(SALT, "escrow2"))}(deployer);
-        restrictionManager = address(new RestrictionManager{salt: SALT}(address(root), deployer));
-        restrictedRedemptions = address(new RestrictedRedemptions{salt: SALT}(address(root), address(escrow), deployer));
-        trancheFactory = address(new TrancheFactory{salt: SALT}(address(root), deployer));
+        restrictionManager = address(new RestrictedTransfers{salt: SALT}(address(root), deployer));
+        restrictedRedemptions = address(new FreelyTransferable{salt: SALT}(address(root), address(escrow), deployer));
```
hieronx: These variables should be renamed too then

script/adapters/Localhost.s.sol
```diff
@@ -63,7 +71,32 @@ contract LocalhostDeployer is FullDeployer {
             bytes32(bytes20(address(asyncVaultFactory))),
             VaultUpdateKind.DeployAndLink
         );
+        // TODO(follow-up): Enable after merging #184
+        // cs[c++] = abi.encodeWithSelector(poolRouter.updateSharePrice.selector, scId, navPerShare);
+        // cs[c++] = abi.encodeWithSelector(poolRouter.notifySharePrice.selector, scId, assetId);
+
+        poolRouter.execute{value: 0.001 ether}(poolId, cs);
```
hieronx: If the above is commented out

```suggestion
        //poolRouter.execute{value: 0.001 ether}(poolId, cs);
```

src/vaults/PoolManager.sol
```diff
@@ -97,30 +97,27 @@ contract PoolManager is Auth, IPoolManager, IUpdateContract, IPoolManagerGateway
...
-        uint128 amount
-    ) external auth {
-        ITranche tranche_ = ITranche(tranche(poolId, trancheId));
-        require(address(tranche_) != address(0), "PoolManager/unknown-token");
-        tranche_.burn(msg.sender, amount);
-
-        sender.sendTransferShares(destinationId, poolId, trancheId, recipient, amount);
-
-        emit TransferTrancheTokens(poolId, trancheId, msg.sender, destinationId, recipient, amount);
+    function transferShares(uint64 poolId, bytes16 scId, uint16 destinationId, bytes32 recipient, uint128 amount)
```
hieronx: Really unrelated but would be nice to be consistent with vaults
```suggestion
    function transferShares(uint64 poolId, bytes16 scId, uint16 destinationId, bytes32 receiver, uint128 amount)
```

hieronx (dismissed):

lemunozm:
> Question for reviewers: In CV, we are 99% referring to the previous tranche token as just token. Do you think this is misleading due to also having tokenId which is linked to assets instead of token? Wondering if we should use shareToken instead.

I agree, I think it's less confusing if we use `shareToken`, not sure how difficult is to add that

deployments/verify.sh
wischli: Good catch! Fixed [1216725](https://github.com/centrifuge/protocol-v3/pull/192/commits/121672507ab4f8ebc80d0e1bf42c14c0119b1691)

script/VaultsDeployer.s.sol
wischli: Fixed: [890c08f](https://github.com/centrifuge/protocol-v3/pull/192/commits/890c08fced5b54905caa0c9e7d09d84d6b5352fe)

wischli:
> > Question for reviewers: In CV, we are 99% referring to the previous tranche token as just token. Do you think this is misleading due to also having tokenId which is linked to assets instead of token? Wondering if we should use shareToken instead.
> 
> I agree, I think it's less confusing if we use `shareToken`, not sure how difficult is to add that

It was a bit painful because of the many `token` collisions but AFAICT, I covered it all: [eb45a9f](https://github.com/centrifuge/protocol-v3/pull/192/commits/eb45a9fe64ad5104eab40696cd787d5e082378fc)

hieronx (dismissed):

hieronx (approved):

PR #194, opened 2025-04-02
Some pending fixes for the gateway

lemunozm:
- Add missing `auth`s: [Manuel thread](https://github.com/xmxanuel/centrifuge-protocol-v3-review/pull/32#discussion_r2011682452)
- Overprotect `topUp()`: [Manuel thread](https://github.com/xmxanuel/centrifuge-protocol-v3-review/pull/5#discussion_r2010687572) 
- Remove old recovery check: [Slack thread](https://kflabs.slack.com/archives/C07PG2EUR9C/p1743579553244239?thread_ts=1743511215.439699&cid=C07PG2EUR9C)
- Protectect against sending empty message. [Manuel thread](https://github.com/xmxanuel/centrifuge-protocol-v3-review/pull/5#discussion_r2012023324)
- Remove unused `isRecovery` condition. 
- Fully support for pool counters to reach `uint48`


wischli (dismissed):
Thanks for the quick fixes!

src/common/Gateway.sol
```diff
@@ -183,11 +182,11 @@ contract Gateway is Auth, IGateway, IRecoverable {
...
-            require(isRecovery || adapter.id != PRIMARY_ADAPTER_ID, "Gateway/non-proof-adapter");
+            require(adapter.id != PRIMARY_ADAPTER_ID, "Gateway/non-proof-adapter");
             messageHash = payload.deserializeMessageProof().hash;
             emit ProcessProof(chainId, messageHash, adapter_);
         } else {
-            require(isRecovery || adapter.id == PRIMARY_ADAPTER_ID, "Gateway/non-message-adapter");
+            require(adapter.id == PRIMARY_ADAPTER_ID, "Gateway/non-message-adapter");
             messageHash = keccak256(payload);
             emit ProcessMessage(chainId, payload, adapter_);
         }
```
lemunozm, line 192: @hieronx could you double-check this?

I'm pretty sure that the second condition shadows `isRecovery` value. It means, any recovered message will be recovered from the adapter that sent in the first instance the `proof` or the `message`, so adding `isRecovery` as an OR does not check anything extra nor bypass anything

src/pools/PoolRegistry.sol
```diff
@@ -12,7 +12,7 @@ import {IShareClassManager} from "src/pools/interfaces/IShareClassManager.sol";
 contract PoolRegistry is Auth, IPoolRegistry {
     using MathLib for uint256;
 
-    uint32 public latestId;
+    uint48 public latestId;
```
lemunozm, line 15: Extra fix to allow pool counter to reach 2^48-1

src/common/Gateway.sol
hieronx: >  It means, any recovered message will be recovered from the adapter that sent in the first instance the proof or the message, so adding isRecovery as an OR does not check anything extra nor bypass anything

No this is inverse. If adapter A fails to send some message, then adapter B will initiate recovery for adapter A.

hieronx: Since the use case is adapter A not being able to deliver messages anymore

src/pools/PoolRegistry.sol
hieronx: Good catch

src/common/Gateway.sol
lemunozm: Yes, adapter B will initiate recovery, but in behalf of AdapterA, it means, adapterA will be inside of the initiate recovery message. And when the message is executed, the handle will process the message as if it was from adapterA

lemunozm: I mean, in the condition `adapter.id != PRIMERY_ADAPTER_ID`, that adapter is not adapterB, is adapterA

hieronx: You're right, this change makes sense 👍

hieronx (approved):

wischli (approved):

PR #195, opened 2025-04-02
Add generic pool dependencies mapping to PoolRegistry

### LINKED ISSUE #172: Add generic pool dependencies mapping to `PoolRegistry`
```solidityi
mapping (PoolId => mapping (bytes32 => address)) poolDependency;
```

hieronx:
Fixes https://github.com/centrifuge/protocol-v3/issues/172


wischli (dismissed):
Very elegant!

src/pools/PoolRouter.sol
```diff
@@ -121,16 +121,16 @@ contract PoolRouter is Auth, Multicall, IPoolRouter, IPoolRouterGatewayHandler {
...
+        poolRegistry.updateDependency(poolId, bytes32("shareClassManager"), address(shareClassManager));
     }
 
     /// @inheritdoc IPoolRouter
     function claimDeposit(PoolId poolId, ShareClassId scId, AssetId assetId, bytes32 investor) external payable {
         _protected();
         _pay();
 
-        IShareClassManager scm = poolRegistry.shareClassManager(poolId);
+        IShareClassManager scm = IShareClassManager(poolRegistry.dependency(poolId, bytes32("shareClassManager")));
```
lemunozm: NIT. Could we create a constant for bytes32("shareClassManager") or an auxiliar method for get such dependency?

lemunozm (commented):
Agree!

src/pools/PoolRouter.sol
hieronx: Definitely. Which of the two options do you prefer?

lemunozm: Maybe the second one abstracts each occurrence, i.e `_shareClassManager(poolId)` but not a strong opinion!

hieronx: Added, as public method since it fits well with `escrow()` and seems useful

src/pools/PoolRouter.sol
```diff
@@ -542,6 +542,11 @@ contract PoolRouter is Auth, Multicall, IPoolRouter, IPoolRouterGatewayHandler {
         return address(bytes20(keccak256(abi.encodePacked("escrow", poolId, scId, escrow_))));
     }
 
+    /// @inheritdoc IPoolRouter
+    function shareClassManager(PoolId poolId) public view returns (IShareClassManager) {
+        return IShareClassManager(poolRegistry.dependency(poolId, bytes32("shareClassManager")));
+    }
```
lemunozm, line 548: Good idea to make it public!

lemunozm (dismissed):

lemunozm (approved):

PR #197, opened 2025-04-02
Simplify gateway payment

lemunozm:
The `Gateway` is now able to inspect the message (if not multi-call) to look for an associated `PoolId`. If the `PoolId` is subsidized, such message will be paid using those funds. This avoid the annoying `setPayableSource` usage into the code and solves some issues.

- Fix [this issue](https://kflabs.slack.com/archives/C07PG2EUR9C/p1743581467342339) where `SyncRequests` are not able to be used by vaults given the `setPayableSource` is not called.
- Fix a payment issue in `BalanceSheetManager` where it was not extracting the payment from the subsidized pool (due `setPayableSource` was not called there neither).
- Remove `setPayableSource()`
- Added `MessageLib.messagePoolId()` to retrieve the PoolId of a message.
- Remove now unused `shouldRepay()` method


src/common/GasService.sol
```diff
@@ -41,9 +41,4 @@ contract GasService is IGasService, Auth {
             return messageGasLimit;
         }
     }
```
hieronx, line 43: Should we just move the remaining logic of `GasService` into `MessageProcessor`? It could return a gas amount per message ID.

lemunozm: `MessageProcessor` doesn't know about batches. This part requires another iteration. Right now, when sending, we're estimating the message gas, even when it is a batch, but we're avoiding the fact that this is a batch, we're just reading the first message of the batch.

I think we should modify this to not check if it's a normal message or not, and instead return the gas based on the length of the message.

lemunozm: And maybe it deserves to move it to the Gateway now (?)

hieronx: Eventually we want to return gas limits to specific to messages. So we don't want this in Gateway since it would require migrations for any message changes

hieronx: Let's make this anyways a followup PR then!

hieronx (approved):

src/common/GasService.sol
lemunozm: Ah, right! Because it depends on the computation done by processing the message in the other side, correct!

Then, I think `messageProcessor` can offer this 👍🏻

Issue #115, opened 2025-02-27
CP: Support storing chain metadata


wischli:
In order to distinguish between the variants of the same underlying asset from different chains, we should prefix assets by the chain symbol. Right now, the CP side has no knowledge about CV chains other than the chain id.

Goal
* Add permissioned call to (TODO) `Guardian` for registering chain metadata (name + symbol)
* When receiving `RegisterAsset`, prefix the asset with the chain metadata
* TBD: Where to store metadata

Ref: https://kflabs.slack.com/archives/C07PG2EUR9C/p1740659952908399?thread_ts=1740654438.583919&cid=C07PG2EUR9C

Linked to/blocked by #98

PR #199, opened 2025-04-02
Add pool creation & chain registration to `Guardian`

### LINKED ISSUE #193: CP: limit pool creation to `Guardian` in the `PoolRouter`
For the initial launch, pool creation should be limited to the `Guardian`.

hieronx:
Fixes https://github.com/centrifuge/protocol-v3/issues/193 and https://github.com/centrifuge/protocol-v3/issues/115


script/CommonDeployer.s.sol
```diff
@@ -95,6 +95,7 @@ abstract contract CommonDeployer is Script, JsonRegistry {
         gateway.file("adapters", chainId, adapters);
         IAuth(address(adapter)).rely(address(root));
         IAuth(address(adapter)).deny(deployer);
+        guardian.file("safe", address(adminSafe));
```
lemunozm: Note, we're not wiring anything now for a Localhost deployer IIRC. I think we need to move this to the `removeCommonDeployerAccess`, or any other part.

hieronx: True! Makes sense to move to `removeCommonDeployerAccess` actually

script/CommonDeployer.s.sol
```diff
@@ -55,7 +55,7 @@ abstract contract CommonDeployer is Script, JsonRegistry {
         messageDispatcher = new MessageDispatcher(chainId, root, gateway, deployer);
 
         adminSafe = adminSafe_;
-        guardian = new Guardian(adminSafe, root, messageDispatcher);
+        guardian = new Guardian(ISafe(deployer), root, messageDispatcher);
```
lemunozm, line 60: NIT. Because `deployer` is not really a `ISafe` implementation, could we add a comment here explaining the why for future readers?

src/common/Guardian.sol
```diff
@@ -31,13 +39,30 @@ contract Guardian is IGuardian {
...
+        onlySafe
+        returns (PoolId poolId)
+    {
+        return poolRouter.createPool(admin, currency, shareClassManager);
+    }
+
+    /// @inheritdoc IGuardian
+    function updateChain(uint16 chainId, string calldata name, string calldata symbol) external onlySafe {
+        assetRegistry.updateChain(chainId, name, symbol);
+    }
```
lemunozm: NIT. Can this be named `setChain` due our naming style for setter methods?: https://www.notion.so/Solidity-style-guide-1152eac24e17809ba9f2f81d5b30ba62?pvs=4#17d2eac24e178016a707c8bc7d0fe5f4

src/pools/AssetRegistry.sol
```diff
@@ -36,24 +37,40 @@ contract AssetRegistry is ERC6909Fungible, IAssetRegistry {
...
+        return string.concat(chain_.name, " ", asset[assetId].name);
     }
 
     /// @inheritdoc IERC6909MetadataExt
     function symbol(uint256 asset_) external view returns (string memory) {
-        return asset[AssetId.wrap(asset_.toUint128())].symbol;
+        AssetId assetId = AssetId.wrap(asset_.toUint128());
+        Chain memory chain_ = chain[assetId.chainId()];
+
+        return string.concat(chain_.symbol, asset[assetId].symbol);
```
lemunozm, line 62: Question, could it be interesting to add a `-` in between to improve readability/organization from UX? 
i.e. `celo-USDC`

test/vaults/BaseTest.sol
```diff
@@ -83,9 +83,9 @@ contract BaseTest is VaultsDeployer, GasSnapshot, Test {
 
         // deploy core contracts
         deployVaults(THIS_CHAIN_ID, adminSafe, address(this));
+        guardian.file("safe", address(adminSafe));
```
lemunozm, line 86: `LocalhostDeployment` is not calling to `removeCommonDeployerAccess`, maybe you want to put this line also in that deployer

lemunozm (commented):
Looks super clean! Just few NIT, and one possible CR

src/pools/AssetRegistry.sol
hieronx: `baseUSDC` is what I've seen used in practice. We could still accomplish yours by adding `-` into the actual symbol, so would rather keep it flexible within the contract

lemunozm: Correct! Makes sense

lemunozm:
Seems like you resolved my nits, but the commit was not pushed (?)

hieronx:
> Seems like you resolved my nits, but the commit was not pushed (?) 

Oops, pushed now!

lemunozm (approved):
Thanks for tackling my nits! 🙌🏻

PR #200, opened 2025-04-02
Simplify ERC165 setup

hieronx:


src/vaults/BaseVaults.sol
```diff
@@ -164,7 +164,7 @@ abstract contract BaseVault is Auth, IBaseVault {
     function supportsInterface(bytes4 interfaceId) public pure virtual override returns (bool) {
         return interfaceId == type(IERC7540Operator).interfaceId || interfaceId == type(IERC7741).interfaceId
             || interfaceId == type(IERC7714).interfaceId || interfaceId == type(IERC7575).interfaceId
-            || interfaceId == type(IRecoverable).interfaceId || interfaceId == type(IERC165).interfaceId;
```
wischli, line 167: Don't understand why we would want to remove that if `IBaseVault` implements `IRecoverable`. Mistake?

hieronx: I don't think we really need to cover all interfaces, I think the only ones that are valuable are those that are standardized. Since the purpose of `supportsInterface` is for other smart contracts to know whether some address supports a certain known i.e. standardized interface. `IRecoverable` is not standardized, only something our contracts need to care about.

wischli (approved):
Thanks for cleaning up my mess!

PR #201, opened 2025-04-02
feat: add queued requests to `MultiShareClass`

wischli:
Ref slack thread: https://kflabs.slack.com/archives/C07PG2EUR9C/p1743364415637219

* Adds queued orders to the `MultiShareClass` (now `ShareClassManager`) for UX reasons to allow users to send requests even though they cannot mutate their pending amount
 * Keep track of requested increments of pending amounts or cancellation when the pending amount must not be mutated due to requiring claim
 * Instead of previously erroring out with `ClaimRequired`, now holds those changes in `queued{Deposit, Redeem}Requests` storage
 * Now, we are only erroring out if we execute a `request{Deposit, Redeem}` on top of a queued {deposit, redeem} cancellation request. Since CV blocks requests after initializing a cancellation, this can only happen due to race condition of an increment and cancellation request of the same user. After the user claimed, we can replay the incrementing request with success.
* Queued amounts are automatically moved to pending amounts after claiming all epochs for a user
  * Due to queued cancellation, claiming can now also trigger `FulfilledCancelled*` on top of `Fulfilled*Request`, which I adjusted in the `PoolRouter` (now `Hub`)

Unfortunately, I had to make some adjustments to the code due to `stack to too deep`. 

Other misc changes
* Moved structs from `MultiShareClass.sol` to `IShareClassManager.sol`
* Applied formatting to `MultiShareClass` (now `ShareClassManager`)
* Added `.t` to `AsyncRequests` test file name, which I missed during the renaming madness (sorry)

# Examples

## Pending > 0, Queued > 0, No queued Cancellation

Imagine the following state for a user
  * 500 pending (of which effectively 200 are approved which is only known during claiming)
  * 1000 queued
  * No queued cancellation

When the user claims
  * Claimed asset amount is 200 (shares depending on NAV and other pending amounts)
  * Pending user amount is updated to 500 - 200 + 1000 = 1300
  * Pending total amount is increased by the queued 1000
  * The `PoolRouter` sends `FulfilledDepositRequest` with asset amount 200

## Pending > 0, Queued > 0, Queued Cancellation

Imagine the following state for a user
  * 500 pending (of which effectively 200 are approved which is only known during claiming)
  * 1000 queued
  * **Queued cancellation**

When the user claims
  * Claimed asset amount is 200 (shares depending on NAV and other pending amounts)
  * Cancelled amount is 500 - 200 + 1000 = 1300
  * **Pending user amount is updated to 0**
  * **Pending total amount is reduced by the cancelled 300**
  * The `PoolRouter` (now `Hub`) sends `FulfilledDepositRequest` with asset amount 200
  * **The `PoolRouter` (now `Hub`) sends `FulfilledCancelledDeposit` with asset amount 1300**


src/pools/MultiShareClass.sol
```diff
@@ -747,4 +617,168 @@ contract MultiShareClass is Auth, IMultiShareClass {
...
+        bytes32 investor,
+        AssetId assetId,
+        UserOrder storage userOrder,
+        bool isDeposit
+    ) private returns (bool skipPendingUpdate) {
+        uint128 latestApproval = isDeposit
+            ? epochPointers[scId_][assetId].latestDepositApproval
+            : epochPointers[scId_][assetId].latestRedeemApproval;
+
+        // Block updates to pending and temporarily store in queued until pending amount does not impact claimable
```
hieronx: This comment seems opposite of what happens here: if this if clause is true, then the updates are applied to pending, rather than to queued.

src/pools/MultiShareClass.sol
```diff
@@ -747,4 +617,168 @@ contract MultiShareClass is Auth, IMultiShareClass {
...
+        require(!(queued.isCancelling == true && amount > 0), CancellationQueued());
+
+        if (!isIncrement) {
+            queued.isCancelling = true;
+        } else {
+            queued.amount += amount;
+        }
+
+        if (isDeposit) {
+            emit UpdatedQueuedDeposit(
```
hieronx: Should we just have a single event, with a `RequestType` enum as one of the fields?

wischli: That's a good idea! I would also apply that to `Updated{Deposit, Redeem}Request` and rename the event `UpdatedPendingRequest`. WDYT?

src/pools/MultiShareClass.sol
wischli: Good point. During the implementation I switched back and forth between the interpretation of the queued bool flag.

src/pools/MultiShareClass.sol
hieronx: Yeah agree!

lemunozm (commented):
What a good feature to have!!

Super clean 💯, although I would let the approval to Jeroen

src/pools/MultiShareClass.sol
wischli: Applied here and united all `Updated` events which should be much easier for Apps: [41cf843](https://github.com/centrifuge/protocol-v3/pull/201/commits/41cf843089555fa012707a6b966a3fe106303449)

Not sure though, if we should index the event by `(poolId, scId, requestType)` instead of `(poolId, scId, epochId)` or is the `indexed` from code not so relevant? Since we index all other SCM by the latter triplet, I didn't want to diverge for now.

src/pools/MultiShareClass.sol
wischli: [f9d69b3](https://github.com/centrifuge/protocol-v3/pull/201/commits/f9d69b363fd11e4d42e3eb37b3ff07822b734986)

src/pools/MultiShareClass.sol
```diff
@@ -366,8 +369,8 @@ contract MultiShareClass is Auth, IMultiShareClass {
             //  1. The sum of pending user orders <= pendingDeposits (instead of equality)
             //  2. The sum of claimable user amounts <= amount of minted share class tokens corresponding to the
             // approved deposit asset amount (instead of equality).
-            //     I.e., it is possible for an epoch to have an excess of a share class token atom which cannot be
-            // claimed by anyone.
+            //     I.e., it is possible for an epoch to have an excess of a share class tokens atom which cannot be
+            // claimed by anyone. This excess is at most n-1 atoms for an epoch with n claimable users.
```
hieronx:
```suggestion
            //     I.e., it is possible for an epoch to have an excess of a share class tokens which cannot be
            // claimed by anyone. This excess is at most n-1 atoms for an epoch with n claimable users.
```

src/pools/interfaces/IShareClassManager.sol
```diff
@@ -65,27 +65,16 @@ struct QueuedOrder {
     uint128 amount;
 }
 
+enum RequestType {
+    /// @dev Whether the request is a deposit one
+    Deposit,
+    /// @dev Whether the request is a redeem one
+    Redeem
+}
```
lemunozm, line 73: ❤️ , much more legible!

src/pools/MultiShareClass.sol
```diff
@@ -624,10 +629,11 @@ contract MultiShareClass is Auth, IMultiShareClass {
         bytes32 investor,
         AssetId assetId,
         UserOrder storage userOrder,
-        bool isDeposit
+        RequestType requestType
     ) private returns (uint128 cancelledAmount) {
-        QueuedOrder storage queued =
-            isDeposit ? queuedDepositRequest[scId_][assetId][investor] : queuedRedeemRequest[scId_][assetId][investor];
+        QueuedOrder storage queued = requestType == RequestType.Deposit
```
hieronx, line 634: We could even index the mappings by the enum. Not sure that is actually better, but wanted to at least share the idea.

hieronx (dismissed):

src/pools/MultiShareClass.sol
hieronx: The `indexed` makes it a bit faster to filter logs but the indexer actually anyways need to pick up all `Updated` events so it doesn't have a big impact.

src/pools/MultiShareClass.sol
wischli: I like the idea because this will simplify the code a lot. Since this PR has lower priority than #183, I will explore that idea.

wischli: I think we should merge this PR and I will explore it in a follow-up. Initially, I wanted to merge #184 first but now that I took that over, I prefer merging this one first.

wischli:
@lemunozm @hieronx Initially, I wanted to hold on merging this PR before #184 but now that I took it over, I would like to merge this PR first.

lemunozm (approved):
Let's merge this beauty!

PR #202, opened 2025-04-02
Remove MessageLib from Gateway

lemunozm:
This PR removes any sign of `MessageLib` from the `Gateway` in order to allow modifying `MessageLib` in the future easily without migrating the `Gateway`.

To accomplish that, the following refactors were done:
- `MessageProcessor` now implements from `IMessageProperties` that is a list of utilities from `MessageLib` that can be used by other contracts in order to not be tied to the `MessageLib` implementation.
- Message recovery handler has been moved to `MessageProcessor`

In a follow-up PR:
- Remove `GasService` and move `estimate()` as a method of `IMessageProperties` (implemented by `MessageProcessor`)
- Fix estimation for batches.
- Allow free batching (outside of the routers), to handle correctly BSM batches


src/common/GasService.sol
```diff
@@ -3,24 +3,27 @@ pragma solidity 0.8.28;
...
     /// @inheritdoc IGasService
     uint64 public proofGasLimit;
 
     /// @inheritdoc IGasService
     uint64 public messageGasLimit;
 
-    constructor(uint64 messageGasLimit_, uint64 proofGasLimit_) Auth(msg.sender) {
+    IMessageProperties public messageProperties;
+
+    constructor(uint64 messageGasLimit_, uint64 proofGasLimit_, IMessageProperties messageProperties_)
```
hieronx, line 21: Just thinking through this being so simplified now, and since the `GasService` is easily swappable: I would make `message/proofGasLimit` both `immutable`, remove the `Auth`, and just make sure we put worst case values for now. This way we have cheaper calls since we remove 2 cold storage reads.

.forge-snapshots/VaultRouter_requestDeposit.snap
```diff
@@ -1 +1 @@
-480749
\ No newline at end of file
+487202
```
hieronx: We're adding several external calls `Gateway` => `MessageProcessor` now, replacing what used to be local calls.

I think you are right this is needed, just noting the increase in gas.

src/common/GasService.sol
hieronx: > Remove GasService and move estimate() as a method of IMessageProperties (implemented by MessageProcessor)


Just saw this comment. That makes sense. Let's make them `immutable` in `MessageProcessor` then?

hieronx (dismissed):
Needs to be rebased but LGTM

.forge-snapshots/VaultRouter_requestDeposit.snap
lemunozm: Yes, that's for sure the reason, but we sadly need it. I hope given the comming improvements with batching this increment is depreciable with the gain of batch everything always

src/common/GasService.sol
lemunozm: I'm doing a refactor in the `GasService` in a follow-up PR, do you mind if I apply these changes there, if finaly need it?

hieronx (approved):

PR #204, opened 2025-04-03
Add `updateRestriction` to `PoolRouter`

### LINKED ISSUE #203: CP: support `UpdateRestriction` in `PoolRouter`

hieronx:
Fixes https://github.com/centrifuge/protocol-v3/issues/203

NOTE: tests should be added in `EndToEnd` in a followup PR

### DIFF
--- a/src/pools/PoolRouter.sol
+++ b/src/pools/PoolRouter.sol
@@ -276,6 +276,16 @@ contract PoolRouter is Auth, Multicall, IPoolRouter, IPoolRouterGatewayHandler {
     }
 
     /// @inheritdoc IPoolRouter
+    function updateRestriction(uint16 chainId, ShareClassId scId, bytes calldata payload)
+        external
+        payable
+    {
+        _protectedAndUnlocked();
+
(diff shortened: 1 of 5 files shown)


wischli (approved):
Thanks!

src/pools/PoolRouter.sol
```diff
@@ -275,6 +275,16 @@ contract PoolRouter is Auth, Multicall, IPoolRouter, IPoolRouterGatewayHandler {
...
 
+    /// @inheritdoc IPoolRouter
+    function updateRestriction(uint16 chainId, ShareClassId scId, bytes calldata payload)
+        external
+        payable
+    {
+        _protectedAndUnlocked();
+
+        sender.sendUpdateRestriction(chainId, unlockedPoolId, scId, payload);
+    }
```
lemunozm, line 286: post-morten question, should we check the existence of scId before sending the message? Or do we assume that if the message fails is enough?

hieronx: Good point, will add in new PR

PR #206, opened 2025-04-03
Gateway: correct payment in all scenarios 

lemunozm:
This PRs adds:

## 1. Payment refactor

The payment has been refactored a bit in order to be able to support batches using subsidizing pools. Main difference is that now inside of start/end batching we're also aggrupating by `PoolId`. That way we're able to determine the associated subsidized pool. 

Start/end batching methods are not permisionless to be able to use them from the controller for BSM actions.

## 2. Payment fix for batching 

Previously the payment was wrongly calculated for batching. Now it takes into account all messages inside of a batch.

## Next steps to finally end up with the Gateway part
- Redo all UTs, given all the changes we do
- Add custom errors


hieronx:
> I'm trying to imagine how a correct set-up will work. Initially, we have ChainA and chainB configured, and we want to configure chainC:

FYI @lemunozm for now I would assume the gas limit of messages/proofs is equivalent on all chains. As long as we deploy the same contracts on all EVM chains, this should be the case.

This only won't be true either if EVMs start to diverge more than they do today (unlikely), or we start to deploy vaults on non-EVM chains but this is far out.

NOTE: this is only the gas limit. The actual cost is chain dependent but this is accounted for by the bridge in `adapter.estimate`

lemunozm:
Interesting!!

Would you want me to revert that, or we keep it as it is for future uses as a Solana CV?

Maybe we can revert for now, I guess. Keeping the `chainId` in the interface

hieronx:
> Maybe we can revert for now, I guess. Keeping the chainId in the interface

I'd do that because it will also be cheaper.

Then you can make these `immutable` as discussed in the other PR.

lemunozm:
> Then you can make these immutable as discussed in the other PR.

Not sure about the immutability. We would still need those to be a mapping for each message code. And they should be able to be modified by governance if some message processing cost change in some upgrade. i.e, we modify `PoolManager.addPool` with some extra logic. We should notify the current deployments to take that extra gas into account when processing a `NotifyPool` message

hieronx:
> Not sure about the immutability. We would still need those to be a mapping for each message code. And they should be able to be modified by governance if some message processing cost change in some upgrade. i.e, we modify PoolManager.addPool with some extra logic. We should notify the current deployments to take that extra gas into account when processing a NotifyPool message

My point was, when we do this, we can easily also swap the `GasService`

lemunozm:
Ok, I see. If it's not much cost to do that then fine.

Just a question. AFAIK, we can not have immutable mapping in solidity. You would add only the message/proof values now and in the future you would change `GasService` to contain a mapping (gas value per message code) ?

hieronx:
>  You would add only the message/proof values now and in the future you would change GasService to contain a mapping (gas value per message code) ?

Yeah exactly

Although we could use something like https://github.com/centrifuge/protocol-v3/blob/main/src/common/libraries/MessageLib.sol#L85 right?

lemunozm:
Yeah, correct. It should be a bit more sophisticated, with `bytes` and decode `uint64` from it, but yes. Sure it will save some gas.

hieronx:
To be clear, really don't think we should do this now. Just that we can easily do this later, by swapping the gas service.

lemunozm:
Fixed here: [430f091](https://github.com/centrifuge/protocol-v3/pull/206/commits/430f0919af4838c26e96b5a656d11388ffd91dd0) 👍🏻

script/utils/JsonRegistry.s.sol
```diff
@@ -9,8 +9,8 @@ contract JsonRegistry is Script {
 
     function register(string memory name, address target) public {
         deploymentOutput = (registeredContracts == 0)
-            ? string(abi.encodePacked(deploymentOutput, '    "', name, '": "0x', _toString(target), '"'))
-            : string(abi.encodePacked(deploymentOutput, ',\n    "', name, '": "0x', _toString(target), '"'));
+            ? string(abi.encodePacked(deploymentOutput, '    "', name, '": "0x', vm.toString(target), '"'))
```
hieronx, line 12: Nice find!

script/CommonDeployer.s.sol
```diff
@@ -43,14 +46,14 @@ abstract contract CommonDeployer is Script, JsonRegistry {
             return; // Already deployed. Make this method idempotent.
         }
 
-        root = new Root(DELAY, deployer);
+        uint64 messageGasLimit = uint64(vm.envOr(MESSAGE_COST_ENV, BASE_MSG_COST));
```
hieronx: Base sounds like it's added on top of the other cost
```suggestion
        uint64 messageGasLimit = uint64(vm.envOr(MESSAGE_COST_ENV, FALLBACK_MSG_COST));
```

src/common/Gateway.sol
```diff
@@ -34,8 +35,8 @@ contract Gateway is Auth, IGateway, IRecoverable {
 
     uint256 public transient fuel;
 
-    /// @dev Says if a send() was performed, but it's still pending to finalize the batch
-    bool public transient pendingBatch;
+    /// @notice Tells is the gateway is actually configured to create batches
```
hieronx: Comment seems confusing, and I think with the new enum the line is pretty self explanatory
```suggestion
```

src/common/Gateway.sol
```diff
@@ -249,32 +250,33 @@ contract Gateway is Auth, IGateway, IRecoverable {
     function send(uint16 chainId, bytes calldata message) external pauseable auth {
         require(message.length > 0, "Gateway/empty-message");
 
+        PoolId poolId = processor.messagePoolId(message);
         if (isBatching) {
-            pendingBatch = true;
+            bytes storage previousMessage = batch[chainId][poolId];
```
hieronx, line 254: I'm a bit worried how complex this will get in assembly, to handle the transient mappings. If we just constrained it to only send messages of 1 pool and 1 chain ID, wouldn't that significantly simplify this?

src/common/Gateway.sol
```diff
@@ -301,52 +301,46 @@ contract Gateway is Auth, IGateway, IRecoverable {
...
     /// @inheritdoc IGateway
     function topUp() external payable {
-        // We only require the top up if:
-        // - We're not in a multicall.
-        // - Or we're in a multicall, but at least one message is required to be sent
-        if (!isBatching || pendingBatch) {
-            require(msg.value != 0, "Gateway/cannot-topup-with-nothing");
-            fuel += msg.value;
-        }
+        require(paymentMethod != PaymentMethod.Subsidized, "Gateway/another-payment-method-already-set");
```
hieronx: From what I see, `PaymentMethod.Subsidized` is never set, and it's always either `TopUp` or `None`. Am I missing something?

src/common/Gateway.sol
lemunozm: The real problem here is to store the `bytes`, that we would still have it. The double mapping can be just a `keccak` of both values as we did in `TransientValuation`.

The good part of doing this for multi pools/chainId is that the high level API becomes super easy to use in all scenarios.

hieronx: That's fair!

src/common/Gateway.sol
lemunozm: You're correct. Maybe I need to refactor this and be more explicit.

What I want to do having 3 states is:
- You can call topUp if Payment == `None`
- You can call again topUp if Payment was already `TopUp` (i.e. you're in a multicall)
- If you never called TopUp, then you can subsidize.

But the state naming and what's happening definitely do not match, so I need to explain this better.

I would try to refactor. What is super important is that if topUp was already called, you NEVER can subsidize. If that's not true, then a user can call `CP.claimDeposit()` and use permissionless subsidized funds. Right now, we protect that by forcing that call to use topUp.

src/common/Gateway.sol
```diff
@@ -318,10 +318,10 @@ contract Gateway is Auth, ReentrancyProtection, IGateway, IRecoverable {
 
     /// @inheritdoc IGateway
     function topUp() external payable {
-        require(paymentMethod != PaymentMethod.Subsidized, "Gateway/another-payment-method-already-set");
+        require(paymentMethod != PaymentMethod.TopUp, "Gateway/another-payment-method-already-set");
 
         paymentMethod = PaymentMethod.TopUp;
-        fuel += msg.value;
+        fuel = msg.value;
```
hieronx: Basically means you cannot top up more than once right?

src/common/Gateway.sol
lemunozm: Simplified here: [06b7608](https://github.com/centrifuge/protocol-v3/pull/206/commits/06b760816192223e2a1db1440d4fdc3e708d1ec6)

Now `Subsidized` is the default payment method unless `TopUp` is called.

I just avoided calling `topUp` twice in the same transaction, which is something that I added that makes no sense, because the payment is made once in the high-level multicall.

I need to fix the tests adding: `/// forge-config: default.isolate = true`

src/common/Gateway.sol
```diff
@@ -318,10 +318,10 @@ contract Gateway is Auth, ReentrancyProtection, IGateway, IRecoverable {
 
     /// @inheritdoc IGateway
     function topUp() external payable {
-        require(paymentMethod != PaymentMethod.Subsidized, "Gateway/another-payment-method-already-set");
+        require(paymentMethod != PaymentMethod.TopUp, "Gateway/another-payment-method-already-set");
 
         paymentMethod = PaymentMethod.TopUp;
```
hieronx, line 323: Should we just rename `TopUp` to `Paid`? Seems cleaner

src/common/Gateway.sol
```diff
@@ -249,32 +250,33 @@ contract Gateway is Auth, IGateway, IRecoverable {
     function send(uint16 chainId, bytes calldata message) external pauseable auth {
         require(message.length > 0, "Gateway/empty-message");
 
+        PoolId poolId = processor.messagePoolId(message);
         if (isBatching) {
```
hieronx, line 253: Could we not remove `isBatching` and just use `batchLocators.length > 0`?

lemunozm: Totally! 🚀 Good catch!

mustermeiszer (commented):
* Can you explain how `subsidize` would work?
* Are there funds specific to a pool or global once? 
* What happens if some logic forgets to call `endBatch()`?

lemunozm:
> Can you explain how subsidize would work?

Yes. subsidize works by calling `gateway.subsidizePool()` adding in another transaction funds for that pool. If poolId == 0, then you subsidize messages than does not use pool.

Later the gateway, will associate the message for the same pool in the same batch and will pay with the funds associated to that pool.

The above is the behavior, except you use routers. Router will not use subsidized pools and instead, they'll use a payment in the same transaction.

>Are there funds specific to a pool or global once?

These are specific to a pool, or if pool==0, they are for messages not associated with a pool. They are never global.

>What happens if some logic forgets to call endBatch()?

Good question. The messages will never be sent. I think this can be an issue for some messages. 

Maybe it makes no sense to make those methods public then. Could we have a wrapper contract that offers something like `execute()` to be able to start/end batches? The BSM controller will add BSM methods to that execution. 

WDYT? @mustermeiszer @hieronx

src/common/Gateway.sol
lemunozm: Finally we cannot do that 😢 , given we need to signal if the batching has started in `startBatching()` when still no messages have been sent.

src/common/Gateway.sol
lemunozm: Removed

script/CommonDeployer.s.sol
lemunozm: Done

src/common/Gateway.sol
hieronx: Right :(

lemunozm:
@hieronx sorry for the back and forth, finally I'll go with `auth` for batching/auth, so the payment can be simplified now as [5bd654d](https://github.com/centrifuge/protocol-v3/pull/206/commits/5bd654d28fe345d3a4415c09a125658142185203)

src/common/GasService.sol
```diff
@@ -3,41 +3,27 @@ pragma solidity 0.8.28;
 
 import {Auth} from "src/misc/Auth.sol";
 import {IGasService} from "src/common/interfaces/IGasService.sol";
+import {MessageLib, MessageType} from "src/common/libraries/MessageLib.sol";
 import {IMessageProperties} from "src/common/interfaces/IMessageProperties.sol";
```
wischli, line 7: Nit: Can be removed because unused

src/common/Gateway.sol
```diff
@@ -249,32 +249,33 @@ contract Gateway is Auth, IGateway, IRecoverable {
...
 
-        uint256 messageGasLimit = gasService.estimate(chainId, message);
-        uint256 proofGasLimit = gasService.estimate(chainId, proof);
+        uint64 messageGasLimit = (isBatching) ? batchGasLimit[chainId][poolId] : gasService.gasLimit(chainId, message);
+        uint64 proofGasLimit = gasService.gasLimit(chainId, proof);
 
-        if (fuel != 0) {
-            uint256 tank = fuel;
+        if (paymentMethod == PaymentMethod.TopUp) {
             for (uint256 i; i < adapters_.length; i++) {
```
wischli: Nit: The "hearder" part is duplicate for both if-else branches. We can move it outside that scope.
```solidity
IAdapter currentAdapter = IAdapter(adapters_[i]);
bool isPrimaryAdapter = i == PRIMARY_ADAPTER_ID - 1;
bytes memory payload = isPrimaryAdapter ? message : proof;

uint256 consumed =
    currentAdapter.estimate(chainId, payload, isPrimaryAdapter ? messageGasLimit : proofGasLimit);
```

test/vaults/integration/VaultRouter.t.sol
```diff
@@ -14,9 +16,10 @@ import {MockERC20Wrapper} from "test/vaults/mocks/MockERC20Wrapper.sol";
 import {MockReentrantERC20Wrapper1, MockReentrantERC20Wrapper2} from "test/vaults/mocks/MockReentrantERC20Wrapper.sol";
 
 contract VaultRouterTest is BaseTest {
+    using MessageLib for *;
+
     uint256 constant GAS_BUFFER = 10 gwei;
-    /// @dev Payload is not taken into account during gas estimation
-    bytes constant PAYLOAD_FOR_GAS_ESTIMATION = "irrelevant_value";
+    bytes PAYLOAD_FOR_GAS_ESTIMATION = MessageLib.NotifyPool(1).serialize();
```
wischli, line 22: Curious why we take this message for gas estimation even though its a CP-to-CV-message.

test/pools/integration/BatchingAndPayment.t.sol
```diff
@@ -0,0 +1,92 @@
...
+    }
+
+    /// Test the following:
+    /// - multicall()
+    ///   - execute(poolA)
+    ///      - notifyPool()
+    ///   - execute(poolA)
+    ///      - notifyPool()
+    ///
+    /// will pay only for one message. The batch sent is [NotifyPool, NotifyPool].
```
wischli, line 44: Doesn't this comment contradict paying `GAS * 2` below?

wischli (commented):
From my limited understanding, looks solid now! Leaving the approval up to others with more in-depth-knowledge.

src/common/GasService.sol
lemunozm: It's used in `messageCode()` and `MessageProof`

src/common/Gateway.sol
lemunozm: I agree this part is not readable. The problem is that the "header" part is based on `i` . But I think I can refactorize it a bit!

test/pools/integration/BatchingAndPayment.t.sol
lemunozm: The comment is wrong, good eye!

hieronx (dismissed):

test/vaults/integration/VaultRouter.t.sol
hieronx: I guess it was changed from `AddPool` which it used to be, to now `NotifyPool`. Ultimately it doesn't matter today since the gas estimates are not defined per message. But could indeed be changed to make the test a bit more accurate.

src/common/Gateway.sol
lemunozm: Factorized here: [ddb0c2d](https://github.com/centrifuge/protocol-v3/pull/206/commits/ddb0c2d287268b67c95197cc3eaa4c71ee46e000). Much easier to read!

test/vaults/integration/VaultRouter.t.sol
lemunozm: Because it's the most simple message from where to extract some properties later in the Gateway. But I agree Vault tests require a huge refactorization/cleaning.

hieronx (approved):

src/common/Gateway.sol
hieronx: Big improvement, great idea!

src/common/GasService.sol
wischli: Github failure, I was referring to `import {IMessageProperties} from "src/common/interfaces/IMessageProperties.sol";
`

test/pools/integration/BatchingAndPayment.t.sol
lemunozm: I forgot to fix it! But I would do in a follow up PR

PR #208, opened 2025-04-03
feat: fully integrate Escrow & change holding update flow

wischli:
Ref: https://kflabs.slack.com/archives/C07PG2EUR9C/p1743689530352989?thread_ts=1743669923.076539&cid=C07PG2EUR9C

### New Deposit + Redeem Flow

Makes CV as only provider of holding updates as a result of invest-redeem flow: 

Deposit
  * `asyncRequests.requestDeposit` -> `escrow.pendingIncrease` + `sendUpdateHoldingsAmount`
  * `hub.approveDeposits` -> `balanceSheet.approvedDeposits` -> `sendUpdateHoldingsAmount` + `escrow.deposit` which decreases pending and increases available cash

Redeem
  * `hub.revokeShares` -> `balanceSheet.revokedShares` -> `escrow.reserveIncrease`
  * `asyncRequests.withdraw` (claimRedeem) -> `escrow.reserveDecrease` + `balanceSheet.withdraw` which transfers funds from escrow to user, reduces available cash and `sendUpdateHoldingsAmount`

### Other Changes

* Fixes some incorrectly declared `price*Per*` variables - apologies for the mistakes and huge shout out to @mustermeiszer for pointing me to the correct ones once again
* Isolates `ISharePriceProvider` from previous `ISyncRequests` which is used in BSM and AsyncRequests to not expose other calls
* Adds message type check to `BSM.update`


src/common/libraries/MessageLib.sol
```diff
@@ -110,6 +112,8 @@ library MessageLib {
         (114 << uint8(MessageType.UpdateHoldingAmount) * 8) +
         (65  << uint8(MessageType.UpdateHoldingValue) * 8) +
         (98  << uint8(MessageType.UpdateShares) * 8) +
+        (57  << uint8(MessageType.ApprovedDeposits) * 8) +
+        (57  << uint8(MessageType.RevokedShares) * 8) +
```
wischli, line 120: @lemunozm Unfortunately these two additions seem to break out current message length because this extends the number of messages to 33 such that `33 * 8 > 256`. Any recommendation on how to best extend?

wischli: Working on a solution, so no need to look into this! Just need to shift all indices by 1 now..

lemunozm: @wischli if we exceed 32 messages probably we should transform the variable to `bytes`

lemunozm: > Just need to shift all indices by 1 now

What do you mean by this?

wischli: Currently, the message kind is represented by a single byte. After the change, it's represented by two bytes which shifts everything by a byte during deserialization.

lemunozm: Here's my proposal: https://github.com/centrifuge/protocol-v3/pull/217

src/vaults/BalanceSheetManager.sol
```diff
@@ -219,6 +219,27 @@ contract BalanceSheetManager is
...
+        // TODO(wischli): Rm hardcode post #187 merge
+        D18 pricePoolPerAsset = d18(1);
+        JournalEntry[] memory journalEntries = new JournalEntry[](0);
+        Meta memory meta = Meta(journalEntries, journalEntries);
+
+        escrow.pendingDepositIncrease(token, tokenId, poolId.raw(), scId.raw(), assetAmount);
+        sender.sendUpdateHoldingAmount(
+            poolId, scId, assetId, address(escrow), assetAmount, pricePoolPerAsset, true, meta
+        );
+        this.updateValue(poolId, scId, token, tokenId, pricePoolPerAsset);
```
wischli: TBD: Maybe remove

src/vaults/BalanceSheetManager.sol
```diff
@@ -219,6 +219,27 @@ contract BalanceSheetManager is
...
 
+    /// @inheritdoc IBalanceSheetManagerGatewayHandler
+    function approvedDeposits(PoolId poolId, ShareClassId scId, AssetId assetId, uint128 assetAmount) external auth {
+        (address token, uint256 tokenId) = poolManager.checkedIdToAsset(assetId.raw());
+        // TODO(wischli): Rm hardcode post #187 merge
+        D18 pricePoolPerAsset = d18(1);
+        JournalEntry[] memory journalEntries = new JournalEntry[](0);
+        Meta memory meta = Meta(journalEntries, journalEntries);
+
+        escrow.pendingDepositIncrease(token, tokenId, poolId.raw(), scId.raw(), assetAmount);
```
mustermeiszer: `escrow.depositIncrease()`

mustermeiszer (commented):
Missing 

* `AsyncInvestmentManager.requestDeposit()` -> `escrow.pendingDepositIncrease`

src/vaults/BalanceSheetManager.sol
```diff
@@ -219,6 +219,27 @@ contract BalanceSheetManager is
...
+        sender.sendUpdateHoldingAmount(
+            poolId, scId, assetId, address(escrow), assetAmount, pricePoolPerAsset, true, meta
+        );
+        this.updateValue(poolId, scId, token, tokenId, pricePoolPerAsset);
+    }
+
+    /// @inheritdoc IBalanceSheetManagerGatewayHandler
+    function revokedShares(PoolId poolId, ShareClassId scId, AssetId assetId, uint128 assetAmount) external auth {
+        (address token, uint256 tokenId) = poolManager.checkedIdToAsset(assetId.raw());
+        escrow.reserveIncrease(token, tokenId, poolId.raw(), scId.raw(), assetAmount);
```
mustermeiszer: NOTE: 

* For new approach: need to find a substitute for reserve action

mustermeiszer: Missing: 

* `AsyncInvestmentManager.redeem` to reduce reservance on perPoolEscrow

src/pools/PoolRouter.sol
```diff
@@ -219,14 +219,7 @@ contract PoolRouter is Auth, Multicall, IPoolRouter, IPoolRouterGatewayHandler {
...
         );
 
-        uint128 valueChange = holdings.increase(unlockedPoolId, scId, paymentAssetId, valuation, approvedAssetAmount);
-
-        accounting.addCredit(
-            holdings.accountId(unlockedPoolId, scId, paymentAssetId, uint8(AccountType.Equity)), valueChange
-        );
-        accounting.addDebit(
-            holdings.accountId(unlockedPoolId, scId, paymentAssetId, uint8(AccountType.Asset)), valueChange
-        );
```
lemunozm: IIUC, this will send the message to CV, and later CV to CP to store the increase holdings, right? Is not to much messaging? Why not update the holdings directly here?

I think there is no need on the CV side when approveDeposit is received to call the entire `approveDeposits()` method. Could not avoid sending back the message and just update holdings here?

wischli: Ref: https://kflabs.slack.com/archives/C07PG2EUR9C/p1743689530352989?thread_ts=1743669923.076539&cid=C07PG2EUR9C

In short: This way, all holdins changes are invoked from CV. The round trips are definitely "an issue". I don't have a solution for that, so for the sake of limited time, better go with that than nothing. Maybe you have a proposal for a better way?

test/pools/integration/Cases.t.sol
```diff
@@ -211,6 +221,19 @@ contract TestCases is PoolsDeployer, Test {
             NAV_PER_SHARE.mulUint128(uint128(valuation.getQuote(APPROVED_SHARE_AMOUNT, USD.addr(), USDC_C2.addr())))
         );
         assertEq(m0.shareAmount, APPROVED_SHARE_AMOUNT);
+
+        // Simulate holding update response messages from CV after user claimed redeem
```
lemunozm: Why do we need to simulate a holding update after a user claims? AFAIK, the holdings are already updated at this point. Would you want to move this section to line 206?

wischli: Because we are moving the holding update from `Hub.revokeShares` ([ref](https://github.com/centrifuge/protocol-v3/pull/208/files#diff-7689ef5d9d6af15897386cbcc13ecfd65e52adbbc611351f49b811594895464fL268-L275)) to `CV.AsyncRequests.withdraw` (correct contract not fully sure yet) which then initiates the holding update messages via the BalanceSheet, see https://kflabs.slack.com/archives/C07PG2EUR9C/p1743689530352989?thread_ts=1743669923.076539&cid=C07PG2EUR9C

> Redeem
> * CP: issuer revokeShares -> message to CV about amount and price
>    * CV: escrow.increaseReserved(..) 
>* CV: User claims their tokens - escrow.decreaseReserve() , escrow.withdraw() and bsm.updateHoldings()
>    * CP: just adapts holdings accordingly
>
> -> Change would be not to adapt holdings inside of revokeShares

lemunozm: Thanks for the ref 🙌🏻. Just a comment regarding this. The returning message will use subsidized with the pool funds. I'm not sure if we want permissionless calls spamming subsided pool messages.

An investor could do small investments that involves small claims to drain the subsidized funds for that pool. cc @mustermeiszer

wischli: Wouldn't it be an easy fix to require claiming redemptions to be paid by the user then?

src/pools/PoolRouter.sol
lemunozm: > so for the sake of limited time, better go with that than nothing

Yeah, I do not want to add noise.

My proposal would be to just increase the holding in CP side just when `approveRedeem` is called. CV side processing will not send any message back, just process them and dispatch any required event. But not sure if I miss something

test/pools/integration/Cases.t.sol
lemunozm: They are paid already by the used, the problem is the update holding message they will dispatch in the other side. But IIUC, @mustermeiszer said that this is not the case. Then I do not understand why they are there in the tests

mustermeiszer: No, it currently definitely is. That is fine for now. We most likely will add a queu to the BSM and then it is batched as a summary at one point.

src/vaults/BalanceSheetManager.sol
wischli: > NOTE:
> 
> * For new approach: need to find a substitute for reserve action

@mustermeiszer Can you elaborate on that? It seems like it requires a code change for this PR and I don't fully know what/how.

wischli: > Missing:
> 
> * `AsyncInvestmentManager.redeem` to reduce reservance on perPoolEscrow

Just to clarify: IIUC, `AsyncRequests.redeem` needs to do two things:
1. `escrow.reserveDecrease` (reverting increase from `bsm.revokedShares`)
2. `bsm.withdraw` (which decreases the deposit amount that is increased as a result of `bsm.approvedDeposits`)

src/vaults/BalanceSheetManager.sol
wischli: Fixed

src/vaults/BalanceSheet.sol
```diff
@@ -199,6 +204,26 @@ contract BalanceSheet is Auth, Recoverable, IBalanceSheet, IBalanceSheetGatewayH
         _revoke(poolId, scId, from, pricePoolPerShare, shares);
     }
 
+    /// @inheritdoc IBalanceSheetGatewayHandler
+    function approvedDeposits(PoolId poolId, ShareClassId scId, AssetId assetId, uint128 assetAmount) external auth {
+        (address token, uint256 tokenId) = poolManager.idToAsset(assetId.raw());
+        D18 priceAssetPerShare = syncRequests.priceAssetPerShare(poolId.raw(), scId.raw(), assetId.raw());
```
mustermeiszer: CR

`poolManager.pricePoolPerAsset`

src/vaults/AsyncRequests.sol
```diff
@@ -439,16 +449,35 @@ contract AsyncRequests is BaseInvestmentManager, IAsyncRequests {
...
+    function _withdraw(address vaultAddr, address receiver, uint128 assets) internal {
+        VaultDetails memory vaultDetails = poolManager.vaultDetails(vaultAddr);
+
+        IAsyncVault vault_ = IAsyncVault(vaultAddr);
+        uint64 poolId = vault_.poolId();
+        bytes16 scId = vault_.trancheId();
+        JournalEntry[] memory journalEntries = new JournalEntry[](0);
+        Meta memory meta = Meta(journalEntries, journalEntries);
+
+        D18 priceAssetPerShare = syncRequests.priceAssetPerShare(poolId, scId, vaultDetails.assetId);
```
mustermeiszer: CR

`poolManager.PoolPerAsset()`

src/vaults/AsyncRequests.sol
```diff
@@ -439,16 +449,35 @@ contract AsyncRequests is BaseInvestmentManager, IAsyncRequests {
         state.maxWithdraw = state.maxWithdraw > assetsUp ? state.maxWithdraw - assetsUp : 0;
 
         if (assetsDown > 0) {
-            VaultDetails memory vaultDetails = poolManager.vaultDetails(address(vault_));
-
-            if (vaultDetails.tokenId == 0) {
-                SafeTransferLib.safeTransferFrom(vaultDetails.asset, address(escrow), receiver, assetsDown);
-            } else {
-                IERC6909(vaultDetails.asset).transferFrom(address(escrow), receiver, vaultDetails.tokenId, assetsDown);
-            }
```
mustermeiszer, line 448: CR

We still need the token transfer here.

mustermeiszer (commented):
Less logic than I thought! Nice!

src/vaults/AsyncRequests.sol
wischli: The transfer is executed in the BalanceSheet: https://github.com/centrifuge/protocol-v3/blob/ed400eaf8693a90d90169688bff37ab1ff96b177/src/vaults/BalanceSheet.sol#L261-L265

mustermeiszer: Ah nice!!

src/vaults/AsyncRequests.sol
wischli: fixed

src/vaults/BalanceSheet.sol
wischli: OMG so sorry for the continued screw ups. Now everything should be perfect. Thanks for your patience!

src/hub/Hub.sol
```diff
@@ -233,14 +233,7 @@ contract Hub is Multicall, Auth, Recoverable, IHub, IHubGatewayHandler {
...
 
-        uint128 valueChange = holdings.increase(unlockedPoolId, scId, paymentAssetId, valuation, approvedAssetAmount);
-
-        accounting.addCredit(
-            holdings.accountId(unlockedPoolId, scId, paymentAssetId, uint8(AccountType.Equity)), valueChange
-        );
-        accounting.addDebit(
-            holdings.accountId(unlockedPoolId, scId, paymentAssetId, uint8(AccountType.Asset)), valueChange
-        );
+        sender.sendApprovedDeposits(unlockedPoolId, scId, paymentAssetId, approvedAssetAmount);
```
wischli: Note: This PR makes CV the single provider of holding updates. When it receives `ApprovedDeposits`, it dispatches the holding update. It's clear this round trip is more costly but it comes with a couple of business side advantages ([more info](https://kflabs.slack.com/archives/C07PG2EUR9C/p1743689530352989?thread_ts=1743669923.076539&cid=C07PG2EUR9C)).

src/hub/Hub.sol
```diff
@@ -267,14 +260,7 @@ contract Hub is Multicall, Auth, Recoverable, IHub, IHubGatewayHandler {
...
 
-        uint128 valueChange = holdings.decrease(unlockedPoolId, scId, payoutAssetId, valuation, payoutAssetAmount);
-
-        accounting.addCredit(
-            holdings.accountId(unlockedPoolId, scId, payoutAssetId, uint8(AccountType.Asset)), valueChange
-        );
-        accounting.addDebit(
-            holdings.accountId(unlockedPoolId, scId, payoutAssetId, uint8(AccountType.Equity)), valueChange
-        );
+        sender.sendRevokedShares(unlockedPoolId, scId, payoutAssetId, payoutAssetAmount);
```
wischli: Holding update is now triggered when a user claims a redemption on CV side inside `AsyncRequests.withdraw`(see `AsyncRequests._withdraw`).

src/vaults/AsyncRequests.sol
```diff
@@ -111,12 +125,9 @@ contract AsyncRequests is BaseInvestmentManager, IAsyncRequests {
...
         require(state.pendingCancelDepositRequest != true, "AsyncRequests/cancellation-is-pending");
 
-        state.pendingDepositRequest = state.pendingDepositRequest + _assets;
-        VaultDetails memory vaultDetails = poolManager.vaultDetails(address(vault_));
-
-        sender.sendDepositRequest(
-            vault_.poolId(), vault_.trancheId(), controller.toBytes32(), vaultDetails.assetId, _assets
-        );
+        state.pendingDepositRequest += assets;
+        balanceSheet.escrow().pendingDepositIncrease(vaultDetails.asset, vaultDetails.tokenId, poolId, scId, assets);
```
wischli, line 129: The escrow access can be improved but since we want to refactor the PerPoolEscrow anyways, I didn't put much effort into the current usage.

lemunozm (dismissed):
I read everything carefully a not a single NIT. Awesome work @wischli. Also thanks a lot for your auto comments, just to the point to understand everything well 🙌🏻

src/vaults/AsyncRequests.sol
```diff
@@ -439,16 +449,36 @@ contract AsyncRequests is BaseInvestmentManager, IAsyncRequests {
...
     }
 
+    /// @dev Transfer funds from escrow to receiver and update holdings
+    function _withdraw(address vaultAddr, address receiver, uint128 assets) internal {
+        VaultDetails memory vaultDetails = poolManager.vaultDetails(vaultAddr);
+
+        IAsyncVault vault_ = IAsyncVault(vaultAddr);
+        uint64 poolId = vault_.poolId();
+        bytes16 scId = vault_.trancheId();
+        JournalEntry[] memory journalEntries = new JournalEntry[](0);
```
hieronx, line 463: Always having to initialize this is imo a strong argument for adding a `deposit` and `withdraw` function to `BalanceSheet` without the `meta` param @mustermeiszer

hieronx (dismissed):

src/vaults/AsyncRequests.sol
lemunozm: We would also need new serialize/deserialize methods I guess. Is there no way in solidity to pass like a null ptr, indicating the array is empty?

hieronx (dismissed):

hieronx (dismissed):

hieronx (dismissed):

foundry.toml
```diff
@@ -7,7 +7,7 @@ solc_version = "0.8.28"
 evm_version = "cancun"
 
 optimizer = true
-optimizer_runs = 90
+optimizer_runs = 1
```
lemunozm, line 10: Why this? Hub size issues?

wischli: Not sure where this comes from. I didnt change anything in the toml.

lemunozm: Ohh, it comes from a @hieronx PR. Is this a local testing artifact, Jeroen?

hieronx: It was while trying to get the `Hub` contract below the limit again. Not sure it's still needed. Let's just add a ticket to update this to the highest possible value just before audit?

hieronx: https://github.com/centrifuge/protocol-v3/issues/162#issuecomment-2800780366

PR #211, opened 2025-04-07
Minor gas optimizations in token

hieronx:


wischli (approved):
The snapshots proved you right, if there was any doubt :rocket:

PR #213, opened 2025-04-07
fix: ShareClassId collision

### LINKED ISSUE #170: Collisions in ShareClassId
The way we create a `ShareClassId` right now could create collisions: https://github.com/centrifuge/protocol-v3/blob/main/src/pools/types/ShareClassId.sol#L21

- If we can have a PoolId 23 and index 1 => ShareClassId 24
- If we can have a PoolId 22 and index 2 => ShareClassId 24

I think we need here: `(uint128(poolId) << 64) + index`

wischli:
Fixes https://github.com/centrifuge/protocol-v3/issues/170

### DIFF
--- a/src/common/types/ShareClassId.sol
+++ b/src/common/types/ShareClassId.sol
@@ -18,7 +18,7 @@ function raw(ShareClassId scId) pure returns (bytes16) {
 }
 
 function newShareClassId(PoolId poolId, uint32 index) pure returns (ShareClassId scId) {
-    return ShareClassId.wrap(bytes16(uint128(PoolId.unwrap(poolId)) + index));
+    return ShareClassId.wrap(bytes16(uint128(PoolId.unwrap(poolId) << 64) + index));
 }
 
 using {isNull, raw, equals as ==} for ShareClassId global;
--- a/src/common/types/ShareClassId.sol
+++ b/src/common/types/ShareClassId.sol
@@ -18,7 +18,7 @@ function raw(ShareClassId scId) pure returns (bytes16) {
 }
 
 function newShareClassId(PoolId poolId, uint32 index) pure returns (ShareClassId scId) {
-    return ShareClassId.wrap(bytes16(uint128(PoolId.unwrap(poolId) << 64) + index));
+    return ShareClassId.wrap(bytes16((uint128(PoolId.unwrap(poolId)) << 64) + index));
 }
 
 using {isNull, raw, equals as ==} for ShareClassId global;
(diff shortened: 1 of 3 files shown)


test/pools/unit/types/ShareClassId.t.sol
```diff
@@ -21,6 +21,18 @@ contract ShareClassIdTest is Test {
         PoolId poolId = PoolId.wrap(poolId_);
         ShareClassId scId = newShareClassId(poolId, index);
 
-        assertEq(scId.raw(), bytes16(uint128(poolId.raw()) + index));
+        assertEq(scId.raw(), bytes16(uint128(poolId.raw() << 64) + index));
+    }
+
+    function testShareClassIdCollosionResistance(uint64 poolId, uint32 index) public pure {
```
hieronx: Wdyt about adding a second `poolId` and `index` input, and ensure that no share class ID of `poolId1` can match no share class ID of `poolId2`?

src/common/types/ShareClassId.sol
```diff
@@ -18,7 +18,7 @@ function raw(ShareClassId scId) pure returns (bytes16) {
 }
 
 function newShareClassId(PoolId poolId, uint32 index) pure returns (ShareClassId scId) {
-    return ShareClassId.wrap(bytes16(uint128(PoolId.unwrap(poolId)) + index));
+    return ShareClassId.wrap(bytes16(uint128(PoolId.unwrap(poolId) << 64) + index));
```
lemunozm: CR: 
```suggestion
    return ShareClassId.wrap(bytes16((uint128(PoolId.unwrap(poolId)) << 64) + index));
```

lemunozm: If not, we're shifting 64 bits over an `uint64`, so we're discarding the bytes. The tests are not notifying this because they assume the same

test/pools/unit/types/ShareClassId.t.sol
```diff
@@ -21,6 +21,27 @@ contract ShareClassIdTest is Test {
...
+        vm.assume(poolId2 != poolId1);
+        vm.assume(index1 != index2);
+
+        assertNotEq(
+            newShareClassId(PoolId.wrap(poolId1), index1).raw(), newShareClassId(PoolId.wrap(poolId2), index2).raw()
+        );
+
+        ShareClassId scId1 = newShareClassId(PoolId.wrap(poolId1), index1);
+        ShareClassId scId2 = newShareClassId(PoolId.wrap(poolId1 - 1), index1 + 1);
+        ShareClassId scId3 = newShareClassId(PoolId.wrap(poolId1 + 1), index1 - 1);
```
lemunozm, line 42: Good checks!

lemunozm (approved):
LGTM!

PR #215, opened 2025-04-07
Cleaning stuff gateway related

lemunozm:
Just some cleanings regarding the Gateway to leave everything done in this regard:
- Rename topUp -> payTrasaction
- Errors as types
- Some other minor stuff


src/common/Gateway.sol
```diff
@@ -233,8 +233,8 @@ contract Gateway is Auth, IGateway, IRecoverable {
         bytes32 messageHash = keccak256(message);
         uint256 recovery = recoveries[chainId][adapter][messageHash];
 
-        require(recovery != 0, "Gateway/message-recovery-not-initiated");
-        require(recovery <= block.timestamp, "Gateway/challenge-period-has-not-ended");
+        require(recovery != 0, MessageRecoveryNotInitiated());
+        require(recovery <= block.timestamp, MessageRecoveryPeriodNotEnded());
```
hieronx: I feel the challenge period terminology here is helpful. Since now it seems like there is a period in which the recovery is handled.

src/pools/PoolRouter.sol
```diff
@@ -573,7 +573,7 @@ contract PoolRouter is Auth, Multicall, IPoolRouter, IPoolRouterGatewayHandler {
     /// @notice Send native tokens to the gateway for transaction payment if it's not in a multicall.
     function _pay() internal {
         if (!gateway.isBatching()) {
-            gateway.topUp{value: msg.value}();
+            gateway.payTransaction{value: msg.value}();
```
hieronx: Maybe just even simpler, `Gateway.pay()`?

src/common/Gateway.sol
```diff
@@ -104,15 +104,15 @@ contract Gateway is Auth, IGateway, IRecoverable {
 
             // Enable new adapters, setting quorum to number of adapters
             for (uint8 j; j < quorum_; j++) {
-                require(_activeAdapters[chainId][addresses[j]].id == 0, "Gateway/no-duplicates-allowed");
+                require(_activeAdapters[chainId][addresses[j]].id == 0, NoDuplicatedAllowed());
```
wischli: Nit
```suggestion
                require(_activeAdapters[chainId][addresses[j]].id == 0, NoDuplicatesAllowed());
```

src/pools/PoolRouter.sol
wischli: Or `Gateway.prepay()`?

wischli (commented):
In the `GasService`, we can remove an unused dependency: https://github.com/centrifuge/protocol-v3/pull/206#discussion_r2030768646

src/common/Gateway.sol
lemunozm: Will add! 🙏🏻

src/pools/PoolRouter.sol
lemunozm: Could `pay` be confused with a more general pay? Using `payTransacion` also helps tie in with `PaymentMethod.Transaction`

src/common/Gateway.sol
lemunozm: I'm always amazed at your eye for these kinds of things!

src/pools/PoolRouter.sol
hieronx: I'm fine with sticking to `payTransaction` if you feel that is clearer!

hieronx (approved):

wischli (approved):

PR #217, opened 2025-04-07
MessageLib: Support 64 messages

lemunozm:
Add support for 64 max messages in `MessageLib`.

I tried using bytes, but I can not operate easily on it being constant/immutable. I think the easier solution is to have two `uint256`


src/common/libraries/MessageLib.sol
```diff
@@ -126,7 +129,9 @@ library MessageLib {
         uint8 kind = message.toUint8(0);
         require(kind <= uint8(type(MessageType).max), UnknownMessageType());
```
wischli, line 130: This will still break with more than 32 messages because be cannot serialize the kind to `uint8` then anymore. I don't think there is a way around using `uint16` here.

wischli (commented):
WDYT about using a function for the message length? See https://github.com/centrifuge/protocol-v3/pull/219

src/common/libraries/MessageLib.sol
lemunozm: Why not? `uint8` can hold 256 different messages

wischli: Sorry, logical mistake on my side. You are right.

wischli (approved):
Sorry for the logical error and thanks for the quick fix!

PR #219, opened 2025-04-07
[DRAFT] Idea: extend message lib

wischli:
@lemunozm Counter proposal for #217


src/common/libraries/MessageLib.sol
```diff
@@ -78,76 +78,77 @@ library MessageLib {
...
+        if (kind == uint16(MessageType.TriggerRedeemRequest)) return 90;
+        if (kind == uint16(MessageType.UpdateHoldingAmount)) return 115; // dynamic
+        if (kind == uint16(MessageType.UpdateHoldingValue)) return 66;
+        if (kind == uint16(MessageType.UpdateShares)) return 99;
+        // if (kind == uint16(MessageType.ApprovedDeposits)) return 58;
+        // if (kind == uint16(MessageType.RevokedShares)) return 58;
+        if (kind == uint16(MessageType.UpdateJournal)) return 10; // dynamic
+        if (kind == uint16(MessageType.TriggerUpdateHoldingAmount)) return 108; // dynamic
+        if (kind == uint16(MessageType.TriggerUpdateShares)) return 92;
+        revert UnknownMessageType();
```
lemunozm, line 117: The problem with this is that it's much more slow if the messages are the last ones in the list, that was the reason to encode it in a single constant

src/common/libraries/MessageLib.sol
```diff
@@ -78,76 +78,77 @@ library MessageLib {
...
 
     function messageType(bytes memory message) internal pure returns (MessageType) {
-        return MessageType(message.toUint8(0));
+        return MessageType(message.toUint16(0));
     }
 
-    function messageCode(bytes memory message) internal pure returns (uint8) {
-        return message.toUint8(0);
+    function messageCode(bytes memory message) internal pure returns (uint16) {
+        return message.toUint16(0);
```
lemunozm, line 125: There is no need to use uint16 since with uint8 we have until 256 messages

PR #220, opened 2025-04-07
Merge asset registry into `HubRegistry`, remove ERC6909 impl

hieronx:


src/misc/interfaces/IERC6909.sol
```diff
@@ -116,6 +116,12 @@ interface IERC6909TotalSupplyExt {
     function totalSupply(uint256 tokenId) external returns (uint256 supply);
 }
 
+interface IERC6909Decimals {
```
lemunozm, line 119: Should `IERC6909MetadataExt` extends from `IERC6909Decimals`?

src/pools/PoolRegistry.sol
```diff
@@ -14,13 +15,24 @@ contract PoolRegistry is Auth, IPoolRegistry {
...
     constructor(address deployer) Auth(deployer) {}
 
+    /// @inheritdoc IPoolRegistry
+    function registerAsset(AssetId assetId, uint8 decimals_) external auth {
+        require(_decimals[assetId] == 0, AssetAlreadyRegistered());
+
+        _decimals[assetId] = decimals_;
+
+        emit NewAsset(assetId, decimals_);
+    }
```
lemunozm: I like the simplification of this PR, but I have just one "disagree" regarding having this pool-unrelated method into the `PoolRegistry`. Given this is global, why not rescue some previous idea from you about have a general contract for general stuff like:

`HubRegistry.setValue(bytes key, uint256 value)` where key here is `keccak("asset", assetId)`? (although not sure about the performance). If not, I think we could have a super small `AssetRegistry` contract where to store this, but having it inside PoolRegistry seems very weird.

src/pools/PoolRegistry.sol
```diff
@@ -29,6 +41,7 @@ contract PoolRegistry is Auth, IPoolRegistry {
     {
         require(admin_ != address(0), EmptyAdmin());
         require(!currency_.isNull(), EmptyCurrency());
+        require(currency[poolId].isNull(), PoolAlreadyRegistered());
```
lemunozm, line 44: Good catch!

src/pools/PoolRegistry.sol
```diff
@@ -76,7 +89,24 @@ contract PoolRegistry is Auth, IPoolRegistry {
         emit UpdateCurrency(poolId, currency_);
     }
 
+    function decimals(PoolId poolId) public view returns (uint8 decimals_) {
```
lemunozm, line 91: NIT: /// @inheritdoc IPoolRegistry

src/pools/PoolRegistry.sol
lemunozm: Another option could be to rename `PoolRegistry` to `HubRegistry`

src/pools/PoolRouter.sol
```diff
@@ -436,21 +415,18 @@ contract PoolRouter is Auth, Multicall, IPoolRouter, IPoolRouterGatewayHandler {
...
+        poolRegistry.registerAsset(assetId, decimals);
     }
 
     /// @inheritdoc IPoolRouterGatewayHandler
     function depositRequest(PoolId poolId, ShareClassId scId, bytes32 investor, AssetId depositAssetId, uint128 amount)
         external
         auth
     {
-        address pendingShareClassEscrow = escrow(poolId, scId, EscrowId.PendingShareClass);
-        assetRegistry.mint(pendingShareClassEscrow, depositAssetId.raw(), amount);
```
lemunozm: Wow! this is a big concept change. Are we sure we do not want to differentiate between deployment & undeployment capital? Not sure, right now, if that information can be retrieved by the `ShareClassManager`

lemunozm (commented):
I really like the idea and how it simplify things.

Nevertheless mixing assets in the `PoolRegistry` seems unnatural. I would rather have 2 contracts or I would rename `PoolRegistry` to `HubRegistry`. I would prefer the former:

- PoolRegistry
- HubRegsitry
  - mapping(assets => decimals)
  - mapping(bytes32 => uint256 value) // for future usages and avoid migrations

src/pools/PoolRegistry.sol
hieronx: I like the idea of the renaming.

I think we shouldn't add another contract, we just will add even more external calls.

src/misc/interfaces/IERC6909.sol
hieronx: Yeah I was still not sure who to set up these interfaces now. That seems like the simplest solution but then the interface ID of `IERC6909MetadataExt` will be wrong, since interface IDs don't account for inherited interfaces.

src/pools/PoolRouter.sol
hieronx: I don't believe that is really relevant on the share class side anymore. That can always be queried on the vaults side, by looking at the available balance of the escrow. I feel that is sufficient now.

src/misc/interfaces/IERC6909.sol
lemunozm: 😲 interesting!

src/pools/PoolRouter.sol
lemunozm: Cool!!

.forge-snapshots/AsyncVault_requestDeposit.snap
```diff
@@ -1 +1 @@
-387182
\ No newline at end of file
+418500
```
hieronx, line 1: I don't understand why these are going up so much

wischli: Are you locally using an the stable foundry version?

src/misc/interfaces/IERC6909.sol
```diff
@@ -116,7 +116,13 @@ interface IERC6909TotalSupplyExt {
...
 }
 
-interface IERC6909MetadataExt {
+interface IERC6909Decimals {
+    /// @notice             Used to retrieve the decimals of an asset
+    /// @dev                address is used but the value corresponds to a AssetId
+    function decimals(uint256 assetId) external view returns (uint8);
+}
+
+interface IERC6909MetadataExt is IERC6909Decimals {
```
lemunozm, line 125: Do we want to keep the interfaces by now? These interfaces already exist in `solmate` and other projects

wischli (approved):
Not even a single nit from my side, great mopping!

lemunozm (commented):
Really really like this simplification!!!

lemunozm (approved):

src/misc/interfaces/IERC6909.sol
hieronx: It is still used by the mock erc6909 implementation, which is used in the vaults tests

.forge-snapshots/AsyncVault_requestDeposit.snap
lemunozm: I can confirm the issue is NOT related to the foundry usage. I got the same values locally 🤔

lemunozm: I think there were already not updated in previous PR, because I got the same without your commit

lemunozm: They seem to be outdated from previous commits, weird!

PR #221, opened 2025-04-07
Rename chainId to centrifugeId

lemunozm:
- Rename `chainId`, `centrifugeChainId`, `domainId`, `destinationId`, `remoteId` to `centrifugeId`.
- Reorganize storages in Gateway

I've not touched many tests to avoid insane conflicts among us. That can be fixed once we leave the test clean.


src/common/Gateway.sol
```diff
@@ -33,39 +33,63 @@ contract Gateway is Auth, IGateway, IRecoverable {
...
+    IRoot public immutable root;
+    IGasService public gasService;
+    IMessageProcessor public processor;
 
-    /// @notice Tells is the gateway is actually configured to create batches
+    //-----------------------------
+    // Batching
+    //-----------------------------
+
+    /// @notice Tells if the gateway is actually configured to create batches
```
hieronx: My personal opinion is that these extensive comments to the state add quite little value (the names + types are pretty explicit on their own), and removing them (as I had proposed on Slack) makes it more readable.

I also think the 3 line section headers within state have a similar effect of making it too extensive and thus less readable.

I don't want to block this PR but would like to align on this before the final audit, in the style guide.

src/common/Gateway.sol
```diff
@@ -216,93 +236,98 @@ contract Gateway is Auth, IGateway, IRecoverable {
...
-            currentAdapter.send{value: consumed}(chainId, payload, gasLimit, address(this));
+            currentAdapter.send{value: consumed}(centrifugeId, payload, gasLimit, address(this));
         }
 
         emit SendMessage(message);
     }
 
+    function subsidizePool(PoolId poolId) external payable {
+        subsidy[poolId] += msg.value;
+        emit ReceiveNativeTokens(poolId, msg.sender, msg.value);
```
hieronx: Think we should just rename this now
```suggestion
        emit SubsidizePool(poolId, msg.sender, msg.value);
```

hieronx (dismissed):

src/common/Gateway.sol
lemunozm: > My personal opinion is that these extensive comments to the state add quite little value (the names + types are pretty explicit on their own), and removing them (as I had proposed on Slack) makes it more readable.

After being honest with me, I think you're right here.

> I also think the 3 line section headers within state have a similar effect of making it too extensive and thus less readable.

However, I have the opposite feeling here. They create free space in the code that allows to visually see the different parts faster. I already used in the Hub and I felt very easy to find the location you're searching for: https://github.com/centrifuge/protocol-v3/blob/main/src/pools/Hub.sol

One-line headers are very easy to confuse with normal comments, and sometimes they keep unrefactored and outdated after some addition/removal methods

hieronx: > However, I have the opposite feeling here. They create free space in the code that allows to visually see the different parts faster. I already used in the Hub and I felt very easy to find the location you're searching for: https://github.com/centrifuge/protocol-v3/blob/main/src/pools/Hub.sol

> One-line headers are very easy to confuse with normal comments, and sometimes they keep unrefactored and outdated after some addition/removal methods

To be clear, I completely agree the multi line section headers are great within the overall code, to add organization. I just feel specifically within the storage list, they add too much separation.

hieronx: So my proposal would be 1 line headers within a section like the storage, and then 3 line headers to separate bigger code sections in the contract.

lemunozm: Ok! I agree then 👍🏻. I've removed the 3 line section of the storage part

src/common/Gateway.sol
lemunozm: Done

src/vaults/interfaces/IVaultRouter.sol
```diff
@@ -151,15 +151,21 @@ interface IVaultRouter is IMulticall, IRecoverable {
...
+    function transferShares(uint16 centrifugeId, address vault, bytes32 receiver, uint128 amount) external payable;
 
     /// @notice This is a more friendly version where the receiver is and EVM address
     /// @dev    The receiver address is padded to 32 bytes internally
-    function transferShares(address vault, uint16 chainId, address receiver, uint128 amount) external payable;
+    function transferShares(uint16 centrifugeId, address vault, address receiver, uint128 amount) external payable;
+
+    /// @notice Register an asset to be used in a pool existing in centrifugeId
+    /// @param centrifugeId Where the asset will be registered
+    /// @param asset If tokenId == 0, an ERC20 address, if tokenId != 0, and ERC6909 address
```
wischli, line 166: Nit:
```suggestion
    /// @param asset If tokenId == 0, an ERC20 address, if tokenId != 0, an ERC6909 address
```

hieronx (approved):
Super clean!

src/vaults/interfaces/IVaultRouter.sol
```diff
@@ -151,15 +151,21 @@ interface IVaultRouter is IMulticall, IRecoverable {
...
+    /// @param tokenId An ERC6909 tokenId
+    function registerAsset(uint16 centrifugeId, address asset, uint256 tokenId) external payable;
```
wischli, line 168: Curious why this was added to the `VaultRouter` interace. Is the current `PoolManager` exposure insufficient?

wischli (commented):
A few seconds too late. Great job! Incredible how many codepaths were touched. I like having the `centrifugeId` as first argument in corresponding calls!

src/vaults/interfaces/IVaultRouter.sol
lemunozm: Without exposing it, we're not able to:
1. Pay for the underlying message
2. batch it

lemunozm: Actually we should make it auth the PoolManager method to avoid any user to call it and drain the pool0 funds (funds intented for unrelated pool method). Will add that in follow up PR

mustermeiszer: Makes sense! Users can still pay for that in the router, right? Great catch!

lemunozm: Will do there, yes

src/vaults/interfaces/IVaultRouter.sol
lemunozm: Fixed in the pools -> hub PR

PR #227, opened 2025-04-08
Unify recoverTokens method, send the message, add TokenRecoverer

lemunozm:
- [x] Add `Recoverable` abstract contract with all the recovery logic
- [x] `Guardian` able to send messages
- [x] Add `TokenRecoverer` contract


src/common/interfaces/IRoot.sol
```diff
@@ -91,5 +78,5 @@ interface IRoot {
 
     /// --- Token Recovery ---
     /// @notice Allows Governance to recover tokens sent to the wrong contract by mistake
-    function recoverTokens(address target, address token, uint256 tokenId, address to, uint256 amount) external;
+    function recoverTokens(address target, address token, address to, uint256 amount) external;
```
lemunozm: I think this can be fully removed if added `TokenRecoverer`.

src/misc/Recoverable.sol
```diff
@@ -0,0 +1,28 @@
...
+
+import {Auth} from "src/misc/Auth.sol";
+import {IERC6909} from "src/misc/interfaces/IERC6909.sol";
+import {SafeTransferLib} from "src/misc/libraries/SafeTransferLib.sol";
+
+import {IRecoverable} from "src/misc/interfaces/IRecoverable.sol";
+
+abstract contract Recoverable is Auth, IRecoverable {
+    /// @inheritdoc IRecoverable
+    function recoverTokens(address token, address receiver, uint256 amount) public auth {
```
hieronx, line 12: What's the reason for having both function definitions here?

lemunozm: We need the old one to allow retro-compatibility with Vaults.

But I've just realized that I need to call the smaller form in the TokenReceiver if tokenId != 0. Will add that line

hieronx: > We need the old one to allow retro-compatibility with Vaults.

True!

lemunozm: Fix here [31d60a5](https://github.com/centrifuge/protocol-v3/pull/227/commits/31d60a5e05a6a825400df7980531eee312da1f21)

src/hub/Hub.sol
```diff
@@ -284,6 +287,8 @@ contract Hub is Auth, Multicall, IHub, IHubGatewayHandler {
     ) public payable {
         _protectedAndUnlocked();
 
+        require(shareClassManager.exists(unlockedPoolId, scId), IShareClassManager.ShareClassNotFound());
```
lemunozm, line 290: Extra: adding those missing checks

hieronx: Good idea

hieronx (approved):

PR #228, opened 2025-04-08
Move estimate() to MesageDispatcher

lemunozm:
`estimate()` method should be better in `MessageDispatcher`, because:
- It's shared across Hub and Vaults.
- It already knows about `localCentrifugeId`, so we can remove the `localCentrifugeId` from VaultRouter too.

### DIFF
--- a/src/common/interfaces/IMessageDispatcher.sol
+++ b/src/common/interfaces/IMessageDispatcher.sol
@@ -1,0 +1,22 @@
+// SPDX-License-Identifier: BUSL-1.1
+pragma solidity >=0.5.0;
+
+import {IVaultMessageSender, IPoolMessageSender, IRootMessageSender} from "src/common/interfaces/IGatewaySenders.sol";
+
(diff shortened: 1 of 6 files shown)


src/common/interfaces/IMessageDispatcher.sol
```diff
@@ -0,0 +1,22 @@
+// SPDX-License-Identifier: BUSL-1.1
+pragma solidity >=0.5.0;
```
wischli, line 2: Nit: 
```suggestion
pragma solidity 0.8.28;
```

wischli (dismissed):

src/common/interfaces/IMessageDispatcher.sol
hieronx: Interface pragmas should be `>=0.5.0`. This allows integrators of Centrifuge contracts to import the interfaces, regardless of the Solidity version they use.

hieronx (dismissed):

hieronx (dismissed):

mustermeiszer (approved):

PR #230, opened 2025-04-09
Fix issue with Gateway proofs

lemunozm:
Proofs were computed by message in the batch, instead of by the hash of the entire batch. This was an error not caught. I definitely need to redo all Gateway tests from scratch doing each case carefully!

### DIFF
--- a/src/common/Gateway.sol
+++ b/src/common/Gateway.sol
@@ -117,16 +117,11 @@ contract Gateway is Auth, IGateway, Recoverable {
     // Incoming methods
     //----------------------------------------------------------------------------------------------
 
-    /// @dev Handle a batch of messages
-    function handle(uint16 centrifugeId, bytes calldata message) external pauseable {
-        for (uint256 pos; pos < message.length;) {
-            bytes calldata inner = message[pos:message.length];
-            _handle(centrifugeId, inner, IAdapter(msg.sender), false);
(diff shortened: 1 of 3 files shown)


src/common/Gateway.sol
```diff
@@ -190,6 +182,18 @@ contract Gateway is Auth, IGateway, Recoverable {
...
 
+    function _handleBatch(uint16 centrifugeId, bytes memory batch_, IAdapter adapter_) internal {
+        bytes memory message = batch_;
+        for (uint256 start; start < batch_.length;) {
+            uint256 length = processor.messageLength(message);
+            message = batch_.slice(start, length);
+            start = length;
+
+            processor.handle(centrifugeId, message);
+            emit ExecuteMessage(centrifugeId, message, adapter_);
```
hieronx, line 193: Now the events won't match. `SendMessage` will include the full batch whereas `ExecuteMessage` per sub message. I would leave `ExecuteMessage` event out of the loop

hieronx (approved):

src/common/Gateway.sol
lemunozm: But we need it inside of the loop to distinguish between a successful message and a failed message in the incoming feature. Would you agree with the other way around? Dispatch a SendMessage per message in the batch?

PR #232, opened 2025-04-09
Feat: Gateway failing support

lemunozm:
Despite the renaming changes and test fixes, the logic itself is just: a96ba21515ca9350924c48d8f067b0dc1b37b2de

I've modified the events a bit to support latest batch changes:

### For sending, we have:
- `SendBatch`: emit when sending a batch by an adapter
- `SendProof`: emit when sending a proof by an adapter
- `PrepareMessage`: emit for each message in the batch

For example, a batch will dispatch:
- `PrepareMessage`
- `PrepareMessage`
- `PrepareMessage`
- `SendBatch`
- `SendProof` * (n-1) adapters

### For receiving, we have:
- `ProcessBatch`: counterpart of SendBatch
- `ProcessProof`: counterpart of SendProof
- `ExecuteMessage` and `FailMessage`: counterpart of PrepareMessage.
  - `ExecuteMessage`: the message was correctly executed
  - `FailMessage`: the message could not be executed and was enqueued

For example, a batch will dispatch:
- `ProcessProof` * (n-1) adapters
- `ProcessBatch`
  - `ExecuteMessage`
  - `FailMessage`
  - `ExecuteMessage`

NOTE: The amount of line test changes is due now the failures are not exposed as a revert.

### DIFF
--- a/foundry.toml
+++ b/foundry.toml
@@ -13,7 +13,7 @@ verbosity = 3
 ffi = true
 fs_permissions = [{ access = "read-write", path = ".forge-snapshots/"}, { access = "read-write", path = "./deployments"}]
 
-gas_snapshot_check = true
+gas_snapshot_check = false
 
 # False-alarm warnings
 ignored_warnings_from = [
--- a/src/common/Gateway.sol
+++ b/src/common/Gateway.sol
@@ -39,7 +39,7 @@ contract Gateway is Auth, IGateway, Recoverable {
     // Batching
     bool public transient isBatching;
     BatchLocator[] public /*transient*/ batchLocators;
-    mapping(uint16 centrifugeId => mapping(PoolId => bytes)) public /*transient*/ batch;
+    mapping(uint16 centrifugeId => mapping(PoolId => bytes)) public /*transient*/ outboundBatch;
     mapping(uint16 centrifugeId => mapping(PoolId => uint64)) public /*transient*/ batchGasLimit;
 
     // Payment
@@ -52,7 +52,7 @@ contract Gateway is Auth, IGateway, Recoverable {
     mapping(uint16 centrifugeId => mapping(IAdapter adapter => Adapter)) internal _activeAdapters;
 
     // Messages
-    mapping(uint16 centrifugeId => mapping(bytes32 messageHash => Message)) internal _messages;
+    mapping(uint16 centrifugeId => mapping(bytes32 messageHash => InboundBatch)) internal _inboundBatch;
     mapping(uint16 centrifugeId => mapping(bytes32 messageHash => bool)) internal _failedMessages;
     mapping(uint16 centrifugeId => mapping(IAdapter adapter => mapping(bytes32 messageHash => uint256 timestamp)))
         public recoveries;
@@ -152,7 +152,7 @@ contract Gateway is Auth, IGateway, Recoverable {
             emit ProcessMessage(centrifugeId, payload, adapter_);
         }
 
-        Message storage state = _messages[centrifugeId][messageHash];
+        InboundBatch storage state = _inboundBatch[centrifugeId][messageHash];
 
         if (adapter.activeSessionId != state.sessionId) {
             // Clear votes from previous session
@@ -168,7 +168,7 @@ contract Gateway is Auth, IGateway, Recoverable {
             state.votes.decreaseFirstNValues(adapter.quorum);
 
             if (isMessageProof) {
-                _handleBatch(centrifugeId, state.pendingMessage);
+                _handleBatch(centrifugeId, state.pendingBatch);
             }
             else {
                 _handleBatch(centrifugeId, payload);
@@ -176,10 +176,10 @@ contract Gateway is Auth, IGateway, Recoverable {
 
             // Only if there are no more pending messages, remove the pending message
             if (state.votes.isEmpty()) {
-                delete state.pendingMessage;
+                delete state.pendingBatch;
             }
         } else if (!isMessageProof) {
-            state.pendingMessage = payload;
+            state.pendingBatch = payload;
         }
     }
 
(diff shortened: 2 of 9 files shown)


src/common/interfaces/IGateway.sol
```diff
@@ -38,29 +38,35 @@ interface IGateway is IMessageHandler, IMessageSender, IGatewayHandler {
...
-    event ProcessMessage(uint16 centrifugeId, bytes message, IAdapter adapter);
+    event ProcessBatch(uint16 centrifugeId, bytes batch, IAdapter adapter);
     event ProcessProof(uint16 centrifugeId, bytes32 messageHash, IAdapter adapter);
-    event ExecuteMessage(uint16 centrifugeId, bytes message, IAdapter adapter);
-    event SendMessage(bytes message);
+    event ExecuteMessage(uint16 centrifugeId, bytes message);
+    event FailMessage(uint16 centrifugeId, bytes message, bytes error);
+    event SendBatch(uint16 centrifugeId, bytes batch, IAdapter adapter);
+    event SendProof(uint16 centrifugeId, bytes proof, IAdapter adapter);
+    event PrepareMessage(uint16 centrifugeId, PoolId poolId, bytes message);
```
lemunozm: I think the events `ExecuteMessage`, `FailMessage` and `PrepareMessage` does not require a `centrifugeId`, parameter because they are always attached to a `ProcessBatch`/`SendBatch` message that already contains that information.

If you agree, I will remove those parameters

hieronx: Indexer looks at individual events rather than groups of events so would keep it.

foundry.toml
```diff
@@ -13,7 +13,7 @@ verbosity = 3
 ffi = true
 fs_permissions = [{ access = "read-write", path = ".forge-snapshots/"}, { access = "read-write", path = "./deployments"}]
 
-gas_snapshot_check = true
+gas_snapshot_check = false
```
hieronx, line 16: Mis understood this param, change makes sense!

src/common/Gateway.sol
```diff
@@ -52,7 +52,8 @@ contract Gateway is Auth, IGateway, Recoverable {
     mapping(uint16 centrifugeId => mapping(IAdapter adapter => Adapter)) internal _activeAdapters;
 
     // Messages
-    mapping(uint16 centrifugeId => mapping(bytes32 messageHash => Message)) internal _messages;
+    mapping(uint16 centrifugeId => mapping(bytes32 messageHash => InboundBatch)) internal _inboundBatch;
+    mapping(uint16 centrifugeId => mapping(bytes32 messageHash => uint256)) internal _failedMessages;
```
hieronx: I think we should likely make these public now

src/common/Gateway.sol
```diff
@@ -167,33 +168,49 @@ contract Gateway is Auth, IGateway, Recoverable {
...
-            start = length;
+            start += length;
 
-            processor.handle(centrifugeId, message);
-            emit ExecuteMessage(centrifugeId, message, adapter_);
+            try processor.handle(centrifugeId, message) {
+                emit ExecuteMessage(centrifugeId, message);
+            }
+            catch (bytes memory err) {
+                bytes32 messageHash = keccak256(batch_);
```
```suggestion
                bytes32 messageHash = keccak256(message);
```
right?

lemunozm: Correct, my fault!

lemunozm: Fixed, and also rename messageHash into batchHash where it needs to be so: [6d55193](https://github.com/centrifuge/protocol-v3/pull/232/commits/6d5519357bf0b05227bce20178eb81bb2ad0fa42)

src/common/Gateway.sol
```diff
@@ -52,8 +52,9 @@ contract Gateway is Auth, IGateway, Recoverable {
     mapping(uint16 centrifugeId => mapping(IAdapter adapter => Adapter)) internal _activeAdapters;
 
     // Messages
-    mapping(uint16 centrifugeId => mapping(bytes32 messageHash => Message)) internal _messages;
-    mapping(uint16 centrifugeId => mapping(IAdapter adapter => mapping(bytes32 messageHash => uint256 timestamp)))
+    mapping(uint16 centrifugeId => mapping(bytes32 batchHash => InboundBatch)) public inboundBatch;
+    mapping(uint16 centrifugeId => mapping(bytes32 messageHash => uint256)) public failedMessages;
+    mapping(uint16 centrifugeId => mapping(IAdapter adapter => mapping(bytes32 batchHash => uint256 timestamp)))
```
hieronx: nit
```suggestion
    mapping(uint16 centrifugeId => mapping(bytes32 messageHash => uint256)) public failedMessages;
    mapping(uint16 centrifugeId => mapping(bytes32 batchHash => InboundBatch)) public inboundBatch;
    mapping(uint16 centrifugeId => mapping(IAdapter adapter => mapping(bytes32 batchHash => uint256 timestamp)))
```

src/common/Gateway.sol
```diff
@@ -135,23 +136,23 @@ contract Gateway is Auth, IGateway, Recoverable {
...
         if (adapter.quorum == 1 && !isMessageProof) {
             // Special case for gas efficiency
-            _handleBatch(centrifugeId, payload, adapter_);
+            _handleBatch(centrifugeId, payload);
             return;
         }
 
         // Verify adapter and parse message hash
-        bytes32 messageHash;
+        bytes32 batchHash;
```
hieronx, line 144: Isn't `batchHash` the same as `proof`?

hieronx (dismissed):

src/common/Gateway.sol
lemunozm: Yes, it's the same, but using later `proof` in the `inboundBatch[centrifugeId][proof]` looks weird (line 155).

I would reserve the word proof for messages that actually are proofs

hieronx (approved):

PR #234, opened 2025-04-09
Fuzz message lib tests

hieronx:


test/common/unit/libraries/MessageLib.t.sol
```diff
@@ -557,24 +604,25 @@ contract TestMessageLibIdentities is Test {
...
-
+    function testTriggerUpdateHoldingAmount(
+        uint64 poolId,
+        bytes16 scId,
+        uint128 assetId,
+        bytes32 who,
+        uint128 amount,
+        uint128 pricePerUnit,
+        bool isIncrease,
+        JournalEntry[] memory debits,
```
lemunozm, line 655: It's great that these can also be fuzzed

lemunozm (dismissed):
LGTM! Thanks for tackling this

test/common/unit/libraries/MessageLib.t.sol
```diff
@@ -113,7 +113,7 @@ contract TestMessageLibIdentities is Test {
         MessageLib.NotifyShareClass memory a = MessageLib.NotifyShareClass({
             poolId: poolId,
             scId: scId,
-            name: name,
+            name: string(abi.encodePacked(name)),
```
lemunozm: We have a `CastLib.toString` for this. Although not sure which is faster TBH.

test/common/unit/libraries/MessageLib.t.sol
```diff
@@ -104,7 +104,7 @@ contract TestMessageLibIdentities is Test {
     function testNotifyShareClass(
         uint64 poolId,
         bytes16 scId,
-        string calldata name,
+        bytes32 name,
```
lemunozm: Good catch, I think it was thanks to `Halmos`, right?

test/common/unit/libraries/MessageLib.t.sol
hieronx: Since it's in test code, don't think it really matters

test/common/unit/libraries/MessageLib.t.sol
hieronx: Yeah although now we are only fuzzing 32 bytes names rather than 128 byte names

test/common/unit/libraries/MessageLib.t.sol
lemunozm: Ah right! correct

test/common/unit/libraries/MessageLib.t.sol
lemunozm: The border case that we need to ensure is that a string of 129 chars is correctly clamped to 128. I think maybe we need special test cases for those

hieronx: Just pushed a commit, think this is tested well now

lemunozm (approved):
LGTM!

PR #236, opened 2025-04-10
Feat: Gateway execute and payment change

lemunozm:
Thread: https://kflabs.slack.com/archives/C07PG2EUR9C/p1744126740591899

### Problem

When we pay for messages, we pay for a message or for a proof. Nevertheless, when the Gateway executes a message, it is not exactly when the message arrives. You could:
- Receive a batch, but not process it due a missing proof => Wasting the adapter gas
- Receive a proof, generation to process the batch => Not having enough adapter gas to tackle it)

### Solution

Each time we receive a proof or a batch, the action dispatched by the adapter does the same: storing the message. So the payment in the other side is the same.

Later, manually, someone needs to call `Gateway.execute()` to actually process the batch. This call is permissionless, and the sender can be refunded with the subsidized pool.


src/common/Gateway.sol
```diff
@@ -198,6 +183,31 @@ contract Gateway is Auth, IGateway, Recoverable {
                 failedMessages[centrifugeId][messageHash]++;
                 emit FailMessage(centrifugeId, message, err);
             }
+
+            PoolId poolId = processor.messagePoolId(message);
+            uint256 refunded = (gasleft() - previousGas) * tx.gasprice;
+
+            if (subsidy[poolId] >= refunded) {
+                subsidy[poolId] -= refunded;
+                payable(msg.sender).transfer(refunded);
```
hieronx: Should use `SafeTransferLib.safeTransferETH`

src/common/Gateway.sol
```diff
@@ -198,6 +183,31 @@ contract Gateway is Auth, IGateway, Recoverable {
...
+            }
+        }
+    }
+
+    function execute(uint16 centrifugeId, bytes32 batchHash) external {
```
hieronx: I'd move this function about `_handleBatch`, order confused me a bit

hieronx: And maybe rename `_handleBatch` to `_process`?

src/common/Gateway.sol
lemunozm: We ensure before reaching this line that, the user will always pay more for this execution than the gas refunded, so there is no attack vector here.

Also, because failing messages are enqueued, this action can not be reexecuted with the same parameters.

src/common/Gateway.sol
```diff
@@ -198,6 +183,31 @@ contract Gateway is Auth, IGateway, Recoverable {
                 failedMessages[centrifugeId][messageHash]++;
                 emit FailMessage(centrifugeId, message, err);
             }
+
+            PoolId poolId = processor.messagePoolId(message);
+            uint256 refunded = (gasleft() - previousGas) * tx.gasprice;
```
hieronx, line 201: Need to check whether there's any ways to manipulate this. Will do some reading

src/common/Gateway.sol
```diff
@@ -107,8 +105,8 @@ contract Gateway is Auth, IGateway, Recoverable {
 
     /// @inheritdoc IGateway
     function file(bytes32 what, address instance) external auth {
-        if (what == "gasService") gasService = IGasService(instance);
-        else if (what == "processor") processor = IMessageProcessor(instance);
+        if (what == "processor") processor = IMessageProcessor(instance);
+        else if (what == "gasLimit") gasLimit = uint64(uint160(bytes20(instance)));
```
hieronx, line 110: `gasLimit` would still be different for `quorum=1` case, which could be the case for some chain IDs but not for others

src/common/Gateway.sol
lemunozm: Fully agree with both

src/common/Gateway.sol
lemunozm: In which sense? `handle()` just stores the message independently of what it is.

hieronx: If `quorum=1`, then the adapter executing `handle()` will pay for the full gas limit of the message processing.

lemunozm: Correct, I forgot about the shortcut we have for quorum == 1.

I think we should remove that shortcut 🤔. The problem is that if not, we have two different ways of handling the entire process in different chains:

1. GasLimit computed in chainA for each message diferently + adapter from chain B synchronously handle and execute the message.

2. GasLimit only for storing a message + adapter from chain B storing the message + permissionless execute()

Could be a mess to ensure both ways always work and to decide which GasLimit will be used in the remote chain.

hieronx: I agree let's do that

lemunozm: BTW, to accuretely measure the `gasLimit` is `max(storing a big batch, handleARecoveryMessage)`.

Do you think we should limit the batch size? for example 4Kb or so

lemunozm:
Closed in favor or https://github.com/centrifuge/protocol-v3/pull/253

PR #239, opened 2025-04-10
Refactor createHoldings

onnovisser:
- Split off `createLiability` from `createHolding`
- Require accounts to be passed into the functions (they should have been created beforehand)
- Remove `kind` from `AccoundId`, accounts should be able to be shared. e.g.: Loss and Gain kinds can share the same account.
- Include accounts in emitted event for indexing


src/hub/Hub.sol
```diff
@@ -337,30 +337,49 @@ contract Hub is Multicall, Auth, Recoverable, IHub, IHubGatewayHandler {
     }
 
     /// @inheritdoc IHub
-    function createHolding(ShareClassId scId, AssetId assetId, IERC7726 valuation, bool isLiability, uint24 prefix)
+    function createHolding(ShareClassId scId, AssetId assetId, IERC7726 valuation, AccountId assetAccount, AccountId equityAccount, AccountId lossAccount, AccountId gainAccount)
```
lemunozm: Not sure what's the reason you removed the kind from the account. That adds a new level to the mapping storages

If you pass an array of `uint24`, you could have the same behavior as before adding the `kind`

src/hub/interfaces/IHoldings.sol
```diff
@@ -15,13 +15,23 @@ struct Holding {
     bool isLiability;
 }
 
+struct CreateHolding {
+    AccountId accountId;
+    uint8 kind;
+}
```
lemunozm: Extracting the `kind` from the account means that now the `Accounting` stores a plain Account without kind information. That was a requirement if it had not changed

src/hub/Hub.sol
lemunozm: The array also would help with the contract size issue in the Hub

src/hub/interfaces/IHoldings.sol
onnovisser: The requirement is now that multiple `kind`s can write to the same `Account` in the `Accounting` contract, so the `kind` can't be embedded in the `accountId` anymore.

lemunozm: That's weird. Is there any thread or issue with that requirement definition? The above would mean that asset & equity accounts can be mixed in the accounting.

Just to double check (from https://www.notion.so/Fix-Accounts-in-createHolding-1c02eac24e1780f0b283fa4f5bc6c7e2?pvs=4#1c02eac24e1780ef90e0eea46b34bde6):

> re-use an existing AccountType.Equity account for all assets

This means that several assets can have the same kind account

onnovisser: Yes, an asset can have their Gain, Loss, and Equity kinds point to the same account. And all assets can also share the same Equity account for example.

src/hub/interfaces/IHoldings.sol
```diff
@@ -15,13 +15,23 @@ struct Holding {
     bool isLiability;
 }
 
+struct CreateHolding {
```
mustermeiszer: NIT

Naming is a bit confusing to me. Maybe something like `HoldingAccountId`

mustermeiszer (dismissed):
Nothing to complain! Awesome work!

src/hub/Hub.sol
```diff
@@ -309,33 +309,55 @@ contract Hub is Multicall, Auth, Recoverable, IHub, IHubGatewayHandler {
         PoolId poolId,
         ShareClassId scId,
         AssetId assetId,
-        IERC7726 valuation,
-        bool isLiability,
-        uint24 prefix
+        IERC7726 valuation,AccountId assetAccount, AccountId equityAccount, AccountId lossAccount, AccountId gainAccount
```
lemunozm: NIT. To fix the format of this file, you can:

- comment `transient` keywords
- Uncomment the file in `foundry.yml`
- save/compile again the fail with the format
- comment the file in `foundry.yml`
- uncomment `transient` keywords

lemunozm (dismissed):

mustermeiszer (approved):

PR #243, opened 2025-04-10
Custom errors in hooks & token

hieronx:


wischli (approved):
Thanks for the smaller tranches of these changes!

PR #244, opened 2025-04-11
Simplify Hub

hieronx:


src/hub/Hub.sol
```diff
@@ -62,6 +59,15 @@ contract Hub is Multicall, Auth, Recoverable, IHub, IHubGatewayHandler {
         transientValuation = transientValuation_;
     }
 
+    /// @inheritdoc IHub
+    modifier unlocked(PoolId poolId) {
+        require(hubRegistry.isAdmin(poolId, msg.sender), IHub.NotAuthorizedAdmin());
+
+        accounting.unlock(poolId);
+        _;
+        accounting.lock();
```
lemunozm: I propose to extract these lines to a new `modifier unlocking()` and add it only to the methods that really need to touch the accounting

src/hub/Hub.sol
```diff
@@ -62,6 +59,15 @@ contract Hub is Multicall, Auth, Recoverable, IHub, IHubGatewayHandler {
         transientValuation = transientValuation_;
     }
 
+    /// @inheritdoc IHub
+    modifier unlocked(PoolId poolId) {
+        require(hubRegistry.isAdmin(poolId, msg.sender), IHub.NotAuthorizedAdmin());
```
lemunozm: To enable paymment:

```suggestion
    modifier unlocked(PoolId poolId) {
        require(hubRegistry.isAdmin(poolId, msg.sender), IHub.NotAuthorizedAdmin());
        _pay();
```

src/hub/Hub.sol
hieronx: Updated!

src/hub/Hub.sol
```diff
@@ -347,95 +365,93 @@ contract Hub is Multicall, Auth, Recoverable, IHub, IHubGatewayHandler {
         accounts[4] = newAccountId(prefix, uint8(AccountType.Expense));
         accounts[5] = newAccountId(prefix, uint8(AccountType.Liability));
 
-        createAccount(accounts[0], true);
-        createAccount(accounts[1], false);
-        createAccount(accounts[2], false);
-        createAccount(accounts[3], false);
-        createAccount(accounts[4], true);
-        createAccount(accounts[5], false);
+        accounting.createAccount(poolId, accounts[0], true);
```
hieronx, line 340: Otherwise you are unlocking and locking again with `createAccount`

src/hub/interfaces/IHub.sol
```diff
@@ -68,130 +65,155 @@ interface IHub {
...
-        VaultUpdateKind kind
-    ) external payable;
+    // function updateVault(
+    //     PoolId poolId,
+    //     ShareClassId scId,
+    //     AssetId assetId,
+    //     bytes32 target,
+    //     bytes32 vaultOrFactory,
+    //     VaultUpdateKind kind
+    // ) external payable;
```
mustermeiszer: NIT 

remove

src/hub/Hub.sol
```diff
@@ -95,32 +92,23 @@ contract Hub is Multicall, Auth, Recoverable, IHub, IHubGatewayHandler {
...
         poolId = hubRegistry.registerPool(admin, sender.localCentrifugeId(), currency);
     }
 
     /// @inheritdoc IHub
-    function claimDeposit(PoolId poolId, ShareClassId scId, AssetId assetId, bytes32 investor) external payable {
-        _protected();
+    function claimDeposit(PoolId poolId, ShareClassId scId, AssetId assetId, bytes32 investor)
+        external
+        payable
+        protected
```
mustermeiszer, line 110: Why protected?

mustermeiszer: I think all of those can be public.

src/hub/Hub.sol
```diff
@@ -347,129 +337,105 @@ contract Hub is Multicall, Auth, Recoverable, IHub, IHubGatewayHandler {
...
-        createAccount(accounts[2], false);
-        createAccount(accounts[3], false);
-        createAccount(accounts[4], true);
-        createAccount(accounts[5], false);
-
-        holdings.create(unlockedPoolId, scId, assetId, valuation, isLiability, accounts);
-    }
-
-    /// @inheritdoc IHub
-    function updateHolding(ShareClassId scId, AssetId assetId) public payable {
```
mustermeiszer, line 361: We can not remove that one. We need to be able to update the value of a holding!

src/hub/Hub.sol
hieronx: `protected` is just the re-entrancy modifier

mustermeiszer (approved):

src/hub/Hub.sol
lemunozm: Maybe we should rename it to reentrancy, I also found this name quite confusing. WDYT @hieronx ?

PR #248, opened 2025-04-13
Update cross-chain transfer logic

hieronx:


script/VaultsDeployer.s.sol
```diff
@@ -97,6 +97,7 @@ contract VaultsDeployer is CommonDeployer {
         IAuth(restrictedTransfers).rely(address(poolManager));
         IAuth(freelyTransferable).rely(address(poolManager));
         messageDispatcher.rely(address(poolManager));
+        gateway.rely(address(poolManager));
```
hieronx, line 100: Required to pay gas from `poolManager`

src/vaults/PoolManager.sol
```diff
@@ -96,13 +98,20 @@ contract PoolManager is Auth, Recoverable, IPoolManager, IUpdateContract, IPoolM
...
     /// @inheritdoc IPoolManager
+    // TODO: change to use address token as input
     function transferShares(uint16 centrifugeId, uint64 poolId, bytes16 scId, bytes32 receiver, uint128 amount)
         external
-        auth
+        payable
     {
         IShareToken shareToken_ = IShareToken(shareToken(poolId, scId));
-        shareToken_.burn(msg.sender, amount);
+        require(
```
hieronx, line 106: This allows hooks to differentiate cross-chain transfers. It means for `RestrictedTransfers`, each chain ID needs to be added as an allowed destination.

src/common/Gateway.sol
```diff
@@ -308,6 +308,7 @@ contract Gateway is Auth, IGateway, Recoverable {
             }
         }
 
+        transactionPayer = address(0);
```
hieronx: @lemunozm previously, a smart contract (including our tests) doing
```solidity
poolManager.registerAsset{value: EXACT_GAS}(..);
someVault.requestDeposit(..);
```

would fail since the `requestDeposit` call would not work subsidized as `transactionPayer` was still set.

src/vaults/PoolManager.sol
```diff
@@ -96,13 +98,20 @@ contract PoolManager is Auth, Recoverable, IPoolManager, IUpdateContract, IPoolM
 
     // --- Outgoing message handling ---
     /// @inheritdoc IPoolManager
+    // TODO: change to use address token as input
```
hieronx: I'm still thinking about the best way to accomplish this. Ideally without adding storage but it seems we don't have this mapping today.

src/vaults/PoolManager.sol
```diff
@@ -98,10 +100,23 @@ contract PoolManager is Auth, Recoverable, IPoolManager, IUpdateContract, IPoolM
...
-        shareToken_.burn(msg.sender, amount);
+        IShareToken share = IShareToken(shareToken(poolId, scId));
+        require(
+            share.checkTransferRestriction(msg.sender, address(uint160(centrifugeId)), amount),
+            CrossChainTransferNotAllowed()
+        );
+
+        gateway.payTransaction{value: msg.value}(msg.sender);
+
+        try share.authTransferFrom(msg.sender, msg.sender, address(this), amount) returns (bool) {}
```
hieronx, line 113: Matching behaviour from the async redeem vault logic

hieronx: We could also have removed the need for an allowance on `burn()` and used this directly, but this would have required migrating the existing deployed share tokens which is not possible.

src/vaults/PoolManager.sol
```diff
@@ -111,7 +126,7 @@ contract PoolManager is Auth, Recoverable, IPoolManager, IUpdateContract, IPoolM
     // @inheritdoc IPoolManagerGatewayHandler
     function registerAsset(uint16 centrifugeId, address asset, uint256 tokenId)
```
hieronx, line 127: Moved this here now that `transferAssets` is also payable.

src/common/Gateway.sol
lemunozm: The issue is the test is missing /// forge-config: default.isolate = true

lemunozm: CR: The issue I think is the test is missing `/// forge-config: default.isolate = true`

Note that here, you're resetting the transaction payer, but you could still be in a batch where more messages needs to be sent. i.e:

- FM.multicall{PAY}()
  - NotifyPoolA // Will be paid with PAY
  - NotifyPoolB // Will be subsidized when it shouldn't

src/vaults/VaultRouter.sol
```diff
@@ -236,37 +236,6 @@ contract VaultRouter is Auth, Multicall, Recoverable, IVaultRouter {
...
-        external
-        payable
-        protected
-    {
-        transferShares(centrifugeId, vault, receiver.toBytes32(), amount);
-    }
-
-    // --- Register Asset ---
-    /// @inheritdoc IVaultRouter
-    function registerAsset(uint16 centrifugeId, address asset, uint256 tokenId) public payable {
```
lemunozm, line 265: These messages can no longer be batched using `VaultRouter.multicall()`, are we ok with this? I think could be use cases where you would want to transfer shares to different locations at once.

src/vaults/interfaces/token/IHook.sol
```diff
@@ -14,6 +14,13 @@ string constant SUCCESS_MESSAGE = "transfer-allowed";
 uint8 constant ERROR_CODE_ID = 1;
 string constant ERROR_MESSAGE = "transfer-blocked";
 
+/// @notice Hook interface to customize share token behaviour
+/// @dev    To detect specific system actions:
+///           Deposit request:      address(0)      -> address(user)
+///           Deposit claim:        address(escrow) -> address(user)
+///           Redeem request:       address(user)   -> address(escrow)
+///           Redeem claim:         address(user)   -> address(0)
+///           Cross-chain transfer: address(user)   -> address(uint160(chainId))
```
lemunozm, line 23: ❤️

lemunozm (commented):
One CR for me. But the cross-transfer logic looks good!!

src/vaults/interfaces/token/IHook.sol
wischli: Love this!

src/vaults/PoolManager.sol
```diff
@@ -80,6 +81,7 @@ contract PoolManager is Auth, Recoverable, IPoolManager, IUpdateContract, IPoolM
     function file(bytes32 what, address data) external auth {
         if (what == "sender") sender = IVaultMessageSender(data);
         else if (what == "tokenFactory") tokenFactory = ITokenFactory(data);
+        else if (what == "gateway") gateway = IGateway(data);
```
wischli, line 84: Nit for potential follow-up PR: Missing test

src/vaults/VaultRouter.sol
hieronx: Yeah I think that is fine. It is a tradeoff but I just don't see a better option.

lemunozm: What was the problem with forcing them to be used in the `VaultRouter`?

hieronx: The `PoolManager` needs to be a ward to call `authTransferFrom`. By doing this, no ERC20 approval is needed.

src/common/Gateway.sol
hieronx: > CR: The issue I think is the test is missing /// forge-config: default.isolate = true

That didn't fix it. I believe because `registerAsset` is called in the setup of the pool, which is part of the test.

> Note that here, you're resetting the transaction payer, but you could still be in a batch where more messages needs to be sent. i.e:

I don't get this, wouldn't `send()` never be triggered until the end of the `multicall`?

lemunozm: There is one send() for each message to a chain/pool.

SO, although all messages are sent at the end, once the first message sets the `transactionPayer` to `address(0)`, the next message will start taken funds from the subsidized pool

hieronx: Gotcha! Should we add a `resetTransactionPayer()` method that is called at the end of `PoolManager.registerAsset/transerShares`?

lemunozm: Mmm... Not sure. That would be a method in the contract for only testing, right?

I think the best fast solution that doesn't imply refactoring all the tests is using `vm.load()` to overwrite that value.

BTW, have you tried adding `/// forge-config: default.isolate = true` to the `setUp()` ?

hieronx: I think this would actually be valuable though, not just for tests.

You might very well have a smart contract integration that wants to do
```solidity
poolManager.registerAsset{value: EXACT_GAS}(..);
someVault.requestDeposit(..);
```

I believe that should be possible. Without some fix for resetting the payment mechanism, the above will never work.

hieronx: Especially with the upcoming wider adoption of smart contract wallets.

lemunozm: Ok, then, let's add it!

Just concerned about the security issues. Anyone could call to it in the middle of a transaction to disable the payment and use the subsidized pool and this kind of things.

I think it should be `auth`, and we should call it after all methods where we already call `_pay()`, probably it should be a modifier as follows:

```solidity
modifier transactionPayment() {
    if (!gateway.isBatching())) gateway.payTransaction(msg.sender);
    _;
    if (!gateway.isBatching())) gateway.resetTransactionPayer();
}

claimRequest() external transactionPayment { 
...
```

WDYT?

Something similar we would need to do in the multicall

hieronx: Yes it should definitely be `auth`

hieronx: I don't think this is needed on the `Hub`, since there is no subsidized transactions possible on the Hub side.

lemunozm: > since there is no subsidized transactions possible on the Hub side

Just to note (although with the latest solution I think it does not apply). There are subsidized messages in some cases on the Hub side. When we a CV message reach CP and dispatch another message to CV, that message is subsidized.

hieronx: Good point!

src/common/Gateway.sol
```diff
@@ -314,6 +314,7 @@ contract Gateway is Auth, IGateway, Recoverable {
             }
         }
 
+        if (!isBatching) transactionPayer = address(0);
```
lemunozm: I think this can be in the `send()` method, and send only in the case where `!isBatching`. Seems a more straightforward place

hieronx: Good idea

hieronx: Updated

lemunozm (approved):
Thanks!! LGTM 💯

PR #249, opened 2025-04-13
Remap pool ids in `AsyncRequests`

hieronx:


src/vaults/AsyncRequests.sol
```diff
@@ -629,4 +633,11 @@ contract AsyncRequests is BaseInvestmentManager, IAsyncRequests {
 
         return VaultPricingLib.calculatePrice(shareToken, shares, vaultDetails.asset, vaultDetails.tokenId, assets);
     }
+
+    function _mapPoolId(uint64 poolId) internal view returns (uint64 mappedPoolId) {
+        // TODO: update legacyCentrifugeId and add all pools before deployment
+        uint16 legacyCentrifugeId = 1;
+        if (poolId == 4139607887) return newPoolId(legacyCentrifugeId, 1).raw();
```
hieronx: Since pool creation is permissioned, we can just ensure the legacy pool(s) are created in the order specified here. This removes the need for storage lookups, so doesn't add extra gas cost.

src/vaults/AsyncRequests.sol
```diff
@@ -629,4 +633,11 @@ contract AsyncRequests is BaseInvestmentManager, IAsyncRequests {
 
         return VaultPricingLib.calculatePrice(shareToken, shares, vaultDetails.asset, vaultDetails.tokenId, assets);
     }
+
+    function _mapPoolId(uint64 poolId) internal pure returns (uint64 mappedPoolId) {
+        // TODO: update v2CentrifugeId and add all pools before deployment
+        uint16 v2CentrifugeId = 1;
```
lemunozm: NIT. Maybe we want to choose another number, like `uint16(-1)`, to avoid colliding this with Ethereum chain Id. That could help later for mapping this to `wormholeId`

lemunozm (dismissed):

wischli (commented):
AFAICT, the touched code paths do not suffice to cover all required remappings. Maybe I am missing something but shouldn't it include all `.poolId()` retrievals  in `BaseInvestmentManager` as well as the one in `VaultRouter.transferShares`?

If so, seems like the remapping should be public. IMO, we should also add a comment to the `IBaseVault.poolId` definition that for async vaults, it should always be wrapped by `AsyncRequests.mapPoolId`?

hieronx:
> AFAICT, the touched code paths do not suffice to cover all required remappings. Maybe I am missing something but shouldn't it include all .poolId() retrievals in BaseInvestmentManager as well as the one in VaultRouter.transferShares?

You are absolutely right, I missed both of those!

> If so, seems like the remapping should be public. IMO, we should also add a comment to the IBaseVault.poolId definition that for async vaults, it should always be wrapped by AsyncRequests.mapPoolId?

Also agree!

src/vaults/AsyncRequests.sol
hieronx: The centrifuge ID will just be the one wherever we deploy the legacy pools.

hieronx:
> AFAICT, the touched code paths do not suffice to cover all required remappings. Maybe I am missing something but shouldn't it include all .poolId() retrievals in BaseInvestmentManager as well as the one in VaultRouter.transferShares?

Actually `PoolManager` doesn't need an update since its' just passed as a param. So users can pass the new mapped iD.

hieronx:
> IMO, we should also add a comment to the IBaseVault.poolId definition that for async vaults, it should always be wrapped by AsyncRequests.mapPoolId?

Actually since this happens automatically in the contracts, I would leave this out of the interface. Integrators shouldn't care about this.

wischli (approved):

PR #252, opened 2025-04-14
Fix deployment

hieronx:


script/utils/JsonRegistry.s.sol
```diff
@@ -15,15 +18,19 @@ contract JsonRegistry is Script {
         registeredContracts += 1;
     }
 
-    function startDeploymentOutput() public {
+    function startDeploymentOutput(bool isTests) public {
         deploymentOutput = '{\n  "contracts": {\n';
+
+        if (!isTests) {
+            console.log(
```
hieronx, line 25: Making it easier to find the latest file

src/common/libraries/MessageLib.sol
```diff
@@ -1086,7 +1086,7 @@ library MessageLib {
         });
     }
 
-    function serialize(UpdateHoldingAmount memory t) public pure returns (bytes memory) {
+    function serialize(UpdateHoldingAmount memory t) internal pure returns (bytes memory) {
```
hieronx, line 1089: This is really important: libraries should only have internal methods. Otherwise they become externally referenced libraries, which are effectively separate contract deployments. These have all sorts of issues and should not be used.

src/vaults/BalanceSheet.sol
```diff
@@ -265,7 +265,7 @@ contract BalanceSheet is Auth, Recoverable, IBalanceSheet, IBalanceSheetGatewayH
             IERC6909(asset).transferFrom(address(escrow), receiver, tokenId, amount);
         }
 
-        sender.sendUpdateHoldingAmount(poolId, scId, assetId, receiver, amount, pricePoolPerAsset, true, m);
+        sender.sendUpdateHoldingAmount(poolId, scId, assetId, receiver, amount, pricePoolPerAsset, false, m);
```
hieronx, line 268: cc @mustermeiszer

src/common/libraries/MessageLib.sol
lemunozm: I think that was me frightening with the annoying deep stack issue that happened when serializing that message. I wanted to force an external call there. Thanks for the eagle eye!

wischli (approved):
Great catches on the libs and BalanceSheet!

script/utils/JsonRegistry.s.sol
lemunozm: NIT. Maybe instead of passing `isTests` for all the files, we can just use an environmental variable? So only from tests we do:

```
vm.setEnv("IS_TESTS", 1);
```

That way we leave deployer interface untouched with testing stuff

PR #253, opened 2025-04-14
Gateway unify payment

lemunozm:
### Problem

When we pay for messages, we pay for a message or for a proof. Nevertheless, when the Gateway executes a message, it is not exactly when the message arrives. You could:
- Receive a batch, but not process it due a missing proof => Wasting the adapter gas
- Receive a proof, generation to process the batch => Not having enough adapter gas to tackle it)

### Solution
Overpay the Gateway to ensure any message can dispatch and process execution. To achieve that, we need to be able to refund later correctly to the correct address associated to pool message


src/common/Gateway.sol
```diff
@@ -271,47 +276,46 @@ contract Gateway is Auth, IGateway, Recoverable {
...
+        uint64 gasLimit = (isBatching) ? batchGasLimit[centrifugeId][poolId] : messageGasLimit;
 
         for (uint256 i; i < adapters_.length; i++) {
-            IAdapter currentAdapter = IAdapter(adapters_[i]);
             bool isPrimaryAdapter = i == PRIMARY_ADAPTER_ID - 1;
-            bytes memory payload = isPrimaryAdapter ? batch : proof;
-            uint64 gasLimit = isPrimaryAdapter ? batchGasLimit_ : proofGasLimit;
 
-            uint256 consumed = currentAdapter.estimate(centrifugeId, payload, gasLimit);
+            uint256 consumed = adapters_[i].estimate(centrifugeId, isPrimaryAdapter ? batch : proof, gasLimit);
```
lemunozm: I needed to play here to avoid stack-deep issues

src/common/Gateway.sol
```diff
@@ -353,22 +357,19 @@ contract Gateway is Auth, IGateway, Recoverable {
...
-        uint256 messageGasLimit = 0;
+        uint256 gasLimit = 0;
         for (uint256 pos; pos < payload.length;) {
             bytes calldata inner = payload[pos:payload.length];
-            messageGasLimit += gasService.gasLimit(centrifugeId, inner);
+            gasLimit += messageGasLimit;
             pos += processor.messageLength(inner);
         }
 
         perAdapter = new uint256[](adapters[centrifugeId].length);
```
lemunozm: @hieronx I think returning this `perAdapter` in `estimate()` now does not add much info. Just a slight variation due to the Adapter in comparison with the big value `messageGasLimit`. I'm in favor of removing this and simplifying tests regarding this WDYT?

lemunozm (commented):
Will adapt with the pool escrows once that's ready, but this can be already merged

src/common/Gateway.sol
```diff
@@ -31,9 +30,11 @@ contract Gateway is Auth, IGateway, Recoverable {
     uint8 public constant PRIMARY_ADAPTER_ID = 1;
     uint256 public constant RECOVERY_CHALLENGE_PERIOD = 7 days;
 
+    // Values
+    uint64 public messageGasLimit;
```
hieronx: This way, we cannot update the gas limit to be per message ID without swapping the entire gateway. Shouldn't we rather keep the gas service for that purpose?

src/common/Gateway.sol
```diff
@@ -107,13 +108,17 @@ contract Gateway is Auth, IGateway, Recoverable {
...
     function file(bytes32 what, address instance) external auth {
-        if (what == "gasService") gasService = IGasService(instance);
+        if (what == "messageGasLimit") messageGasLimit = uint64(uint160(bytes20(instance)));
         else if (what == "processor") processor = IMessageProcessor(instance);
         else revert FileUnrecognizedParam();
 
         emit File(what, instance);
     }
 
+    receive() external payable {
```
hieronx, line 123: Like it!

src/common/Gateway.sol
hieronx: This will add another storage lookup. Wonder if we can accomplish this another way 🤔

src/common/interfaces/IGateway.sol
```diff
@@ -41,6 +41,11 @@ interface IGateway is IMessageHandler, IMessageSender, IGatewayHandler {
         bytes pendingBatch;
     }
 
+    struct Funds {
+        uint256 value;
```
hieronx: Should we store `value` as `uint64` (and just discard remaining decimals), such that this struct is packed into 1 storage slot?

src/common/Gateway.sol
hieronx: Yeah think that is fine now!

src/common/Gateway.sol
lemunozm: True 👍🏻

src/common/Gateway.sol
lemunozm: `adapters_` is already in memory, though. So I think it should not be from the storage. I tried everything there and that was the only way I found.

src/common/interfaces/IGateway.sol
lemunozm: Good point

src/common/Gateway.sol
hieronx: That's true, then it seems fine!

src/common/Gateway.sol
```diff
@@ -280,45 +284,45 @@ contract Gateway is Auth, IGateway, Recoverable {
...
 
-    function subsidizePool(PoolId poolId) external payable {
-        subsidy[poolId] += msg.value;
+    function setRefundAddress(PoolId poolId, address refund) public auth {
+        subsidy[poolId].refund = refund;
+        emit SetRefundAddress(poolId, refund);
+    }
+
+    function subsidizePool(PoolId poolId) public payable {
+        subsidy[poolId].value += uint64(msg.value);
```
hieronx: Actually, I believe this will overflow at 18 ETH (`type(uint64).max / 10**18` since ether has 18 decimals).

If we move to `uint96`, which would afaik be 12 bytes and thus with `address` being 20 bytes should fit, then it will only overflow at `7.9228162514264337593543950335 × 10^10` ETH which is infeasible.

hieronx: Alternatively we could divide by `10**x` to lose the latter bits of precision.

But using `uint96` seems cleaner, if I didn't make a mistake in the logic.

src/common/WormholeAdapter.sol
```diff
@@ -18,19 +19,19 @@ import {
...
 
-    constructor(IMessageHandler gateway_, address relayer_, uint16 refundWormholeId_, address deployer)
-        Auth(deployer)
-    {
+    constructor(IMessageHandler gateway_, address relayer_, address deployer) Auth(deployer) {
         gateway = gateway_;
         relayer = IWormholeRelayer(relayer_);
-        refundWormholeId = refundWormholeId_;
+
+        IWormholeDeliveryProvider deliveryProvider = IWormholeDeliveryProvider(relayer.getDefaultDeliveryProvider());
```
hieronx, line 33: Suggestion from https://github.com/burrasec/2025-04-Centrifuge-Audit/issues/5

src/common/Gateway.sol
```diff
@@ -267,58 +272,75 @@ contract Gateway is Auth, IGateway, Recoverable {
...
+                if (!success) {
+                    // If refund fails, move remaining fuel to global pot
+                    subsidy[PoolId.wrap(0)].value += uint96(fuel);
+                    emit SubsidizePool(PoolId.wrap(0), address(0), fuel);
+                }
+                    
+                fuel = 0;
+            }
+
+            transactionPayer = address(0);
```
hieronx: Fixes https://github.com/burrasec/2025-04-Centrifuge-Audit/issues/4

src/common/Gateway.sol
```diff
@@ -267,58 +272,75 @@ contract Gateway is Auth, IGateway, Recoverable {
...
             }
         }
     }
 
-    function subsidizePool(PoolId poolId) external payable {
-        subsidy[poolId] += msg.value;
+    function _closeTransaction() internal {
+        if (transactionPayer != address(0)) {
+            if (fuel > 0) {
+                (bool success,) = transactionPayer.call{value: fuel}(new bytes(0));
```
hieronx: Fixes https://github.com/burrasec/2025-04-Centrifuge-Audit/issues/2

src/common/Gateway.sol
```diff
@@ -117,6 +118,10 @@ contract Gateway is Auth, IGateway, Recoverable {
         emit File(what, instance);
     }
 
+    receive() external payable {
```
hieronx, line 123: Fixes https://github.com/burrasec/2025-04-Centrifuge-Audit/issues/1

src/common/WormholeAdapter.sol
lemunozm: Agree!

lemunozm: Do you know if we have something similar for Axelar?

test/common/unit/GasService.t.sol
```diff
@@ -10,31 +10,19 @@ import {GasService, IGasService} from "src/common/GasService.sol";
...
-        uint256 messageGasLimit = service.gasLimit(CHAIN_ID, message);
+        uint256 messageGasLimit = service.gasLimit(CENTRIFUGE_ID, message);
         assertEq(messageGasLimit, MESSAGE_GAS_LIMIT);
+    }
 
-        uint256 proofGasLimit = service.gasLimit(CHAIN_ID, proof);
-        assertEq(proofGasLimit, PROOF_GAS_LIMIT);
+    function testMaxBatchSize(bytes calldata) public view {
+        uint256 messageGasLimit = service.maxBatchSize(CENTRIFUGE_ID);
+        assertEq(messageGasLimit, MAX_BATCH_SIZE);
```
lemunozm:
```suggestion
        uint256 maxBatchSize = service.maxBatchSize(CENTRIFUGE_ID);
        assertEq(maxBatchSize, MAX_BATCH_SIZE);
```

src/common/Gateway.sol
```diff
@@ -267,58 +275,75 @@ contract Gateway is Auth, IGateway, Recoverable {
...
+    function _closeTransaction() internal {
+        if (transactionPayer == address(0)) return;
+
+        if (fuel > 0) {
+            (bool success,) = transactionPayer.call{value: fuel}(new bytes(0));
+
+            if (!success) {
+                // If refund fails, move remaining fuel to global pot
+                subsidy[PoolId.wrap(0)].value += uint96(fuel);
+                emit SubsidizePool(PoolId.wrap(0), address(0), fuel);
```
lemunozm:
```suggestion
                emit SubsidizePool(PoolId.wrap(0), gateway, fuel);
```
Here sender should not be the `Gateway`? Or even better, the `transactionPayer`?

lemunozm (commented):
A couple of minors. Love the latest changes!

src/common/Gateway.sol
hieronx: `gateway` makes sense! Will be `address(this)` then

I don't really like `transactionPayer` since I feel it should be separated between direct subsidies from a user and a repayment after a tx.

lemunozm (commented):
I can not approve, but you have mine!

hieronx (approved):

PR #254, opened 2025-04-14
Add batch IDs

hieronx:


src/common/Gateway.sol
```diff
@@ -132,10 +132,12 @@ contract Gateway is Auth, IGateway, Recoverable {
             return processor.handle(centrifugeId, payload);
         }
 
+        bytes32 batchId = keccak256(abi.encodePacked(centrifugeId, payload));
         bytes32 messageProofHash = processor.messageProofHash(payload);
         bool isMessageProof = messageProofHash != bytes32(0);
         if (adapter.quorum == 1 && !isMessageProof) {
             // Special case for gas efficiency
+            emit ProcessBatch(centrifugeId, batchId, payload, adapter_);
             _handleBatch(centrifugeId, payload);
```
lemunozm, line 144: Good catch!

lemunozm (commented):
Right now the centrifugeIds will not match, because the sending part uses the remote one, and the receiving part uses the remote.

It means, when sending from A to B:
- SendBatch(B)
- ProcessBatch(A)

So they has different hashes. I think that would be fixed using both centrifugeIds local and remote

hieronx:
> Right now the centrifugeIds will not match, because the sending part uses the remote one, and the receiving part uses the remote.

Fixed in the last commit (just before your review)

src/common/Gateway.sol
```diff
@@ -201,7 +206,7 @@ contract Gateway is Auth, IGateway, Recoverable {
         }
     }
 
-    function retry(uint16 centrifugeId, bytes memory message) external {
+    function retry(uint16 centrifugeId, bytes memory message) external pauseable {
```
hieronx, line 209: Flagged by cross-chain security review, that this should also be pauseable.

lemunozm (approved):

PR #255, opened 2025-04-14
Hub: Remove add debit/credit and add updateJournal

lemunozm:
- Remove add debit/credit
- Add UpdateJournal
- Add tests for `isAdmin` check


src/hub/Hub.sol
```diff
@@ -370,20 +370,13 @@ contract Hub is Multicall, Auth, Recoverable, IHub, IHubGatewayHandler {
...
 
         accounting.unlock(poolId);
-        accounting.addDebit(account, amount);
-        accounting.lock();
-    }
 
-    /// @inheritdoc IHub
-    function addCredit(PoolId poolId, AccountId account, uint128 amount) external payable {
-        _protected(poolId);
+        _updateJournal(debits, credits);
```
hieronx: Maybe `_updateJournal` should just be a method exposed by `Accounting`? Would also make the `Hub` a bit smaller

hieronx (dismissed):

src/hub/Hub.sol
lemunozm: Interesting! If we can assume this will never will change. I agree in doing it

hieronx: @mustermeiszer thoughts?

mustermeiszer: Ahhh. How do we envision calling into that from externals?

mustermeiszer (dismissed):
One remark: 

Is it possible to merge `hub.updateJournal` and `hub.updateJournalEntries`?

src/hub/Hub.sol
hieronx: Just like it is now, `Hub.updateJournal` and `Hub.updateJournalEntries`. Only difference is that those call `Accounting.updateJournal` rather than `Hub._updateJournal` internal method.

hieronx:
As discussed in sync:

Add `updateJournal` to `Accounting`, remove `_updateJournal` internal method from `Hub`.

Remove BalanceSheet.journalEntry, sendJournalEntry + Message, updateJournalEntry handler in Hub.

And also remove the `meta` param in the BalanceSheet methods.

src/vaults/BalanceSheet.sol
```diff
@@ -124,21 +120,10 @@ contract BalanceSheet is Auth, Recoverable, IBalanceSheet, IBalanceSheetGatewayH
...
 
-    /// @inheritdoc IBalanceSheet
-    function updateValue(PoolId poolId, ShareClassId scId, address asset, uint256 tokenId, D18 pricePoolPerAsset)
-        external
-        auth
-    {
-        uint128 assetId = poolManager.assetToId(asset, tokenId);
-        sender.sendUpdateHoldingValue(poolId, scId, AssetId.wrap(assetId), pricePoolPerAsset);
-        emit UpdateValue(poolId, scId, asset, tokenId, pricePoolPerAsset, uint64(block.timestamp));
-    }
```
lemunozm, line 140: Talked with Frederik about removing this and avoid concurrent issues with the counterpart in CP side.

src/hub/Hub.sol
```diff
@@ -397,20 +430,19 @@ contract Hub is Multicall, Auth, Recoverable, IHub, IHubGatewayHandler {
...
-    /// @inheritdoc IHub
-    function addCredit(PoolId poolId, AccountId account, uint128 amount) external payable {
-        _protected(poolId);
+        for (uint256 i; i < debits.length; i++) {
+            accounting.addDebit(debits[i].accountId, debits[i].amount);
+        }
+
+        for (uint256 i; i < credits.length; i++) {
+            accounting.addCredit(credits[i].accountId, credits[i].amount);
+        }
```
lemunozm, line 444: TODO: Move it to the Accounting

src/hub/Accounting.sol
```diff
@@ -63,7 +63,7 @@ contract Accounting is Auth, IAccounting {
         debited = 0;
         credited = 0;
         _currentPoolId = poolId;
-        
+
         if (TransientJournal.journalId(poolId) == 0) {
             TransientJournal.setJournalId(poolId, _generateJournalId(poolId));
         }
```
lemunozm, line 69: TODO: remove this. it seems unused: https://kflabs.slack.com/archives/C07PG2EUR9C/p1744712036752739

src/common/libraries/MessageLib.sol
```diff
@@ -138,22 +129,13 @@ library MessageLib {
         uint8 kind = message.toUint8(0);
         require(kind <= uint8(type(MessageType).max), UnknownMessageType());
 
-        length = (kind <= 31)
-            ? uint16(uint8(bytes32(MESSAGE_LENGTHS_1)[31 - kind]))
-            : uint16(uint8(bytes32(MESSAGE_LENGTHS_2)[63 - kind]));
+        length = uint16(uint8(bytes32(MESSAGE_LENGTHS_1)[31 - kind]));
```
hieronx, line 132: Yay

src/hub/Accounting.sol
lemunozm: Unless we still want to link message in batches using the JournalId, not sure if it's still a feature

hieronx (approved):
🔥

PR #258, opened 2025-04-15
fix: assetId consitency

GalloDaSballo:
Attempts to rename every instance of `AssetId` to `assetId` to confirm that all of those are the same asset

Using different names for the same thing is condusize to thinking it can be something else


wischli:
While we totally understand your point we decided against it as we prefer being more explicit, at least for call arguments and event parameters. But then it doesn't make sense for us to have a generic `assetId` for storages and more explicit `{deposit, payout}AssetId` elsewhere. We want to express that even though a user might deposit with asset `A`, the can redeem into asset `B`.

PR #260, opened 2025-04-15
Fix batch hash in `SendProof` event, apply gas optimization

hieronx:


src/common/Gateway.sol
```diff
@@ -289,8 +292,7 @@ contract Gateway is Auth, IGateway, Recoverable {
             (isBatching) ? batchGasLimit[centrifugeId][poolId] : gasService.gasLimit(centrifugeId, batch);
 
         for (uint256 i; i < adapters_.length; i++) {
-            bool isPrimaryAdapter = i == PRIMARY_ADAPTER_ID - 1;
```
hieronx, line 292: Needed to avoid stack too deep

src/common/Gateway.sol
```diff
@@ -137,13 +137,14 @@ contract Gateway is Auth, IGateway, Recoverable {
         Adapter memory adapter = _activeAdapters[centrifugeId][adapter_];
         require(adapter.id != 0, InvalidAdapter());
 
-        if (processor.isMessageRecovery(payload)) {
+        IMessageProcessor processor_ = processor;
```
lemunozm, line 140: Does it save some read? I though this was automatically cached

lemunozm (approved):

PR #262, opened 2025-04-15
feat: pool escrow

wischli:
Adds per pool escrow(s) that internally separate by share class ids
* Deployed via new factory `PoolEscrowFactory` during `poolManager.addShareClass`
* Exposes deterministic retrieval of escrow address for a given pool id (`poolEscrowFactory.escrow(poolId)`) which is more lightweight (roughly 600-900 gas) than a storage read (2100-2600 gas).

## Backwards compatibility with Legacy v2 vaults

* Legacy vaults retrieve the global escrow from the async investment manager `manager.escrow()` ([ref](https://github.com/centrifuge/vaults-internal/blob/5e8262b26f8b3b488fa11855bfc2ee2e9943c09c/src/ERC7540Vault.sol#L102))
* Therefore, the previous implicit `BaseInvestmentManager.escrow()` from the public escrow storage was replaced with a custom function that assumes the message sender to be `vault` such that we can derive the pool id via `IBaseVault(vault).poolId()`. We do this using `staticcall` which reverts if the `msg.sender` does not support `poolId()`
* Nevertheless, we use `poolEscrowProvider.escrow(poolId)` in v3 vaults to save gas due to not requiring calling into the underlying investment manager 

TODOs

- [x] Add tests for escrow factory 
- [x] Ensure solution is lightweight
- [x] Ensure not breaking for legacy vaults + add docs for that


src/hooks/RestrictedTransfers.sol
```diff
@@ -67,13 +70,13 @@ contract RestrictedTransfers is Auth, IRestrictedTransfers, IHook {
...
     {
-        if (uint128(hookData.from).getBit(FREEZE_BIT) == true && !root.endorsed(from)) {
+        if (uint128(hookData.from).getBit(FREEZE_BIT) == true && !root.endorsed(from) && from != ESCROW_IDENTIFIER) {
             // Source is frozen and not endorsed
             return false;
         }
 
-        if (root.endorsed(to) || to == address(0)) {
-            // Destination is endorsed and source was already checked, so the transfer is allowed
+        if (root.endorsed(to) || to == address(0) || to == ESCROW_IDENTIFIER) {
```
wischli: NOTE: **All** changes in this contract were necessary because previously the global escrow contract was endorsed by root during deployment which we cannot do anymore as escrows are deployed dynamically by the corresponding factory. If we still wanted to endorse, the factory would have to have auth permissions for root which is an attack vector we don't want to expose.

src/vaults/PoolManager.sol
```diff
@@ -191,7 +186,18 @@ contract PoolManager is Auth, Recoverable, IPoolManager, IUpdateContract, IPoolM
...
+        address escrow = poolEscrowFactory.escrows(poolId);
+        // Deploy new escrow only on first added share class for pool
+        if (poolEscrowFactory.escrows(poolId) == address(0)) {
+            escrow = poolEscrowFactory.newEscrow(poolId);
+        }
+
+        // Update escrow as member to enable minting, burning and transfers on deposit and redeem
+        bytes memory updateMember =
+            MessageLib.UpdateRestrictionMember({user: escrow.toBytes32(), validUntil: type(uint64).max}).serialize();
+        IHook(hook).updateRestriction(shareToken_, updateMember);
```
wischli: NOTE: This was the second necessary change ([ref first](https://github.com/centrifuge/protocol-v3/pull/262#discussion_r2046335604)) due to the removal of `root.endorse` during deployment. Without setting the escrow as member, minting/burning and transferring from it will fail in the hooks ([ref](https://kflabs.slack.com/archives/C07PG2EUR9C/p1744735665106169?thread_ts=1744705032.092819&cid=C07PG2EUR9C)).

src/vaults/Escrow.sol
```diff
@@ -6,157 +6,147 @@ import {IERC20} from "src/misc/interfaces/IERC20.sol";
...
-contract Escrow is Auth, IPerPoolEscrow, IEscrow {
-    mapping(uint64 poolId => mapping(bytes16 scId => mapping(address token => mapping(uint256 tokenId => uint256))))
-        internal reservedAmount;
-    mapping(uint64 poolId => mapping(bytes16 scId => mapping(address token => mapping(uint256 tokenId => uint256))))
-        internal pendingDeposit;
-    mapping(uint64 poolId => mapping(bytes16 scId => mapping(address token => mapping(uint256 tokenId => uint256))))
-        internal holding;
+import {IPoolEscrow, IEscrow} from "src/vaults/interfaces/IEscrow.sol";
 
+contract Escrow is Auth, IEscrow {
```
wischli, line 16: NOTE: Escrow contract still necessary, because we use that for the `routerEscrow`. However, the `routerEscrow` is only `IEscrow` which is why I separated both contracts.

src/hooks/FreelyTransferable.sol
```diff
@@ -32,13 +32,14 @@ contract FreelyTransferable is Auth, IRestrictedTransfers, IHook {
 
     /// @dev Least significant bit
     uint8 public constant FREEZE_BIT = 0;
+    /// @dev Magic address denoting a transfer to the escrow
+    /// @dev Solely used for gas saving since escrow is per pool
+    address public constant ESCROW_IDENTIFIER = address(uint160(uint8(0xce)));
```
wischli: Ref: https://kflabs.slack.com/archives/C07PG2EUR9C/p1744705032092819

src/vaults/BaseInvestmentManager.sol
```diff
@@ -16,23 +16,24 @@ import {IBaseVault} from "src/vaults/interfaces/IERC7540.sol";
...
+import {IPoolEscrowProvider} from "src/vaults/interfaces/factories/IPoolEscrowFactory.sol";
 
 abstract contract BaseInvestmentManager is Auth, Recoverable, IBaseInvestmentManager {
     using MathLib for uint256;
 
     address public immutable root;
-    address public immutable escrow;
 
     IPoolManager public poolManager;
+    IPoolEscrowProvider public poolEscrowProvider;
```
wischli, line 27: TODO: Discuss if we rather want the provider to be immutable

src/vaults/Escrow.sol
```diff
@@ -6,157 +6,147 @@ import {IERC20} from "src/misc/interfaces/IERC20.sol";
...
+
+    // TODO: Endorse root and rely on it for auth?
+    constructor(uint64 poolId_, address deployer) Escrow(deployer) {
+        poolId = poolId_;
+    }
 
-    /// @inheritdoc IPerPoolEscrow
-    function pendingDepositIncrease(address token, uint256 tokenId, uint64 poolId, bytes16 scId, uint256 value)
+    /// @inheritdoc IPoolEscrow
+    function pendingDepositIncrease(bytes16 scId, address asset, uint256 tokenId, uint256 value)
```
wischli: NOTE: I didn't change any logic for the calls, just reordered the input arguments to have `scId` as first argument (to be in sync with the rest of the codebase). I did not change the order of variables in the events to not create any breaking changes.

src/vaults/BaseVaults.sol
```diff
@@ -76,7 +79,9 @@ abstract contract BaseVault is Auth, Recoverable, IBaseVault {
         share = token_;
         _shareDecimals = IERC20Metadata(share).decimals();
         root = IRoot(root_);
+        // TODO: Redundant due to filing?
         manager = IBaseInvestmentManager(manager_);
+        poolEscrowProvider = poolEscrowProvider_;
```
wischli: I was curious why for vaults we set the manager in the constructor as well as supporting `filing` for it. I extended that approach vor the escrow factory.

test/vaults/fuzzing/recon-core/Setup.sol
```diff
@@ -27,9 +28,11 @@ abstract contract Setup is BaseSetup, SharedStorage {
     TokenFactory tokenFactory;
 
     // Handled //
+    // TODO(wischli): Remove
     Escrow public escrow; // NOTE: Restriction Manager will query it
```
wischli: NOTE: Will work on this while dealing with reviews as the escrow is quite tightly coupled into the fuzzing.

test/vaults/integration/PoolManager.t.sol
```diff
@@ -365,11 +375,6 @@ contract PoolManagerTest is BaseTest, PoolManagerTestHelper {
             poolId, scId, MessageLib.UpdateRestrictionMember(randomUser.toBytes32(), validUntil).serialize()
         );
         assertTrue(shareToken.checkTransferRestriction(address(0), randomUser, 0));
-
-        vm.expectRevert(IRestrictedTransfers.EndorsedUserCannotBeUpdated.selector);
-        poolManager.updateRestriction(
-            poolId, scId, MessageLib.UpdateRestrictionMember(address(escrow).toBytes32(), validUntil).serialize()
-        );
```
wischli: NOTE: The global escrow was endorsed and could not be updated. AFAICT, we cannot forbid this anymore, as we utilize it ourselves in `PoolManager.addShareClass` to set the pool escrow as a member ([ref](https://kflabs.slack.com/archives/C07PG2EUR9C/p1744737734698919?thread_ts=1744705032.092819&cid=C07PG2EUR9C))

test/vaults/mocks/MockHook.sol
```diff
@@ -9,4 +9,6 @@ contract MockHook is Mock {
     function supportsInterface(bytes4 interfaceId) external pure returns (bool) {
         return interfaceId == type(IHook).interfaceId || interfaceId == type(IERC165).interfaceId;
     }
+
+    function updateRestriction(address token, bytes memory update) external {}
```
wischli, line 13: NOTE: Required for `PoolManager` test which just needs to be able to call it.

test/vaults/unit/Escrow.t.sol
```diff
@@ -40,7 +40,9 @@ contract EscrowTest is BaseTest {
         escrow.unapprove(address(erc20), spender);
         assertEq(erc20.allowance(address(escrow), spender), 0);
     }
+}
 
+contract PoolEscrowTest is BaseTest {
```
wischli: NOTE: Added full support for ERC6909

src/vaults/AsyncRequests.sol
```diff
@@ -45,21 +47,26 @@ contract AsyncRequests is BaseInvestmentManager, IAsyncRequests {
     using BytesLib for bytes;
     using MathLib for uint256;
 
+    /// @dev Magic address denoting a transfer to the escrow
+    /// @dev Solely used for gas saving since escrow is per pool
+    address public constant ESCROW_IDENTIFIER = address(uint160(uint8(0xce)));
```
lemunozm: We have this constant in several places. Can we move to a common location?

src/vaults/PoolManager.sol
```diff
@@ -210,7 +203,18 @@ contract PoolManager is Auth, Recoverable, IPoolManager, IUpdateContract, IPoolM
...
+        pool.shareClasses[scId].shareToken = shareToken_;
+
+        address escrow = poolEscrowFactory.deployedPoolEscrow(poolId);
+        // Deploy new escrow only on first added share class for pool
+        if (escrow == address(0)) {
+            escrow = poolEscrowFactory.newEscrow(poolId);
+        }
+
+        // Update escrow as member to enable minting, burning and transfers on deposit and redeem
+        bytes memory updateMember =
```
hieronx: This won't apply to every hook, since some might not support the `UpdateRestrictionMember` message.

Also some might not have a hook at all (`address(0)`), for fully permissionless hooks.

I think it is just up to the pool admin to ensure this message is called. Our UI can batch it.

src/vaults/SyncRequests.sol
```diff
@@ -315,7 +317,13 @@ contract SyncRequests is BaseInvestmentManager, ISyncRequests {
         balanceSheet.issue(poolId, scId, receiver, priceData.poolPerShare, shares);
 
         balanceSheet.deposit(
-            poolId, scId, vaultDetails.asset, vaultDetails.tokenId, escrow, depositAssetAmount, priceData.poolPerAsset
+            poolId,
+            scId,
+            vaultDetails.asset,
+            vaultDetails.tokenId,
+            address(poolEscrowProvider.poolEscrow(poolId.raw())),
```
hieronx: Nit: if you put `address escrow =  address(poolEscrowProvider.poolEscrow(poolId.raw()))` on the line above this, I believe the rest of the call would fit on 1 line.

src/hooks/FreelyTransferable.sol
hieronx: I'd add this to `IHook.sol`, since that is the shared interface for hooks

src/vaults/BaseVaults.sol
```diff
@@ -178,6 +188,12 @@ abstract contract BaseVault is Auth, Recoverable, IBaseVault {
         assets = manager.convertToAssets(address(this), shares);
     }
 
+    // --- Legacy views ---
+    /// @inheritdoc IBaseVault
+    function escrow() public view returns (address) {
```
hieronx: Doesn't this call need to be in `BaseInvestmentManager`? Since the v2.1 vaults call `manager.escrow()`

src/hooks/FreelyTransferable.sol
```diff
@@ -32,19 +32,21 @@ contract FreelyTransferable is Auth, IRestrictedTransfers, IHook {
...
+    constructor(address root_, address deployer) Auth(deployer) {
         root = IRoot(root_);
-        escrow = escrow_;
     }
 
     // --- Callback from share token ---
     /// @inheritdoc IHook
     function onERC20Transfer(address from, address to, uint256 value, HookData calldata hookData)
         external
+        view
```
hieronx: I don't think we should do this, see reasoning on Discord

src/vaults/factories/PoolEscrowFactory.sol
```diff
@@ -0,0 +1,85 @@
...
+
+        emit DeployPoolEscrow(poolId, address(escrow_));
+        return address(escrow_);
+    }
+
+    // --- View methods ---
+    /// @inheritdoc IEscrowProvider
+    function escrow(uint64 poolId) external view returns (address) {
+        if (poolId == V2_POOL_ID) {
+            return V2_GLOBAL_ESCROW;
```
hieronx: I likely am missing something: why are we still keeping the old global escrow around? I believe we can migrate this entirely.

src/vaults/BaseVaults.sol
wischli: Yes, you're right

src/vaults/factories/PoolEscrowFactory.sol
```diff
@@ -0,0 +1,85 @@
...
+    function file(bytes32 what, address data) external auth {
+        if (what == "poolManager") poolManager = data;
+        else if (what == "balanceSheet") balanceSheet = data;
+        else if (what == "asyncRequests") asyncRequests = data;
+        else revert FileUnrecognizedParam();
+        emit File(what, data);
+    }
+
+    /// @inheritdoc IPoolEscrowFactory
+    function newEscrow(uint64 poolId) public auth returns (address) {
```
lemunozm: NIT. Could we return here a `IPoolEscrow`?

src/vaults/factories/PoolEscrowFactory.sol
wischli: Do we need to support vaults from more than one pool id? If so, the `BaseInvestmentManager.escrow()` would be lacking the pool id information such that it can only map to one new global v2 escrow. Otherwise, good point that we could just migrate it.

lemunozm (commented):
I couple of NITs, but I need still to review this

src/vaults/factories/PoolEscrowFactory.sol
hieronx: > would be lacking the pool id information such that it can only map to one new global v2 escrow

It can just use `msg.sender` to look up the pool ID based on the vault address.

src/vaults/factories/PoolEscrowFactory.sol
wischli: Due to the refactor in [d72ad0a](https://github.com/centrifuge/protocol-v3/pull/262/commits/d72ad0a1e9fe0998e58731d1f0ea155a38fc0835) which removed `poolEscrow` and `IEscrowProvider`, I decided to just return the address and require consumers to type cast if necessary.

src/vaults/AsyncRequests.sol
wischli: [5991500](https://github.com/centrifuge/protocol-v3/pull/262/commits/599150076de04b9aa97c6729d096fba0c5d9800a)

src/hooks/FreelyTransferable.sol
wischli: [5991500](https://github.com/centrifuge/protocol-v3/pull/262/commits/599150076de04b9aa97c6729d096fba0c5d9800a)

src/vaults/PoolManager.sol
wischli: [91ae272](https://github.com/centrifuge/protocol-v3/pull/262/commits/91ae272baf0a209af7f80ee7dbe2c8e9b787aba4)

src/vaults/factories/PoolEscrowFactory.sol
wischli: [d72ad0a](https://github.com/centrifuge/protocol-v3/pull/262/commits/d72ad0a1e9fe0998e58731d1f0ea155a38fc0835)

src/vaults/BaseInvestmentManager.sol
```diff
@@ -67,6 +68,17 @@ abstract contract BaseInvestmentManager is Auth, Recoverable, IBaseInvestmentMan
...
 
+    /// @inheritdoc IBaseInvestmentManager
+    function escrow() public view returns (address) {
+        (bool success, bytes memory result) = msg.sender.staticcall(abi.encodeWithSelector(IBaseVault.poolId.selector));
+
+        require(success && result.length == 32, SenderNotVault());
+
+        uint64 poolId = abi.decode(result, (uint64));
+        return poolEscrowProvider.escrow(poolId);
+    }
```
lemunozm: Really smart solution! 🙌🏻

src/vaults/AsyncVault.sol
```diff
@@ -46,9 +51,9 @@ contract AsyncVault is AsyncRedeemVault, IAsyncVault {
         );
 
         if (tokenId == 0) {
-            SafeTransferLib.safeTransferFrom(asset, owner, asyncManager().escrow(), assets);
+            SafeTransferLib.safeTransferFrom(asset, owner, _poolEscrowProvider.escrow(poolId), assets);
```
lemunozm: Ok, I understand we do not break compatibility because `BaseInvestmentManager.escrow()` still exists.

src/vaults/factories/PoolEscrowFactory.sol
lemunozm: Ok. Just to notify, if you return an address instead of `IPoolEscrow` to avoid breaking retro-compatibility, both types, AFAIK, have the same abi. 

Is there any reason to keep the `address` as type when it always is a `IPoolEscrow` interface?

lemunozm (dismissed):
Everything looks amazing!!

Maybe we can add the link to the exact old deployed vault contract, to be able to understand the changes and in the future avoid breaking the old vaults. Not sure where to add this, maybe a retrocompatibility doc file. But not for now! Good job! 🥇

src/vaults/AsyncRequests.sol
```diff
@@ -259,7 +265,7 @@ contract AsyncRequests is BaseInvestmentManager, IAsyncRequests {
 
         // Burn redeemed share class tokens from escrow
         IShareToken shareToken = IShareToken(IAsyncVault(vault_).share());
-        shareToken.burn(address(escrow), shares);
+        shareToken.burn(poolEscrowProvider.escrow(poolId), shares);
```
mustermeiszer: QUESTION

Won't this call into the `hook` too and then block as it is not he `ESCROW_HOOK_ID`?

src/vaults/AsyncRequests.sol
```diff
@@ -230,7 +236,7 @@ contract AsyncRequests is BaseInvestmentManager, IAsyncRequests {
 
         // Mint to escrow. Recipient can claim by calling deposit / mint
         IShareToken shareToken = IShareToken(IAsyncVault(vault_).share());
-        shareToken.mint(address(escrow), shares);
+        shareToken.mint(poolEscrowProvider.escrow(poolId), shares);
```
mustermeiszer: QUESTION

Won't this call into the `hook` too and then block as it is not he `ESCROW_HOOK_ID`?

src/vaults/BaseInvestmentManager.sol
mustermeiszer: Thats really smart! Great solution

src/vaults/BaseVaults.sol
```diff
@@ -87,6 +95,8 @@ abstract contract BaseVault is Auth, Recoverable, IBaseVault {
     // --- Administration ---
     function file(bytes32 what, address data) external auth {
         if (what == "manager") manager = IBaseInvestmentManager(data);
+        /// @dev NOT supported in legacy v2 vaults
```
mustermeiszer: ❤️

src/vaults/factories/PoolEscrowFactory.sol
```diff
@@ -0,0 +1,64 @@
...
+        require(escrows[poolId] == address(0), EscrowAlreadyDeployed());
+        PoolEscrow escrow_ = new PoolEscrow{salt: bytes32(uint256(poolId))}(poolId, address(this));
+
+        escrow_.rely(root);
+        escrow_.rely(poolManager);
+        escrow_.rely(balanceSheet);
+        escrow_.rely(asyncRequests);
```
mustermeiszer, line 43: QUESTION

Do we need to rely them? Isn't the approval enough?

mustermeiszer (commented):
Amazing work! 

Do we check that `mint/burn` work as the `CentrifugeToken` seems to check the `onTransfer` logic that calls into the hook. I would expect those to fail if we receiver or sender is the new pool escrow. Happy if it works. Just wanted to highlight that. 
If it fails, couldn't we store the escrow in the hook itself as a storage item? 

Added the requirement for updating the hook of legacy vaults [here](https://www.notion.so/Deployment-Legacy-1b92eac24e178042a6d1d1881807d717?pvs=4#1d82eac24e178087b0f9cf96c459eb02) to our list in notion.

src/vaults/factories/PoolEscrowFactory.sol
wischli: I have no strong opinion about this. I have to admit that I was under the assumption that type casting actually costs gas but since it's just a pure compile-time instruction, I was wrong. Then, I prefer your path actually.

src/vaults/AsyncRequests.sol
wischli: Yes, this is the reason why we have to add the pool escrow as a member, [see above](https://github.com/centrifuge/protocol-v3/pull/262#discussion_r2046338528). I removed doing that inside `PoolManagar.addShareClass`, because Jeroen pointed out that this should rather be done manually by the FM ([ref](https://github.com/centrifuge/protocol-v3/pull/262#discussion_r2047689497)).

Therefore, we do this in vault integration tests inside our mock `centrifugeChain.addShareClass`.

src/vaults/AsyncRequests.sol
wischli: Same response as [above](https://github.com/centrifuge/protocol-v3/pull/262#discussion_r2048810761).

src/vaults/factories/PoolEscrowFactory.sol
wischli: We need to rely all of these contracts because they are executing escrow calls, i.e. it doesn't suffice for them to have read access.

src/vaults/factories/PoolEscrowFactory.sol
wischli: [99cb82a](https://github.com/centrifuge/protocol-v3/pull/262/commits/99cb82a42bf8505c80782b08cfcd664d803b9fd6)

src/vaults/factories/PoolEscrowFactory.sol
mustermeiszer: Ah, yes! the increase and so on! Perfect!

src/vaults/AsyncRequests.sol
mustermeiszer: Perfect! THanks for the explainer.

mustermeiszer: Just a thought @hieronx couldn't somebody use the escrow as a provider of funds? IIRC this is possible as we use `transferFrom` in the `SyncReyest` contract, right? IN the `AsyncRequest` it is the vault, this one does not have permission, but in the `SyncRequest` it is the manager.

mustermeiszer: CR

Also just checked, the given flow will not work with the current managers - as `balanceSheet.deposit()` expects the tokens to already be there -part of the `manager.mint()` flow , but the `BaseVault` only transfers them after `manager.mint()` call. 

we probably need to do change the logic in `baseVault.mint()` to 
1.  `manager.preMint()` - issuance and transfer os shares
2. `SafeTransferFrom()` - token to escrow
3.  `manager.mint()` - `balanceSheet.deposit()` logic

wischli: Unfortunately, we cannot change the `mint()` behaviour of legacy vaults, so it appears to me that we need to find another solution?

Just a quick question: Since this flow wasn't changed, it appears to me that it was broken before as well?

mustermeiszer: That is just a requirement for the sync vaults. So we can change that. The async ones work with pending and increase just fine

wischli: Side question after looking deeper into the sync mint flow: I don't fully understand our `BaseSyncDepositVault.mint`: When we execute `syncDepositManager.mint(address(this, shares, receiver, msg.sender))` we 
1. Mint shares to the receiver via `balanceSheet.issue`
2. Execute `balanceSheet.deposit()` with **zero asset amount**. I.e., we neither increase the pending deposit, nor transfer anything and also send a holding update with zero increased amount. However, we then do the actual asset transfer afterwards in the sync vault `SafeTransferLib.safeTransferFrom(asset, msg.sender, address(_poolEscrowProvider.escrow(poolId)), assets)`.

So shouldn't `syncRequests.mint -> _issueShares` actually do 

```diff
        balanceSheet.deposit(
-            poolId, scId, vaultDetails.asset, vaultDetails.tokenId, escrow, depositAssetAmount = 0, priceData.poolPerAsset
+            poolId, scId, vaultDetails.asset, vaultDetails.tokenId, owner, depositAssetAmount > 0, priceData.poolPerAsset
        );
```

Where `owner` is the `msg.sender` of `BaseSyncDepositVault` such that we can also decrease the `BaseSyncDepositVault.mint` to

```diff
        assets = syncDepositManager.mint(address(this), shares, receiver, msg.sender);
-        SafeTransferLib.safeTransferFrom(asset, msg.sender, address(_poolEscrowProvider.escrow(poolId)), assets);
        emit Deposit(receiver, msg.sender, assets, shares);
    }
```

wischli: @mustermeiszer Committed my proposal, curious about your thoughts: https://github.com/centrifuge/protocol-v3/pull/262/commits/b679255c000eee1d424d874cdff80de980911c6a

Not sure if this is acceptable (assuming it's not) because then the transfer of the assets is not initiated from the vault but the BSM such that it needs approval.

The idea was that we previously (red) did what IMO can be shortened (green):
```diff

Sync vault deposit: 
-  1. Transfer assets `msg.sender -> escrow` and then
   2. Sync request `deposit` which calls `_issueShares` with non-zero asset amount and 
-     * `provider = escrow`, such that we transfer from escrow to escrow in BSM
+     * `provider = msg.sender`, such that we transfer from msg.sender to escrow in BSM

Sync vault mint:
  1. Sync request `mint` which called `_issueShares` with 
-    * zero amount and `provider = escrow`, such that we transfer from escrow to escrow in BSM
+    * non-zero asset amount and `provider = msg.sender`, such that we transfer from msg.sender to escrow in BSM
- 2. Transfer assets `msg.sender -> escrow`
```

src/vaults/AsyncRequests.sol
```diff
@@ -128,7 +128,7 @@ contract AsyncRequests is BaseInvestmentManager, IAsyncRequests {
         require(state.pendingCancelDepositRequest != true, CancellationIsPending());
 
         state.pendingDepositRequest += assets;
-        IPoolEscrow(poolEscrowProvider.escrow(poolId)).pendingDepositIncrease(
+        IPoolEscrow(address(poolEscrowProvider.escrow(poolId))).pendingDepositIncrease(
```
lemunozm, line 131: NIT. THis can be:
```suggestion
        poolEscrowProvider.escrow(poolId).pendingDepositIncrease(
```
If poolEscrowProvider already store IPoolEscrow in the mapping

src/vaults/AsyncRequests.sol
hieronx: @wischli this doesn't work because with ERC-4626 `deposit`/`mint`, the approval of assets needs to be to the vault address (as that is the only address an integrator knows). With your alternative proposal, the ERC20 allowance would need to be given to the `BalanceSheet`.

hieronx: > That is just a requirement for the sync vaults. So we can change that. The async ones work with pending and increase just fine

For re-entrancy protection it is actually better if the transfer is at the end of the transaction. See the discussions with Alex about the hook security considerations.

Maybe we have a ` bool externalTransfer` in `deposit()` which basically doesn't check the balance? Not super clean but seems like the safest approach.

hieronx: Or we just add `protected` reentrancy guard to `syncDepositVault.deposit/mint`, maybe cleaner.

src/vaults/AsyncRequests.sol
```diff
@@ -150,8 +153,8 @@ contract AsyncRequests is BaseInvestmentManager, IAsyncRequests {
         );
 
         require(
-            _canTransfer(vaultAddr, owner, address(escrow), shares)
-                && _canTransfer(vaultAddr, controller, address(escrow), shares),
+            _canTransfer(vaultAddr, owner, ESCROW_HOOK_ID, shares)
+                && _canTransfer(vaultAddr, controller, ESCROW_HOOK_ID, shares),
```
mustermeiszer, line 154: Just a side note: 

* IF we anuyways need to whitelist the escrow in order for this too work, why have the `ESCROW_HOOK_ID` at all?

lemunozm:
Maybe we can merge this already to avoid future conflicts with others refactorings (?)

src/vaults/AsyncRequests.sol
wischli: Seems feasible to me. @hieronx WDYT?

src/vaults/AsyncRequests.sol
wischli: Fixed in two commits:
1. [2f8d13b](https://github.com/centrifuge/protocol-v3/pull/262/commits/2f8d13ba54728e6c55507bcfe154edad662f3fc7): Adds `BalanceSheet.noteDeposit` but does not perform transfers at the end of the vault calls
2. [aa7ba3a](https://github.com/centrifuge/protocol-v3/pull/262/commits/aa7ba3abb83096af2e23e9b1caa7b0dcf80b10d6): Adds `Escrow.noteDeposit` such that we can execute transfers last

As discussed in the sync, this solution is hacky but I also couldn't come up with something better that achieves the same result this effectively.

src/vaults/BalanceSheet.sol
```diff
@@ -229,19 +245,26 @@ contract BalanceSheet is Auth, Recoverable, IBalanceSheet, IBalanceSheetGatewayH
         uint256 tokenId,
         address provider,
         uint128 amount,
-        D18 pricePoolPerAsset
+        D18 pricePoolPerAsset,
```
mustermeiszer, line 248: QUESTION

No, blocker. But for follow-up. I would propose instead of adding all of that here, we should just trigger the message and the deposit call to the escrow directly FROM the `SyncRequyest` manager.

src/vaults/AsyncRequests.sol
hieronx: To be able to identify redemption requests (https://github.com/centrifuge/protocol-v3/blob/main/src/vaults/interfaces/token/IHook.sol#L21)

lemunozm (approved):
LGTM!!

Good job to both of you dealing with this retro-compatible stuff!

src/vaults/SyncRequests.sol
```diff
@@ -298,24 +300,31 @@ contract SyncRequests is BaseInvestmentManager, ISyncRequests {
...
         // Mint shares for receiver & notify CP about issued shares
         balanceSheet.issue(poolId, scId, receiver, priceData.poolPerShare, shares);
 
-        balanceSheet.deposit(
-            poolId, scId, vaultDetails.asset, vaultDetails.tokenId, escrow, depositAssetAmount, priceData.poolPerAsset
+        // Notice deposit of shares and notify CP about holding update
+        // NOTE: Does not transfer shares from escrow to owner and assumes that to be handled elsewhere, i.e. in
+        // SyncDepositVault.{mint, deposit}
+        balanceSheet.noteDeposit(
+            poolId, scId, vaultDetails.asset, vaultDetails.tokenId, owner, depositAssetAmount, priceData.poolPerAsset
```
mustermeiszer, line 327: QUESTION

Same as above, for follow-up. Just trigger message and deposit here directly. I also would like to move the message receiving to the `AsyncRequest` manager in this PR. Makes it more obvious IMO that it is handled there as it is tiglhlity coupled to the logic there. Also for reviewers seeing a deposit there instead of the BSM.

mustermeiszer (approved):
LGTM! 

I can do the follow-up.

src/vaults/BalanceSheet.sol
wischli: But then the `BalanceSheet` is not the single source of truth for writing to the escrow anymore which I thought was one principe we wanted to follow. I suppose that changed given the latest work we did including this PR.

Would gladly let you do the follow up!

src/vaults/BaseVaults.sol
```diff
@@ -366,7 +373,8 @@ abstract contract BaseSyncDepositVault is BaseVault {
     /// @inheritdoc IERC7575
     function mint(uint256 shares, address receiver) public returns (uint256 assets) {
         assets = syncDepositManager.mint(address(this), shares, receiver, msg.sender);
-        SafeTransferLib.safeTransferFrom(asset, msg.sender, syncDepositManager.escrow(), assets);
+        // NOTE: For security reasons, transfer must stay at end of call
+        SafeTransferLib.safeTransferFrom(asset, msg.sender, address(_poolEscrowProvider.escrow(poolId)), assets);
```
wischli, line 377: Post mortem: Fixes https://github.com/Recon-Fuzz/centrifuge-review-protocol-v3/issues/36

GalloDaSballo: LG, swaps the order to perform the unsafer call after

PR #263, opened 2025-04-16
Ref: Give contract support to create3

lemunozm:
Fix: https://www.notion.so/Use-deployer-instead-of-msg-sender-in-all-contracts-1d62eac24e17800393b3e5ed8a2b8760

### DIFF
--- a/script/CommonDeployer.s.sol
+++ b/script/CommonDeployer.s.sol
@@ -60,7 +60,7 @@ abstract contract CommonDeployer is Script, JsonRegistry {
         messageProcessor = new MessageProcessor(root, tokenRecoverer, deployer);
 
         gasService = new GasService(maxBatchSize, messageGasLimit);
-        gateway = new Gateway(centrifugeId, root, gasService);
+        gateway = new Gateway(centrifugeId, root, gasService, deployer);
 
         messageDispatcher = new MessageDispatcher(centrifugeId, root, gateway, tokenRecoverer, deployer);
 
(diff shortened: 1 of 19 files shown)


lemunozm:
I think I can remove the change for contracts created by factories, but not sure if we want to leave everything following the same pattern

wischli:
> I think I can remove the change for contracts created by factories, but not sure if we want to leave everything following the same pattern

IIUC, in practice there would be no difference right? Because contracts deployed by factories have the factory as `msg.sender` and thus also as deployer if provided as `address(this)`.

src/vaults/factories/TokenFactory.sol
```diff
@@ -46,7 +46,11 @@ contract TokenFactory is Auth, ITokenFactory {
                 bytes1(0xff),
                 address(this),
                 salt,
-                keccak256(abi.encodePacked(type(CentrifugeToken).creationCode, abi.encode(decimals)))
+                keccak256(
+                    abi.encodePacked(
+                        type(CentrifugeToken).creationCode, abi.encode(decimals), abi.encode(address(this))
+                    )
+                )
```
wischli: I think this change serves as argument to revert the changes for factory-deployed contracts. We must not change `getAddress` for factories which we already use in practice because this breaking change is not backwards compatible with our legacy Maker and Morpho vaults, i.e. `getAddress` now would return different a different deterministic address.

lemunozm: I think you're right. I'm gonna revert the changes that have to do with factories. We only need to support Create3 in contracts directly deployed by our scripts

lemunozm: Also it reduces a bit of noise

wischli: Yes, factories can still deploy with `Create2` confirmed by Jeroen ([ref](https://kflabs.slack.com/archives/C07PG2EUR9C/p1744709568426489?thread_ts=1744709101.655979&cid=C07PG2EUR9C)).

wischli (approved):
LGTM!

PR #266, opened 2025-04-16
Feat: Refund per escrow

lemunozm:
This finalize the refund gateway cycle. No offchain calls are need it.

### "Missing" points

- The old vaults uses the old escrow. That would mean that we can not refund that pool right now. My idea would be to just manually deploy a `PoolEscrow` for the living vault. The living vault will still use the old escrow but from Gateway perspective everything is well configured


src/vaults/factories/PoolEscrowFactory.sol
```diff
@@ -76,7 +80,8 @@ contract PoolEscrowFactory is IPoolEscrowFactory, Auth {
     // --- Internal methods ---
     function _deterministicAddress(uint64 poolId) internal view returns (address) {
         bytes32 salt = bytes32(uint256(poolId));
-        bytes memory bytecode = abi.encodePacked(type(PoolEscrow).creationCode, abi.encode(poolId, address(this)));
+        bytes memory bytecode =
+            abi.encodePacked(type(PoolEscrow).creationCode, abi.encode(poolId, sharedGateway, address(this)));
```
wischli: IIUC, `sharedGateway` cannot be assumed to be static right? If it is indeed not guaranteed to be static, this would be a breaking change for this deterministic address retrieval.

wischli:
Maybe I am misunderstanding something due to fatigue. I believe the escrows should stay decoupled from the shared gateway. Two questions:

1. Can you elaborate on why you want to link all existing escrows to a new gateway, once it emerges, from our deployed v3 code and not within a migration spell, if necessary?

2. Shouldn't the gateway gain access to escrows via the `IPoolEscrow` as other contracts do? And is write access required or does view suffice for retrieving the address based on an input pool id?

src/vaults/factories/PoolEscrowFactory.sol
lemunozm: Yes! it's guaranteed to be static. That's why I added the `SharedDependency` contract. If `Gateway` changes, we just change the gateway address inside `SharedDependency`, without touching the rest of the consumers, aka, pool escrows.

lemunozm:
> Can you elaborate on why you want to link all existing escrows to a new gateway, once it emerges, from our deployed v3 code and not within a migration spell, if necessary?

It's the only way to transmit the funds deposited through `receive()` back to the `Gateway`. If we do not that, we would need some external call to `withdrawEth` from the `escrow`, and later subsidize the pool. I wanted to avoid that external call. Of course open to discussing PROS/CONS or other solutions.

> Shouldn't the gateway gain access to escrows via the IPoolEscrow as other contracts do? And is write access required or does view suffice for retrieving the address based on an input pool id?

The problem is that the `Gateway` doesn't know anything about the escrow. What happens here is that the adapter will send back asynchronously some funds to the escrow. And we need to move those funds back to the `Gateway`.

lemunozm:
Thinking in what you say, maybe the Gateway can try to withdrawFunds from the escrow each time subsidizing a pool is required 🤔. Will think about that

lemunozm:
I've re-done the current PR @william. Thanks for make me think in another possible solutions.

The only downside with the new solution is that the UX should do `subsidized[pool] + escrow.balance` to show the correct subsidized pool available to sending message. But I think it's totally feasible.

src/common/Gateway.sol
```diff
@@ -347,13 +349,22 @@ contract Gateway is Auth, IGateway, Recoverable {
         transactionPayer = address(0);
     }
 
-    function setRefundAddress(PoolId poolId, address refund) public auth {
+    function _requestPoolFunding(PoolId poolId) internal {
+        IRecoverable refund = subsidy[poolId].refund;
+        if (!poolId.isNull() && address(refund) != address(0)) {
+            _receivePoolTarget = poolId;
```
hieronx: I don't think this is really needed, if there is just a permissionless method on the `Escrow` to call `gateway.subsidizePool` from it

hieronx: That would make every single `send()` cheaper, which is important.

hieronx: Actually why can't we just call `gateway.subsidizePool` in `Escrow.receive`? 🤔

lemunozm: That was my first approach (see [here](https://github.com/centrifuge/protocol-v3/commit/50c2ca496ea23824203d5b39b32af0fe4694a8e5)). But it seemed to introduce more noise to the Escrow. I needed another contract in between to be able to migrate easily the Gateway for each Escrow in the future, and the factory was more complex too.

WDYT, if we only call the refund method in case there is no money in the subsidized pool? I mean, you tried to send as usual, an only if there is no enough money, you try to get it from the Escrow

lemunozm: I think it makes sense, at least conceptually, that the Escrows don't know about a Gateway existance. It should be the Gateway who recover the tokens from them.

hieronx: Alright spent some more time thinking through this. I understand your point about the other approach creating more noise.

> WDYT, if we only call the refund method in case there is no money in the subsidized pool? I mean, you tried to send as usual, an only if there is no enough money, you try to get it from the Escrow

I agree this is a good idea.

I would add to that: I think the logic with `_receivePoolTarget` is quite complex and hard to read/audit. I would rather do something like this:
```solidity
function _requestPoolFunding(PoolId poolId) internal {
        IRecoverable refund = subsidy[poolId].refund;
        if (!poolId.isNull() && address(refund) != address(0)) {
            uint256 refundBalance = address(refund).balance;
            refund.recoverTokens(ETH_ADDRESS, address(this), refundBalance);
            subsidy[poolId].value += uint96(refundBalance);
            emit SubsidizePool(poolId, msg.sender, refundBalance);
        }
    }
```

lemunozm: Will do!

hieronx: I think we might need to add something like
```solidity

uint256 public constant ESCROW_REQUEST_THRESHOLD = 0.001 ether;

function _requestPoolFunding(PoolId poolId) internal {
    ...
    if (address(refund).balance < ESCROW_REQUEST_THRESHOLD) return;
    ...
}
```

otherwise, someone can make every gateway call more expensive by sending 1 wei to the pool escrow.

lemunozm: Why not have:

```solidity
if (subsidy[poolId].value < consumed) {
    // Not enough to pay the message, try to recover all Eth from the escrow
    _requestPoolFunding(..);
}
```

That way, no matter the Eth the escrow have, it will only try to recover if funds are needed. That should be quite eventual because when you recover, you recover all funds.

hieronx: Yeah that should work!

lemunozm:
@hieronx I realize that if we migrate the gateway, we would need to update `escrow.rely(newGateway)` for all escrows. Is this something we need to avoid? Or can we deal with it?

The same will happen with any dependency, such as the AsyncRequest migration.

src/common/Gateway.sol
```diff
@@ -362,15 +365,33 @@ contract Gateway is Auth, Recoverable, IGateway {
...
+            uint256 refundBalance = address(refund).balance;
+
+            // Send to the gateway GLOBAL_POT
+            refund.recoverTokens(ETH_ADDRESS, address(this), refundBalance);
+
+            // Extract from the GLOBAL_POT
+            subsidy[GLOBAL_POT].value -= uint96(refundBalance);
+            _subsidizePool(poolId, address(refund), refundBalance);
+        }
+    }
```
lemunozm, line 380: Finally, this is the request method. Not sure if it's more legible than using the previous `_transientRefundPool`.

1. Add funds to gateway.subsidy[Pool0]
2. Decrement gateway.subsidy[Pool0]
3. Increment gateway.subsidy[PoolA]

hieronx:
> @hieronx I realize that if we migrate the gateway, we would need to update escrow.rely(newGateway) for all escrows. Is this something we need to avoid? Or can we deal with it?
> The same will happen with any dependency, such as the AsyncRequest migration.

I think that is unavoidable and acceptable.

src/common/Gateway.sol
hieronx: Hm didn't realize that before, but yeah I do think this is still better than the transient option. Simpler control flow imo.

src/vaults/factories/PoolEscrowFactory.sol
```diff
@@ -34,6 +37,7 @@ contract PoolEscrowFactory is IPoolEscrowFactory, Auth {
         PoolEscrow escrow_ = new PoolEscrow{salt: bytes32(uint256(poolId))}(poolId, address(this));
 
         escrow_.rely(root);
+        escrow_.rely(gateway);
         escrow_.rely(poolManager);
         escrow_.rely(balanceSheet);
         escrow_.rely(asyncRequests);
```
wischli, line 43: Note to myself: Missing test coverage for deployed escrow wards

wischli (approved):
LGTM!

src/common/Gateway.sol
```diff
@@ -36,6 +36,7 @@ contract Gateway is Auth, Recoverable, IGateway {
     using TransientStorageLib for bytes32;
     using MessageProofLib for *;
 
+    PoolId public constant GLOBAL_POT = PoolId.wrap(0);
     uint8 public constant MAX_ADAPTER_COUNT = 8;
     uint8 public constant PRIMARY_ADAPTER_ID = 1;
     uint256 public constant RECOVERY_CHALLENGE_PERIOD = 7 days;
```
hieronx, line 42: Ultra nit
```suggestion
    uint8 public constant PRIMARY_ADAPTER_ID = 1;
    uint8 public constant MAX_ADAPTER_COUNT = 8;
    PoolId public constant GLOBAL_POT = PoolId.wrap(0);
    uint256 public constant RECOVERY_CHALLENGE_PERIOD = 7 days;
```

hieronx (approved):

src/vaults/factories/PoolEscrowFactory.sol
lemunozm: They are already added to the wards IIUC your note

lemunozm:
I'm unable to decrypt the failing test case in [the job](https://github.com/centrifuge/protocol-v3/actions/runs/14591701885/job/40928222306?pr=266). It happens in `BaseInvestmentManager.escrow()`, but it should work fine 🤔.

Seems like some special (0x0.....03) address makes fail the staticcall without returning the success value

PR #267, opened 2025-04-16
feat: Deploy using create3

gpmayorga:


wischli:
@gpmayorga Can we close this in favor of #419?

gpmayorga:
> @gpmayorga Can we close this in favor of #419?

Nope. This will be for our production deployments. #419 will have to be merged first as I loaded a lot of my ideas in there already but did not tackle deterministic deployments yet. And deploying with catapulta will also be tried as part of this PR

PR #268, opened 2025-04-16
Gateway UTs and fixes

lemunozm:
WIP

### Changes in Gateway

- Architectural "fix": move `MessageProof` from `MessageLib` and `MessageProcessor` to `Gateway`. This message is special. It is different from any other message and is fully handled by the Gateway. No `MessageDispatcher` or `MessageProcessor` are involved. It has no relation to a pool, nor is it required to have a length for processing it. Moving to the Gateway simplify some stuff and testing.
- Minor renames to adjust better to what each message really is.

### Fixes in Gateway

- Fix handling batches. Batches were not correctly handled. It was working due all messages used had the same length.
- Fix batchId used in `SendProof` event: https://kflabs.slack.com/archives/C07PG2EUR9C/p1744874375552539


hieronx:
I merged `main` into this to have a clean diff in https://github.com/centrifuge/protocol-v3/pull/272.

lemunozm:
For reviewers. Assume no diff with the previous Gateway tests. Just like a new file

src/common/Gateway.sol
```diff
@@ -16,21 +19,36 @@ import {IMessageSender} from "src/common/interfaces/IMessageSender.sol";
 import {IGateway} from "src/common/interfaces/IGateway.sol";
 import {PoolId} from "src/common/types/PoolId.sol";
 import {IGatewayHandler} from "src/common/interfaces/IGatewayHandlers.sol";
+import {MessageLib, MessageType} from "src/common/libraries/MessageLib.sol";
+
+uint8 constant MESSAGE_PROOF_ID = 1;
+
+function deserializeMessageProof(bytes memory data) pure returns (bytes32) {
```
hieronx: Actually this seems unorganized now that I see it.. Maybe we stick to the library

src/common/Gateway.sol
```diff
@@ -39,29 +57,23 @@ contract Gateway is Auth, IGateway, Recoverable {
...
-    address public transient transactionPayer;
+    address public transient transactionRefund;
     mapping(PoolId => Funds) public subsidy;
 
     // Adapters
     mapping(uint16 centrifugeId => IAdapter[]) public adapters;
     mapping(uint16 centrifugeId => mapping(IAdapter adapter => Adapter)) internal _activeAdapters;
 
-    // Messages
+    // Imbound & recoveries
```
```suggestion
    // Inbound & recoveries
```

src/common/Gateway.sol
```diff
@@ -39,29 +57,23 @@ contract Gateway is Auth, IGateway, Recoverable {
...
     mapping(uint16 centrifugeId => mapping(bytes32 messageHash => uint256)) public failedMessages;
     mapping(uint16 centrifugeId => mapping(bytes32 batchHash => InboundBatch)) public inboundBatch;
-    mapping(uint16 centrifugeId => mapping(IAdapter adapter => mapping(bytes32 batchHash => uint256 timestamp))) public
-        recoveries;
+    mapping(uint16 centrifugeId => mapping(IAdapter adapter => mapping(bytes32 payloadHash => uint256 timestamp)))
```
hieronx, line 64: Is `payloadHash == payloadId`?

hieronx (dismissed):
I didn't review the tests but the src changes LGTM!

src/common/Gateway.sol
lemunozm: No, if payload can be a batch or a proof, then payloadHash is hash(batch) or hash(proof)

src/common/Gateway.sol
lemunozm: > Actually this seems unorganized

In what sense do you say it?

hieronx: I think reading the code, it is harder to get lost and not understand the organization of the logic now.

But totally fine to revisit this later.

hieronx (dismissed):

src/common/Gateway.sol
lemunozm: You were right, I'm more happy with: [de87d6f](https://github.com/centrifuge/protocol-v3/pull/268/commits/de87d6fcb54c20eaedbe560d7d57fac3d10fdb90)

hieronx (approved):

Issue #165, opened 2025-03-19
Transient support for batching


lemunozm:
Make the bach storages transient unless the solidity version adds support for them before we need to deploy it (I do not think so). 

It means:
  - Code in assembly how to store a transient map of `bytes`: https://github.com/centrifuge/protocol-v3/blob/main/src/common/Gateway.sol#L60
  - Code in assembly a transient array: https://github.com/centrifuge/protocol-v3/blob/main/src/common/Gateway.sol#L63

PR #272, opened 2025-04-19
Transient support for batching

hieronx:
Fixes https://github.com/centrifuge/protocol-v3/issues/165

### DIFF
--- a/src/common/Gateway.sol
+++ b/src/common/Gateway.sol
@@ -67,7 +67,6 @@ contract Gateway is Auth, IGateway, Recoverable {
     // Outbound processing (batching)
     bool public transient isBatching;
     mapping(uint16 centrifugeId => mapping(PoolId => bytes)) public /*transient*/ outboundBatch;
-    mapping(uint16 centrifugeId => mapping(PoolId => uint128)) public /*transient*/ batchGasLimit;
 
     // Payment
     uint256 public transient fuel;
@@ -316,7 +315,7 @@ contract Gateway is Auth, IGateway, Recoverable {
         require(adapters[centrifugeId].length != 0, EmptyAdapterSet());
 
         uint128 batchGasLimit_ =
-            (isBatching) ? batchGasLimit[centrifugeId][poolId] : gasService.gasLimit(centrifugeId, batch);
+            (isBatching) ? keccak256(abi.encode(centrifugeId, poolId)).tloadUint128() : gasService.gasLimit(centrifugeId, batch);
 
         for (uint256 i; i < adapters_.length; i++) {
             uint256 consumed =
@@ -409,7 +408,9 @@ contract Gateway is Auth, IGateway, Recoverable {
             
             _send(centrifugeId, poolId, outboundBatch[centrifugeId][poolId]);
             delete outboundBatch[centrifugeId][poolId];
-            delete batchGasLimit[centrifugeId][poolId];
+
+            bytes32 slot = keccak256(abi.encode(centrifugeId, poolId));
+            slot.tstore(uint256(0));
         }
 
         TransientArrayLib.clear(BATCH_LOCATORS_SLOT);
--- a/src/common/Gateway.sol
+++ b/src/common/Gateway.sol
@@ -390,8 +390,7 @@ contract Gateway is Auth, Recoverable, IGateway {
 
         bytes32[] memory locators = TransientArrayLib.getBytes32(BATCH_LOCATORS_SLOT);
         for (uint256 i; i < locators.length; i++) {
-            uint16 centrifugeId = uint16(bytes2(locators[i]));
-            PoolId poolId = PoolId.wrap(uint64(bytes8(locators[i] << 16)));
+            (uint16 centrifugeId, PoolId poolId) = _parseLocator(locators[i]);
             
             bytes32 gasLimitSlot = keccak256(abi.encode("batchGasLimit", centrifugeId, poolId));
             bytes32 outboundBatchSlot = keccak256(abi.encode("outboundBatch", centrifugeId, poolId));
@@ -406,6 +405,11 @@ contract Gateway is Auth, Recoverable, IGateway {
         isBatching = false;
 
         _closeTransaction();
+    }
+
+    function _parseLocator(bytes32 locator) internal pure returns (uint16 centrifugeId, PoolId poolId) {
+        centrifugeId = uint16(bytes2(locator));
+        poolId = PoolId.wrap(uint64(bytes8(locator << 16)));
     }
 
     //----------------------------------------------------------------------------------------------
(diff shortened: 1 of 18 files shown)


src/common/Gateway.sol
```diff
@@ -449,3 +444,20 @@ contract Gateway is Auth, IGateway, Recoverable {
         return inboundBatch[centrifugeId][batchHash].votes;
     }
 }
+
+library MessageProofLib {
```
hieronx, line 455: Given it's only used within the same file, I would actually move this to internal methods in the contract, rather than a library.

src/common/interfaces/IGateway.sol
```diff
@@ -51,10 +42,25 @@ interface IGateway is IMessageHandler, IMessageSender, IGatewayHandler {
...
-    event SendProof(uint16 indexed centrifugeId, bytes32 proofId, bytes32 batchHash, IAdapter adapter);
-    event ProcessBatch(uint16 indexed centrifugeId, bytes32 batchId, bytes batch, IAdapter adapter);
-    event ProcessProof(uint16 indexed centrifugeId, bytes32 proofId, bytes32 batchHash, IAdapter adapter);
+    event SendBatch(
+        uint16 indexed centrifugeId,
+        bytes32 payloadId,
+        bytes batch,
+        IAdapter adapter,
+        bytes32 adapterData,
+        address transactionPayer,
```
hieronx, line 51: `adapterData` can be used to link a Centrifuge payload ID to a message ID On the adapter side.

`underpaid` can be used to know which transactions still need a manual payment to the adapter's relayer.

src/common/Gateway.sol
lemunozm: Will do

src/misc/libraries/TransientArrayLib.sol
```diff
@@ -0,0 +1,39 @@
+// SPDX-License-Identifier: AGPL-3.0-only
```
lemunozm: Is this license intended?

src/misc/libraries/TransientBytesLib.sol
```diff
@@ -0,0 +1,42 @@
+// SPDX-License-Identifier: AGPL-3.0-only
```
lemunozm: Same here

src/common/Gateway.sol
```diff
@@ -397,14 +388,21 @@ contract Gateway is Auth, IGateway, Recoverable {
...
 
-        for (uint256 i; i < batchLocators.length; i++) {
-            BatchLocator memory locator = batchLocators[i];
-            _send(locator.centrifugeId, locator.poolId, outboundBatch[locator.centrifugeId][locator.poolId]);
-            delete outboundBatch[locator.centrifugeId][locator.poolId];
-            delete batchGasLimit[locator.centrifugeId][locator.poolId];
+        bytes32[] memory locators = TransientArrayLib.getBytes32(BATCH_LOCATORS_SLOT);
+        for (uint256 i; i < locators.length; i++) {
+            uint16 centrifugeId = uint16(bytes2(locators[i]));
+            PoolId poolId = PoolId.wrap(uint64(bytes8(locators[i] << 16)));
```
lemunozm: I would move this and its counterpart method as internal methods that we can easily test, ensuring always behave identically. If you agree, I can do in the gateway-uts once we merge this.

src/misc/libraries/TransientBytesLib.sol
```diff
@@ -0,0 +1,42 @@
...
+    function set(bytes32 key, bytes memory value) internal {
+        uint256 chunks = value.length / 32 + 1;
+
+        bytes32 lengthSlot = keccak256(abi.encodePacked(key, type(uint256).max));
+        lengthSlot.tstore(value.length);
+
+        for (uint256 i = 0; i < chunks; i++) {
+            bytes32 slot = keccak256(abi.encodePacked(key, i));
+            slot.tstore(bytes32(value.sliceZeroPadded(i * 32, 32)));
+        }
```
lemunozm, line 21: Love the idea of reusing `TransientStorageLib` as building block! 🚀

src/common/Gateway.sol
```diff
@@ -286,16 +271,19 @@ contract Gateway is Auth, IGateway, Recoverable {
...
+            gasLimitSlot.tstore(uint256(newGasLimit));
 
             if (previousMessage.length == 0) {
-                batchLocators.push(BatchLocator(centrifugeId, poolId));
-                outboundBatch[centrifugeId][poolId] = message;
+                TransientArrayLib.push(BATCH_LOCATORS_SLOT, bytes32(abi.encodePacked(bytes2(centrifugeId), bytes8(poolId.raw()))));
+                TransientBytesLib.set(batchSlot, message);
             } else {
-                outboundBatch[centrifugeId][poolId] = bytes.concat(previousMessage, message);
+                TransientBytesLib.set(batchSlot, bytes.concat(previousMessage, message));
```
lemunozm, line 286: I think it should not be complicated to do a `TransientBytesLib.append()` to avoid concatenating the message here, and just insert the new message at the end of the bytes.

Will try to do it if I have time before Wednesday!

lemunozm (approved):
Stellar work! 🚀 

Just a couple of NITs that I can take in the `gateway-uts` branch

src/misc/libraries/TransientArrayLib.sol
hieronx: Nope! Will update

src/common/Gateway.sol
hieronx: Good idea! Great if you can do it

src/common/Gateway.sol
hieronx: Quickly looked at it, it is a little complex since you need to either add just new chunks, or edit the last chunk if `length % 32 != 0`.

lemunozm: Yeah, I think you should slice() the first chunk and the last stored chunk, join both, and then continue as usual

lemunozm (approved):
LGTM!

src/common/Gateway.sol
```diff
@@ -340,14 +326,19 @@ contract Gateway is Auth, IGateway, Recoverable {
                     centrifugeId,
                     keccak256(abi.encodePacked(localCentrifugeId, centrifugeId, batchHash)),
                     batch,
-                    adapters_[i]
+                    adapters_[i],
+                    adapterData,
+                    transactionPayer,
```
lemunozm, line 331: @hieronx Postmorten comment. Should not be:

```solidity
transactionRefund != address(0) ? transactionRefund : subsidy[poolId].refund
```
instead?

hieronx: True!

PR #273, opened 2025-04-21
Add PoolId, ShareClassId and AssetId types to common module

lemunozm:
This PR is the first chunk of this refactoring.

It only does the required changes in `vaults` to get rid of non-typed values in the `common` module.

It increases a bit the test lines, but those lines should be removed again once everything uses custom types.

### DIFF
--- a/src/vaults/AsyncRequests.sol
+++ b/src/vaults/AsyncRequests.sol
@@ -20,6 +20,7 @@ import {IVaultMessageSender} from "src/common/interfaces/IGatewaySenders.sol";
 import {IInvestmentManagerGatewayHandler} from "src/common/interfaces/IGatewayHandlers.sol";
 import {PoolId} from "src/common/types/PoolId.sol";
 import {ShareClassId} from "src/common/types/ShareClassId.sol";
+import {AssetId} from "src/common/types/AssetId.sol";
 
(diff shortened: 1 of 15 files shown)


src/vaults/AsyncRequests.sol
```diff
@@ -168,7 +171,11 @@ contract AsyncRequests is BaseInvestmentManager, IAsyncRequests {
         VaultDetails memory vaultDetails = poolManager.vaultDetails(address(vault_));
 
         sender.sendRedeemRequest(
-            mapPoolId(vault_.poolId()), vault_.trancheId(), controller.toBytes32(), vaultDetails.assetId, shares
+            PoolId.wrap(mapPoolId(vault_.poolId())),
+            ShareClassId.wrap(vault_.trancheId()),
+            controller.toBytes32(),
+            vaultDetails.assetId,
+            shares
```
lemunozm, line 178: These increased lines should be removed again once vault returns a `PoolId` (future PR)

lemunozm: NOTE: The Abi of a custom type is the same.

> The data-representation of the values of such types are inherited from the underlying type and the underlying type is also used in the ABI

Ref: https://soliditylang.org/blog/2021/09/27/user-defined-value-types/

mustermeiszer (approved):

PR #275, opened 2025-04-21
Simplify escrow

hieronx:


src/vaults/Escrow.sol
```diff
@@ -75,88 +76,54 @@ contract PoolEscrow is Escrow, Recoverable, IPoolEscrow {
...
     }
 
     /// @inheritdoc IPoolEscrow
     function reserveIncrease(bytes16 scId, address asset, uint256 tokenId, uint256 value) external auth {
-        uint256 newValue = reservedAmount[scId][asset][tokenId] + value;
-        reservedAmount[scId][asset][tokenId] = newValue;
+        uint128 newValue = holding[scId][asset][tokenId].reserved + value.toUint128();
+        holding[scId][asset][tokenId].reserved = newValue;
 
+        // NOTE: think we should either include the delta here, or emit 2 separate events IncreaseReserve and
```
wischli: In favor of including delta.

src/vaults/Escrow.sol
```diff
@@ -75,88 +76,54 @@ contract PoolEscrow is Escrow, Recoverable, IPoolEscrow {
...
     function availableBalanceOf(bytes16 scId, address asset, uint256 tokenId) public view returns (uint256) {
-        uint256 holding_ = holding[scId][asset][tokenId];
-        uint256 reservedAmount_ = reservedAmount[scId][asset][tokenId];
-
-        if (holding_ < reservedAmount_) {
-            return 0;
-        } else {
-            return holding_ - reservedAmount_;
-        }
+        Holding storage holding_ = holding[scId][asset][tokenId];
```
wischli, line 123: Nit question: Should be cheaper to init as `memory` as we don't perform a write, right?

```suggestion
        Holding memory holding_ = holding[scId][asset][tokenId];
```

wischli (approved):
LGTM!

src/vaults/Escrow.sol
hieronx: I believe copying to memory from storage is actually more expensive in this case, but not 100% sure. Let's snapshot this later?

wischli (approved):

src/vaults/Escrow.sol
wischli: Ah yes you might be right.

PR #277, opened 2025-04-21
feat: Legacy vault mapping

mustermeiszer:
The idea is simple

* Adapter is a Centrifuge V2 `IInvestmentManager` for the legacy vaults
* Adapter is a Centrifuge V3 `IAsyncVault` for the system

image

As seen in graphic, the `LegacyVaultAdapter` implements the old `IInvestmentManager` interface and forwards all relevant calls to the new system, exchanging the vault address with `address(this)`. Furthermore, the adapter inherets the `AsyncVault` and is simply a normal vault for the new system. 

This approach has the following advantages
* System does not need to know about the adapter 
  &rarr; Exception: we can NOT deploy the vault normally but need to migrate it to the `PoolManager` storage with a spell
  &rarr; Link and unlink work as expected
* Existing users of the old vaults can seamlessly migrate if needed
* Existing users of the old vaults can still use the full `IInvestmentManager` interface if integrations rely on that
* UI does not need to have exception cases for any vaults


mustermeiszer:
@hieronx the only open question for me would be whether we want to overrie the the `callback` functions of the vaults and call both into the legacy AND the new vault?

test/vaults/integration/SyncDeposit.t.sol
```diff
@@ -35,18 +35,18 @@ contract SyncDepositTestHelper is BaseTest {
...
-            syncVault.poolId(), syncVault.trancheId(), assetId, pricePoolPerAsset.inner(), uint64(block.timestamp)
+            syncVault.poolId(), syncVault.scId(), assetId, pricePoolPerAsset.inner(), uint64(block.timestamp)
         );
     }
 
     function _assertDepositEvents(SyncDepositVault vault, uint128 shares, D18 pricePoolPerShare, D18 priceAssetPerShare)
         internal
     {
         PoolId poolId = PoolId.wrap(vault.poolId());
-        ShareClassId scId = ShareClassId.wrap(vault.trancheId());
```
lemunozm, line 49: Thanks for the renaming

src/vaults/legacy/LegacyVaultAdapter.sol
```diff
@@ -0,0 +1,205 @@
...
+        shares = asyncManager().deposit(address(this), assets, receiver, owner);
+    }
+
+    /// @inheritdoc IInvestmentManager
+    function mint(address, /* vault */ uint256 shares, address receiver, address owner)
+        public
+        legacy
+        returns (uint256 assets)
+    {
+        assets = asyncManager().mint(address(this), shares, receiver, owner);
```
lemunozm, line 169: Should not be the vault?

```suggestion
        assets = asyncManager().mint(vault, shares, receiver, owner);
```

lemunozm (commented):
LGTM! Just a CR, IIUC

src/vaults/legacy/LegacyVaultAdapter.sol
lemunozm: Ahh, unless you want to forward to the async manager the adapter address to be able to later handle it. Then it's not a CR

mustermeiszer:
```
    /// @inheritdoc IDepositManager
    function mint(address vaultAddr, uint256 shares, address receiver, address controller)
        public
        auth
        returns (uint256 assets)
    {
        AsyncInvestmentState storage state = investments[vaultAddr][controller];
        uint128 shares_ = shares.toUint128();
        _processDeposit(state, shares_, shares_, vaultAddr, receiver);

        assets = uint256(_calculateAssets(vaultAddr, shares_, state.depositPrice, MathLib.Rounding.Down));
    }
```

Luckily, there is never a token transfer between manager and vault. Always, between manager and escrow. Hence, we can shadow the legacy vault address completely from the new system through the adapter.

lemunozm (dismissed):

lemunozm (dismissed):

lemunozm (approved):

PR #278, opened 2025-04-22
Feat: add transient bytes append

lemunozm:
It avoids setting the whole batch each time a new message is added

### DIFF
--- a/src/common/Gateway.sol
+++ b/src/common/Gateway.sol
@@ -280,7 +280,7 @@ contract Gateway is Auth, Recoverable, IGateway {
                 TransientArrayLib.push(BATCH_LOCATORS_SLOT, _encodeLocator(centrifugeId, poolId));
                 TransientBytesLib.set(batchSlot, message);
             } else {
-                TransientBytesLib.set(batchSlot, bytes.concat(previousMessage, message));
+                TransientBytesLib.append(batchSlot, message);
             }
         } else {
             _send(centrifugeId, poolId, message);
--- a/src/common/Gateway.sol
+++ b/src/common/Gateway.sol
@@ -278,10 +278,9 @@ contract Gateway is Auth, Recoverable, IGateway {
 
             if (previousMessage.length == 0) {
                 TransientArrayLib.push(BATCH_LOCATORS_SLOT, _encodeLocator(centrifugeId, poolId));
-                TransientBytesLib.set(batchSlot, message);
-            } else {
-                TransientBytesLib.append(batchSlot, message);
             }
+
+            TransientBytesLib.append(batchSlot, message);
         } else {
             _send(centrifugeId, poolId, message);
             _refundTransaction();
--- a/src/misc/libraries/TransientBytesLib.sol
+++ b/src/misc/libraries/TransientBytesLib.sol
@@ -9,18 +9,6 @@ library TransientBytesLib {
     using TransientStorageLib for bytes32;
     using BytesLib for bytes;
 
-    function set(bytes32 key, bytes memory value) internal {
-        uint256 chunks = value.length / 32 + 1;
-
-        bytes32 lengthSlot = keccak256(abi.encodePacked(key, type(uint256).max));
-        lengthSlot.tstore(value.length);
-
-        for (uint256 i = 0; i < chunks; i++) {
-            bytes32 slot = keccak256(abi.encodePacked(key, i));
-            slot.tstore(bytes32(value.sliceZeroPadded(i * 32, 32)));
-        }
-    }
-
     function append(bytes32 key, bytes memory value) internal {
         bytes32 lengthSlot = keccak256(abi.encodePacked(key, type(uint256).max));
         uint256 prevLength = lengthSlot.tloadUint256();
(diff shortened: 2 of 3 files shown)


src/misc/libraries/TransientBytesLib.sol
```diff
@@ -21,6 +21,27 @@ library TransientBytesLib {
...
+
+        bytes32 joinSlot = keccak256(abi.encodePacked(key, startChunk));
+        bytes memory firstPart = abi.encodePacked(joinSlot.tloadBytes32()).sliceZeroPadded(0, offset);
+        bytes memory secondPart = value.sliceZeroPadded(0, 32 - offset);
+        joinSlot.tstore(bytes32(bytes.concat(firstPart, secondPart)));
+
+        for (uint256 i = startChunk + 1; i < chunks; i++) {
+            bytes32 slot = keccak256(abi.encodePacked(key, i));
+            slot.tstore(bytes32(value.sliceZeroPadded((32 - offset) + i * 32, 32)));
+        }
```
wischli: I think this includes a subtle bug: The first 32 bytes of `value` are partly written in the merged chunk. So when you hit the loop, the next data starts at `value[(32 - offset):]`. But `i` starts from `startChunk + 1`, such that `i * 32` already skips part of the value.

Example
* prevLength = 50
* value.length = 80
* offset = 50 % 32 = 18
* 32 - offset = 14
* totalLength = 50 + 80 = 130
* chunks = 130 / 32 + 1 = 5
* startChunk = 50 / 32 = 1
  * loop runs from `i = 2` to `i = 4 < chunks`
* However, for `i = 2` we write to `(32 - offset) + i * 32 = 14 + 2 * 32 = 78` when it should be 46 right?



```suggestion
        uint256 valueOffset = 32 - offset;
        uint256 chunkIndex = startChunk + 1;
        for (; valueOffset < value.length; chunkIndex++) {
            bytes32 slot = keccak256(abi.encodePacked(key, chunkIndex));
            slot.tstore(bytes32(value.sliceZeroPadded(valueOffset, 32)));
            valueOffset += 32;
        }
```

wischli (commented):
IIUC, I think the implementation has a minor bug.

Also curious if we plan on gas optimizing this lib utilizing assembly @hieronx.

lemunozm:
> Also curious if we plan on gas optimizing this lib utilizing assembly @hieronx.

I think the idea is just to wait for https://github.com/ethereum/solidity/issues/15007

wischli:
Should come with a unit test and most ideally a snapshot one

src/misc/libraries/TransientBytesLib.sol
lemunozm: You're completely right! And the unitary tests confirm that. Thanks to put your mathematical brain into this 🙌🏻

lemunozm:
Will leave the snapshot for post-audit. But I think the improvement is significant.

A batch process rewrites all the time its first part several times. i.e: to write a batch with messages A,B,C,D,E, the slots writen would be:

- A
- A,B
- A,B,C
- A,B,C,D
- A,B,C,D,E

Apart from concatenating everything each time from the Gateway

src/common/Gateway.sol
```diff
@@ -280,7 +280,7 @@ contract Gateway is Auth, Recoverable, IGateway {
                 TransientArrayLib.push(BATCH_LOCATORS_SLOT, _encodeLocator(centrifugeId, poolId));
                 TransientBytesLib.set(batchSlot, message);
```
hieronx: Should we remove `.set()` and just use `append()` every time?

test/misc/unit/libraries/TransientBytesLib.t.sol
```diff
@@ -17,4 +17,14 @@ contract TransientBytesLibTest is Test {
...
         assertEq(TransientBytesLib.get(invalidKey).length, 0);
     }
+
+    function testAppend(bytes calldata data1, bytes calldata data2) public {
+        bytes32 key = keccak256(abi.encode("key"));
+
+        TransientBytesLib.set(key, data1);
+        assertEq(TransientBytesLib.get(key), data1);
+
+        TransientBytesLib.append(key, data2);
```
hieronx, line 15: Can we add a second `append` in here with `data3`, to check appending multiple times works well?

src/common/Gateway.sol
lemunozm: I think as a lib, having `set()` could be beneficial (?). Not sure if you ask for saving audit lines

hieronx: Yeah I mainly ask for purpose of simpler audit exactly

lemunozm: yes, done!

wischli (approved):

test/misc/unit/libraries/TransientBytesLib.t.sol
hieronx: @lemunozm could you still add this in a followup PR?

lemunozm: I added it. The current test uses `append `after `append`

hieronx: My bad I misread!

PR #279, opened 2025-04-22
Gateway: renamings to more accurate names

lemunozm:
- Rename events `ProcessBatch/Proof` to `HandleBatch/Proof` to avoid confusing it with the fact that those messages are processed by the `MessageProcessor`. They are just handled by the Gateway
- Remove the name "message" from recovery actions to avoid thinking that it recovers messages. It recovers batches or proofs.


hieronx (approved):

wischli (approved):
Great simplification!

lemunozm:
This is just a renaming, so it could be out of the incoming Gateway audit

PR #290, opened 2025-04-22
Hub: Remove valuation

### LINKED ISSUE #250: TransientValuation handling issue
Currently, this test is working: https://github.com/centrifuge/protocol-v3/blob/main/test/hub/integration/Cases.t.sol#L234 when it shouldn't.

It's missing `/// forge-config: default.isolate = true`, so in real scenarions the transientValuation will be reset and the price missing. We should instead either:

- Add `TransientValuation` to the Hub.
- Use `transientValuation.setPrice()` in a generic multicall
- Remove TransientValuation and use prices in Hub side.

lemunozm:
Fixes https://github.com/centrifuge/protocol-v3/issues/250

This is an daring solution, happy to discuss or even not merge it. But I think it deserves to evaluate it.

The point is that every time we use the `holding.valuation`, we use it from an admin role which has the same level of privileges for setting the valuation. It means: the own actor can retriver the price from the oracle and pass it to the method in the same moment.

We only need to automate this through an IERC7726 if the valuation should be set by an admin to be used later by a lower-privileged user. **This is not the case for any of our methods in the Hub.**

That means that we can simplify it by:
- Removing transient valuation at all.
- Removing the `holding.valuation`
- Using D18 prices as parameters

In the same line, we can also remove the `valuation` parameter in `approveDeposits` and `revokeShares` by a price

NOTE: The PR is ready except for adapting the UTs from `Holding.t.sol`

### DIFF
--- a/src/hub/interfaces/IHoldings.sol
+++ b/src/hub/interfaces/IHoldings.sol
@@ -1,7 +1,7 @@
 // SPDX-License-Identifier: BUSL-1.1
 pragma solidity >=0.5.0;
 
-import {IERC7726} from "src/misc/interfaces/IERC7726.sol";
+import {D18} from "src/misc/types/D18.sol";
 
 import {PoolId} from "src/common/types/PoolId.sol";
 import {AssetId} from "src/common/types/AssetId.sol";
@@ -11,8 +11,8 @@ import {ShareClassId} from "src/common/types/ShareClassId.sol";
 struct Holding {
     uint128 assetAmount;
     uint128 assetAmountValue;
-    IERC7726 valuation; // Used for existence
     bool isLiability;
+    bool existence;
 }
 
 struct HoldingAccount {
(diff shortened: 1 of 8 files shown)


src/hub/interfaces/IHoldings.sol
```diff
@@ -26,41 +26,26 @@ interface IHoldings {
...
 
     /// @notice Emitted when a holding is increased
     event Increase(
-        PoolId indexed,
-        ShareClassId indexed scId,
-        AssetId indexed assetId,
-        IERC7726 valuation,
-        uint128 amount,
-        uint128 increasedValue
+        PoolId indexed, ShareClassId indexed scId, AssetId indexed assetId, uint128 amount, uint128 increasedValue
```
mustermeiszer, line 34: CR 

Should contain the price used for changing

src/hub/interfaces/IHoldings.sol
```diff
@@ -26,41 +26,26 @@ interface IHoldings {
...
 
     /// @notice Emitted when a holding is decreased
     event Decrease(
-        PoolId indexed,
-        ShareClassId indexed scId,
-        AssetId indexed assetId,
-        IERC7726 valuation,
-        uint128 amount,
-        uint128 decreasedValue
+        PoolId indexed, ShareClassId indexed scId, AssetId indexed assetId, uint128 amount, uint128 decreasedValue
```
mustermeiszer, line 39: CR 

Should contain the price used for changing

src/hub/interfaces/IHoldings.sol
```diff
@@ -26,41 +26,26 @@ interface IHoldings {
...
 
     /// @notice Emitted when the holding is updated
     event Update(
         PoolId indexed, ShareClassId indexed scId, AssetId indexed assetId, bool isPositive, uint128 diffValue
     );
 
-    /// @notice Emitted when a holding valuation is updated
-    event UpdateValuation(PoolId indexed, ShareClassId indexed scId, AssetId indexed assetId, IERC7726 valuation);
+    /// @notice Emitted when a holding is updated
+    event UpdateValuation(PoolId indexed, ShareClassId indexed scId, AssetId indexed assetId);
```
mustermeiszer, line 48: QUESTION 

Do we even need that now? `Update` would be the one, right?

mustermeiszer (commented):
@hieronx @lemunozm totally aggree that this seems to be a good change. The only con is that higher level integrations need to have a price oracle if they want to automate that. But that seems fine. Given, that we anyways need custom `Valuaitons` for our assets as those are not "FROM" the chain in most cases and have custom addresses.

PR #291, opened 2025-04-22
Hub improvements, permissionless hook

hieronx:


src/common/Guardian.sol
```diff
@@ -50,8 +50,8 @@ contract Guardian is IGuardian {
 
     // --- Admin actions ---
     /// @inheritdoc IGuardian
-    function createPool(address admin, AssetId currency) external onlySafe returns (PoolId poolId) {
-        return hub.createPool(admin, currency);
+    function createPool(uint48 poolId_, address admin, AssetId currency) external onlySafe returns (PoolId poolId) {
```
lemunozm, line 53: Curious, why finally going with giving the poolId?

lemunozm (approved):
LGTM!!

wischli (approved):

src/common/Guardian.sol
hieronx: @mustermeiszer 's proposal to make pool creation more deterministic and easier to batch actions

lemunozm: Don't we found and issue with that @mustermeiszer? Some attacker could front run your poolId and prevent you to create any pool?

mustermeiszer: Yes, but that can also happen with the onchain one. Need to resubmit in any case. This makes it easier for the UI though

src/hub/HubRegistry.sol
```diff
@@ -34,16 +32,16 @@ contract HubRegistry is Auth, IHubRegistry {
...
         external
         auth
         returns (PoolId poolId)
     {
         require(manager_ != address(0), EmptyAccount());
         require(!currency_.isNull(), EmptyCurrency());
-        require(currency[poolId].isNull(), PoolAlreadyRegistered());
 
-        poolId = newPoolId(centrifugeId, ++latestId);
+        poolId = newPoolId(centrifugeId, poolId_);
```
mustermeiszer, line 43: CR 

We should check whether the `PoolId` is well formed instead of adapting it here with the `CentrifugeId`. The UI should just create it and pass it on here IMO.

src/common/Guardian.sol
lemunozm: But if we use `+1`, nobody can frontrun the id.

lemunozm: The UI can:
- createPool
- read the Event and show the successful creation of the pool with the new PoolId
- batch any number of actions

PR #295, opened 2025-04-23
Organize some vault interface files and imports

### LINKED ISSUE #287: Move IERC* interfaces to misc
Our IERC* interfaces should be public and centrifuge-agnostic. So they should be placed in `misc` folder.

lemunozm:
Fixes https://github.com/centrifuge/protocol-v3/issues/287

- Move IERC* interfaces to `misc`: https://github.com/centrifuge/protocol-v3/issues/287
- Move out vault's dependent interfaces to `vaults`. => `IBaseVaults.sol` file. 
- Organize imports


src/vaults/legacy/LegacyVaultAdapter.sol
```diff
@@ -4,7 +4,7 @@ pragma solidity 0.8.28;
 import {PoolId} from "src/common/types/PoolId.sol";
 import {ShareClassId} from "src/common/types/ShareClassId.sol";
 
-import {IERC7540} from "src/vaults/legacy/interfaces/IERC7540.sol";
+import {IVault} from "src/vaults/legacy/interfaces/IVault.sol";
```
wischli: Nit: Not sure if we want to keep IERC7540 here. If not, maybe `ILegacyVault` might be more appropriate?

wischli (dismissed):

src/vaults/legacy/LegacyVaultAdapter.sol
lemunozm: Agree with ILegacyVault name. I removed the IERC7540, because the interface adds poolId and scId, which are not IERC7540 compatible.

lemunozm: Done!

wischli (approved):
Thanks for applying my suggestion!

PR #297, opened 2025-04-23
Ref: Factory addresss typed

### LINKED ISSUE #284: Use Interfaces instead of plain address
There are several places where we can use interfaces instead of addresses, which improves the readability and avoid castings:

- ~ShareToken~
- All factories
- Some Escrows
- BaseVault

lemunozm:
Use the Interface where it's known the address is always the interface. 

Fixes https://github.com/centrifuge/protocol-v3/issues/284

### DIFF
--- a/src/vaults/factories/AsyncVaultFactory.sol
+++ b/src/vaults/factories/AsyncVaultFactory.sol
@@ -10,19 +10,23 @@ import {AsyncVault} from "src/vaults/AsyncVault.sol";
 import {IVaultFactory} from "src/vaults/interfaces/factories/IVaultFactory.sol";
 import {IPoolEscrowProvider} from "src/vaults/interfaces/factories/IPoolEscrowFactory.sol";
 import {IShareToken} from "src/vaults/interfaces/token/IShareToken.sol";
+import {IAsyncRedeemManager} from "src/vaults/interfaces/investments/IAsyncRedeemManager.sol";
 
 /// @title  ERC7540 Vault Factory
 /// @dev    Utility for deploying new vault contracts
 contract AsyncVaultFactory is Auth, IVaultFactory {
     address public immutable root;
-    address public immutable investmentManager;
+    IAsyncRedeemManager public immutable asyncRedeemManager;
     IPoolEscrowProvider public immutable poolEscrowProvider;
 
-    constructor(address root_, address investmentManager_, IPoolEscrowProvider poolEscrowProvider_, address deployer)
-        Auth(deployer)
-    {
+    constructor(
+        address root_,
+        IAsyncRedeemManager asyncRedeemManager_,
+        IPoolEscrowProvider poolEscrowProvider_,
+        address deployer
+    ) Auth(deployer) {
         root = root_;
-        investmentManager = investmentManager_;
+        asyncRedeemManager = asyncRedeemManager_;
         poolEscrowProvider = poolEscrowProvider_;
     }
 
@@ -33,14 +37,13 @@ contract AsyncVaultFactory is Auth, IVaultFactory {
         address asset,
         uint256 tokenId,
         IShareToken token,
-        address, /* escrow */
         address[] calldata wards_
     ) public auth returns (address) {
         AsyncVault vault =
-            new AsyncVault(poolId, scId, asset, tokenId, token, root, investmentManager, poolEscrowProvider);
+            new AsyncVault(poolId, scId, asset, tokenId, token, root, asyncRedeemManager, poolEscrowProvider);
 
         vault.rely(root);
-        vault.rely(investmentManager);
+        vault.rely(address(asyncRedeemManager));
         uint256 wardsCount = wards_.length;
         for (uint256 i; i < wardsCount; i++) {
             vault.rely(wards_[i]);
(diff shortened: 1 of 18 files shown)


src/vaults/factories/SyncDepositVaultFactory.sol
```diff
@@ -39,15 +41,15 @@ contract SyncDepositVaultFactory is Auth, IVaultFactory {
...
     ) public auth returns (address) {
         SyncDepositVault vault = new SyncDepositVault(
-            poolId, scId, asset, tokenId, token, root, syncRequests, asyncRequests, poolEscrowProvider
+            poolId, scId, asset, tokenId, token, root, syncDepositManager, asyncRedeemManager, poolEscrowProvider
         );
 
         vault.rely(root);
-        vault.rely(syncRequests);
+        vault.rely(address(syncDepositManager));
+        vault.rely(address(asyncRedeemManager));
```
lemunozm, line 52: I'm pretty sure we need to add here `asyncRedeemManager` due it ends up using auth methods from the vault

mustermeiszer: @lemunozm those reky should be Done in addVault. Can you Check? Or in approve Manager?

mustermeiszer (approved):
I think the Change to the Factory is wrong but appeocing in case is blockig

lemunozm:
Happy to double-check that, why do you think so?

Thanks for the approval!

src/vaults/factories/AsyncVaultFactory.sol
```diff
@@ -10,19 +10,23 @@ import {AsyncVault} from "src/vaults/AsyncVault.sol";
...
 import {IPoolEscrowProvider} from "src/vaults/interfaces/factories/IPoolEscrowFactory.sol";
 import {IShareToken} from "src/vaults/interfaces/token/IShareToken.sol";
+import {IAsyncRedeemManager} from "src/vaults/interfaces/investments/IAsyncRedeemManager.sol";
 
 /// @title  ERC7540 Vault Factory
 /// @dev    Utility for deploying new vault contracts
 contract AsyncVaultFactory is Auth, IVaultFactory {
     address public immutable root;
-    address public immutable investmentManager;
+    IAsyncRedeemManager public immutable asyncRedeemManager;
```
wischli, line 19: I think this should be `IAsyncRequests` instead because this is a fully async vault factory. No need to restrict the underlying investment manager.

src/vaults/factories/SyncDepositVaultFactory.sol
wischli: @mustermeiszer We are doing the other direction in `addVault`: https://github.com/centrifuge/protocol-v3/blob/818362323b53de92b45aebd50023580bcc31c483/src/vaults/SyncRequests.sol#L116

AFAICT, both are necessary. So good catch @lemunozm on the missing rely of the async redeem manager.

src/vaults/legacy/LegacyVaultAdapter.sol
```diff
@@ -36,7 +37,7 @@ contract LegacyVaultAdapter is AsyncVault, ILegacyVaultAdapter, IInvestmentManag
         address asset,
         IShareToken token,
         address root,
-        address manager,
+        IAsyncRedeemManager manager,
```
wischli, line 40: CR: Should be `IInvestmentManager `

wischli (commented):
Post mortem: Two invest manager interfaces are incorrect

src/vaults/factories/AsyncVaultFactory.sol
lemunozm: AFAIK, if it compiles, it's because internally the usage of the vault does not use more capabilities than `IAsyncRedeemManager` offer. It seems like the actions from `IAsyncDepositManager` are only used from the `PoolManager`, not used by the vault itself. (EDIT: checked)

Ideally, with time, would be cool to differentiate that logic

src/vaults/legacy/LegacyVaultAdapter.sol
lemunozm: Same answer as above, the vault does not use more capabilities than `IAsyncRedeemManager` offer

lemunozm: But correct it can be simplified to IBaseInvestmentManager, good catch!

lemunozm: 🤔 After tried I realized I can not use `IBaseInvestmentManager` because the underlying vault is the `AsyncVault` who use `IAsyncRedeemManager` actions

wischli: This is the legacy vault, so it must be `IInvestmentManager`

lemunozm: Ahh, ok, using the old interface
Correct! I read you bad

src/vaults/factories/AsyncVaultFactory.sol
wischli: > It seems like the actions from `IAsyncDepositManager` are only used from the `PoolManager`, not used by the vault itself. (EDIT: checked)

I disagree. The `AsyncVault` uses the `IAsyncDepositManager` via the `asyncManager()` in numerous codepaths: https://github.com/centrifuge/protocol-v3/blob/bd8cf05bc296c518b9811215e8d6880e9eff3ff7/src/vaults/AsyncVault.sol#L56

I can fix this with the below legacy vault investment manager interface and the merge of `IBaseInvestmentManager` and `IVaultManager`.

lemunozm: Correct, there is a cast in `asyncManager()`, I missed that in my review. Sorry!

If we add `IAsyncRequests` as manager, we can then remove the `asyncManager()` method from the `AsyncVault`, at least for the internal usage.

PR #299, opened 2025-04-23
Escrow creation at addPool phase

lemunozm:
Move the escrow creation from addShareClass to addPool


wischli (commented):
In our Centrifuge Chain mock, we are setting the pool escrow as member in our `notifyShareClass` mock: https://github.com/centrifuge/protocol-v3/blob/818362323b53de92b45aebd50023580bcc31c483/test/vaults/mocks/MockCentrifugeChain.sol#L115

It would be great to move that to `addPool` mock as part of this PR ❤️

lemunozm:
It seems like we can't because the member is per sharecLassId

mustermeiszer (approved):

PR #300, opened 2025-04-23
Repay mechanism

hieronx:
Fixes https://github.com/burrasec/2025-04-Centrifuge-Audit/issues/9, https://github.com/burrasec/2025-04-Centrifuge-Audit/issues/10, https://github.com/burrasec/2025-04-Centrifuge-Audit/issues/12, https://github.com/burrasec/2025-04-Centrifuge-Audit/issues/13

### DIFF
--- a/src/common/Gateway.sol
+++ b/src/common/Gateway.sol
@@ -289,48 +289,57 @@ contract Gateway is Auth, Recoverable, IGateway {
         }
     }
 
-    function _send(uint16 centrifugeId, PoolId poolId, bytes memory batch) private {
-        bytes32 batchHash = keccak256(batch);
+    function _send(uint16 centrifugeId, PoolId poolId, bytes memory batch) internal returns (bool succeeded) {
         IAdapter[] memory adapters_ = adapters[centrifugeId];
         require(adapters[centrifugeId].length != 0, EmptyAdapterSet());
 
-        uint128 batchGasLimit_ =
-            (isBatching) ? _gasLimitSlot(centrifugeId, poolId).tloadUint128() : gasService.gasLimit(centrifugeId, batch);
+        SendData memory data = SendData(keccak256(batch), (isBatching) ? _gasLimitSlot(centrifugeId, poolId).tloadUint128() : gasService.gasLimit(centrifugeId, batch), bytes32(""), new uint256[](MAX_ADAPTER_COUNT));
+        data.payloadId = keccak256(abi.encodePacked(localCentrifugeId, centrifugeId, data.batchHash));
 
+        // Estimate gas usage
+        uint256 total;
         for (uint256 i; i < adapters_.length; i++) {
-            uint256 consumed = adapters_[i].estimate(
+            data.gasCost[i] = adapters_[i].estimate(
                 centrifugeId,
-                i == PRIMARY_ADAPTER_ID - 1 ? batch : batchHash.serializeMessageProof(),
-                batchGasLimit_
+                i == PRIMARY_ADAPTER_ID - 1 ? batch : data.batchHash.serializeMessageProof(),
+                data.batchGasLimit
             );
 
-            if (transactionRefund != address(0)) {
-                require(consumed <= fuel, NotEnoughTransactionGas());
-                fuel -= consumed;
-            } else {
-                if (consumed > subsidy[poolId].value) {
-                    _requestPoolFunding(poolId);
-                }
-                if (consumed <= subsidy[poolId].value) {
-                    subsidy[poolId].value -= uint96(consumed);
-                } else {
-                    underpaid[centrifugeId][batchHash]++;
-                    emit UnderpaidBatch(centrifugeId, batch);
-                    return;
-                }
+            total += data.gasCost[i];
+        }
+
+        // Ensure sufficient funds are available
+        if (transactionRefund != address(0)) {
+            require(total <= fuel, NotEnoughTransactionGas());
+            fuel -= total;
+        } else {
+            if (total > subsidy[poolId].value) {
+                _requestPoolFunding(poolId);
             }
 
-            bytes32 adapterData = adapters_[i].send{value: consumed}(
+            if (total <= subsidy[poolId].value) {
+                subsidy[poolId].value -= uint96(total);
+            } else {
+                underpaid[centrifugeId][data.batchHash]++;
+                emit UnderpaidBatch(centrifugeId, batch);
+                return false;
+            }
+        }
+        delete total;
+
+        // Send batch and proofs
+        for (uint256 i; i < adapters_.length; i++) {
+            bytes32 adapterData = adapters_[i].send{value: data.gasCost[i]}(
                 centrifugeId,
-                i == PRIMARY_ADAPTER_ID - 1 ? batch : batchHash.serializeMessageProof(),
-                batchGasLimit_,
+                i == PRIMARY_ADAPTER_ID - 1 ? batch : data.batchHash.serializeMessageProof(),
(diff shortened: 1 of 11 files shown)


src/common/Gateway.sol
```diff
@@ -288,63 +289,88 @@ contract Gateway is Auth, Recoverable, IGateway {
...
+            require(total <= fuel, NotEnoughTransactionGas());
+            fuel -= total;
+        } else {
+            if (total > subsidy[poolId].value) {
+                _requestPoolFunding(poolId);
+            }
+
+            if (total <= subsidy[poolId].value) {
+                subsidy[poolId].value -= uint96(total);
             } else {
```
wischli, line 322: Currently, this branch is unreachable because we handle > and <= above.

src/common/Gateway.sol
```diff
@@ -288,63 +289,88 @@ contract Gateway is Auth, Recoverable, IGateway {
...
-                    subsidy[poolId].value -= uint96(consumed);
-                } else {
-                    consumed = 0;
-                }
+                underpaid[centrifugeId][data.batchHash]++;
+                emit UnderpaidBatch(centrifugeId, batch);
+                return false;
             }
+        }
+        delete total;
```
wischli: Q: Is it Common practice to delete a scoped variable instead of setting it to 0?

src/common/Gateway.sol
hieronx: I don't think so, `if (total > subsidy[poolId].value) {` does not return so it continues to this second else statement as well.

src/common/Gateway.sol
hieronx: Was unnecessary ,removed

src/common/Gateway.sol
wischli: Ah yes sorry, mistook it for returning

test/vaults/integration/Deposit.t.sol
```diff
@@ -158,25 +158,6 @@ contract DepositTest is BaseTest {
         assertTrue(vault.maxDeposit(self) <= amount * 0.01e18);
     }
 
-    function testDepositWithPrepaymentFromGateway(uint256 amount) public {
```
hieronx, line 161: This test is hard to fix right now, and i don't believe it adds much: it really just tests the gateway subsidy logic instead of the deposit flow, which is already tested well within the Gateway unit tests.

src/common/Gateway.sol
```diff
@@ -288,86 +289,113 @@ contract Gateway is Auth, Recoverable, IGateway {
...
 
-    function _send(uint16 centrifugeId, PoolId poolId, bytes memory batch) private {
-        bytes32 batchHash = keccak256(batch);
+    function _send(uint16 centrifugeId, PoolId poolId, bytes memory batch) internal returns (bool succeeded) {
         IAdapter[] memory adapters_ = adapters[centrifugeId];
         require(adapters[centrifugeId].length != 0, EmptyAdapterSet());
 
-        uint128 batchGasLimit_ =
-            (isBatching) ? _gasLimitSlot(centrifugeId, poolId).tloadUint128() : gasService.gasLimit(centrifugeId, batch);
+        SendData memory data = SendData(keccak256(batch), (isBatching) ? _gasLimitSlot(centrifugeId, poolId).tloadUint128() : gasService.gasLimit(centrifugeId, batch), bytes32(""), new uint256[](MAX_ADAPTER_COUNT));
```
wischli, line 296: Nit: Could we use the explicit constructor? It's incredibly hard to read. I have formatted it properly:

```suggestion
        SendData memory data = SendData({
            batchHash: keccak256(batch),
            batchGasLimit: (isBatching)
                ? _gasLimitSlot(centrifugeId, poolId).tloadUint128()
                : gasService.gasLimit(centrifugeId, batch),
            payloadId: bytes32(""),
            gasCost: new uint256[](MAX_ADAPTER_COUNT)
        });
```

src/common/Gateway.sol
```diff
@@ -288,86 +289,113 @@ contract Gateway is Auth, Recoverable, IGateway {
...
+        bytes32 batchHash = keccak256(batch);
+        require(underpaid[centrifugeId][batchHash] > 0, NotUnderpaidBatch());
+
+        PoolId poolId = processor.messagePoolId(batch);
+        if (msg.value > 0) this.subsidizePool{value: msg.value}(poolId);
+
+        require(_send(centrifugeId, poolId, batch), InsufficientFundsForRepayment());
+        underpaid[centrifugeId][batchHash]--;
+
+        emit RepayBatch(centrifugeId, batch);
```
wischli, line 372: Nit: I would have liked to see this event in a test

src/common/Gateway.sol
```diff
@@ -288,86 +289,113 @@ contract Gateway is Auth, Recoverable, IGateway {
...
+    /// @inheritdoc IGateway
+    function repay(uint16 centrifugeId, bytes memory batch) external payable pauseable {
+        bytes32 batchHash = keccak256(batch);
+        require(underpaid[centrifugeId][batchHash] > 0, NotUnderpaidBatch());
+
+        PoolId poolId = processor.messagePoolId(batch);
+        if (msg.value > 0) this.subsidizePool{value: msg.value}(poolId);
+
+        require(_send(centrifugeId, poolId, batch), InsufficientFundsForRepayment());
+        underpaid[centrifugeId][batchHash]--;
```
wischli, line 370: Q: Is there any way to know how much payment needs to be provided for fixing an unpaid message or is it sufficient to provide large funds and then get back the remainder which is not consumed automatically?

wischli (approved):
AFAICT, solves the raised flags.

Since I haven't been that involved on the Gateway low level logic, I don't fully understand all of it in detail though.

src/common/Gateway.sol
hieronx: Transactions can only be underpaid if they were supposed to be subsidized. This is a case where already exact payments do not happen, rather the fund manager or other incentivized party calls `subsidizePool`. So when transactions become underpaid, the solution would be the fund manager subsidizing with a sufficient amount for several transactions.

The only reason I included the `subsidizePool` call into the `repay` call is that otherwise you could have a potential griefing attack, where when a fund manager subsidizes, all the funds are gone again before `repay` is called.

wischli: Thanks for the explanation, makes sense to me!

src/common/Gateway.sol
hieronx: Fixed in https://github.com/centrifuge/protocol-v3/pull/302

src/common/Gateway.sol
hieronx: Fixed in https://github.com/centrifuge/protocol-v3/pull/302

test/integration/EndToEnd.t.sol
```diff
@@ -234,6 +236,10 @@ contract TestEndToEnd is Test {
                 kind: uint8(VaultUpdateKind.DeployAndLink)
             }).serialize()
         );
+
+        vm.stopPrank();
+        vm.deal(address(this), DEFAULT_SUBSIDY);
+        cv.gateway.subsidizePool{value: DEFAULT_SUBSIDY}(poolId);
```
lemunozm, line 242: Maybe it's interesting to put this per test case, to add the exact required gas, in order to ensure the messages are correctly paid as expected. Otherwise, we could not notice some breaking change in the payment

lemunozm (commented):
Post-morten LGTM!!

I bit worried about the cost/complexity of the send method, but not sure what workaround exists to avoid stacking issues

PR #301, opened 2025-04-24
Refactor pool creation, remove trigger redeem request

hieronx:


src/hub/HubRegistry.sol
```diff
@@ -76,18 +78,27 @@ contract HubRegistry is Auth, IHubRegistry {
...
-        emit UpdateCurrency(poolId, currency_);
+        emit UpdateCurrency(poolId_, currency_);
+    }
+
+    //----------------------------------------------------------------------------------------------
+    // PermissViewionless methods
+    //----------------------------------------------------------------------------------------------
+
+    /// @inheritdoc IHubRegistry
+    function poolId(uint16 centrifugeId, uint48 postfix) public pure returns (PoolId poolId_) {
```
mustermeiszer, line 95: CR

The `HubRegistry` MUST have a storage for the `centrifugeId`

src/hub/HubRegistry.sol
```diff
@@ -76,18 +78,27 @@ contract HubRegistry is Auth, IHubRegistry {
...
+        poolId_ = newPoolId(centrifugeId, postfix);
```
mustermeiszer, line 96:
```suggestion
    function poolId(uint48 postfix) public pure returns (PoolId poolId_) {
        poolId_ = newPoolId(centrifugeId, postfix);
        require(currency[poolId_].isNull(), PoolAlreadyRegistered())
```

src/hub/HubRegistry.sol
hieronx: Why @mustermeiszer ?

src/hub/HubRegistry.sol
hieronx: I disagree, think this is a useful lookup without the require. UI can easily check it still

src/hub/HubRegistry.sol
hieronx: I think it can just be looked up in the other contract

mustermeiszer: That is essential for a correctly formed poolId. Otherwise, the pool is on ChainA but has id from ChainB

hieronx: Of course I understand why it needs the right chain ID. I am asking, why does it need to be in the `HubRegistry` vs in the `MessageDispatcher` where it already is now? And for the UI to look it up.

Anyways if you feel strongly I can add it here.

mustermeiszer: I just think you should not be able to create a pool in the first place. Or am I missing the check during pool creation?

mustermeiszer: That would be a pool that is saying I am on Chain B while it is on Ch.ain A

src/hub/HubRegistry.sol
mustermeiszer: Sounds good.

mustermeiszer: Just think the `centrifugeId` should not be part of the parameter. Or why do you think it should be?

src/hub/HubRegistry.sol
hieronx: There's a check in the `Hub`

mustermeiszer (dismissed):

mustermeiszer (approved):

PR #303, opened 2025-04-24
fix: investment manager types

wischli:
Fixes two minor post mortems from #297

### DIFF
--- a/src/vaults/AsyncVault.sol
+++ b/src/vaults/AsyncVault.sol
@@ -16,7 +16,6 @@ import {IAsyncVault} from "src/vaults/interfaces/IBaseVaults.sol";
 import {IAsyncRequests} from "src/vaults/interfaces/investments/IAsyncRequests.sol";
 import {IPoolEscrowProvider} from "src/vaults/interfaces/factories/IPoolEscrowFactory.sol";
 import {IShareToken} from "src/vaults/interfaces/token/IShareToken.sol";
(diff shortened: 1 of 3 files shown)


src/vaults/legacy/LegacyVaultAdapter.sol
```diff
@@ -94,12 +94,12 @@ contract LegacyVaultAdapter is AsyncVault, ILegacyVaultAdapter, IInvestmentManag
     // --- IInvestmentManager - View functions ---
     /// @inheritdoc IInvestmentManager
     function convertToShares(address, /* vault */ uint256 _assets) public view returns (uint256 shares) {
-        shares = manager.convertToShares(this, _assets);
+        shares = asyncManager().convertToShares(this, _assets);
```
wischli, line 97: While `manager` is correct as well for fully asynchronous vaults for which `asyncManager == manager` holds true, all other calls use `asyncManager()`.

src/vaults/AsyncVault.sol
```diff
@@ -36,7 +35,7 @@ contract AsyncVault is AsyncRedeemVault, IAsyncVault {
         uint256 tokenId_,
         IShareToken token_,
         address root_,
-        IAsyncRedeemManager manager_,
+        IAsyncRequests manager_,
```
wischli, line 38: Ref: https://github.com/centrifuge/protocol-v3/pull/297#discussion_r2056254626

hieronx (approved):

PR #305, opened 2025-04-25
Minor SCM cleanups

hieronx:


src/hub/ShareClassManager.sol
```diff
@@ -768,7 +793,7 @@ contract ShareClassManager is Auth, IShareClassManager {
             poolId,
             scId_,
             assetId,
-            issueEpochId[scId_][assetId] + 1,
+            epochId[scId_][assetId].deposit + 1,
```
hieronx: @wischli @mustermeiszer I believe this was a mistake and should be deposit rather than issue ID?

src/common/Gateway.sol
```diff
@@ -371,7 +371,7 @@ contract Gateway is Auth, Recoverable, IGateway {
         require(underpaid[centrifugeId][batchHash] > 0, NotUnderpaidBatch());
 
         PoolId poolId = processor.messagePoolId(batch);
-        if (msg.value > 0) this.subsidizePool{value: msg.value}(poolId);
+        if (msg.value > 0) subsidizePool(poolId);
```
hieronx, line 374: Unrelated to the PR but minor comment from the auditors

src/hub/ShareClassManager.sol
```diff
@@ -35,31 +36,29 @@ contract ShareClassManager is Auth, IShareClassManager {
 
     IHubRegistry public hubRegistry;
```
hieronx: Personally think we should make this `immutable`, would you agree @wischli @mustermeiszer ?

src/hub/ShareClassManager.sol
mustermeiszer: Yes, good catch! There are also the `nowDepositEpoch()` getters.

src/hub/ShareClassManager.sol
```diff
@@ -312,6 +323,58 @@ contract ShareClassManager is Auth, IShareClassManager {
...
+        string calldata symbol,
+        bytes32 salt,
+        bytes calldata
+    ) external auth {
+        require(exists(poolId, scId_), ShareClassNotFound());
+
+        _updateMetadata(scId_, name, symbol, salt);
+
+        emit UpdateMetadata(poolId, scId_, name, symbol, salt);
+    }
```
mustermeiszer: NIT

Can we also expose that in the hub? We wanted that, right?

src/hub/ShareClassManager.sol
```diff
@@ -312,6 +323,58 @@ contract ShareClassManager is Auth, IShareClassManager {
...
     }
 
+    /// @inheritdoc IShareClassManager
+    function updatePricePerShare(PoolId poolId, ShareClassId scId_, D18 navPoolPerShare) external auth {
+        require(exists(poolId, scId_), ShareClassNotFound());
+
+        ShareClassMetrics storage m = metrics[scId_];
+        m.navPerShare = navPoolPerShare;
+        emit UpdateShareClass(
+            poolId, scId_, navPoolPerShare.mulUint128(m.totalIssuance), navPoolPerShare, m.totalIssuance
```
mustermeiszer: NIT

Can you remove the `navPoolPerShare.mulUint128(m.totalIssuance)` and the `m.totalIssuance`? Those will not be ensured to be up-to-date. I think that is misleading. But not required.

src/hub/ShareClassManager.sol
wischli: Yeah, its part of the `Increase Hub interface` ticket: https://www.notion.so/Feat-Increase-Hub-interface-1cb2eac24e17801f9f60d673927ef4eb?pvs=4#1d62eac24e178085a267c7ff2f54809e

src/vaults/interfaces/token/IHook.sol
```diff
@@ -21,8 +21,8 @@ address constant ESCROW_HOOK_ID = address(uint160(uint8(0xce)));
 /// @notice Hook interface to customize share token behaviour
 /// @dev    To detect specific system actions:
 ///           Deposit request:      address(0)      -> address(user)
-///           Deposit claim:        address(escrow) -> address(user)
-///           Redeem request:       address(user)   -> address(escrow)
+///           Deposit claim:        ESCROW_HOOK_ID  -> address(user)
+///           Redeem request:       address(user)   -> ESCROW_HOOK_ID
```
wischli, line 25: Thanks for cleaning up my mess!

wischli (dismissed):

src/hub/ShareClassManager.sol
hieronx: I'm adding it here now actually! Also simplifying the salt logic a bit then

src/hub/ShareClassManager.sol
hieronx: Fixed

src/hub/ShareClassManager.sol
```diff
@@ -554,30 +563,33 @@ contract ShareClassManager is Auth, IShareClassManager {
...
-        require(salt != bytes32(0), InvalidSalt());
-
         ShareClassMetadata storage meta = metadata[scId_];
 
-        // Ensure that the salt remains unchanged if it's already set,
-        // or that it is being set for the first time.
-        require(salt == meta.salt || meta.salt == bytes32(0), InvalidSalt());
+        // Ensure that the salt is not being updated or is being set for the first time
+        require(
+            (salt == bytes32(0) && meta.salt != bytes32(0)) || (salt != bytes32(0) && meta.salt == bytes32(0)),
```
mustermeiszer, line 581: So one could update with `salt != meta.salt` but `meta.salt != bytes(0)` and `salt != bytes(0)` and it will succeed? Isn't that a bit odd?

mustermeiszer (approved):
Think the salt update is rather making it worse, but maybe I missed something. Approving anyways, rest is good!

src/hub/ShareClassManager.sol
hieronx: `updateMetadata` doesn't pass a salt anymore, only on `addShareClass`, so it anyways never can be updated

hieronx: I agree we should maybe change the logic but didn't really see a better way, without updating salt directly in `addShareClass` which would be more expensive

PR #304, opened 2025-04-24
Fix: Majors

mustermeiszer:
High level changes
1. Escrow `authTransferTo` instead of BSM having infinite allowance on PoolEscrow
2. Burn / mint shares not in `Fulfilled*` but in `{Issued, Revoked}Shares`
3. For async flow, pending amounts are held in a global escrow and transferred to the user during claiming

PR #306, opened 2025-04-25
fix: Escrow logic and other majors

mustermeiszer:
#### High level changes
1. Escrow `authTransferTo` instead of BSM having infinite allowance on PoolEscrow
2. Burn / mint shares not in `Fulfilled*` but in `{Issued, Revoked}Shares`
3. For async flow, pending amounts are held in a global escrow and transferred to the user during claiming

#### Asset Flows 
1. `AsyncVault.requestDeposit`: Transfer from owner to global escrow
2. `AsyncRequests.approvedDeposits`: Transfer from global escrow to pool escrow

#### Shares Flows
1. `AsyncVault.requestRedeem`: Transfer from owner to global
2. `AsyncRequests.issuedShares`: BS mints in global
3. `AsyncRequests.revokedShares`: Transfer from global to BS, BS burns in BS


mustermeiszer:
> https://github.com/centrifuge/protocol-v3/pull/304#pullrequestreview-2793211807

@wischli answering here

FMs will access funds by calling `bs.withdraw()`

script/VaultsDeployer.s.sol
```diff
@@ -104,10 +108,12 @@ contract VaultsDeployer is CommonDeployer {
         // Rely async investment manager
         balanceSheet.rely(address(asyncRequests));
         messageDispatcher.rely(address(asyncRequests));
+        vaultsGlobalEscrow.rely(address(asyncRequests));
 
         // Rely sync investment manager
```
hieronx, line 115: Can you add
- `register` call for `vaultsGlobalEscrow`
- `deny(deployer)` call for `vaultsGlobalEscrow`

src/vaults/AsyncRequests.sol
```diff
@@ -471,13 +498,7 @@ contract AsyncRequests is BaseInvestmentManager, IAsyncRequests {
 
         if (assets > 0) {
             VaultDetails memory vaultDetails = poolManager.vaultDetails(vault_);
-
-            address escrow = address(poolEscrowProvider.escrow(vault_.poolId()));
-            if (vaultDetails.tokenId == 0) {
```
hieronx, line 426: Quite nice to not need 6909 compatibility everywhere anymore

src/vaults/Escrow.sol
```diff
@@ -57,6 +57,22 @@ contract Escrow is Auth, IEscrow {
             emit Approve(asset, tokenId, spender, 0);
```
hieronx: I believe we should remove all `approve` related methods from `Escrow` now?

src/vaults/AsyncRequests.sol
```diff
@@ -198,6 +200,46 @@ contract AsyncRequests is BaseInvestmentManager, IAsyncRequests {
...
+        PoolId poolId,
+        ShareClassId scId,
+        AssetId assetId,
+        uint128 assetAmount,
+        uint128 shareAmount,
+        D18 pricePoolPerShare
+    ) external auth {
+        (address asset, uint256 tokenId) = poolManager.idToAsset(assetId);
+        // Lock assets to ensure they are not withdrawn and are available for the redeeming user
+        poolEscrowProvider.escrow(poolId).reserveIncrease(scId, asset, tokenId, assetAmount);
```
hieronx, line 230: Since we have a global escrow now again, should we just transfer to global escrow here? Then the reserve logic is not needed at all anymore in the per pool escrow.

mustermeiszer: What if the pool does not have the funds? Then this would fail? We would need to replay that.

src/vaults/Escrow.sol
mustermeiszer: Good point

script/VaultsDeployer.s.sol
mustermeiszer: Done

src/vaults/Escrow.sol
mustermeiszer: We can't the router escrow needs approval

mustermeiszer: Or we force transfer there too

src/vaults/AsyncRequests.sol
hieronx: Ah true, good point!

src/vaults/Escrow.sol
```diff
@@ -11,51 +11,31 @@ import {Recoverable} from "src/misc/Recoverable.sol";
...
-        if (IERC20(asset).allowance(address(this), spender) == 0) {
-            SafeTransferLib.safeApprove(asset, spender, type(uint256).max);
-            emit Approve(asset, spender, type(uint256).max);
-        }
-    }
-
     /// @inheritdoc IEscrow
-    function approveMax(address asset, uint256 tokenId, address spender) external auth {
+    function authTransferTo(address asset, uint256 tokenId, address receiver, uint256 amount) external auth {
+        emit AuthTransferTo(asset, tokenId, receiver, amount);
```
hieronx, line 21: Why emit events here first?

hieronx (approved):

wischli (approved):
Blind signing

hieronx (approved):

PR #309, opened 2025-04-25
Remove ERC6909 support in vaults

hieronx:


src/vaults/PoolManager.sol
```diff
@@ -350,7 +360,38 @@ contract PoolManager is Auth, Recoverable, IPoolManager, IUpdateContract, IPoolM
...
 
-    // --- Public functions ---
+    /// @inheritdoc IPoolManager
+    function linkVault(PoolId poolId, ShareClassId scId, AssetId assetId, IBaseVault vault) public auth {
+        _shareClass(poolId, scId);
+
+        AssetIdKey memory assetIdKey = _idToAsset[assetId];
+
+        IBaseInvestmentManager manager = vault.manager();
+        IVaultManager(address(manager)).addVault(poolId, scId, vault, assetIdKey.asset, assetId);
```
mustermeiszer, line 370: I don't like that. 

Nothing prevents the issuer to create vaults WITH an tokenId. We need to check here, that the `tokenId` is zero. 

What I would like even more is check that in the vault factory and error out there. Keeps the rest clean and erc6909 compatible

hieronx: You mean like https://github.com/centrifuge/protocol-v3/pull/309/files#diff-45a4b4e2ecc7b448c01ed1cf3873a557613d81c880dddbb87c7447c96b421980R40 and https://github.com/centrifuge/protocol-v3/pull/309/files#diff-6696e599e559b01d3287004a63f9f2f108f5e757886f2d888234ea91d9778133R47 ? ;-)

mustermeiszer: NEeeeeevermind!! Haha.

mustermeiszer (approved):
LGTM! Suprinsingly small

PR #311, opened 2025-04-25
Remove transient valuation

hieronx:


wischli (dismissed):
Agree the transient valuation makes solely sense for testing at this point

wischli (approved):

PR #312, opened 2025-04-25
refactor: pricing

wischli:
1. Merges `ConversionLib` and `VaultPricingLib` into new `PricingLib` which fundamentally still contains all methods of both libs
  * Methods from previous `VaultPricingLib` are `view` because they require querying decimals
  * Methods from previous `ConversionLib` are `pure` because they assume knowledge of assets

____________

 2. Removes all asset <-> share amount conversions which rely on `priceAssetPerShare := pPAPS` 
  * Exception: Some `AsyncRequest` calculations are based on `AsyncInvestState.{deposit, redeem}Price` which is pAPS ([ref](https://github.com/centrifuge/protocol-v3/blob/86c349cad45f0faa3a0b2e005145a63b607e5b55/src/vaults/interfaces/investments/IAsyncRequests.sol#L17)) and as result of receiving `Fulfilled{Deposit, Redeem}`. So it is set dynamically and doesn't require a separate message. IMO no reason to refactor this part as well.
  * WIN: Wherever (Hub + Vault) we convert assets <-> shares, we always use the same calculations based on pPPS and pPPA . Before,  the vault relied on pAPS and the `VaultPricingLib` vs. SCM relied on pPPS + pPPA and `ConversionLib` because it proved to be more precise
____________

3. Refactors `UpdateContractValuation` ([ref](https://kflabs.slack.com/archives/C07PG2EUR9C/p1745528306160669?thread_ts=1745528288.157159&cid=C07PG2EUR9C))
  * Previously provided `IERC7726` for querying `priceAssetPerShare` (i.e. being able to retrieve a quote amount for two assets and a base amount)
  * Now provides `ISyncDepositValuation` which returns a `D18 pricePoolPerShare` ([ref](https://kflabs.slack.com/archives/C07PG2EUR9C/p1745563421933209?thread_ts=1745528288.157159&cid=C07PG2EUR9C))

____________

4. Adds rounding to `D18.mul` and `D18.reciprocalMul` as well as pricing functions to be more explicit and allow for switching the rounding in case we realize we need to round differently somewhere

____________

#### Follow-up tasks for next week

- [x] ~~Always return `uint128` amounts in `PricingLib`~~ -> Reverted because in vaults, we sometimes rely on `uint256` amounts (e.g. `maxMint`) which cause overflow issues during tests
- [ ] 100% Coverage for `PricingLib`
- [x] Investigate https://github.com/Recon-Fuzz/centrifuge-review-protocol-v3/issues/43 - ~~breaks a lot of tests~~, luckily not anymore after all refactors 🎉


src/common/libraries/PricingLib.sol
```diff
@@ -0,0 +1,263 @@
...
+            return 0;
+        }
+
+        uint8 assetDecimals = getAssetDecimals(asset, tokenId);
+        uint8 shareDecimals = IERC20Metadata(shareToken).decimals();
+
+        // NOTE: More precise than d18(assets * 10 ** assetDecimals, shares * 10 ** shareDecimals)
+        return toPriceDecimals(assets, assetDecimals).mulDiv(
+            10 ** PRICE_DECIMALS, toPriceDecimals(shares, shareDecimals), rounding
+        );
```
wischli, line 134: Spend quite some time on this because I wanted to get rid the price decimal conversions but none of my attempts were as precise as the unchanged calculation

src/hub/Hub.sol
```diff
@@ -3,12 +3,12 @@ pragma solidity 0.8.28;
 
 import {D18, d18} from "src/misc/types/D18.sol";
 import {MathLib} from "src/misc/libraries/MathLib.sol";
-import {ConversionLib} from "src/misc/libraries/ConversionLib.sol";
 import {IERC7726} from "src/misc/interfaces/IERC7726.sol";
 import {Auth} from "src/misc/Auth.sol";
 import {Multicall, IMulticall} from "src/misc/Multicall.sol";
 import {Recoverable} from "src/misc/Recoverable.sol";
 
+import {PricingLib} from "src/common/libraries/PricingLib.sol";
```
hieronx, line 11: Appears unused in this file

hieronx (approved):
I can't find any issues, looks perfect! Great sign how few test changes there are

src/hub/Hub.sol
hieronx: Fixed in https://github.com/centrifuge/protocol-v3/pull/313/files

wischli: Yeah artifact of merging both libs

mustermeiszer (commented):
Looks all good to me!

PR #313, opened 2025-04-25
Transient price handling in `BalanceSheet`, manager per pool ID

hieronx:


src/vaults/AsyncRequests.sol
```diff
@@ -428,8 +429,9 @@ contract AsyncRequests is BaseInvestmentManager, IAsyncRequests {
         PoolId poolId = vault_.poolId();
         ShareClassId scId = vault_.scId();
 
-        // TODO: Enable transient pricing for BS
-        // (D18 pricePoolPerAsset,) = poolManager.pricePoolPerAsset(poolId, scId, vaultDetails.assetId, true);
+        (D18 pricePoolPerAsset,) = poolManager.pricePoolPerAsset(poolId, scId, vaultDetails.assetId, true);
```
mustermeiszer: NIT 

BS will use this Price.

src/vaults/BalanceSheet.sol
```diff
@@ -148,11 +146,19 @@ contract BalanceSheet is Auth, Recoverable, IBalanceSheet, IBalanceSheetGatewayH
...
     function noteRevoke(PoolId poolId, ShareClassId scId, address from, uint128 shares)
         external
-        authOrManager(poolId, scId)
+        authOrManager(poolId)
     {
         _noteRevoke(poolId, scId, from, shares);
     }
 
+    function setPricePoolPerAsset(PoolId poolId, ShareClassId scId, D18 value) external authOrManager(poolId) {
+        TransientStorageLib.tstore(keccak256(abi.encode("pricePoolPerAsset", poolId, scId)), value.raw());
```
mustermeiszer: Should we maybe include the assetID here? Not sure if otherwise would be quite easy to override here.

src/vaults/BalanceSheet.sol
```diff
@@ -325,4 +326,12 @@ contract BalanceSheet is Auth, Recoverable, IBalanceSheet, IBalanceSheetGatewayH
         queue.increase = 0;
         queue.decrease = 0;
     }
+
+    function _pricePoolPerAsset(PoolId poolId, ShareClassId scId) internal view returns (D18) {
+        return d18(TransientStorageLib.tloadUint128(keccak256(abi.encode("pricePoolPerAsset", poolId, scId))));
```
mustermeiszer: CR

If price is 0 must default to pool manager price

src/vaults/BalanceSheet.sol
```diff
@@ -325,4 +326,12 @@ contract BalanceSheet is Auth, Recoverable, IBalanceSheet, IBalanceSheetGatewayH
...
+    }
+
+    function _pricePoolPerShare(PoolId poolId, ShareClassId scId) internal view returns (D18) {
+        return d18(TransientStorageLib.tloadUint128(keccak256(abi.encode("pricePoolPerShare", poolId, scId))));
```
mustermeiszer: CR

Same as above

src/vaults/BalanceSheet.sol
```diff
@@ -148,11 +146,19 @@ contract BalanceSheet is Auth, Recoverable, IBalanceSheet, IBalanceSheetGatewayH
...
+        authOrManager(poolId)
     {
         _noteRevoke(poolId, scId, from, shares);
     }
 
+    function setPricePoolPerAsset(PoolId poolId, ShareClassId scId, D18 value) external authOrManager(poolId) {
+        TransientStorageLib.tstore(keccak256(abi.encode("pricePoolPerAsset", poolId, scId)), value.raw());
+    }
+
+    function setPricePoolPerShare(PoolId poolId, ShareClassId scId, D18 value) external authOrManager(poolId) {
```
mustermeiszer: NIT

Maybe call overridePricePoolPerShare

src/vaults/BalanceSheet.sol
wischli: Just to be sure: We dont want to default here to the sync manager which might have a custom pPPS valuation right?

mustermeiszer: Yes. The sync manager should use this override instead of the default pool manager price.

src/vaults/BalanceSheet.sol
hieronx: I wonder how we should handle when technically a price is actually 0.0.

src/vaults/AsyncRequests.sol
hieronx: I don't understand this comment?

src/vaults/BalanceSheet.sol
hieronx: Updated

src/vaults/BalanceSheet.sol
hieronx: Fixed

src/vaults/AsyncRequests.sol
mustermeiszer: If we have a default price in the BS, this will be it. So we technically dont need to override it here.

src/vaults/BalanceSheet.sol
wischli: My question was rather the opposite way: In my PR, we change the custom valuation to pPPS in the sync manager. 

So should BS fallback to `syncRequests.pricePoolPerShare` instead of poolmanager?

src/vaults/BalanceSheet.sol
wischli: I asked myself the same thing yesterday and decided on just returning 0 then. This was influenced by the implementation of VaultPricingLib which defaulted to 0 when the price was 0. So if you had `price = 0 = pPAPS = pPPS / pPPA`, that also applied.

wischli: So maybe the order of priority for pPPS is 

1. BS
2. SR = SyncRequests
3. PM

When the SR defaults to BS, this should be applied to the entire Vault Repository because SR is the SharePriceProvider. Though I dont remember if any contract still relies on that.

mustermeiszer (dismissed):
LGTM!Looks nice and clean!

For the null price case we could introduce two additional bool transient storages and if true we would use the zero price. 

Essentially something like `overridePricePoolPerAssetZero()`

Then the if clause could be extended to account for that storage

hieronx:
> For the null price case we could introduce two additional bool transient storages and if true we would use the zero price.
> Essentially something like overridePricePoolPerAssetZero()
> Then the if clause could be extended to account for that storage

Good idea, updated!

src/vaults/BalanceSheet.sol
hieronx: @mustermeiszer wdyt?

hieronx: Merging for now, let's consider in a followup PR

src/vaults/BalanceSheet.sol
mustermeiszer: I think we should never use the sync price anywhere as default besides in the sync one. The less implicit the action the easier. My main reason is not that it doesn't make sense. But it makes it harder to predict, simulate and reason about.

If somebody wants to have the sync price, they should overwrite the pool manager one before interacting with the bs.

mustermeiszer: I also would make the pool manager the default sharePriceProvider.

mustermeiszer: @hieronx and @wischli where is this provider actually used?

src/vaults/BalanceSheet.sol
mustermeiszer: Answered above. I don't think the sync manager should be the default price provider. Too opaque.

PR #315, opened 2025-04-26
Add slither based cleanups

hieronx:


src/hub/ShareClassManager.sol
```diff
@@ -31,9 +31,6 @@ contract ShareClassManager is Auth, IShareClassManager {
     using CastLib for *;
     using BytesLib for bytes;
 
-    uint32 constant META_NAME_LENGTH = 128;
```
hieronx, line 34: Unused. Wonder if we might need to add checks for this though

src/vaults/BaseInvestmentManager.sol
```diff
@@ -24,11 +24,10 @@ abstract contract BaseInvestmentManager is Auth, Recoverable, IBaseInvestmentMan
     using MathLib for uint256;
 
     address public immutable root;
+    IEscrow public immutable globalEscrow;
```
hieronx, line 22: No update method so should be immutable

src/vaults/BaseVaults.sol
```diff
@@ -200,7 +200,7 @@ abstract contract BaseVault is Auth, Recoverable, IBaseVault {
 }
 
 abstract contract BaseAsyncRedeemVault is BaseVault, IAsyncRedeemVault {
-    IAsyncRedeemManager public asyncRedeemManager;
+    IAsyncRedeemManager public immutable asyncRedeemManager;
```
hieronx, line 201: No update method so should be immutable

src/vaults/BaseVaults.sol
```diff
@@ -327,7 +327,7 @@ abstract contract BaseAsyncRedeemVault is BaseVault, IAsyncRedeemVault {
 }
 
 abstract contract BaseSyncDepositVault is BaseVault {
-    ISyncDepositManager public syncDepositManager;
+    ISyncDepositManager public immutable syncDepositManager;
```
hieronx, line 328: No update method so should be immutable

src/hub/Hub.sol
```diff
@@ -114,7 +114,7 @@ contract Hub is Multicall, Auth, Recoverable, IHub, IHubGatewayHandler {
 
             totalPayoutShareAmount += payoutShareAmount;
             totalPaymentAssetAmount += paymentAssetAmount;
-            totalCancelledAssetAmount = cancelledAssetAmount;
+            totalCancelledAssetAmount += cancelledAssetAmount;
```
hieronx, line 116: Good one to make sure we cover in end-to-end tests in the next few weeks

hieronx: Although I believe technically this can never be set multiple times. Since you need to claim before submitting a new request after a cancelation.

hieronx: Might make sense to rename `totalCancelledAssetAmount`. @wischli let me know what you think

wischli: Yeah good catch. It should be non-zero at most once, yes.

lemunozm:
Curious, did you use some tool to highlight the unused imports?

PR #319, opened 2025-04-26
Use `authTransferFrom` everywhere, add sync support to `VaultRouter`

hieronx:
- Remove ability for hooks to block `authTransferFrom`
- Some renamings in `BalanceSheet`
- Use `authTransferFrom` in `BalanceSheet.revoke` to not need an allowance
- Remove ERC6909 support from `VaultRouter`
- Add sync deposit flow support to `VaultRouter`

### DIFF
--- a/src/vaults/AsyncRequests.sol
+++ b/src/vaults/AsyncRequests.sol
@@ -108,27 +108,22 @@ contract AsyncRequests is BaseInvestmentManager, IAsyncRequests {
         auth
         returns (bool)
     {
-        uint128 _assets = assets.toUint128();
-        require(_assets != 0, ZeroAmountNotAllowed());
+        uint128 assets_ = assets.toUint128();
+        require(assets_ != 0, ZeroAmountNotAllowed());
 
-        return _processDepositRequest(vault_, _assets, controller);
-    }
-
-    /// @dev Necessary because of stack-too-deep
-    function _processDepositRequest(IBaseVault vault_, uint128 assets, address controller) internal returns (bool) {
         VaultDetails memory vaultDetails = poolManager.vaultDetails(vault_);
         PoolId poolId = vault_.poolId();
         ShareClassId scId = vault_.scId();
 
         require(poolManager.isLinked(poolId, scId, vaultDetails.asset, vault_), AssetNotAllowed());
 
-        require(_canTransfer(vault_, address(0), controller, convertToShares(vault_, assets)), TransferNotAllowed());
+        require(_canTransfer(vault_, address(0), controller, convertToShares(vault_, assets_)), TransferNotAllowed());
 
         AsyncInvestmentState storage state = investments[vault_][controller];
         require(state.pendingCancelDepositRequest != true, CancellationIsPending());
 
-        state.pendingDepositRequest += assets;
-        sender.sendDepositRequest(poolId, scId, controller.toBytes32(), vaultDetails.assetId, assets);
+        state.pendingDepositRequest += assets_;
+        sender.sendDepositRequest(poolId, scId, controller.toBytes32(), vaultDetails.assetId, assets_);
 
         return true;
     }
@@ -142,8 +137,11 @@ contract AsyncRequests is BaseInvestmentManager, IAsyncRequests {
         uint128 shares_ = shares.toUint128();
         require(shares_ != 0, ZeroAmountNotAllowed());
 
-        // You cannot redeem using a disallowed asset, instead another vault will have to be used
-        require(poolManager.isLinked(vault_.poolId(), vault_.scId(), vault_.asset(), vault_), AssetNotAllowed());
+        VaultDetails memory vaultDetails = poolManager.vaultDetails(vault_);
+        PoolId poolId = vault_.poolId();
+        ShareClassId scId = vault_.scId();
+
+        require(poolManager.isLinked(poolId, scId, vaultDetails.asset, vault_), AssetNotAllowed());
 
         require(
             _canTransfer(vault_, owner, ESCROW_HOOK_ID, shares)
@@ -155,9 +153,7 @@ contract AsyncRequests is BaseInvestmentManager, IAsyncRequests {
         require(state.pendingCancelRedeemRequest != true, CancellationIsPending());
 
         state.pendingRedeemRequest = state.pendingRedeemRequest + shares_;
-        VaultDetails memory vaultDetails = poolManager.vaultDetails(vault_);
-
-        sender.sendRedeemRequest(vault_.poolId(), vault_.scId(), controller.toBytes32(), vaultDetails.assetId, shares_);
+        sender.sendRedeemRequest(poolId, scId, controller.toBytes32(), vaultDetails.assetId, shares_);
 
         return true;
     }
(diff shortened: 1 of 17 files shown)


src/vaults/BalanceSheet.sol
```diff
@@ -293,7 +299,8 @@ contract BalanceSheet is Auth, Recoverable, IBalanceSheet, IBalanceSheetGatewayH
 
     function _executeRevoke(PoolId poolId, ShareClassId scId, address from, uint128 shares) internal {
         IShareToken token = poolManager.shareToken(poolId, scId);
-        token.burn(from, shares);
+        token.authTransferFrom(from, from, address(this), shares);
```
hieronx, line 302: Means a manager does not need to provide an allowance before calling `BalanceSheet.revoke`. Similar to how no allowance is required for `requestRedeem` and `transferShares`

src/vaults/BalanceSheet.sol
```diff
@@ -141,6 +146,7 @@ contract BalanceSheet is Auth, Recoverable, IBalanceSheet, IBalanceSheetGatewayH
         external
         authOrManager(poolId)
     {
+        require(!root.endorsed(from), CannotTransferFromEndorsedContract());
```
hieronx, line 149: Blocks balance sheet managers from transferring shares out of the global escrow

src/vaults/interfaces/token/IHook.sol
```diff
@@ -38,8 +38,8 @@ interface IHook is IERC165 {
         returns (bytes4);
 
     /// @notice Callback on authorized ERC20 transfer.
-    /// @dev    MUST return bytes4(keccak256("onERC20AuthTransfer(address,address,address,uint256,(bytes16,bytes16))"))
-    ///         if successful
+    /// @dev    Cannot be blocked, can only be used to update state.
```
hieronx, line 41: I've decided we should just not allow blocking `authTransferFrom`. No real reason to, and it simplifies several paths (removing try-catches)

src/vaults/Escrow.sol
```diff
@@ -60,35 +58,35 @@ contract PoolEscrow is Escrow, Recoverable, IPoolEscrow {
...
         holding[scId][asset][tokenId].reserved = newValue;
 
         emit IncreaseReserve(asset, tokenId, poolId, scId, value, newValue);
     }
 
     /// @inheritdoc IPoolEscrow
-    function reserveDecrease(ShareClassId scId, address asset, uint256 tokenId, uint256 value) external auth {
+    function reserveDecrease(ShareClassId scId, address asset, uint256 tokenId, uint128 value) external auth {
         uint128 prevValue = holding[scId][asset][tokenId].reserved;
-        uint128 value_ = value.toUint128();
```
hieronx, line 91: No casting needed if we just use `uint128` params

src/vaults/VaultRouter.sol
```diff
@@ -102,13 +102,29 @@ contract VaultRouter is Auth, Multicall, Recoverable, IVaultRouter {
...
-            _approveMax(vaultDetails.asset, vaultDetails.tokenId, address(vault));
+            _approveMax(vaultDetails.asset, address(vault));
         }
 
         _pay();
         vault.requestDeposit(amount, controller, owner);
     }
 
+    /// @inheritdoc IVaultRouter
+    function deposit(BaseSyncDepositVault vault, uint256 assets, address receiver, address owner)
```
hieronx, line 113: We didn't support this in the router before. By doing that, we can 1) use permits for deposits, 2) automatically pay, 3) use ERC20 wrapping.

src/vaults/VaultRouter.sol
```diff
@@ -120,12 +136,7 @@ contract VaultRouter is Auth, Multicall, Recoverable, IVaultRouter {
         lockedRequests[controller][vault] += amount;
 
         VaultDetails memory vaultDetails = poolManager.vaultDetails(vault);
-
-        if (vaultDetails.tokenId == 0) {
```
hieronx, line 124: Not anymore needed

src/vaults/Escrow.sol
```diff
@@ -34,20 +34,14 @@ contract Escrow is Auth, IEscrow {
 
     /// @inheritdoc IEscrow
     function authTransferTo(address asset, address receiver, uint256 amount) external auth {
-        uint256 balance = IERC20(asset).balanceOf(address(this));
-        require(balance >= amount, InsufficientBalance(asset, 0, amount, balance));
-
-        SafeTransferLib.safeTransfer(asset, receiver, amount);
-        emit AuthTransferTo(asset, receiver, amount);
+        authTransferTo(asset, 0, receiver, amount);
```
hieronx, line 37: Seemed a bit simpler

src/vaults/BalanceSheet.sol
hieronx: 1 bug fix for this, to block revoking shares in the global escrow: https://github.com/centrifuge/protocol-v3/pull/321/commits/5ce3316e7de9ff008841700f49790227e1ca3f3d

src/vaults/VaultRouter.sol
hieronx: Fix related to allowances in https://github.com/centrifuge/protocol-v3/pull/322/files#diff-b718607f380b901292cff484622241096a53434bcdc641336afa997895920c24R123

src/vaults/AsyncRequests.sol
```diff
@@ -142,8 +137,11 @@ contract AsyncRequests is BaseInvestmentManager, IAsyncRequests {
         uint128 shares_ = shares.toUint128();
         require(shares_ != 0, ZeroAmountNotAllowed());
 
-        // You cannot redeem using a disallowed asset, instead another vault will have to be used
-        require(poolManager.isLinked(vault_.poolId(), vault_.scId(), vault_.asset(), vault_), AssetNotAllowed());
+        VaultDetails memory vaultDetails = poolManager.vaultDetails(vault_);
+        PoolId poolId = vault_.poolId();
+        ShareClassId scId = vault_.scId();
+
+        require(poolManager.isLinked(poolId, scId, vaultDetails.asset, vault_), AssetNotAllowed());
```
mustermeiszer, line 144: Nice one!

src/vaults/VaultRouter.sol
```diff
@@ -102,13 +102,29 @@ contract VaultRouter is Auth, Multicall, Recoverable, IVaultRouter {
...
+    function deposit(BaseSyncDepositVault vault, uint256 assets, address receiver, address owner)
+        external
+        payable
+        protected
+    {
+        require(owner == msg.sender || owner == address(this), InvalidOwner());
+        require(!vault.supportsInterface(type(IERC7540Deposit).interfaceId), NonSyncDepositVault());
+
+        VaultDetails memory vaultDetails = poolManager.vaultDetails(vault);
+        SafeTransferLib.safeTransferFrom(vaultDetails.asset, owner, address(this), assets);
```
mustermeiszer, line 122: QUESTION

Is this CEI?

mustermeiszer (commented):
Great improvement

src/vaults/VaultRouter.sol
hieronx: No but here there is no way around it as we need to transfer to the vault first. From what I see the re-entrancy `protected` modifier should be sufficient though.

src/vaults/BalanceSheet.sol
wischli: Great catch!

src/vaults/AsyncRequests.sol
wischli: I am wondering whether we should add the same check to the `SyncRequests.deposit`?

PR #320, opened 2025-04-27
refactor: rm ISharePriceProvider

wischli:
* Removes `ISharePriceProvider` as it is not and should not be used outside of the impl `SyncRequests` ([ref comment](https://github.com/centrifuge/protocol-v3/pull/313#discussion_r2061317723))
* Moves `ISyncDepositValuation` and `struct` Prices to `ISyncRequests.sol`


hieronx (approved):
LGTM!

PR #323, opened 2025-04-27
Fix auth in accounting, cleanups

hieronx:


src/hub/Accounting.sol
```diff
@@ -58,7 +87,7 @@ contract Accounting is Auth, IAccounting {
     }
 
     /// @inheritdoc IAccounting
-    function addJournal(JournalEntry[] memory debits, JournalEntry[] memory credits) external {
+    function addJournal(JournalEntry[] memory debits, JournalEntry[] memory credits) external auth {
```
hieronx, line 108: Oops!

src/hub/Hub.sol
```diff
@@ -169,15 +169,15 @@ contract Hub is Multicall, Auth, Recoverable, IHub, IHubGatewayHandler {
     //----------------------------------------------------------------------------------------------
     /// @inheritdoc IHub
     function notifyPool(PoolId poolId, uint16 centrifugeId) external payable {
-        _protectedAndPaid(poolId);
+        _isManagerAndPaid(poolId);
```
hieronx, line 172: I felt `_protectedAndPaid`/`_protected` was confusing wrt the re-entrancy modifier also called `protected`

src/vaults/PoolManager.sol
```diff
@@ -137,6 +137,7 @@ contract PoolManager is
     function registerAsset(uint16 centrifugeId, address asset, uint256 tokenId)
         external
         payable
+        protected
```
hieronx, line 140: Don't see a clear risk, but since there is external calls, seems worth adding the re-entrancy modifier.

src/hub/Hub.sol
lemunozm: Much more legible!

PR #324, opened 2025-04-27
Add transient storage lib tests, gateway tests

hieronx:


test/common/unit/Gateway.t.sol
```diff
@@ -986,6 +986,9 @@ contract GatewayTestSend is GatewayTest {
         bytes32 batchId = keccak256(abi.encodePacked(LOCAL_CENT_ID, REMOTE_CENT_ID, batchHash));
         gateway.setRefundAddress(POOL_A, POOL_REFUND);
 
+        vm.expectRevert(IGateway.NotUnderpaidBatch.selector);
+        gateway.repay(REMOTE_CENT_ID, message);
```
lemunozm, line 990: NIT. I would move this to another error test case. I can do this in another PR

PR #325, opened 2025-04-27
Remove triggering deposits/withdrawals/revocations, add docs

hieronx:


mustermeiszer (approved):

src/hub/Hub.sol
```diff
@@ -225,31 +229,13 @@ contract Hub is Multicall, Auth, Recoverable, IHub, IHubGatewayHandler {
...
-    /// @inheritdoc IHub
-    function sendTriggerUpdateShares(
-        uint16 centrifugeId,
-        PoolId poolId,
-        ShareClassId scId,
-        address who,
-        uint128 shares,
-        bool isIssuance
-    ) public payable {
+    function sendTriggerIssueShares(uint16 centrifugeId, PoolId poolId, ShareClassId scId, address who, uint128 shares)
```
lemunozm, line 232: post-morten. The API should be just: `triggerIssueShares`?

hieronx: Seems inconsistent with all the other functions having `send`? But also fine with me, then let's do for all the trigger messages. Do you want to put up a PR?

lemunozm: Yeah, I can do it

PR #329, opened 2025-04-28
ref: Rename (A)SyncRequests to (A)SyncRequestManager

wischli:
[Ref Slack discussion](https://kflabs.slack.com/archives/C07PG2EUR9C/p1745827676690659?thread_ts=1745827350.227039&cid=C07PG2EUR9C)

Also adds linked vault check to `SyncRequestManager.deposit` which as introduced for async counterpart in 319: https://github.com/centrifuge/protocol-v3/pull/319#discussion_r2063078290


src/vaults/SyncRequestManager.sol
```diff
@@ -177,6 +181,9 @@ contract SyncRequests is BaseInvestmentManager, ISyncRequests {
         auth
         returns (uint256 shares)
     {
+        VaultDetails memory vaultDetails = poolManager.vaultDetails(vault_);
+        require(poolManager.isLinked(vault_.poolId(), vault_.scId(), vaultDetails.asset, vault_), AssetNotAllowed());
+
```
wischli, line 186: Probably should have split up PR due to rename noise. Here is the missing linked vault check for which I also added tests to both managers

test/vaults/integration/AsyncRequestManager.t.sol
```diff
@@ -0,0 +1,118 @@
...
+
+    // --- Simple Errors ---
+    function testRequestDepositUnlinkedVault() public {
+        (, address vault_, uint128 assetId) = deploySimpleVault(VaultKind.Async);
+        IAsyncVault vault = IAsyncVault(vault_);
+
+        VaultDetails memory vaultDetails = poolManager.vaultDetails(vault);
+        poolManager.unlinkVault(vault.poolId(), vault.scId(), AssetId.wrap(assetId), vault);
+
+        vm.expectRevert(IBaseInvestmentManager.AssetNotAllowed.selector);
```
mustermeiszer, line 107: NIT

Error will be `NoAuth` I guess.

lemunozm (approved):
Thanks so much for the renaming!

test/vaults/integration/AsyncRequestManager.t.sol
wischli: If that was the case, than the `vm.expectRevert` would fail due to a different error. However, since we execute the call from the `asyncRequestManager` itself, it should never fail with unauthorized.

lemunozm (approved):

PR #331, opened 2025-04-28
Rename claim to notify in Hub side

lemunozm:


src/hub/Hub.sol
```diff
@@ -101,7 +101,7 @@ contract Hub is Multicall, Auth, Recoverable, IHub, IHubGatewayHandler {
     //----------------------------------------------------------------------------------------------
 
     /// @inheritdoc IHub
```
hieronx, line 103: Just noticed I accidentally changed the section header above this to FreezeOnly methods, should be Permissionless methods. Could you fix?

lemunozm: Done

mustermeiszer (approved):

PR #333, opened 2025-04-29
Docs: add diagrams

lemunozm:
Note: following the UML standard, black arrows mean dependencies and white arrows mean implementations

## General messaging (`messaging.puml`)

How the messages are dispatched/processed in the system

image

## Common overview (`common.puml`)

image

## Hub overview (`hub.puml`)

image

## Vaults overview (`vaults/overview.puml`)

image

## Vaults (`vaults/vaults.puml`)

image

## Managers (`vaults/managers.puml`)

image

## Deployer scripts (`deployers.puml`)

image


docs/architecture/common.puml
```diff
@@ -0,0 +1,63 @@
...
+
+Guardian ---> Hub
+Guardian -left---> Root
+Guardian ---> MessageDispatcher
+
+Gateway -up--> GasService
+Gateway -up--|> Recoverable
+Gateway -right-> Root
+Gateway -down-->"n" IAdapter
+Gateway -left-> PoolEscrow
```
wischli: Q: This dependency is achieved by setting the pool escrow as refund address when we add a pool in the pool manager, right? If I am right, can we maybe add a note or additional information because AFAICT the other dependencies are pretty clear from looking at each of the contracts.

docs/architecture/vaults/overview.puml
```diff
@@ -0,0 +1,88 @@
...
+
+VaultRouter -left-> PoolManager
+VaultRouter -right-> Escrow
+
+PoolEscrowFactory -up-> PoolEscrow
+PoolEscrow --|> Escrow
+
+BaseVault --> ShareToken
+BaseVault --> ERC20
+BaseVault --> ERC6909
```
wischli, line 79: CR: Remove.

Sadly, we had to remove ERC6909 support for vaults in #309. IIRC the main reason was that the ERC4626 interface for the SyncDepositVault does not support it.

wischli (commented):
Thanks for these great diagrams! 

Do you think it is possible to make the Vaults overview more readable, i.e. have a graph that is a bit more planar? I understand why it is done the way it is but I am not sure if it brings that much value, to be honest. But maybe that's just my personal preference which is a fan of displaying "contract category buckets" to the reader. I understand this is not easily achievable and might not be worth the time spent. I just think an external person will be overwhelmed by this.

docs/architecture/common.puml
lemunozm: I agree! Will do

lemunozm: Done. Updated diagram

docs/architecture/vaults/overview.puml
lemunozm: Done. Updated diagram

wischli (approved):
Thanks again for the great work!

wischli:
Merging now such that I can apply the removal of `IVaultManager` in my PR.

PR #337, opened 2025-04-30
Fix: Gateway repay batches

lemunozm:
When a batch is repaid, it will send again. Nevertheless, when sending the gas limit is computed using some transient storages that are cleaned at the moment of repay, giving a free cost for sending batches.

Scenario:
- We send a batch that can not be paid.
- The transaction ends, and transient storage is removed.
- We repay the batch, but now the `isBatching` and `batchGasLimit` are incorrectly computed.

Solution is to store the `gasLimit` along with the underpaid message


src/common/Gateway.sol
```diff
@@ -283,52 +283,55 @@ contract Gateway is Auth, Recoverable, IGateway {
...
+                if (total <= subsidy[poolId].value) {
+                    subsidy[poolId].value -= uint96(total);
+                } else {
+                    underpaid[centrifugeId][data.batchHash].counter++;
+                    underpaid[centrifugeId][data.batchHash].gasLimit = batchGasLimit;
+                    emit UnderpaidBatch(centrifugeId, batch);
+                    return false;
+                }
             }
         }
```
lemunozm, line 336: Added `{` `}` to the whole block to solve the stack too deep issue but no logic has changed

src/common/Gateway.sol
```diff
@@ -283,52 +283,55 @@ contract Gateway is Auth, Recoverable, IGateway {
...
-                return false;
+                if (total > subsidy[poolId].value) {
+                    _requestPoolFunding(poolId);
+                }
+
+                if (total <= subsidy[poolId].value) {
+                    subsidy[poolId].value -= uint96(total);
+                } else {
+                    underpaid[centrifugeId][data.batchHash].counter++;
+                    underpaid[centrifugeId][data.batchHash].gasLimit = batchGasLimit;
```
lemunozm, line 331: Despite this line

src/common/Gateway.sol
```diff
@@ -283,52 +283,55 @@ contract Gateway is Auth, Recoverable, IGateway {
...
 
-    function _send(uint16 centrifugeId, PoolId poolId, bytes memory batch) internal returns (bool succeeded) {
+    function _send(uint16 centrifugeId, PoolId poolId, bytes memory batch, uint128 batchGasLimit)
+        internal
+        returns (bool succeeded)
+    {
         IAdapter[] memory adapters_ = adapters[centrifugeId];
         require(adapters[centrifugeId].length != 0, EmptyAdapterSet());
 
         SendData memory data = SendData({
```
hieronx, line 298: I wonder if the struct might not be necessary anymore, with your `{ }` trick and the removal of the gas limit?

hieronx (approved):

src/common/Gateway.sol
lemunozm: I tried but it's still needed 😢, There is also a new parameter `batchGasLimit`

PR #339, opened 2025-04-30
Fix: issueShares payment

lemunozm:
`hub.issueShares` did not take into account the payment, so all value paid was skipped, making the message fail, and the user's funds lost.

Along with this fix, I added some test simplifications to `Cases.t.sol`

### DIFF
--- a/src/hub/Hub.sol
+++ b/src/hub/Hub.sol
@@ -346,10 +346,11 @@ contract Hub is Multicall, Auth, Recoverable, IHub, IHubGatewayHandler {
         uint32 nowIssueEpochId,
         D18 navPoolPerShare
     ) external payable returns (uint128 issuedShareAmount, uint128 depositAssetAmount, uint128 depositPoolAmount) {
-        _isManager(poolId);
+        _isManagerAndPaid(poolId);
 
         (issuedShareAmount, depositAssetAmount, depositPoolAmount) =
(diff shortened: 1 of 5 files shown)


src/hub/Hub.sol
```diff
@@ -346,10 +346,11 @@ contract Hub is Multicall, Auth, Recoverable, IHub, IHubGatewayHandler {
         uint32 nowIssueEpochId,
         D18 navPoolPerShare
     ) external payable returns (uint128 issuedShareAmount, uint128 depositAssetAmount, uint128 depositPoolAmount) {
-        _isManager(poolId);
```
lemunozm, line 349: This was the issue

test/hub/mocks/MockVaults.sol
```diff
@@ -130,19 +123,22 @@ contract MockVaults is Test, Auth, IAdapter {
...
-        bytes memory popped = lastMessages[lastMessages.length - 1];
+        message = lastMessages[0];
+
+        for (uint256 i = 1; i < lastMessages.length; i++) {
+            lastMessages[i - 1] = lastMessages[i];
+        }
+
         lastMessages.pop();
-        return popped;
     }
```
lemunozm, line 143: FYI. I've modified the `popMessage()` method to perform instead a pop first, in order to test the messages in order as they arrive

hieronx (approved):

PR #341, opened 2025-04-30
Fix Axelar address format + a few minors

hieronx:
Fixes https://github.com/centrifuge/protocol-v3/issues/340


test/common/unit/AxelarAdapter.t.sol
```diff
@@ -169,12 +172,31 @@ contract AxelarAdapterTest is Test {
...
+        stringBytes[1] = "x";
+
+        for (uint256 i; i < 20; ++i) {
+            stringBytes[2 + i * 2] = characters[uint8(addressBytes[i] >> 4)];
+            stringBytes[3 + i * 2] = characters[uint8(addressBytes[i] & 0x0f)];
+        }
+
+        return string(stringBytes);
+    }
+}
```
lemunozm, line 202: Q: If this is not used in production, do we really need this for testing? 

I think we can use `vm.toString(address)` or just a plain string (?)

hieronx: It is what Axelar uses in production. Just in our code, we were doing `string(abi.encodePacked(addr))` and that wasn't equivalent to Axelar's usage of toString. See Cantina Discord thread.

src/hub/Hub.sol
```diff
@@ -414,8 +414,8 @@ contract Hub is Multicall, Auth, Recoverable, IHub, IHubGatewayHandler {
         IERC7726 valuation,
         AccountId assetAccount,
         AccountId equityAccount,
-        AccountId lossAccount,
-        AccountId gainAccount
+        AccountId gainAccount,
```
hieronx, line 417: Minor nitpick: think gain before loss makes more sense

mustermeiszer (approved):

PR #343, opened 2025-05-02
`VaultManager` refactor

hieronx:


src/vaults/SyncRequestManager.sol
```diff
@@ -282,6 +250,15 @@ contract SyncRequestManager is BaseInvestmentManager, ISyncRequestManager {
         priceData.assetPerShare = PricingLib.priceAssetPerShare(priceData.poolPerShare, priceData.poolPerAsset);
     }
 
+    /// @inheritdoc IVaultManager
+    function vaultKind(IBaseVault vault_) public view returns (VaultKind, address) {
```
mustermeiszer, line 254: QUESTION

IMO this methid should in general just live in the vault itself... Much easier and possible now as we have the adapter...

mustermeiszer (approved):
I am fine with this change, but wanted to propose some changes if we are at it. 

namely
* Hardcode `SyncRequestManager` and `AsyncRequestManager` in the `PoolManger`
* Make `vaultKind` a vault method of the `IBaseVault`

wischli:
Shoot. I had these changes pending but didnt push them yet.

wischli:
> I am fine with this change, but wanted to propose some changes if we are at it. 
> 
> 
> 
> namely
> 
> * Hardcode `SyncRequestManager` and `AsyncRequestManager` in the `PoolManger`
> 
> * Make `vaultKind` a vault method of the `IBaseVault` 
> 
> 

I can include this in my changes which also add more tests.

src/vaults/SyncRequestManager.sol
wischli: Im not sure if that works @mustermeiszer because our SyncDepositVaults are actually 2 vaults: A SyncDepositVault and a BaseAsyncRedeemVault.

So each of these vaults doesnt know about the other one but the request manager does. So for a SyncDepositVault, the impl of vaultKind is clear but for async vaults, its not as it could be either.

PR #344, opened 2025-05-03
Simplify escrow factory, minor pool manager + gateway cleanups

hieronx:


src/vaults/PoolManager.sol
```diff
@@ -119,6 +115,7 @@ contract PoolManager is
         protected
     {
         IShareToken share = IShareToken(shareToken(poolId, scId));
+        require(centrifugeId != sender.localCentrifugeId(), LocalTransferNotAllowed());
```
lemunozm, line 118: What happens if the receiver is in the same chain? Should be a normal ERC20 transfer?

mustermeiszer (approved):
LGTM

test/vaults/integration/PoolManager.t.sol
```diff
@@ -87,11 +87,9 @@ contract PoolManagerTest is BaseTest, PoolManagerTestHelper {
                 && nonWard != address(gateway)
         );
 
-        IVaultFactory[] memory vaultFactories = new IVaultFactory[](1);
-        vaultFactories[0] = asyncVaultFactory;
-
         // redeploying within test to increase coverage
-        new PoolManager(tokenFactory, vaultFactories, address(this));
+        new PoolManager(tokenFactory, address(this));
```
lemunozm, line 91: Curious, offtopic. Why the coverage didn't work without it?

lemunozm (approved):

PR #345, opened 2025-05-04
Add source chain ID checks, change share transfers into two-step process

hieronx:


src/common/MessageDispatcher.sol
```diff
@@ -539,19 +539,24 @@ contract MessageDispatcher is Auth, IMessageDispatcher {
...
+    function sendTransferShares(
+        uint16 centrifugeId,
+        PoolId poolId,
+        ShareClassId scId,
+        bytes32 receiver,
+        uint128 amount,
+        bool forward
+    ) external auth {
+        gateway.send(
+            forward ? centrifugeId : poolId.centrifugeId(),
```
lemunozm: Should we not dispatch an error if centrifugeId == localCentrifugeId?
It means, the user must do ERC20.transfer instead

src/common/libraries/MessageLib.sol
```diff
@@ -76,6 +76,12 @@ enum VaultUpdateKind {
     Unlink
 }
 
+enum MessageDirection {
+    Bidirectional,
```
lemunozm: I think the name is AnyToAny. Bidirectional for me means HubToVaults + VaultsTohub, but for some messages here we can do i.e VaultsToVaults or HubToHub for some recovery messages

src/common/libraries/MessageLib.sol
```diff
@@ -159,6 +165,24 @@ library MessageLib {
         }
     }
 
+    function messageDirection(bytes memory message) internal pure returns (MessageDirection) {
```
lemunozm: As discussed, we can encode each direction into a uint256. I can do this in a PR on top of this if you want.

src/common/MessageDispatcher.sol
lemunozm: I think we're actually forwarding when `!forward`, right? The logic is ok, but the name and where `true` and `false` are placed are the opposite IMO.

src/vaults/PoolManager.sol
```diff
@@ -306,13 +306,20 @@ contract PoolManager is
...
+        bytes32 receiver,
+        uint128 amount
+    ) public auth {
         IShareToken shareToken_ = shareToken(poolId, scId);
 
-        shareToken_.mint(destinationAddress, amount);
+        if (centrifugeId == sender.localCentrifugeId()) {
+            shareToken_.mint(receiver.toAddress(), amount);
+        } else {
+            sender.sendTransferShares(centrifugeId, poolId, scId, receiver, amount, true);
```
lemunozm: Should not add some check here to validate the message before sending it to the correct chain?

src/hub/Hub.sol
```diff
@@ -615,16 +627,33 @@ contract Hub is Multicall, Auth, Recoverable, IHub, IHubGatewayHandler, IHubGuar
...
     }
 
     /// @inheritdoc IHubGatewayHandler
-    function decreaseShareIssuance(PoolId poolId, ShareClassId scId, uint128 amount) external {
+    function decreaseShareIssuance(uint16 centrifugeId, PoolId poolId, ShareClassId scId, uint128 amount) external {
         _auth();
 
+        require(hubRegistry.chain(poolId, centrifugeId), DisabledChain());
         shareClassManager.decreaseShareClassIssuance(poolId, scId, amount);
     }
```
lemunozm: I think we have the same issue with these messages. This message does not act over an assetId. We're checking here the poolId matches with the centrifugeId, but this messages is Vault to Hub, so any bad adapter can sent this message and pass the check.

lemunozm: I think this issue will disappear once we configure adapters by pool ID. Actually do we really need this PR, if we go with adapters-per-pool?

wischli (commented):
Makes sense to me so far without any red flags on top of your ongoing conversation.

src/vaults/PoolManager.sol
```diff
@@ -130,6 +130,8 @@ contract PoolManager is
 
         emit TransferShares(centrifugeId, poolId, scId, msg.sender, receiver, amount);
         sender.sendInitiateTransferShares(poolId, scId, centrifugeId, receiver, amount);
+
+        gateway.endTransactionPayment();
```
lemunozm, line 134: Post-mortem note. Maybe it makes sense to have a `transactionPay` modifier too, to avoid missing some calling end in some refactor/new feature. I will add to some PR if aggreed

hieronx: Makes sense!

PR #346, opened 2025-05-05
Automated NAV manager + simple price manager

hieronx:
TODO:
- [x] Bundle approve/issue and approve/redeem, to not allow slippage between assets and shares on the hub side
- [x] Integrate snapshots
- [ ] Limit price delta per day
- [x] Figure out authentication setup: 1) re-use `HubRegistry.manager()`, 2) custom mapping in each contract
- [ ] Figure out how to automate gas calculations for price updates: 1) call gas service directly, 2) store fixed values, 3) allow to set by manager

### Updated Hub diagram

image


src/managers/NAVManager.sol
```diff
@@ -0,0 +1,170 @@
...
+        uint128 issuance = shareClassManager.issuance(scId, centrifugeId);
+
+        globalIssuance = globalIssuance + issuance - networkMetrics.issuance;
+        globalNetAssetValue = globalNetAssetValue + netAssetValue_ - networkMetrics.netAssetValue;
+        D18 price = globalNetAssetValue / d18(globalIssuance);
+
+        networkMetrics.netAssetValue = netAssetValue_;
+        networkMetrics.issuance = issuance;
+
+        hub.updateSharePrice(poolId, scId, price);
```
mustermeiszer: QUESTION

Would we potentially also issue/revoke in this step?
Would we potentially also distribute prices to other domains in this step?

src/managers/NAVManager.sol
```diff
@@ -0,0 +1,170 @@
...
+    /// @inheritdoc ISnapshotHook
+    function onSync(PoolId poolId_, ShareClassId scId_, uint16 centrifugeId) external {
+        require(poolId == poolId_ && scId == scId_);
+        require(msg.sender == holdings, NotAuthorized());
+
+        NetworkMetrics storage networkMetrics = metrics[centrifugeId];
+        D18 netAssetValue_ = netAssetValue(centrifugeId);
+        uint128 issuance = shareClassManager.issuance(scId, centrifugeId);
+
+        globalIssuance = globalIssuance + issuance - networkMetrics.issuance;
```
mustermeiszer: QUESTION 

Why can't we use the issuance from the shareClassManager?

hieronx: That might already be updated from issuances/revokations from another network that were not actually in sync yet (aka that was not a snapshot on that network)

src/managers/NAVManager.sol
hieronx: > Would we potentially also issue/revoke in this step?

I guess we could do that yes!

> Would we potentially also distribute prices to other domains in this step?

Might become too expensive if there are too many networks, but we could have an optional, limited array of network IDs to update, yes.

test/integration/EndToEnd.t.sol
```diff
@@ -39,11 +39,13 @@ import {UpdateContractMessageLib} from "src/spoke/libraries/UpdateContractMessag
```
hieronx: Will need to discuss more how to adapt the end-to-end tests. Currently they heavily override share prices. I think it would make sense to do the overriding of prices at a higher layer, and use the NAV manager + simple price manager in the base end to end tests, since that will (in the future) be the most common setup.

mustermeiszer (commented):
First look, it does look really good! Why is the initialize holding/liabilityu needed here through? I would prefer to just add accounts w/ a function manyually.

onnovisser:
> Limit price delta per day

How do you imagine this @hieronx ?

src/managers/NAVManager.sol
```diff
@@ -0,0 +1,306 @@
...
+    constructor(PoolId poolId_, IHub hub_) {
+        poolId = poolId_;
+
+        hub = hub_;
+        hubRegistry = hub_.hubRegistry();
+        holdings = hub.holdings();
+        accounting = hub.accounting();
+    }
+
+    /// @dev Check if the msg.sender is a manager
```
hieronx: Nit: this is pretty obvious:
```suggestion
```

src/managers/NAVManager.sol
```diff
@@ -0,0 +1,306 @@
...
+    modifier onlyManager() {
+        require(manager[msg.sender], NotAuthorized());
+        _;
+    }
+
+    /// @dev Check if the msg.sender is a hub manager
```
hieronx:
```suggestion
```

src/managers/NAVManager.sol
```diff
@@ -0,0 +1,306 @@
...
+    IHoldings public immutable holdings;
+    IAccounting public immutable accounting;
+
+    INAVHook public navHook;
+    mapping(uint16 centrifugeId => uint16) public accountCounter;
+    mapping(uint16 centrifugeId => mapping(AssetId => AccountId)) public assetIdToAccountId;
+    mapping(address => bool) public manager;
+
+    constructor(PoolId poolId_, IHub hub_) {
+        poolId = poolId_;
```
hieronx: I think we briefly discussed this but I believe we can make this contract reusable across pools, by removing the poolId param here, adding it to all functions, and storing managers per pool ID.

This would make it much simpler to setup for new pools.

src/managers/NAVManager.sol
```diff
@@ -0,0 +1,306 @@
...
+        emit Transfer(scId_, fromCentrifugeId, toCentrifugeId, sharesTransferred);
+    }
+
+    /// @inheritdoc INAVManager
+    function updateHoldingValue(ShareClassId scId, AssetId assetId) public onlyManager {
+        hub.updateHoldingValue(poolId, scId, assetId);
+        (bool isSnapshot,) = holdings.snapshot(poolId, scId, assetId.centrifugeId());
+        if (isSnapshot) {
+            _onSync(scId, assetId.centrifugeId());
+        }
```
hieronx: Great idea to combine this!

Just a nit:
```suggestion
        if (isSnapshot) _onSync(scId, assetId.centrifugeId());
```

src/managers/SimplePriceManager.sol
```diff
@@ -0,0 +1,211 @@
...
+
+    uint16[] public networks;
+    uint128 public globalIssuance;
+    uint128 public globalNetAssetValue;
+    mapping(uint16 centrifugeId => NetworkMetrics) public metrics;
+
+    mapping(address => bool) public manager;
+    mapping(address => bool) public caller;
+
+    constructor(PoolId poolId_, IHub hub_) {
```
hieronx: Likewise as the nav manager, I believe this can be made to work across pool IDs.

src/managers/SimplePriceManager.sol
```diff
@@ -0,0 +1,211 @@
...
+
+    /// @inheritdoc ISimplePriceManager
+    function approveDepositsAndIssueShares(AssetId depositAssetId, uint128 approvedAssetAmount, uint128 extraGasLimit)
+        external
+        onlyManager
+    {
+        uint32 nowDepositEpochId = shareClassManager.nowDepositEpoch(scId, depositAssetId);
+        uint32 nowIssueEpochId = shareClassManager.nowIssueEpoch(scId, depositAssetId);
+
+        require(nowDepositEpochId == nowIssueEpochId, MismatchedEpochs());
```
hieronx: Great check!

hieronx (commented):
Looks already really good!

script/HubDeployer.s.sol
```diff
@@ -95,6 +98,8 @@ contract HubDeployer is CommonDeployer, HubConstants {
     ShareClassManager public shareClassManager;
     HubHelpers public hubHelpers;
     Hub public hub;
+    NAVManagerFactory public navManagerFactory;
+    SimplePriceManagerFactory public simplePriceManagerFactory;
```
lemunozm: I think this should be in `ManagersDeployer.s.sol`

src/hub/Hub.sol
```diff
@@ -748,7 +748,13 @@ contract Hub is Multicall, Auth, Recoverable, IHub, IHubGatewayHandler, IHubGuar
...
         _auth();
 
-        emit ForwardTransferShares(targetCentrifugeId, poolId, scId, receiver, amount);
+        shareClassManager.updateShares(originCentrifugeId, poolId, scId, amount, false);
+        shareClassManager.updateShares(targetCentrifugeId, poolId, scId, amount, true);
+
+        ISnapshotHook hook = holdings.snapshotHook(poolId);
+        if (address(hook) != address(0)) hook.onTransfer(poolId, scId, originCentrifugeId, targetCentrifugeId, amount);
+
+        emit ForwardTransferShares(originCentrifugeId, targetCentrifugeId, poolId, scId, receiver, amount);
```
lemunozm: This means two things regarding gas:
- We need to increase the gasService value for this message.
- We need to also provide an `extraGasLimit` to `InitialTransferShares` for the hook called here. Note the actual `extraGasLimit` is for the Execute case

src/managers/interfaces/INAVManagerFactory.sol
```diff
@@ -0,0 +1,15 @@
...
+import {INAVManager} from "./INAVManager.sol";
+
+import {PoolId} from "../../common/types/PoolId.sol";
+
+interface INAVManagerFactory {
+    event DeployNavManager(PoolId indexed poolId, address indexed manager);
+
+    error InvalidPoolId();
+
+    /// @notice Deploys new merkle proof manager.
```
lemunozm: doc is off

src/managers/NAVManager.sol
lemunozm: That would also mean we would control this contract, so we should add `root` to it IIUC

src/managers/NAVManager.sol
```diff
@@ -0,0 +1,302 @@
...
+
+    /// @inheritdoc INAVManager
+    function assetAccount(uint16 centrifugeId, AssetId assetId) public view returns (AccountId) {
+        return assetIdToAccountId[centrifugeId][assetId];
+    }
+
+    /// @inheritdoc INAVManager
+    function expenseAccount(uint16 centrifugeId, AssetId assetId) public view returns (AccountId) {
+        return assetAccount(centrifugeId, assetId);
+    }
```
lemunozm, line 228: Q. `asset` and `expense` account are the same?

src/managers/NAVManager.sol
```diff
@@ -0,0 +1,302 @@
...
+
+    /// @inheritdoc INAVManager
+    function setNAVHook(INAVHook navHook_) external onlyHubManager {
+        navHook = navHook_;
+        emit SetNavHook(address(navHook_));
+    }
+
+    /// @inheritdoc INAVManager
+    function updateManager(address manager_, bool canManage) external onlyHubManager {
+        require(manager_ != address(0), EmptyAddress());
```
lemunozm: I think this check is not needed, 0 is as wrong as any other wrong address.

src/managers/NAVManager.sol
```diff
@@ -0,0 +1,302 @@
...
+    //----------------------------------------------------------------------------------------------
+
+    /// @inheritdoc INAVManager
+    function initializeNetwork(uint16 centrifugeId) external onlyManager {
+        require(accountCounter[centrifugeId] == 0, AlreadyInitialized());
+
+        hub.createAccount(poolId, equityAccount(centrifugeId), false);
+        hub.createAccount(poolId, liabilityAccount(centrifugeId), false);
+        hub.createAccount(poolId, gainAccount(centrifugeId), false);
+        hub.createAccount(poolId, lossAccount(centrifugeId), false);
```
lemunozm, line 75: Q. Are all these accounts `creditNormal`? `liabilityAccount` is `debitNormal` inside `initializeLiability()`

src/managers/NAVManager.sol
```diff
@@ -0,0 +1,302 @@
...
+        require(index > 0, NotInitialized());
+        require(index < type(uint16).max, ExceedsMaxAccounts());
+
+        AccountId assetAccount_ = assetIdToAccountId[centrifugeId][assetId];
+        if (assetAccount_.isNull()) {
+            assetAccount_ = withCentrifugeId(centrifugeId, index);
+            assetIdToAccountId[centrifugeId][assetId] = assetAccount_;
+        }
+
+        hub.createAccount(poolId, assetAccount_, true);
```
lemunozm: Why is not `hub.createAccount` in the `if` branch?

src/managers/NAVManager.sol
```diff
@@ -0,0 +1,302 @@
...
+    function initializeLiability(ShareClassId scId, AssetId assetId, IValuation valuation) external onlyManager {
+        uint16 centrifugeId = assetId.centrifugeId();
+        uint16 index = accountCounter[centrifugeId];
+        require(index > 0, NotInitialized());
+        require(index < type(uint16).max, ExceedsMaxAccounts());
+
+        AccountId expenseAccount_ = assetIdToAccountId[centrifugeId][assetId];
+        if (expenseAccount_.isNull()) {
+            expenseAccount_ = withCentrifugeId(centrifugeId, index);
+            assetIdToAccountId[centrifugeId][assetId] = expenseAccount_;
```
lemunozm: Because `assetIdToAccountId` is used in both `initializeHolding` and `initializeLiability`, couldn't the manager mess up things? Should we use two different storage instead?

src/managers/NAVManager.sol
```diff
@@ -0,0 +1,302 @@
...
+
+        emit Transfer(scId_, fromCentrifugeId, toCentrifugeId, sharesTransferred);
+    }
+
+    /// @inheritdoc INAVManager
+    function updateHoldingValue(ShareClassId scId, AssetId assetId) public onlyManager {
+        hub.updateHoldingValue(poolId, scId, assetId);
+        (bool isSnapshot,) = holdings.snapshot(poolId, scId, assetId.centrifugeId());
+        if (isSnapshot) _onSync(scId, assetId.centrifugeId());
+    }
```
lemunozm: IIUC, `_onSync()`, will be already called by the hub inside `hub.updateHoldingValue()`, so you don't need the last two lines of this method

src/managers/NAVManager.sol
```diff
@@ -0,0 +1,302 @@
...
+
+    //----------------------------------------------------------------------------------------------
+    // Calculations
+    //----------------------------------------------------------------------------------------------
+
+    /// @inheritdoc INAVManager
+    function netAssetValue(uint16 centrifugeId) public view returns (uint128) {
+        // TODO: how to handle when one of the accounts is not positive (or positive for loss account)
+        // Which should never happen, but still...
```
lemunozm: If this should never happen, I think an error in a view method is ok

test/integration/ThreeChainEndToEnd.t.sol
```diff
@@ -101,17 +102,22 @@ contract ThreeChainEndToEndDeployment is EndToEndFlows {
...
+        h.hub.updateSharePrice(POOL_A, SC_1, d18(1, 1));
+        h.hub.notifySharePrice{value: GAS}(POOL_A, SC_1, sB.centrifugeId);
+
         // B: Mint shares
-        vm.startPrank(address(sB.root));
+        vm.startPrank(BSM);
         IShareToken shareTokenB = IShareToken(sB.spoke.shareToken(POOL_A, SC_1));
-        shareTokenB.mint(INVESTOR_A, amount);
+        sB.balanceSheet.issue(POOL_A, SC_1, INVESTOR_A, amount);
+        sB.balanceSheet.submitQueuedShares(POOL_A, SC_1, 0);
```
lemunozm: Q: Do we need to call this method because later initializeTransferShares will need to have the shares registered to transfer them, right?

test/managers/integration/NAVManager.t.sol
```diff
@@ -0,0 +1,245 @@
...
+        navManager.initializeHolding(scId, asset2, IValuation(address(valuation)));
+        navManager.initializeHolding(scId, asset3, IValuation(address(valuation)));
+        navManager.initializeLiability(scId, liabilityAsset, IValuation(address(valuation)));
+
+        cv.updateHoldingAmount(POOL_A, scId, asset1, uint128(1000 * 10 ** asset1Decimals), d18(1, 1), true, false, 0);
+        cv.updateHoldingAmount(POOL_A, scId, asset2, uint128(2300 * 10 ** asset2Decimals), d18(1, 1), true, false, 1);
+
+        vm.expectCall(address(hub), abi.encodeWithSelector(hub.updateSharePrice.selector, POOL_A, scId, d18(1, 1)));
+        vm.expectCall(address(hub), abi.encodeWithSelector(hub.notifySharePrice.selector, POOL_A, scId, CHAIN_CP));
+        vm.expectCall(address(hub), abi.encodeWithSelector(hub.notifySharePrice.selector, POOL_A, scId, CHAIN_CV));
```
lemunozm, line 94: Q. Are these methods called by the code below? Where?

lemunozm (commented):
A lot of great stuff here! Left a partial review, I still need to dig more in the `SimplePriceManager`

script/HubDeployer.s.sol
onnovisser: `ManagersDeployer` is only for Spoke managers I think.  Maybe we need a `HubManagersDeployer`?

lemunozm: Ah, that's correct. Yeah, probably we need a `HubManagersDeployer` and rename `ManagerDeployer` to `SpokeManagerDeployer`

src/managers/NAVManager.sol
onnovisser: it's only called in `hub.updateHoldingAmount`

src/managers/interfaces/INAVManagerFactory.sol
onnovisser: oops!

test/integration/ThreeChainEndToEnd.t.sol
onnovisser: the issue was that the share tokens were minted, but the ShareClassManager didn't have any issuance for them

test/managers/integration/NAVManager.t.sol
onnovisser: yes, in `updateHoldingAmount` or `updateShares` it may call the snapshot hook which calls the `SimplePriceManager` which calls `updateSharePrice` and `notifySharePrice`

lemunozm: Wow! Crazy 👍🏻

src/managers/NAVManager.sol
mustermeiszer: Do we think this manager will ever hold tokens? For paying for fees etc? 

If yes, I would keep it per pool. If no, fine for me to make reusable.

src/managers/NAVManager.sol
```diff
@@ -0,0 +1,302 @@
...
+
+    constructor(IHub hub_) {
+        hub = hub_;
+    }
+
+    /// @inheritdoc INAVManagerFactory
+    function newManager(PoolId poolId) external returns (INAVManager) {
+        require(hub.hubRegistry().exists(poolId), InvalidPoolId());
+
+        NAVManager manager = new NAVManager{salt: bytes32(uint256(poolId.raw()))}(poolId, hub);
```
mustermeiszer: CR 

We need to pass an initial manager to the manager, otherwise nobody will be able to use it. Or am I missing something?

src/managers/NAVManager.sol
mustermeiszer: Liability account should be `creditNormal` - [see](https://github.com/centrifuge/protocol-v3/blob/1e886d6399739d0ff6bf100bba5379bdaed7be3b/src/hub/HubHelpers.sol#L129)

src/managers/NAVManager.sol
```diff
@@ -0,0 +1,302 @@
...
+
+        emit UpdateManager(manager_, canManage);
+    }
+
+    //----------------------------------------------------------------------------------------------
+    // Account creation
+    //----------------------------------------------------------------------------------------------
+
+    /// @inheritdoc INAVManager
+    function initializeNetwork(uint16 centrifugeId) external onlyManager {
```
mustermeiszer: QUESTION

I am not a fan of hardcoding the initialization here. I know we want them to use it without trusting the submitter. Maybe lets keep in the first iteration.

src/managers/NAVManager.sol
```diff
@@ -0,0 +1,302 @@
...
+        hub.createAccount(poolId, gainAccount(centrifugeId), false);
+        hub.createAccount(poolId, lossAccount(centrifugeId), false);
+
+        accountCounter[centrifugeId] = 5;
+
+        emit InitializeNetwork(centrifugeId);
+    }
+
+    /// @inheritdoc INAVManager
+    function initializeHolding(ShareClassId scId, AssetId assetId, IValuation valuation) external onlyManager {
```
mustermeiszer: QUESTION

I really think this method should never error if the `AssetId` had an acocunt beforehand. Passing on a valuation is anyways such a big trust already IMO, then we can also trust the account setup to be passed here.

src/managers/NAVManager.sol
```diff
@@ -0,0 +1,302 @@
...
+
+    /// @inheritdoc INAVManager
+    function setHoldingAccountId(ShareClassId scId, AssetId assetId, uint8 kind, AccountId accountId)
+        external
+        onlyManager
+    {
+        hub.setHoldingAccountId(poolId, scId, assetId, kind, accountId);
+        // TODO: update assetIdToAccountId mapping and update value?
+        // Do we need to do something with the old account/value?
+    }
```
mustermeiszer: QUESTION

If we have these here, shouldn't we rely rather on 

* HUB -> sets up accounts for assets
* NAVManager -> Consumes these accounts and expectes them to be rightly formed

For the networks I see the point of automating, but if managers can change accounts here, I see no gain in setting them up automatically in the init methods. Would rather use account parameter and view function to create accounts for using as argument.

src/managers/SimplePriceManager.sol
```diff
@@ -0,0 +1,211 @@
...
+    {
+        uint32 nowRedeemEpochId = shareClassManager.nowRedeemEpoch(scId, payoutAssetId);
+        uint32 nowRevokeEpochId = shareClassManager.nowRevokeEpoch(scId, payoutAssetId);
+
+        require(nowRedeemEpochId == nowRevokeEpochId, MismatchedEpochs());
+
+        D18 navPoolPerShare = _navPerShare();
+        hub.approveRedeems(poolId, scId, payoutAssetId, nowRedeemEpochId, approvedShareAmount);
+        hub.revokeShares(poolId, scId, payoutAssetId, nowRevokeEpochId, navPoolPerShare, extraGasLimit);
+    }
```
mustermeiszer: QUESTION

* This makes it unusable for de-tokens as these usually follow 
  1.  approve(de)
  2. invest(underlying)
  3. updatePrice(underlying)
  4. issue(de)

* Shouldn't this only be possible if NAV is updated?

src/managers/NAVManager.sol
hieronx: Don't see why it would hold tokens no.

src/managers/NAVManager.sol
hieronx: Two reasons:
1. Cost is very little (just some bytecode we deploy once), and it makes setup a lot easier.
2. Means that the automated nav manager can be a less trusted role, as it has limited actions, compared to the hub manager role

src/managers/NAVManager.sol
hieronx: I don't really get what you mean. Hub will anyways error out if you try to initialize a holding for the same asset ID a second time, right?

src/managers/NAVManager.sol
hieronx: I don't think we should be designing these higher level manager contracts to be as generic as possible. There is a specific reason this manager is supposed to be used, it is valuable to add helper methods for this use case, at basically no cost (just bytecode).

src/managers/hub/SimplePriceManager.sol
```diff
@@ -0,0 +1,171 @@
...
+    mapping(PoolId poolId => Metrics) public metrics;
+    mapping(PoolId poolId => mapping(uint16 centrifugeId => NetworkMetrics)) public networkMetrics;
+    mapping(PoolId poolId => mapping(address => bool)) public manager;
+
+    constructor(IHub hub_, address deployer) Auth(deployer) {
+        hub = hub_;
+        hubRegistry = hub_.hubRegistry();
+        shareClassManager = hub_.shareClassManager();
+
+        // TODO: where to check share class count?
```
wischli: Reminder for this TODO

src/managers/hub/SimplePriceManager.sol
```diff
@@ -0,0 +1,171 @@
...
+    //----------------------------------------------------------------------------------------------
+
+    /// @inheritdoc ISimplePriceManager
+    function networks(PoolId poolId) external view returns (uint16[] memory) {
+        return metrics[poolId].networks;
+    }
+
+    /// @inheritdoc ISimplePriceManager
+    function setNetworks(PoolId poolId, uint16[] calldata centrifugeIds) external onlyHubManager(poolId) {
+        metrics[poolId].networks = centrifugeIds;
```
wischli: Nit: Suppose we should emit an event here?

src/managers/hub/SimplePriceManager.sol
```diff
@@ -0,0 +1,171 @@
...
+    // Updates
+    //----------------------------------------------------------------------------------------------
+
+    /// @inheritdoc INAVHook
+    function onUpdate(PoolId poolId, ShareClassId scId, uint16 centrifugeId, uint128 netAssetValue) external auth {
+        NetworkMetrics storage networkMetrics_ = networkMetrics[poolId][centrifugeId];
+        uint128 issuance = shareClassManager.issuance(scId, centrifugeId);
+
+        metrics[poolId].issuance = metrics[poolId].issuance + issuance - networkMetrics_.issuance;
+        metrics[poolId].netAssetValue = metrics[poolId].netAssetValue + netAssetValue - networkMetrics_.netAssetValue;
```
wischli: I suppose these should never underflow because `metrics[poolId].*` is always geq than the network counterpart? But then, `metrics[..].issuance` is not updated as part of `hub.updateShares`.

src/managers/hub/SimplePriceManager.sol
```diff
@@ -0,0 +1,171 @@
...
+        bytes[] memory cs = new bytes[](networkCount + 1);
+        cs[0] = abi.encodeWithSelector(hub.updateSharePrice.selector, poolId, scId, price);
+
+        for (uint256 i; i < networkCount; i++) {
+            cs[i + 1] = abi.encodeWithSelector(hub.notifySharePrice.selector, poolId, scId, metrics[poolId].networks[i]);
+        }
+
+        IMulticall(address(hub)).multicall{value: MAX_MESSAGE_COST * (cs.length)}(cs);
+
+        emit Update(poolId, metrics[poolId].netAssetValue, metrics[poolId].issuance, price);
```
wischli: Nit: We read `metrics[poolId]` multiple times (especially the `networks` field in the loop) and write once to `netAssetValue` as well as `issuane`. Could improve gas cost, which makes sense for a heavy operation such as this one, by caching `metrics[poolId]` read in the header.

src/managers/hub/SimplePriceManager.sol
```diff
@@ -0,0 +1,171 @@
...
+        D18 price = _navPerShare(poolId);
+
+        networkMetrics_.netAssetValue = netAssetValue;
+        networkMetrics_.issuance = issuance;
+
+        uint256 networkCount = metrics[poolId].networks.length;
+        bytes[] memory cs = new bytes[](networkCount + 1);
+        cs[0] = abi.encodeWithSelector(hub.updateSharePrice.selector, poolId, scId, price);
+
+        for (uint256 i; i < networkCount; i++) {
```
wischli, line 123: Q: Should we add a safety check for `networkCount < MAX_NETWORK_COUNT_CONST`?

src/managers/hub/SimplePriceManager.sol
```diff
@@ -0,0 +1,171 @@
...
+        uint16 fromCentrifugeId,
+        uint16 toCentrifugeId,
+        uint128 sharesTransferred
+    ) external auth {
+        NetworkMetrics storage fromMetrics = networkMetrics[poolId][fromCentrifugeId];
+        NetworkMetrics storage toMetrics = networkMetrics[poolId][toCentrifugeId];
+        fromMetrics.issuance -= sharesTransferred;
+        toMetrics.issuance += sharesTransferred;
+
+        emit Transfer(poolId, fromCentrifugeId, toCentrifugeId, sharesTransferred);
```
wischli: Nit: Shouldn't we include the `scId` here despite the `SimplePriceManager` only supporting a unique scId per pool id? I can imagine for indexer, it might still be helpful, especially with other nav managers that will have that field.

src/managers/hub/SimplePriceManager.sol
```diff
@@ -0,0 +1,171 @@
...
+    }
+
+    //----------------------------------------------------------------------------------------------
+    // Helpers
+    //----------------------------------------------------------------------------------------------
+
+    function _navPerShare(PoolId poolId) internal view returns (D18) {
+        return metrics[poolId].issuance == 0
+            ? d18(1, 1)
+            : d18(metrics[poolId].netAssetValue) / d18(metrics[poolId].issuance);
```
wischli: Note to myself to investigate this division because I think there are higher precision variants to achieve the same. No blocker!

src/managers/hub/SimplePriceManager.sol
```diff
@@ -0,0 +1,171 @@
...
+        emit UpdateManager(poolId, manager_, canManage);
+    }
+
+    //----------------------------------------------------------------------------------------------
+    // Updates
+    //----------------------------------------------------------------------------------------------
+
+    /// @inheritdoc INAVHook
+    function onUpdate(PoolId poolId, ShareClassId scId, uint16 centrifugeId, uint128 netAssetValue) external auth {
+        NetworkMetrics storage networkMetrics_ = networkMetrics[poolId][centrifugeId];
```
wischli, line 107: Q: Can you help me understand why `networkMetrics[..].issuance` necessary to keep track of here? AFAIU, it replicates `shareClassManager.issuance` most of the times
1. When shares are transferred, `scm.issuance` is updated in `Hub.initiateTransferShares` and then `networkMetrics` via `onTransfer`
2. However, `hub.updateShares` does not update the `networkMetrics`.

src/managers/hub/SimplePriceManager.sol
```diff
@@ -0,0 +1,171 @@
...
+    // Administration
+    //----------------------------------------------------------------------------------------------
+
+    /// @inheritdoc ISimplePriceManager
+    function networks(PoolId poolId) external view returns (uint16[] memory) {
+        return metrics[poolId].networks;
+    }
+
+    /// @inheritdoc ISimplePriceManager
+    function setNetworks(PoolId poolId, uint16[] calldata centrifugeIds) external onlyHubManager(poolId) {
```
wischli: Just want to make sure this was thought through: Is it realistic to remove a previously supported network for a pool at some point @hieronx ? If so, its issuance is still potentially represented in `metrics[poolId]` without us cleaning it up.

src/managers/hub/NAVManager.sol
```diff
@@ -0,0 +1,269 @@
...
+
+        emit InitializeLiability(poolId, scId, assetId);
+    }
+
+    //----------------------------------------------------------------------------------------------
+    // INAVHook updates
+    //----------------------------------------------------------------------------------------------
+
+    /// @inheritdoc ISnapshotHook
+    function onSync(PoolId poolId, ShareClassId scId, uint16 centrifugeId) external auth {
```
wischli, line 128: Just to be sure: Are we fine with just the generic `auth` safeguard here without pool-permission-granularity? AFAICT, the current implementation is safe because only trusted contracts from us call `onSync` but since this can easily provide access to other pools, I wanted to raise it.

src/managers/hub/SimplePriceManager.sol
```diff
@@ -0,0 +1,171 @@
...
+        hub.approveRedeems(poolId, scId, payoutAssetId, nowRedeemEpochId, approvedShareAmount);
+        hub.revokeShares(poolId, scId, payoutAssetId, nowRevokeEpochId, navPoolPerShare, extraGasLimit);
+    }
+
+    //----------------------------------------------------------------------------------------------
+    // Helpers
+    //----------------------------------------------------------------------------------------------
+
+    function _navPerShare(PoolId poolId) internal view returns (D18) {
+        return metrics[poolId].issuance == 0
```
wischli: Nit: Can remove an `SLOAD` by storing `metrics[poolId].issuance` on a scoped variable and re-using that.

src/managers/hub/NAVManager.sol
```diff
@@ -0,0 +1,269 @@
...
+import {IHoldings} from "../../hub/interfaces/IHoldings.sol";
+import {IAccounting} from "../../hub/interfaces/IAccounting.sol";
+import {IHubRegistry} from "../../hub/interfaces/IHubRegistry.sol";
+
+/// @dev Assumes all assets in a pool are shared across all share classes, not segregated.
+contract NAVManager is INAVManager, Auth {
+    IHub public hub;
+    IHubRegistry public hubRegistry;
+    IHoldings public holdings;
+    IAccounting public accounting;
```
wischli: Q: Is it desired for these contracts to be `immutable` in SimplePriceManager but mutable here?

src/managers/hub/NAVManager.sol
```diff
@@ -0,0 +1,269 @@
...
+    }
+
+    /// @inheritdoc INAVManager
+    function setHoldingAccountId(PoolId poolId, ShareClassId scId, AssetId assetId, uint8 kind, AccountId accountId)
+        external
+        onlyManager(poolId)
+    {
+        hub.setHoldingAccountId(poolId, scId, assetId, kind, accountId);
+        // TODO: update assetIdToAccountId mapping and update value?
+        // Do we need to do something with the old account/value?
```
wischli: Reminder for this TODO

src/managers/hub/NAVManager.sol
```diff
@@ -0,0 +1,269 @@
...
+    }
+
+    /// @inheritdoc INAVManager
+    function expenseAccount(AssetId assetId) public pure returns (AccountId) {
+        return withAssetId(assetId, uint16(AccountType.Expense));
+    }
+
+    /// @inheritdoc INAVManager
+    function equityAccount(uint16 centrifugeId) public pure returns (AccountId) {
+        return withCentrifugeId(centrifugeId, 1);
```
wischli: Nit: Maybe we can move the magic numbers 1-4 to contract consts?

src/common/types/AccountId.sol
```diff
@@ -1,14 +1,28 @@
 // SPDX-License-Identifier: BUSL-1.1
 pragma solidity 0.8.28;
 
-type AccountId is uint32;
+import {AssetId} from "./AssetId.sol";
 
-function raw(AccountId accountId_) pure returns (uint32) {
+type AccountId is uint256;
```
wischli, line 6: I suppose it was discussed this impacts `Accounting` and `Holdings` storages which will need to be migrated/reset?

test/managers/hub/integration/NAVManager.t.sol
```diff
@@ -0,0 +1,234 @@
...
+        valuation.setPrice(POOL_A, scId, asset1, d18(1, 1));
+        valuation.setPrice(POOL_A, scId, asset2, d18(1, 1));
+        valuation.setPrice(POOL_A, scId, asset3, d18(1, 1));
+        valuation.setPrice(POOL_A, scId, liabilityAsset, d18(1, 1));
+
+        vm.deal(address(simplePriceManager), 1 ether);
+    }
+
+    /// forge-config: default.isolate = true
+    function testSuccess() public {
```
wischli: Ultra-nit: Great test! Would be easier to follow if the different actions were isolated into separate tests though.

wischli (commented):
Great work on this beast and the tests @onnovisser 🎉  Sorry for the many questions I have 😅

src/common/types/AccountId.sol
onnovisser: yes this was discussed in a sync. keeping it as uint32 would be very inflexible and holdings and accounting aren't used yet on mainnet, so we now still have a chance to change it

src/managers/NAVManager.sol
mustermeiszer: Then lets go with fixed setup.

src/managers/NAVManager.sol
mustermeiszer: Ignore

src/managers/SimplePriceManager.sol
mustermeiszer: This is actually a hard requirement - we need to allow this flow. 

The flow should either be

i. Approve(de)
ii. invest(underlying)
iii. fulfillInvest(underlying) - de receives tokens
iv. updatePrice(underlying) 
v. udpatePrice(de)
iv. issue(de)

The main problem is us hardcoding that, as this means that the liquidty of the undelrying defines the liquiduty of the de. For the fully async flow I would argue that is fine. 

If we want to go with more liquidity on the de we can have sync invest vaults with a manual price in between liquiudyt intervals defined by us.

mustermeiszer: @hieronx do you agree?

src/managers/NAVManager.sol
mustermeiszer: Lets go with fixed then

src/managers/NAVManager.sol
mustermeiszer: Solved as code checks for hub manager

src/managers/hub/NAVManager.sol
onnovisser: made them all immutable (except newly added gateway)

src/managers/hub/SimplePriceManager.sol
onnovisser: updated

src/managers/hub/SimplePriceManager.sol
hieronx: Definitely!

src/managers/hub/SimplePriceManager.sol
hieronx: Yeah this is a good point, we should clean this up. Possibly easier by having `addNetwork` and `removeNetwork` methods.

src/managers/hub/SimplePriceManager.sol
onnovisser: Do we have such a restriction when deploying pools to new networks?

src/managers/hub/NAVManager.sol
onnovisser: I think it's also safe because, in order for anything bad to potentially happen, pool managers first need to explicitly make this contract a hub manager, because this contract (or the SimplePriceManager) does not have auth on core contracts.

src/managers/hub/SimplePriceManager.sol
```diff
@@ -0,0 +1,276 @@
...
+
+        emit UpdateManager(poolId, manager_, canManage);
+    }
+
+    //----------------------------------------------------------------------------------------------
+    // Updates
+    //----------------------------------------------------------------------------------------------
+
+    /// @inheritdoc INAVHook
+    function onUpdate(PoolId poolId, ShareClassId scId, uint16 centrifugeId, uint128 netAssetValue) external auth {
```
hieronx: I would separate this into two contracts:
- `SimplePriceManager` with a `virtual` onUpdate method
- `BatchPriceManager` which inherits from `SimplePriceManager`, which overrides onUpdate and adds the epoch checks, and adds the `approve/issue/revoke` methods.

onnovisser: named them `SimplePriceManagerBase` and `SimplePriceManager` for now. But open to other names.

hieronx: Why is the `Base` needed at all? I think you can just have `SimplePriceManager`, and then in the future `BatchPriceManager` can inherit from that.

onnovisser: I just wasn't sure about the naming of `BatchPriceManager`. Maybe `SimplePriceManager` and `BatchSimplePriceManager` ?

hieronx: Yeah that seems fine too

src/managers/hub/NAVManager.sol
```diff
@@ -0,0 +1,259 @@
...
+        (bool equityIsPositive, uint128 equity) = accounting.accountValue(poolId, equityAccount(centrifugeId));
+        (bool gainIsPositive, uint128 gain) = accounting.accountValue(poolId, gainAccount(centrifugeId));
+        (bool lossIsPositive, uint128 loss) = accounting.accountValue(poolId, lossAccount(centrifugeId));
+        (bool liabilityIsPositive, uint128 liability) = accounting.accountValue(poolId, liabilityAccount(centrifugeId));
+
+        require(
+            equityIsPositive && gainIsPositive && liabilityIsPositive && (!lossIsPositive || loss == 0), InvalidNAV()
+        );
+
+        return equity + gain - loss - liability;
```
wischli, line 213: Can we assume that under normal conditions this never underflows?

src/managers/hub/SimplePriceManager.sol
```diff
@@ -0,0 +1,171 @@
...
+
+    /// @inheritdoc INAVHook
+    function onUpdate(PoolId poolId, ShareClassId scId, uint16 centrifugeId, uint128 netAssetValue)
+        public
+        virtual
+        auth
+    {
+        require(scId.index() == 1, InvalidShareClass());
+
+        gateway.withBatch(
```
wischli, line 96: @lemunozm Im fuzzy on the re-entrancy protection with the latest changes. Could you check if this call should be `protected`? Currently the `SimplePriceManager` doesn't implement `ReentrancyGuard`.

src/managers/hub/SimplePriceManager.sol
```diff
@@ -0,0 +1,171 @@
...
+/// @notice Base share price calculation manager for single share class pools.
+contract SimplePriceManager is ISimplePriceManager, Auth {
+    IGateway public gateway;
+    IHub public immutable hub;
+    IHubRegistry public immutable hubRegistry;
+    IShareClassManager public immutable shareClassManager;
+
+    mapping(PoolId poolId => Metrics) public metrics;
+    mapping(PoolId poolId => mapping(uint16 centrifugeId => NetworkMetrics)) public networkMetrics;
+    mapping(PoolId poolId => mapping(address => bool)) public manager;
```
wischli: Nit: Could you move this one line down to have storage order by line length?
```suggestion
    mapping(PoolId poolId => mapping(address => bool)) public manager;
    mapping(PoolId poolId => mapping(uint16 centrifugeId => NetworkMetrics)) public networkMetrics;
```

src/managers/hub/BatchSimplePriceManager.sol
```diff
@@ -0,0 +1,194 @@
...
+        uint32 nowIssueEpochId = requestManager.nowIssueEpoch(scId, depositAssetId);
+
+        require(nowDepositEpochId == nowIssueEpochId, MismatchedEpochs());
+
+        D18 pricePoolPerAsset = hub.pricePoolPerAsset(poolId, scId, depositAssetId);
+        D18 navPoolPerShare = _navPerShare(poolId);
+        requestManager.approveDeposits(
+            poolId, scId, depositAssetId, nowDepositEpochId, approvedAssetAmount, pricePoolPerAsset, address(0)
+        );
+        requestManager.issueShares(poolId, scId, depositAssetId, nowIssueEpochId, navPoolPerShare, extraGasLimit, address(0));
```
wischli: IIUC, any value of the refund address here is ineffective due to being called with `withBatch`, right?

src/managers/hub/BatchSimplePriceManager.sol
```diff
@@ -0,0 +1,194 @@
...
+    }
+
+    /// @inheritdoc IBatchSimplePriceManager
+    function approveRedeemsAndRevokeShares(
```
wischli: Curious why we don't follow the callback flow of `approveDepositAndIssueShares`.

src/managers/hub/NAVManager.sol
```diff
@@ -0,0 +1,259 @@
...
+/// @dev Assumes all assets in a pool are shared across all share classes, not segregated.
+contract NAVManager is INAVManager, Auth {
+    IHub public immutable hub;
+    IHubRegistry public immutable hubRegistry;
+    IHoldings public immutable holdings;
+    IAccounting public immutable accounting;
+
+    mapping(PoolId => INAVHook) public navHook;
+    mapping(PoolId poolId => mapping(uint16 centrifugeId => bool)) public initialized;
+    mapping(PoolId poolId => mapping(address => bool)) public manager;
```
wischli: Same nit as above
```suggestion
    mapping(PoolId poolId => mapping(address => bool)) public manager;
    mapping(PoolId poolId => mapping(uint16 centrifugeId => bool)) public initialized;
```

src/managers/hub/NAVManager.sol
```diff
@@ -0,0 +1,259 @@
...
+import {IHub, AccountType} from "../../hub/interfaces/IHub.sol";
+import {IAccounting} from "../../hub/interfaces/IAccounting.sol";
+import {IHubRegistry} from "../../hub/interfaces/IHubRegistry.sol";
+
+/// @dev Assumes all assets in a pool are shared across all share classes, not segregated.
+contract NAVManager is INAVManager, Auth {
+    IHub public immutable hub;
+    IHubRegistry public immutable hubRegistry;
+    IHoldings public immutable holdings;
+    IAccounting public immutable accounting;
```
wischli: Nit
```suggestion
    IHoldings public immutable holdings;
    IAccounting public immutable accounting;
    IHubRegistry public immutable hubRegistry;
```

wischli (commented):
A few observations. Overall looks good on the first glance. Great job!

src/managers/hub/BatchSimplePriceManager.sol
```diff
@@ -0,0 +1,194 @@
...
+
+import {PoolId} from "../../common/types/PoolId.sol";
+import {AssetId} from "../../common/types/AssetId.sol";
+import {ShareClassId} from "../../common/types/ShareClassId.sol";
+import {IHub} from "../../hub/interfaces/IHub.sol";
+
+import {IBatchRequestManager} from "../../vaults/interfaces/IBatchRequestManager.sol";
+
+/// @notice Simple price manager for single share class pools with async request management.
+contract BatchSimplePriceManager is SimplePriceManager, IBatchSimplePriceManager {
```
hieronx: @onnovisser could we move this file out of this PR? As discussed in last sync, I think this has still too many open questions, and it makes more sense to spend a bit more time on it still.

src/managers/hub/NAVManager.sol
```diff
@@ -0,0 +1,259 @@
...
+    /// @inheritdoc INAVManager
+    function setNAVHook(PoolId poolId, INAVHook navHook_) external onlyHubManager(poolId) {
+        navHook[poolId] = navHook_;
+        emit SetNavHook(poolId, address(navHook_));
+    }
+
+    /// @inheritdoc INAVManager
+    function updateManager(PoolId poolId, address manager_, bool canManage) external onlyHubManager(poolId) {
+        manager[poolId][manager_] = canManage;
+
```
hieronx:
```suggestion
```

src/managers/hub/BatchSimplePriceManager.sol
onnovisser: that's right

src/managers/hub/BatchSimplePriceManager.sol
onnovisser: approveRedeems doesn't send a message so it doesn't need to be batched

src/managers/hub/BatchSimplePriceManager.sol
onnovisser: removed

src/managers/hub/NAVManager.sol
```diff
@@ -0,0 +1,259 @@
...
+        emit InitializeLiability(poolId, scId, assetId);
+    }
+
+    //----------------------------------------------------------------------------------------------
+    // INAVHook updates
+    //----------------------------------------------------------------------------------------------
+
+    /// @inheritdoc ISnapshotHook
+    function onSync(PoolId poolId, ShareClassId scId, uint16 centrifugeId) external auth {
+        _onSync(poolId, scId, centrifugeId);
```
hieronx: Why do we need the separate internal method?

src/managers/hub/NAVManager.sol
```diff
@@ -0,0 +1,259 @@
...
+        PoolId poolId,
+        ShareClassId scId,
+        uint16 fromCentrifugeId,
+        uint16 toCentrifugeId,
+        uint128 sharesTransferred
+    ) external auth {
+        require(address(navHook[poolId]) != address(0), InvalidNAVHook());
+
+        navHook[poolId].onTransfer(poolId, scId, fromCentrifugeId, toCentrifugeId, sharesTransferred);
+
```
hieronx:
```suggestion
```

src/managers/hub/SimplePriceManager.sol
```diff
@@ -0,0 +1,171 @@
...
+    //----------------------------------------------------------------------------------------------
+
+    /// @inheritdoc ISimplePriceManager
+    function file(bytes32 what, address data) external auth {
+        if (what == "gateway") gateway = IGateway(data);
+        else revert ISimplePriceManager.FileUnrecognizedParam();
+        emit File(what, data);
+    }
+
+    /// @inheritdoc ISimplePriceManager
```
hieronx: This should be moved to a View methods section at the bottom

hieronx: Actually since it's anyways not packed together with the other values in `Metrics`, because uint128 + uint128 is already 1 slot, it might be cleaner to just have a separate mapping for networks directly, rather than having this in the Metrics struct.

src/managers/hub/SimplePriceManager.sol
```diff
@@ -0,0 +1,171 @@
...
+                return;
+            }
+        }
+        revert NetworkNotFound();
+    }
+
+    /// @inheritdoc ISimplePriceManager
+    function updateManager(PoolId poolId, address manager_, bool canManage) external onlyHubManager(poolId) {
+        manager[poolId][manager_] = canManage;
+
```
hieronx:
```suggestion
```

hieronx (commented):
Some minor comments. Overall, amazing work! Especially great to see the extensive test coverage 🔥

src/managers/hub/NAVManager.sol
onnovisser: Under normal conditions yes. At least, Asset will always be equal to Equity + Gain - Loss. The only way it could potentially underflow is if liabilities exceeds assets

src/managers/hub/NAVManager.sol
onnovisser: leftover from when there were 2 methods calling _onSync. Will remove the internal method

src/managers/hub/SimplePriceManager.sol
onnovisser: updated

src/managers/hub/SimplePriceManager.sol
onnovisser: Added the `lockCallback` that I believe was needed to prevent this

src/hub/Hub.sol
```diff
@@ -390,6 +390,13 @@ contract Hub is BatchedMulticall, Auth, Recoverable, IHub, IHubRequestManagerCal
 
         (bool isPositive, uint128 diff) = holdings.update(poolId, scId, assetId);
         _updateAccountingValue(poolId, scId, assetId, isPositive, diff);
+
+        (bool isSnapshot,) = holdings.snapshot(poolId, scId, assetId.centrifugeId());
+
+        if (isSnapshot) {
+            ISnapshotHook hook = holdings.snapshotHook(poolId);
+            if (address(hook) != address(0)) hook.onSync(poolId, scId, assetId.centrifugeId());
+        }
```
lemunozm, line 399: Shouldn't we call `holding.setSnapshot` as we do in `HubHandler`? If not, should not add this hook call inside `holding.update()`?

My point is that we're calling to `hook.onSync()` from two different places in code: holdings and hub, and that's weird, and also involves rely on two contracts for integrators of hooks

src/managers/hub/NAVManager.sol
```diff
@@ -0,0 +1,249 @@
...
+    /// @inheritdoc INAVManager
+    function netAssetValue(PoolId poolId, uint16 centrifugeId) public view returns (uint128) {
+        (bool equityIsPositive, uint128 equity) = accounting.accountValue(poolId, equityAccount(centrifugeId));
+        (bool gainIsPositive, uint128 gain) = accounting.accountValue(poolId, gainAccount(centrifugeId));
+        (bool lossIsPositive, uint128 loss) = accounting.accountValue(poolId, lossAccount(centrifugeId));
+        (bool liabilityIsPositive, uint128 liability) = accounting.accountValue(poolId, liabilityAccount(centrifugeId));
+
+        require(
+            equityIsPositive && gainIsPositive && liabilityIsPositive && (!lossIsPositive || loss == 0), InvalidNAV()
```
lemunozm, line 210: Should take into accoutn `gain == 0` as we do with `loss == 0` in this check?

I mean `.. && (gainIsPositive || gain == 0) && ..`

src/managers/hub/SimplePriceManager.sol
```diff
@@ -0,0 +1,176 @@
...
+    mapping(PoolId poolId => uint16[]) internal _networks;
+
+    constructor(IHub hub_, address deployer) Auth(deployer) {
+        hub = hub_;
+        gateway = hub_.gateway();
+        hubRegistry = hub_.hubRegistry();
+        shareClassManager = hub_.shareClassManager();
+    }
+
+    modifier onlyManager(PoolId poolId) {
```
lemunozm: This seems not used in `SimplePriceManager`

src/managers/hub/SimplePriceManager.sol
```diff
@@ -0,0 +1,176 @@
...
+
+/// @notice Base share price calculation manager for single share class pools.
+contract SimplePriceManager is ISimplePriceManager, Auth {
+    IGateway public gateway;
+    IHub public immutable hub;
+    IHubRegistry public immutable hubRegistry;
+    IShareClassManager public immutable shareClassManager;
+
+    mapping(PoolId poolId => Metrics) public metrics;
+    mapping(PoolId poolId => mapping(address => bool)) public manager;
```
lemunozm: `manager` storage seems not used

lemunozm (commented):
Amazing job. So sorry for the late reviews and the continue rebasing. Should not be the way of work 🙏🏻 

Left some comments

src/managers/hub/SimplePriceManager.sol
onnovisser: ah yes, was from when it had manager actions. removed

src/managers/hub/NAVManager.sol
onnovisser: They're considered `isPositive` if they are 0 or more. However loss is expected to be 0 or less.

lemunozm: Thanks for clarifying this!

src/hub/Hub.sol
onnovisser: Good point. However, there's a couple issues:

- The hook needs to be called after the accounting update, so it can't be called in `Holding.update`
- The Hub can't call `Holding.setSnapshot` here, because it needs a nonce

onnovisser: So I'm not sure where else to put it

src/managers/hub/NAVManager.sol
wischli: > They're considered `isPositive` if they are 0 or more. However loss is expected to be 0 or less.

Nit: Mathematically, that's not correct and not following our logic of the BSM. I'd like to fix this, but definitely not in this PR.

lemunozm (dismissed):
LGTM!

src/managers/hub/NAVManager.sol
```diff
@@ -0,0 +1,249 @@
...
+        AccountId expenseAccount_ = expenseAccount(assetId);
+
+        hub.createAccount(poolId, expenseAccount_, true);
+        hub.initializeLiability(poolId, scId, assetId, valuation, expenseAccount_, liabilityAccount(centrifugeId));
+
+        emit InitializeLiability(poolId, scId, assetId);
+    }
+
+    //----------------------------------------------------------------------------------------------
+    // INAVHook updates
```
hieronx:
```suggestion
    // ISnapshotHook updates
```

src/managers/hub/SimplePriceManager.sol
```diff
@@ -0,0 +1,164 @@
...
+/// @notice Base share price calculation manager for single share class pools.
+contract SimplePriceManager is ISimplePriceManager, Auth {
+    IGateway public gateway;
+    IHub public immutable hub;
+    IHubRegistry public immutable hubRegistry;
+    IShareClassManager public immutable shareClassManager;
+
+    mapping(PoolId poolId => Metrics) public metrics;
+    mapping(PoolId poolId => mapping(uint16 centrifugeId => NetworkMetrics)) public networkMetrics;
+    mapping(PoolId poolId => uint16[]) internal _networks;
```
hieronx:
```suggestion
    mapping(PoolId poolId => uint16[]) internal _networks;
    mapping(PoolId poolId => mapping(uint16 centrifugeId => NetworkMetrics)) public networkMetrics;
```

test/managers/hub/integration/NAVManager.t.sol
onnovisser: have split up the test

lemunozm (approved):

PR #350, opened 2025-05-05
On/offramp manager

hieronx:
Introduces a higher level BSM that enables controlled on-ramping and off-ramping of ERC20 assets into pools.

* On-ramp: Allows depositing whitelisted ERC20 assets into the balance sheet. Assets must be explicitly enabled via the corresponding `UpdateContract` msg (setting `onramp` storage) before deposits are permitted.
* Off-ramp: Enables withdrawing assets from the balance sheet to specified receiver addresses. Each asset-relayer pair can be configured with a dedicated receiver address via `UpdateContract` which sets `relayer` storage

This manager integrates with the existing BS while providing granular control over which assets can enter/exit and who can facilitate these operations.


src/managers/OnOfframpManager.sol
```diff
@@ -0,0 +1,68 @@
...
+        }
+
+        balanceSheet.deposit(poolId, scId, asset, 0, address(this), amount);
+    }
+
+    function offramp(address asset) external {
+        address offrampDestination_ = offrampDestination[asset];
+        require(offrampDestination_ != address(0), NoOfframpDestinationSet());
+
+        IPoolEscrow escrow = balanceSheet.poolEscrowProvider().escrow(poolId);
```
lemunozm: How is it possible a manager uses the balanceSheet if balanceSheet is the vaults side but managers in the hub side?

src/managers/interfaces/IBalanceSheetManager.sol
```diff
@@ -0,0 +1,12 @@
+// SPDX-License-Identifier: BUSL-1.1
+pragma solidity >=0.5.0;
+
+import {IERC165} from "src/misc/interfaces/IERC165.sol";
+
+interface IDepositManager is IERC165 {
```
hieronx, line 6: Idea of these interfaces is to make it easier for the UI, to support any balance sheet manager that uses these interfaces, rather than have to specifically support `OnOfframpManager`

src/managers/OnOfframpManager.sol
```diff
@@ -0,0 +1,132 @@
...
+    }
+
+    //----------------------------------------------------------------------------------------------
+    // Owner actions
+    //----------------------------------------------------------------------------------------------
+
+    /// @inheritdoc IUpdateContract
+    function update(PoolId poolId_, ShareClassId, /* scId */ bytes calldata payload) external {
+        require(poolId == poolId_, InvalidPoolId());
+        require(msg.sender == spoke, NotAuthorized());
```
hieronx: This means we need to deploy a new on off ramp manager for every pool if we ever migrate the `Spoke`. I'd say this is ok since we should aim to not need to do that anyways

src/spoke/libraries/UpdateContractMessageLib.sol
```diff
@@ -64,4 +66,53 @@ library UpdateContractMessageLib {
     function serialize(UpdateContractSyncDepositMaxReserve memory t) internal pure returns (bytes memory) {
         return abi.encodePacked(UpdateContractType.SyncDepositMaxReserve, t.assetId, t.maxReserve);
     }
+
+    //---------------------------------------
+    //   UpdateContract.Toggle (submsg)
```
hieronx: Idea is to try to design these as generic, reusable update contract messages, that serve multiple purposes

src/managers/OnOfframpManager.sol
```diff
@@ -0,0 +1,132 @@
...
+
+    /// @inheritdoc IDepositManager
+    function deposit(address asset, uint256, /* tokenId */ uint128 amount, address owner) external {
+        require(manager[msg.sender] || permissionlessOnramp, IAuth.NotAuthorized());
+
+        require(onramp[asset], NotAllowedOnrampAsset());
+        require(owner == address(this) || owner == msg.sender);
+        require(amount <= IERC20(asset).balanceOf(owner), InvalidAmount());
+
+        balanceSheet.deposit(poolId, scId, asset, 0, amount);
```
hieronx, line 96: Only supports ERC20 assets. That is intentional, I think it's good to keep this simple, and focus on the main use cases (like JTRSY). We can adapt or build a new version later, for ERC6909 assets.

src/managers/OnOfframpManager.sol
```diff
@@ -0,0 +1,134 @@
...
+
+        uint8 kind = uint8(UpdateContractMessageLib.updateContractType(payload));
+
+        if (kind == uint8(UpdateContractType.UpdateAddress)) {
+            UpdateContractMessageLib.UpdateContractUpdateAddress memory m =
+                UpdateContractMessageLib.deserializeUpdateContractUpdateAddress(payload);
+            address who = m.who.toAddress();
+
+            if (m.kind == "onramp") {
+                onramp[who] = m.isEnabled;
```
lemunozm: From the storage name, this should named `asset` instead of `who` IIUC

src/managers/OnOfframpManager.sol
```diff
@@ -0,0 +1,134 @@
...
+    //----------------------------------------------------------------------------------------------
+    // Deposit & withdraw actions
+    //----------------------------------------------------------------------------------------------
+
+    /// @inheritdoc IDepositManager
+    function deposit(address asset, uint256, /* tokenId */ uint128 amount, address owner) external {
+        require(manager[msg.sender] || permissionlessOnramp, IAuth.NotAuthorized());
+
+        require(onramp[asset], NotAllowedOnrampAsset());
+        require(owner == address(this) || owner == msg.sender);
```
lemunozm: In which case `owner == address(this)` will be true?

src/managers/OnOfframpManager.sol
```diff
@@ -0,0 +1,134 @@
...
+        require(amount <= IERC20(asset).balanceOf(owner), InvalidAmount());
```
lemunozm: NIT. Can we consider that the ERC20 will already do that check in the implementation [here](https://github.com/centrifuge/protocol-v3/blob/main/src/misc/ERC20.sol#L106)? I think that check is a ERC20 duty.

src/managers/OnOfframpManager.sol
```diff
@@ -0,0 +1,134 @@
...
+
+        if (owner == msg.sender) SafeTransferLib.safeTransferFrom(asset, owner, address(this), amount);
+
+        balanceSheet.deposit(poolId, scId, asset, 0, amount);
+    }
+
+    /// @inheritdoc IWithdrawManager
+    function withdraw(address asset, uint256, /* tokenId */ uint128 amount, address receiver) external {
+        require(manager[msg.sender] || permissionlessOfframp, IAuth.NotAuthorized());
+        require(offramp[asset][receiver], InvalidOfframpDestination());
```
lemunozm:
```suggestion
        require(manager[msg.sender] || offramp[asset][receiver], IAuth.NotAuthorized());
```
Is it my suggestion equivalent to what exists without requiring explicitly to enable `permissionlessOfframp`?

It means, if it's the manager, it can withdraw without restriction, but if not, it should be in the offramping configuration.

src/managers/OnOfframpManager.sol
```diff
@@ -0,0 +1,134 @@
...
+    using MathLib for uint256;
+
+    PoolId public immutable poolId;
+    address public immutable spoke;
+    ShareClassId public immutable scId;
+    IBalanceSheet public immutable balanceSheet;
+
+    bool public permissionlessOnramp;
+    bool public permissionlessOfframp;
+    mapping(address => bool) public manager;
```
lemunozm: Nobody sets this storage. Why not use the manager from the BSM?

lemunozm (commented):
Looks good!

src/managers/OnOfframpManager.sol
hieronx: If you already transferred assets to the on off ramp manager. That is helpful for onboarding assets from Circle, where you just send USDC directly to the on off ramp manager

src/managers/OnOfframpManager.sol
hieronx: Agree!

src/managers/OnOfframpManager.sol
hieronx: No if it's the manager, it should still be limted to the offramp destinations

The idea is that the manager can still be a single cold wallet, and doens't need to be a multisig. If we don't restrict the offramp destinations, it would need to be heavily restricted, and then we might as well make it a manager on the balance sheet.

src/managers/OnOfframpManager.sol
hieronx: It should be a separate permission layer. This is indeed missing a setter, and we should possibly rename this

src/managers/OnOfframpManager.sol
lemunozm: Ah right

lemunozm (commented):
LGTM, despite the missing setter for the `manager`!

lemunozm (approved):
Looks great!!

Do you still want to merge this to spearbit-review? Or should we create another branch from the last commit shared with spearbit?

hieronx:
> Do you still want to merge this to spearbit-review? Or should we create another branch from the last commit shared with spearbit?

We should create another branch, I'll do that

src/managers/OnOfframpManager.sol
```diff
@@ -86,20 +75,15 @@ contract OnOfframpManager is IOnOfframpManager {
     //----------------------------------------------------------------------------------------------
 
     /// @inheritdoc IDepositManager
-    function deposit(address asset, uint256, /* tokenId */ uint128 amount, address owner) external {
-        require(relayer[msg.sender] || permissionlessOnramp, IAuth.NotAuthorized());
-
+    function deposit(address asset, uint256, /* tokenId */ uint128 amount, address /* owner */) external {
         require(onramp[asset], NotAllowedOnrampAsset());
```
lemunozm, line 94: Ok, I understand if we want to set some maximum, we should create another manager that supports that, right?

lemunozm (approved):
Great simplification!

src/managers/OnOfframpManager.sol
hieronx: Yeah indeed

src/managers/OnOfframpManager.sol
```diff
@@ -0,0 +1,125 @@
...
+
+        if (kind == uint8(UpdateContractType.UpdateAddress)) {
+            UpdateContractMessageLib.UpdateContractUpdateAddress memory m =
+                UpdateContractMessageLib.deserializeUpdateContractUpdateAddress(payload);
+
+            if (m.kind == "onramp") {
+                address asset = m.what.toAddress();
+                onramp[asset] = m.isEnabled;
+
+                if (m.isEnabled) SafeTransferLib.safeApprove(asset, address(balanceSheet), type(uint256).max);
```
hieronx, line 66: Since `BalanceSheet` now does not allow specifying `owner`, it is safe to leave around the approval. This is cheaper.

test/managers/integration/OnOfframpManager.t.sol
```diff
@@ -0,0 +1,131 @@
...
+import {ShareClassId} from "src/common/types/ShareClassId.sol";
+
+import {UpdateContractMessageLib} from "src/spoke/libraries/UpdateContractMessageLib.sol";
+
+import {UpdateRestrictionMessageLib} from "src/hooks/libraries/UpdateRestrictionMessageLib.sol";
+
+import {OnOfframpManager} from "src/managers/OnOfframpManager.sol";
+import {IOnOfframpManager} from "src/managers/interfaces/IOnOfframpManager.sol";
+
+abstract contract OnOfframpManagerBaseTest is BaseTest {
```
lemunozm, line 23: I think it can be easier to test and understand the test without all the IT overhead (and also do future refactors) in a UT environment, just by mocking deposit/withdraw from `BalanceSheet`

hieronx: Yeah I need to check that, but would still argue on this level integration tests are critical, and potentially more important than mocked unit tests. E.g. I wouldn't have caught the missing approval without it.

lemunozm: Yeah I agree, for me the ideal is:
- Everything cover by UT (in terms of coverage, 100%)
- Just one happy test case in IT level to check that all connections work as expected. That can be done in an already set environment. In this case, just a deposit-and-then-withdraw test case.

I know it seems more "work" at first, but I actually think everything gets simpler and easier to understand.

ALso having this separation can helps integrators to just check one test case to understand the contract and avoid all the burden from the little details covered in UT

lemunozm (approved):
LGTM!

test/managers/integration/OnOfframpManager.t.sol
hieronx: Yeah that makes sense! I'll try to refactor these tests early next week

src/managers/OnOfframpManager.sol
```diff
@@ -0,0 +1,125 @@
...
+        require(msg.sender == spoke, NotAuthorized());
+
+        uint8 kind = uint8(UpdateContractMessageLib.updateContractType(payload));
+
+        if (kind == uint8(UpdateContractType.UpdateAddress)) {
+            UpdateContractMessageLib.UpdateContractUpdateAddress memory m =
+                UpdateContractMessageLib.deserializeUpdateContractUpdateAddress(payload);
+
+            if (m.kind == "onramp") {
+                address asset = m.what.toAddress();
```
mustermeiszer: QUESTION

Would make this `AssetId` based and not `address` based.

src/managers/OnOfframpManager.sol
```diff
@@ -0,0 +1,125 @@
...
+contract OnOfframpManager is IOnOfframpManager {
+    using CastLib for *;
+
+    PoolId public immutable poolId;
+    address public immutable spoke;
+    ShareClassId public immutable scId;
+    IBalanceSheet public immutable balanceSheet;
+
+    mapping(address asset => bool) public onramp;
+    mapping(address asset => mapping(address relayer => address receiver)) public offramp;
```
mustermeiszer: QUESTION 

would map as `mapping(address asset => address receiver)` offramp and as `mapping(address relayer => bool)` relayer

src/managers/OnOfframpManager.sol
```diff
@@ -0,0 +1,125 @@
...
+        require(poolId == poolId_, InvalidPoolId());
+        require(msg.sender == spoke, NotAuthorized());
+
+        uint8 kind = uint8(UpdateContractMessageLib.updateContractType(payload));
+
+        if (kind == uint8(UpdateContractType.UpdateAddress)) {
+            UpdateContractMessageLib.UpdateContractUpdateAddress memory m =
+                UpdateContractMessageLib.deserializeUpdateContractUpdateAddress(payload);
+
+            if (m.kind == "onramp") {
```
mustermeiszer, line 60: QUESTION

Would remove completely. If we make this permissionless we should just allow to onramp any kind of asset permissionlessly. Or what is your worry here?

src/managers/OnOfframpManager.sol
hieronx: Why? Would just be 2 storage reads on every transaction.

src/managers/OnOfframpManager.sol
hieronx: Idea was that `UpdateAddress` is a relatively generic message that could be reused for different purposes. For that reason I think asset address fits better.

src/managers/OnOfframpManager.sol
hieronx: I think one potential issue would be that you could spam 1000s of assets into a balance sheet of a pool, and make the submission of queued assets until it's a snapshot really expensive.

hieronx: Generally seems like something that the pool manager should still control.

src/managers/OnOfframpManager.sol
mustermeiszer: But whitelisting a relayer for every asset is really, really annoying.

src/managers/OnOfframpManager.sol
mustermeiszer: I disagree. All our core internally is build around the asset id. I feel rather strong about his. External is address + token, internal is asset id

src/managers/OnOfframpManager.sol
mustermeiszer: Makes sense.

src/managers/OnOfframpManager.sol
hieronx: That's a really good point! Will change.

src/managers/OnOfframpManager.sol
hieronx: OK I don't feel as strongly, will adapt.

src/managers/OnOfframpManager.sol
```diff
@@ -0,0 +1,127 @@
...
+                else SafeTransferLib.safeApprove(asset, address(balanceSheet), 0);
+
+                emit UpdateOnramp(asset, m.isEnabled);
+            } else if (m.kind == "offramp") {
+                address asset = m.what.toAddress();
+                address relayer = m.who.toAddress();
+                address receiver = m.where.toAddress();
+                offramp[asset][relayer] = receiver;
+                emit UpdateOfframp(asset, relayer, receiver);
+            }
```
mustermeiszer: CR

I think we need a third message to add relayers now.

hieronx: Yeah was still working on this!

src/managers/OnOfframpManager.sol
wischli: Agreeing with @mustermeiszer despite this being resolved already

src/managers/OnOfframpManager.sol
```diff
@@ -0,0 +1,136 @@
...
+import {CastLib} from "src/misc/libraries/CastLib.sol";
+import {IERC165} from "src/misc/interfaces/IERC165.sol";
+import {SafeTransferLib} from "src/misc/libraries/SafeTransferLib.sol";
+
+import {PoolId} from "src/common/types/PoolId.sol";
+import {AssetId} from "src/common/types/AssetId.sol";
+import {ShareClassId} from "src/common/types/ShareClassId.sol";
+import {UpdateContractType, UpdateContractMessageLib} from "src/spoke/libraries/UpdateContractMessageLib.sol";
+
+import {IBalanceSheet} from "src/spoke/interfaces/IBalanceSheet.sol";
```
wischli, line 13: Nit
```suggestion

import {UpdateContractType, UpdateContractMessageLib} from "src/spoke/libraries/UpdateContractMessageLib.sol";
import {IBalanceSheet} from "src/spoke/interfaces/IBalanceSheet.sol";
```

src/managers/OnOfframpManager.sol
wischli: Q: The OnOfframpManager spell then also would have to replicate all storage map entries. IIUC, your decision of immutable pool id, spoke, scId and balancesheet (vs. file) is based on the fact that migrating the `Spoke` is unlikely?

src/managers/OnOfframpManager.sol
```diff
@@ -0,0 +1,136 @@
...
+    IBalanceSheet public immutable balanceSheet;
+
+    constructor(address spoke_, IBalanceSheet balanceSheet_) {
+        spoke = spoke_;
+        balanceSheet = balanceSheet_;
+    }
+
+    /// @inheritdoc IOnOfframpManagerFactory
+    function newManager(PoolId poolId, ShareClassId scId) external returns (IOnOfframpManager) {
+        OnOfframpManager manager = new OnOfframpManager{salt: keccak256(abi.encode(poolId.raw(), scId.raw()))}(
```
wischli, line 129: Nit: We can save ~118 gas without losing too much collision resistance with a `bytes32` + bit shifting approach
```suggestion
        OnOfframpManager manager = new OnOfframpManager{salt: bytes32((uint256(poolId.raw()) << 128) | uint256(ShareClassId.unwrap(scId)))}(
```

The current approach costs ~138 gas
* `abi.encode` cost: ~96 gas (32 gas + 2 words × 32 gas per word)
* `keccak256` cost: ~42 gas (30 gas + 6 gas per word for 2 words)

My proposal only ~20 gas as its a basic arithmetic operation

wischli (commented):
I love the simplicity of the manager, great work! Apart from a minor gas saving, nothing really new from my side luckily. 

I definitely agree with the points made by other reviewers which I will just summarize here so we can come back to this before merging this PR:
1. Switch from `address`-based to `AssetId`-based approach ([ref](https://github.com/centrifuge/protocol-v3/pull/350#discussion_r2120685079))
2. Nit: Test refactoring ([ref](https://github.com/centrifuge/protocol-v3/pull/350#discussion_r2113804918))

src/managers/OnOfframpManager.sol
hieronx: It is a neat idea but I think given the code readability + how uncommon this action is, I would lean to keeping it as is.

hieronx:
@lemunozm @wischli I would propose to merge this PR as is, and refactor the tests in a follow up PR. Would you agree that's fine?

lemunozm (approved):
LGTM! Let's merge it then!

PR #351, opened 2025-05-05
Extend balance sheet interface

hieronx:


src/vaults/BalanceSheet.sol
```diff
@@ -172,6 +172,25 @@ contract BalanceSheet is Auth, Recoverable, IBalanceSheet, IBalanceSheetGatewayH
         TransientStorageLib.tstore(keccak256(abi.encode("pricePoolPerShareIsSet", poolId, scId)), true);
     }
 
+    //----------------------------------------------------------------------------------------------
+    // View methods
+    //----------------------------------------------------------------------------------------------
+
+    /// @inheritdoc IBalanceSheet
+    function availableBalanceOf(PoolId poolId, ShareClassId scId, address asset, uint256 tokenId)
```
hieronx, line 182: Reasoning: from a balance sheet manager perspective, you really don't need to know about the escrow contract. You only need to know the available balance.

mustermeiszer (approved):

PR #352, opened 2025-05-07
Fix unlinked vault checks

hieronx:


src/vaults/SyncRequestManager.sol
```diff
@@ -275,11 +273,13 @@ contract SyncRequestManager is BaseRequestManager, ISyncRequestManager {
         balanceSheet.noteDeposit(poolId, scId, vaultDetails.asset, vaultDetails.tokenId, receiver, assets);
     }
 
-    function _maxDeposit(PoolId poolId, ShareClassId scId, address asset, uint256 tokenId)
+    function _maxDeposit(PoolId poolId, ShareClassId scId, address asset, uint256 tokenId, IBaseVault vault_)
         internal
         view
         returns (uint128 maxDeposit_)
     {
+        if (!poolManager.isLinked(poolId, scId, asset, vault_)) return 0;
```
hieronx, line 281: I think if the vault is unlinked, `maxDeposit` and `maxMint` should return 0 as you cannot deposit/mint.

wischli: Just to be sure: For our previews, we don't factor this in right now. Is that desired or should we return 0 there as well?

wischli (approved):
Great catch and lean solution!

src/vaults/SyncRequestManager.sol
hieronx: Actually preview should not account for this, per the original 4626 spec

> MUST NOT account for deposit limits like those returned from maxDeposit and should always act as though the deposit would be accepted, regardless if the user has enough tokens approved, etc.

PR #353, opened 2025-05-07
Reduce re-entrancy in gateway

hieronx:


src/common/Gateway.sol
```diff
@@ -448,6 +448,9 @@ contract Gateway is Auth, Recoverable, IGateway {
     /// @inheritdoc IGateway
     function endBatching() external auth {
         require(isBatching, NoBatched());
+        
+        TransientArrayLib.clear(BATCH_LOCATORS_SLOT);
+        isBatching = false;
```
lemunozm: We're cleaning the state before reading here isn't?

wischli: Yeah I think it should be 2 lines lower, after
```solidity
        bytes32[] memory locators = TransientArrayLib.getBytes32(BATCH_LOCATORS_SLOT);
```

lemunozm (approved):

lemunozm (approved):

PR #354, opened 2025-05-07
Optimize transient array

hieronx:


lemunozm (approved):
LGTM!

xmxanuel:
@hieronx @lemunozm 
If you want to follow the exact dynamic array approach for Solidity storage. 
```solidity
// pseudo code
lengthSlot = hash(key);
transientStorage[lengthSlot] = currLength;
valueSlot = hash(lengthSlot);
```

Basically, the entire array will be stored starting from `valueSlot`.

xmxanuel:
here is my issue from the report:



#### I2. TransientArrayLib implementation could be optimized to match Solidity's storage pattern
[TransientArrayLib.sol](https://github.com/centrifuge/protocol-v3/blob/f2ef1edefaa2d22e8c00ab70b647dbce3bf859a6/src/misc/libraries/TransientArrayLib.sol)

**Description:** The `TransientArrayLib` library implements a transient storage array using a custom storage pattern. While the implementation is functionally correct, it doesn't follow the same pattern that Solidity uses for dynamic arrays in storage, which could lead to confusion and 
has higher gas costs.

Solidity's native implementation for dynamic arrays computes the slot for array elements differently. For a dynamic array at storage slot `p`, Solidity:

1. Stores the array length at slot `p`
2. Stores the array elements starting at slot `keccak256(p) + i`, where `i` is the index

This is more gas-efficient than the current implementation which computes a new hash for each element access.

The current TransientArrayLib implementation:
- Stores length at `keccak256(abi.encodePacked(key, type(uint256).max))`
- Stores each element at `keccak256(abi.encodePacked(key, i))`

**Recommendation:** Consider refactoring the library to match Solidity's storage pattern for better gas efficiency and consistency with standard patterns:
- Store length at `keccak256(abi.encode(key))` (similar to how namespaced storage slots work)
- Store elements at `keccak256(keccak256(abi.encode(key))) + i`

**References:**
- [Solidity Storage Layout Documentation](https://docs.soliditylang.org/en/v0.8.20/internals/layout_in_storage.html#mappings-and-dynamic-arrays)

PR #355, opened 2025-05-07
Spearbit fix review

hieronx:
### Fix unlinked vault checks

https://github.com/centrifuge/protocol-v3/pull/355/commits/72369c6f6a338ea4394d7457f7a1e94301b89d9a

Right now, once a sync vault is unlinked, the `max*` values are not set to 0. This fixes that.

### Extend balance sheet interface (#351)

https://github.com/centrifuge/protocol-v3/pull/355/commits/44f82fd89f8dba4ccda5a25e8197480ad018cbae

This commit just updates the interface and adds a single helper method for balance sheet managers to look up the available balance for an asset.

### Reduce re-entrancy in gateway (#353)

https://github.com/centrifuge/protocol-v3/pull/355/commits/bf6b8d3d14e7e2c0d10a9727640d6cf9fdc6d1c0

This commit ensures all external calls in gateway are CEI compatible, reducing the probability of re-entrancy issues.

### Optimize transient array (#354)

https://github.com/centrifuge/protocol-v3/pull/355/commits/915655a54499a1f851d65b75b5c23547a19d3108

This commit optimizes the transient array implementation to avoid a keccak256 hash for each array item, mimicking the Solidity dynamic storage behaviour.

### Optimize transient bytes (#356)

https://github.com/centrifuge/protocol-v3/pull/355/commits/e52ee2a6ac9137ba4c41aa789bb56bbc78d42574

Same as #354, for transient bytes.

### Remove owner param from BalanceSheet.deposit (#359)

https://github.com/centrifuge/protocol-v3/pull/355/commits/60b34a00132188445cf611047e1d2d870a7bc6a7

This commit removes the `owner` parameter from `BalanceSheet.deposit`. Otherwise, if any manager leaves a non-zero approval, a manager from another pool could take this.

### Allow holding updates before initialization (#361)

https://github.com/centrifuge/protocol-v3/pull/355/commits/f3135b5bffbc702754f1e95d8225d0cffdc63b8b

This commit changes the holding behaviour on the Hub side, to allow `updateHoldingAmount` messages to come in from Vaults before `initializeHolding/Liability` is called. Effectively, this means a balance sheet manager can deposit assets, and the hub manager can only later call initialize to recognize this on its books.

### Remove vault managers as wards on share tokens (#362)

https://github.com/centrifuge/protocol-v3/pull/355/commits/49d2d6ae2a76e4f5ccddf98d57cb57764143d387

This commit reduces the coupling between request managers, `PoolManager` and share tokens. The set up of permissions for new vaults is now done in `PoolManager`, and the managers are no longer a ward on every share token.

### Simplify ERC20, balance sheet, and misc cleanups (#360)

https://github.com/centrifuge/protocol-v3/pull/355/commits/9c45f4ecc3614d73c2c17b94847e7d9c09090fae

This commit is some overall cleanups: removing an unused param from `updateHoldingAmount` message, fixing some comments and interfaces, and better organizing the code especially in the `BalanceSheet`.

It also removes the `from` parameter in `BalanceSheet.revoke`, for the same reason as #359.

And it fixes Spearbit issue 12.

### Remove recovery messages (#358)

https://github.com/centrifuge/protocol-v3/pull/355/commits/ff60d33571d96f4bc4bd33dc9cbb8337dd47e47a

This commit removes the recovery messages. Now, recoveries can only be triggered locally by Guardian, simplifying the behaviour of the `Gateway`.

### fix: fulfillment grief protection (#367)

https://github.com/centrifuge/protocol-v3/pull/355/commits/f07bc63db9d070845cebd1bbef68965f1e2c4978

This commit only sends fulfilled messages if the amount is non-zero, removing a potential griefing issue.

### Remove gateway estimation (#366)

https://github.com/centrifuge/protocol-v3/pull/355/commits/66a07449a6407db9d42606a3a7ca97c95f44edf9

This commit removes the estimate method from the gateway, since correct estimations are too complex onchain, and should rather be computed offchain.

### refactor: merged fulfilled messages (#369)

https://github.com/centrifuge/protocol-v3/pull/355/commits/36d96a9018d344a6c8cca19094f161f69a876b85

This commit merges the request and cancelation fulfillment messages, to remove the gateway refund issue as well as simplify the request handlers.

Fixes Spearbit issue 11.

### Fix: start/end transactional payment (#370)

https://github.com/centrifuge/protocol-v3/pull/355/commits/a32d13ad2f1cca9552fcd8b9f12d801ef05a636a

This commit adds a method to reset transaction payment method, allowing to pay for multiple transactions in 1 go.

It also introduces a new `HubHelpers` contract that takes some code from `Hub`. Reason is that the `Hub` has become too large for the max bytecode limit.

### fix vault claim-redeem payment (#365)

https://github.com/centrifuge/protocol-v3/pull/355/commits/7025ef85088eb9bf1afa66f3db5492712145472c

This commit adds a payment for the `BalanceSheet.withdraw` call on `claimRedeem`.

### ref: remove deprecated gateway msgs (https://github.com/centrifuge/protocol-v3/pull/374)

https://github.com/centrifuge/protocol-v3/pull/355/commits/12067a1e919f504545e83ceb6dcd081ca55c3e47

Followup of #358.

### Add source chain ID checks, change share transfers into two-step process (https://github.com/centrifuge/protocol-v3/pull/345)

https://github.com/centrifuge/protocol-v3/pull/355/commits/e9806a82d871b660854c7f4378c1a842df3be478

This commit adds extra checks to the message processor, that the source chain is either the hub chain for the pool (for messages hub => vaults), or the message is using the asset id from the source chain.

It also changes transfer shares into a two step process. The reason for this is that if you have hub chain A and chains B and C, it is likely that B <=> C is not set up as an adapter.

It also adds a new method to add an underpaid method, used for the 2nd step of the share transfer, in case it is not going to the hub chain. This is because we cannot easily charge fees for the 2nd step on the originating chain. We plan to implement a higher level gas relayer function to enable paying this automatically.

### Add methods to reset the price overrides #375

https://github.com/centrifuge/protocol-v3/pull/355/commits/48125f730cf2b168bbe458511af6f336f8fd01e4

This commit adds reset methods for the price overrides, since otherwise the transient storage stays and it might be possible to reuse this.

Follow-up in #382.

### Rename vaults to spokes, decouple vault managers and implementations #377

https://github.com/centrifuge/protocol-v3/pull/355/commits/859869fafb9e552bd32a92d9848880ac500c4d2d

This commit renames `src/vaults` to `src/spokes`, and moves the vault contracts into `src/spokes/vaults`, to better organize up this part of the code and reduce coupling between vaults and the rest.

Renaming references the hub-and-spoke architeture of the system. The spokes part can also be used without ever deploying a vault, e.g. just issuing a token directly based on offchain deposits (like a fiat backed stablecoin).

### Fix sliceZeroPadded #376

https://github.com/centrifuge/protocol-v3/pull/355/commits/3f632296b21038bbcbddd8b6b6f885396af7f2d0

This commit fixes Spearbit issues 1 and 6.

### Remove ERC20 wrapping feature (https://github.com/centrifuge/protocol-v3/pull/380)

https://github.com/centrifuge/protocol-v3/pull/355/commits/162b66c8e5e1521c1288d3d2dbce4d472458854e

This commit removes the ERC20 wrapping feature, fixing Spearbit issues 2 and 7.

It also adds a few comments to make it clear that `tokenId=0` cannot be used with ERC6909 assets. This fixes Spearbit issue 8.

### Reset price in managers after overrides (https://github.com/centrifuge/protocol-v3/pull/382)

https://github.com/centrifuge/protocol-v3/pull/355/commits/6756cdfa46a4fb11793e1612fec6e9d560f4744a

This commit actually integrates the price reset methods into the request managers.

### Fix licenses (https://github.com/centrifuge/protocol-v3/pull/383)

https://github.com/centrifuge/protocol-v3/pull/355/commits/9ff5eb54b5571a7fdea6162dcb90761a8b13d138

No code changes, just updating the license of `src/misc` to GPL2 or later.

### Rename PoolManager to Spoke

https://github.com/centrifuge/protocol-v3/pull/355/commits/fe074139d340b22e1524719ea7bd3cf0697de358

No code changes, just renaming `src/spokes/PoolManager` to `src/spokes/Spoke` (similar to `src/hub/Hub`, it's the central contract)

### Modify index from MessageProofId (https://github.com/centrifuge/protocol-v3/pull/381)

https://github.com/centrifuge/protocol-v3/pull/355/commits/a7316fe59d611025124fd816ca0b4a59a4c45710

Minor change, moving the proof ID in the messages to make it cleaner.

### MultiAdapter (https://github.com/centrifuge/protocol-v3/pull/372)

https://github.com/centrifuge/protocol-v3/pull/355/commits/ebae20248655a60f32b4a8d87f40be7f34b7647c

This PR splits up the gateway into 2 separate contracts, for better readability. There is no logic changes.

### custom message for updating the BSM (https://github.com/centrifuge/protocol-v3/pull/387)

https://github.com/centrifuge/protocol-v3/pull/355/commits/fce58cc761553206c18a5106e75375e5a31b19c5

This commit adds a separate message for updating balance sheet managers, to not need `UpdateContract`

### Pass tokenWards directly to the factory once. (https://github.com/centrifuge/protocol-v3/pull/390)

https://github.com/centrifuge/protocol-v3/pull/355/commits/78c92608b70d3a79f36c36f32fa4aee03da0cda1

This commit goes further in decoupling the spoke codebase, just changing how wards are set up on tokens.

### Fix: Correct check before disputeRecovery (https://github.com/centrifuge/protocol-v3/pull/391)

https://github.com/centrifuge/protocol-v3/pull/355/commits/2f13df1ecba3e3e674e74af100e31d3db50ded19

This commit adds a check in `disputeRecovery` to avoid calling it when no recovery is pending.

### Missing renames from Vaults to spoke (https://github.com/centrifuge/protocol-v3/pull/388)

https://github.com/centrifuge/protocol-v3/pull/355/commits/3ad4b69c032a1427825aee61589d3bc741506706

This commit follows up on the previous renaming of `vaults` to `spoke`

### paths spokes to spoke (https://github.com/centrifuge/protocol-v3/pull/389)

https://github.com/centrifuge/protocol-v3/pull/355/commits/7762b38519e2f77cd0f03ff7e43ff765bfb630ad

This commit also follows up by renaming the folder from `spokes` to `spoke`.

### ref: improve pricing precision (https://github.com/centrifuge/protocol-v3/pull/392)

https://github.com/centrifuge/protocol-v3/pull/355/commits/76a09f1732cd9d9e20e92cfbe124699a8a8f1b7f

This commit updates the pricing calculations to be more precise.

### Ref: UpdateVault as independent message (https://github.com/centrifuge/protocol-v3/pull/394)

https://github.com/centrifuge/protocol-v3/pull/355/commits/1332e3f5b21b0d51dd37c2ec8a7b033cbbdcab06

This commit moves the vault update logic into a separate message, to enable a simpler API on the `Hub` side.

### Add token & vault update methods (https://github.com/centrifuge/protocol-v3/pull/393)

https://github.com/centrifuge/protocol-v3/pull/355/commits/2df37e552565034d716bb52c3e1da3319fd09aa6

This commit separates some logic in `Spoke`, to allow the migration of legacy vaults.

### fix transferShares issue & some minor refactors (https://github.com/centrifuge/protocol-v3/pull/396)

https://github.com/centrifuge/protocol-v3/pull/355/commits/8e64f71447e438c43d42df8c9d304aaa9f987b23

This commit fixes an issue with the new transfer shares logic introduced in https://github.com/centrifuge/protocol-v3/pull/345.

### Use wrapped errors in safe transfer lib

https://github.com/centrifuge/protocol-v3/pull/355/commits/4dd193b917f5241a3ed01699193194a983e124ca

This commit just adds wrapped errors to `SafeTransferLib`, to better discover why a transfer failed.

### Fix common wards on deployment, add file method for hub helpers (https://github.com/centrifuge/protocol-v3/pull/399)

https://github.com/centrifuge/protocol-v3/pull/355/commits/290bf9beb7b25e4c8509dac3a5c5079e71929268

This PR fixes the wards set up in the common deploy scripts, and adds a `file` method for `hubHelpers` in `hub`.

### Tests: EndToEnd additions (https://github.com/centrifuge/protocol-v3/pull/400)

https://github.com/centrifuge/protocol-v3/pull/355/commits/d1ec5f5c4c547def5bed93b6671b071916958923

This PR adds end-to-end tests, and makes a few minor interface changes inspired by this.

### Restrict account IDs (https://github.com/centrifuge/protocol-v3/pull/403)

https://github.com/centrifuge/protocol-v3/pull/355/commits/247282e10f3a90eb060e12e2fd1d866b3210dae6

This PR blocks the asset account ID from being the same as the equity/gain/loss account ID, and likewise for expense & liability account IDs. This prevents potential accounting issues.

### Holding & share snapshots (https://github.com/centrifuge/protocol-v3/pull/385)

https://github.com/centrifuge/protocol-v3/pull/355/commits/4ec9c26a75d7f17b2e6bd5e45a93102ea5e5e3de

### Ref: Custom valuation interface: IERC7726 -> IValuation (https://github.com/centrifuge/protocol-v3/pull/402)

https://github.com/centrifuge/protocol-v3/pull/355/commits/1ee99a25c2993dff71f5f065a7ed8cce203e902a

Removes the IERC7726 interface as it was a bit hacky to rely on this.

### File reorganization (spoke and vaults modules) (https://github.com/centrifuge/protocol-v3/pull/405)

https://github.com/centrifuge/protocol-v3/pull/355/commits/f22e214a5f9689ee101e458190dfb1f3cc701160

Moves `vaults` into a separate root folder, to further decouple.

### Ref: max price age messages as independent messages (https://github.com/centrifuge/protocol-v3/pull/407)

https://github.com/centrifuge/protocol-v3/pull/355/commits/7b2645601f643b3f6fdc2bee6b8f1945c17fe874

Uses separate messages for the max price setters, so `UpdateContract` is only used for higher level contracts, not the core system actions.

### Ref: split submessage files (https://github.com/centrifuge/protocol-v3/pull/408)

https://github.com/centrifuge/protocol-v3/pull/355/commits/0803b4d7910e5f845f2f93f494bf5c6d3cde647e

Splits message lib into 2 separate libs

(out of scope)

### fix: blocked requests edge case (https://github.com/centrifuge/protocol-v3/pull/398)

https://github.com/centrifuge/protocol-v3/pull/355/commits/ee717dd26534736a71ad0c78ea620bf2b7d1acf6

Fixes Spearbit issue 9.

### Ref: generic vault <-> spoke interface (https://github.com/centrifuge/protocol-v3/pull/409)

https://github.com/centrifuge/protocol-v3/pull/355/commits/27bff514918cd8a3b889b93985887d75077de767

Further decouples vaults from spoke by introducing a lighter interface.

### Organize Spoke (https://github.com/centrifuge/protocol-v3/pull/410)

https://github.com/centrifuge/protocol-v3/pull/355/commits/58618ebf345b9ccfd6210de4130744e1675bc167

Adds section headers and generally organizes the spoke contract a bit more.

### Remove priceAssetPerShare (https://github.com/centrifuge/protocol-v3/pull/411)

https://github.com/centrifuge/protocol-v3/pull/355/commits/40903a25fe10c1c2ed28765e4193b3b100965622

Remove a method that was now unused.


lemunozm:
README. Do not merge this as `squash` to avoid losing track of all commits. I would just rebase

src/misc/libraries/TransientArrayLib.sol
```diff
@@ -8,32 +8,34 @@ library TransientArrayLib {
...
 
     function push(bytes32 key, bytes32 value) internal {
-        bytes32 lengthSlot = keccak256(abi.encodePacked(key, type(uint256).max));
+        bytes32 lengthSlot = keccak256(abi.encodePacked(key));
         uint256 length_ = lengthSlot.tloadUint256();
         lengthSlot.tstore(length_ + 1);
 
-        bytes32 slot = keccak256(abi.encodePacked(key, length_));
+        uint256 baseSlot = uint256(keccak256(abi.encodePacked(key)));
+        bytes32 slot = bytes32(baseSlot + length_ + 1);
```
DrasticWatermelon: second keccak isnt necessary, why aren't we using `baseSlot = uint256(lengthSlot)`?

src/misc/libraries/TransientBytesLib.sol
```diff
@@ -10,44 +10,46 @@ library TransientBytesLib {
...
         uint256 prevLength = lengthSlot.tloadUint256();
 
         uint256 startChunk = prevLength / 32;
         uint256 offset = prevLength % 32;
 
         lengthSlot.tstore(prevLength + value.length);
 
-        bytes32 joinSlot = keccak256(abi.encodePacked(key, startChunk));
+        uint256 baseSlot = uint256(keccak256(abi.encodePacked(key)));
+        bytes32 joinSlot = bytes32(baseSlot + startChunk + 1);
```
DrasticWatermelon: second keccak can be avoided by just using `baseSlot = uint256(lengthSlot)`, same as TransientArrayLib.sol comment

src/common/Gateway.sol
```diff
@@ -285,89 +142,57 @@ contract Gateway is Auth, Recoverable, IGateway {
...
+            batchGasLimit,
+            transactionRefund != address(0) ? transactionRefund : address(subsidy[poolId].refund)
+        );
 
         return true;
     }
 
+    /// @inheritdoc IGateway
+    function addUnpaidMessage(uint16 centrifugeId, bytes memory message) external auth {
+        _addUnpaidBatch(centrifugeId, message, gasService.gasLimit(centrifugeId, message));
```
DrasticWatermelon, line 183: Are we fine with `gasService.gasLimit(centrifugeId, message)` potentially not being enough gas to execute the message? 
AFAIU the call will return a constant which should overshoot the worst case's gas consumption, but I'm wondering whether it makes sense to possibly add an override to also specify the gas limit

src/spoke/Spoke.sol
```diff
@@ -62,61 +46,59 @@ contract PoolManager is
...
         emit File(what, factory, status);
     }
 
     //----------------------------------------------------------------------------------------------
     // Outgoing methods
     //----------------------------------------------------------------------------------------------
 
-    /// @inheritdoc IPoolManager
+    /// @inheritdoc ISpoke
     function transferShares(uint16 centrifugeId, PoolId poolId, ShareClassId scId, bytes32 receiver, uint128 amount)
```
DrasticWatermelon, line 98: looks to me like if I were to provide an invalid `centrifugeId` here (i.e. one not used by the system), then the `InitiateTransferShares` message will be sent to the pool chain correctly to the pool chain which will store an `ExecuteTransferShares` msg to the chain with provided `centrifugeId`.

Given that you stated you plan on having a keeper to execute such unpaid messages automatically, I'm wondering on whether you've already foreseen the need to filter out messages targeted to such invalid chain ids in order to avoid spending tx fees on msgs that will never be sent correctly.

src/hub/ShareClassManager.sol
```diff
@@ -286,8 +294,6 @@ contract ShareClassManager is Auth, IShareClassManager {
         EpochRedeemAmounts storage epochAmounts = epochRedeemAmounts[scId_][payoutAssetId][nowRevokeEpochId];
         epochAmounts.navPoolPerShare = navPoolPerShare;
 
-        require(epochAmounts.approvedShareAmount <= metrics[scId_].totalIssuance, RevokeMoreThanIssued());
-
```
DrasticWatermelon, line 290: why is this check not needed anymore?

src/spoke/BalanceSheet.sol
```diff
@@ -0,0 +1,348 @@
...
+    /// @inheritdoc IBalanceSheetGatewayHandler
+    function updateManager(PoolId poolId, address who, bool canManage) external auth {
+        manager[poolId][who] = canManage;
+        emit UpdateManager(poolId, who, canManage);
+    }
+
+    /// @inheritdoc IBalanceSheetGatewayHandler
+    function setQueue(PoolId poolId, ShareClassId scId, bool enabled) external auth {
+        queueDisabled[poolId][scId] = !enabled;
+    }
```
DrasticWatermelon, line 281: flipping the param for a setter is awkward IMO, I'd rather have the param be called `disabled` and write it to the mapping as is

src/misc/libraries/TransientArrayLib.sol
hieronx: We are pushing to the `(i + 1)th` slot here which is separate from the length slot.

hieronx: True!

src/spoke/Spoke.sol
hieronx: The share tokens hooks can implement logic for limiting destination `centrifugeId`s for a given token. E.g. if `FullRestriction` hook is used, every target chain ID needs to be whitelisted.

I don't think this is a major issue though, since passing an invalid `centrifugeId` is basically the same as passing an invalid receiver.

src/hub/ShareClassManager.sol
hieronx: Because the total issuance is not updated synchronously anymore, rather asynchronously from the spoke side, we cannot check based on the total issuance here. It is rather on the hub manager to ensure they limit this. It is still somewhat limited since the epoch amount is limited by the incoming redemptions.

src/spoke/BalanceSheet.sol
hieronx: Yeah fair, will update.

src/spoke/Spoke.sol
DrasticWatermelon: I agree its a minor nit for the smart contract. I'm more worried with your keeper spending gas fees on txs that will never execute successfully

hieronx: Keeper should be able to have a list of valid `centrifugeId`s yeah, definitely!

src/spoke/BalanceSheet.sol
lemunozm: From the user of the API, is it not more "friendly" to use `enable` as it is the common pattern everywhere? If I read `setQueue(true)` I would expect to enable the queue not to disable it.

Actually the current name should be `enableQueue()` I think 🤔

src/common/Gateway.sol
lemunozm: > AFAIU the call will return a constant which should overshoot the worst case's gas consumption

Yes, that's the case. 

> Are we fine with gasService.gasLimit(centrifugeId, message) potentially not being enough gas to execute the message?

Under which circumstances could that happen? The gas limit depends on the message itself

src/misc/libraries/TransientArrayLib.sol
xmxanuel: The idea is to implement it in the same way as **namespaced storage** with a dynamic array would work in Solidity.

The `keccak of the key` 
- stores the length

the`keccak(keccak of the key)`
- stores the actual array elements

I wrote a PoC to illustrate
- look at  `readFirstElement` how to access it low level in assembly

```solidity
pragma solidity ^0.8.13;

import {Test, console} from "forge-std/Test.sol";

contract MyContract {
    struct StorageV0 {
        uint[] myArray;
    }

    bytes32 public constant StorageV0Location =
        bytes32(keccak256("mykey"));

    function storagev0() internal pure returns (StorageV0 storage storageV0) {
        bytes32 position = StorageV0Location;
        assembly {
            storageV0.slot := position
        }
    }

    function push(uint element) external {
        StorageV0 storage $ = storagev0();
        $.myArray.push(element);
    }

    function readFirstElement() public view returns(uint) {
        bytes32 storagePos = keccak256("mykey");
        uint firstElement;
        assembly {
            mstore(0, storagePos)
            let slot := keccak256(0x0, 0x20)
            firstElement := sload(slot)
        }
        return firstElement;
    }
} 

contract MyContractTest is Test {
    MyContract public myContract;

    function setUp() public {
        myContract = new MyContract();
    }

    function testPush() public {
        uint element = 42;
        myContract.push(element);
        assertEq(myContract.readFirstElement(), 42);
    }
}
```

xmxanuel: I mean, we could still argue if this is the best way or a bit too complicated here.

src/spoke/BalanceSheet.sol
hieronx: `setQueue(true)` would enable it, `setQueue(false)` would disable it. I actually think that is preferred, since otherwise we need two methods, `enableQueue()` and `disableQueue()`

lemunozm: Sorry I wanted to mean `allowQueue(bool enable)` 😆

src/common/Gateway.sol
DrasticWatermelon: I dont have a concrete example in mind but it'd be via remotely likely stuff like opcode reprices

src/misc/libraries/TransientArrayLib.sol
DrasticWatermelon: Yeah I just wanted to point out an extra optimization available given we're already optimizing the lib.

I like the idea of having transient arrays laid out in the same way as Solidity would (tbh it makes me a little angry at solc that we dont have `transient` applicable to reference types), and the impl LGTM.

tbh I dont see any particular gain in using a non-standard layout a part from possibly not having to execute a hash but then we'd need a collision-safe map from a key to a transient storage slot

hieronx: Fixed in https://github.com/centrifuge/protocol-v3/pull/355/commits/a75f1d9300674958b4e8aaa54af225b1cbaf4ec8

src/misc/libraries/TransientBytesLib.sol
hieronx: Fixed in https://github.com/centrifuge/protocol-v3/pull/355/commits/a75f1d9300674958b4e8aaa54af225b1cbaf4ec8

src/spoke/BalanceSheet.sol
hieronx: I'm not sure why we'd use the term allow rather than set, which we are using everywhere else?

src/common/Gateway.sol
lemunozm: I think the same issue already exists with the failing messages. As far as I remember, we assume the gasLimit is high enough to handle future changes.

src/spoke/BalanceSheet.sol
lemunozm: I've seen that pattern when setting a boolean in other projects. But true, we're not using it now (I remember previous it was in the registry/hub). Let's go as it is now

PR #357, opened 2025-05-07
[Boat coding]: estimation by action

lemunozm:
Estimating by sent message is very dificult to use from the outside because there is no clue of which messages are sent for a given action by the UX. Instead here we map by action.

### DIFF
--- a/src/common/interfaces/IGasService.sol
+++ b/src/common/interfaces/IGasService.sol
@@ -1,6 +1,11 @@
 // SPDX-License-Identifier: BUSL-1.1
 pragma solidity >=0.5.0;
 
+struct MessageCost {
+    uint32 messageSize;
+    uint128 gasCost;
+}
+
 interface IGasService {
     /// @notice Gas limit for the execution cost of an individual message in a remote chain.
     /// @dev    NOTE: In the future we could want to dispatch:
@@ -16,4 +21,6 @@ interface IGasService {
     /// @param centrifugeId Where to the cost is defined
     /// @return Max cost in WEI units
     function maxBatchSize(uint16 centrifugeId) external view returns (uint128);
+
+    function actionCost(bytes4 action) external view returns (uint32, uint128);
 }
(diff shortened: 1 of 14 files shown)


src/common/interfaces/IGasService.sol
```diff
@@ -1,6 +1,11 @@
 // SPDX-License-Identifier: BUSL-1.1
 pragma solidity >=0.5.0;
 
+struct MessageCost {
+    uint32 messageSize;
+    uint128 gasCost;
+}
```
lemunozm, line 7: FYI, I can not know the message size from the code, because some messages has dynamic size based on its content. This is why I need to add the size per action.

src/common/interfaces/IGateway.sol
```diff
@@ -221,10 +221,10 @@ interface IGateway is IMessageHandler, IMessageSender, IGatewayHandler {
     /// @notice Used to calculate overall cost for bridging a payload on the first adapter and settling
     ///         on the destination chain and bridging its payload proofs on n-1 adapter
     ///         and settling on the destination chain.
-    /// @param  payload Used in gas cost calculations.
+    /// @param  actions Used in gas cost calculations.
     /// @dev    Currenly the payload is not taken into consideration.
     /// @return total Total cost for sending one message and corresponding proofs on through all adapters
-    function estimate(uint16 centrifugeId, bytes calldata payload) external view returns (uint256 total);
+    function estimate(uint16 centrifugeId, bytes4[] memory actions) external view returns (uint256 total);
```
lemunozm, line 227: I think sadly bytes4 is not enough. For example, for `Hub.updateContract.selector` I do not know the message size/cost associated to it because it's dynamic. 

By now, we can assume the worst case, but maybe we can also parameterize and use some tuple `(bytes4, kind)` to specify some special behavior in the action.

lemunozm: I think is not horrible to assume the worst case, because the unused gas is refunded

lemunozm:
Close in favor of https://github.com/centrifuge/protocol-v3/pull/366

PR #358, opened 2025-05-08
Remove recovery messages

hieronx:


lemunozm (commented):
Like it so much!

I think we can remove `IGatewayHandler` too

src/common/MessageProcessor.sol
```diff
@@ -51,8 +49,7 @@ contract MessageProcessor is Auth, IMessageProcessor {
 
     /// @inheritdoc IMessageProcessor
     function file(bytes32 what, address data) external auth {
-        if (what == "gateway") gateway = IGatewayHandler(data);
```
lemunozm, line 54: NOTE: The current failing tests are because of this I guess, not sure why CI is not run

lemunozm (approved):
LGTM!

lemunozm (approved):

PR #359, opened 2025-05-09
Remove `owner` param from `BalanceSheet.deposit`

hieronx:


lemunozm (approved):
Curious, does it comes from some review issue?

hieronx:
> Curious, does it comes from some review issue?

Technically not a security issue but it is an integration footgun: https://github.com/xmxanuel/centrifuge-protocol-v3-may-review/pull/3#discussion_r2079687984

PR #360, opened 2025-05-09
Simplify ERC20, balance sheet, and misc cleanups

hieronx:


src/misc/ERC20.sol
```diff
@@ -49,35 +43,38 @@ contract ERC20 is Auth, IERC20Metadata, IERC20Permit {
...
+
     /// @inheritdoc IERC20
     function balanceOf(address user) public view virtual returns (uint256) {
         return _balanceOf(user);
     }
 
-    function _setBalance(address user, uint256 value) internal virtual {
-        balances[user] = value;
+    function _balanceOf(address user) internal view virtual returns (uint256) {
+        return balances[user];
```
lemunozm, line 67: Both are exactly the same, maybe the internal one is not need it?

hieronx: It is because it's `virtual` and overridden in `ShareToken.sol`

src/common/MessageDispatcher.sol
```diff
@@ -641,7 +641,6 @@ contract MessageDispatcher is Auth, IMessageDispatcher {
         PoolId poolId,
         ShareClassId scId,
         AssetId assetId,
-        address provider,
```
hieronx, line 644: This param seems unused entirely. @mustermeiszer do you agree it can be removed?

src/common/interfaces/IGatewayHandlers.sol
```diff
@@ -274,14 +274,8 @@ interface IRequestManagerGatewayHandler {
         external;
 }
 
-/// @notice Interface for CV methods related to epoch called by messages
+/// @notice Interface for Vaults methods related to epoch called by messages
 interface IBalanceSheetGatewayHandler {
-    function triggerDeposit(PoolId poolId, ShareClassId scId, AssetId assetId, address provider, uint128 amount)
```
hieronx, line 279: These were already unused

src/common/libraries/PricingLib.sol
```diff
@@ -238,26 +238,19 @@ library PricingLib {
...
-    function toPriceDecimals(uint128 _value, uint8 decimals) private pure returns (uint256) {
+    ///      NOTE: We require all assets to have 2 <= decimals <= 18, see `PoolManager.registerAsset`
+    function _toPriceDecimals(uint128 _value, uint8 decimals) private pure returns (uint256) {
         if (PRICE_DECIMALS == decimals) return uint256(_value);
         return uint256(_value) * 10 ** (PRICE_DECIMALS - decimals);
     }
-
-    /// @dev Converts decimals of the value from the price decimals back to the intended decimals
-    /// @dev NOTE: We require all assets to have 2 <= decimals <= 18, see `PoolManager.registerAsset`
-    function fromPriceDecimals(uint256 _value, uint8 decimals) private pure returns (uint128) {
```
hieronx, line 259: This was unused

src/vaults/BalanceSheet.sol
```diff
@@ -126,27 +141,48 @@ contract BalanceSheet is Auth, Recoverable, IBalanceSheet, IBalanceSheetGatewayH
...
     /// @inheritdoc IBalanceSheet
     function revoke(PoolId poolId, ShareClassId scId, address from, uint128 shares) external authOrManager(poolId) {
-        _noteRevoke(poolId, scId, from, shares);
-        _executeRevoke(poolId, scId, from, shares);
-    }
+        emit Revoke(poolId, scId, from, _pricePoolPerShare(poolId, scId), shares);
 
-    /// @inheritdoc IBalanceSheet
-    /// @dev This function is mostly useful to keep higher level integrations CEI adherent.
-    function noteRevoke(PoolId poolId, ShareClassId scId, address from, uint128 shares)
```
hieronx, line 145: @mustermeiszer I don't think `noteRevoke` can work, since it would require the manager to be able to burn shares aftewrads, which is not possible. I also think for CEI compatibility, managers should just be able to call `revoke` at the end of their tx. I would remove it, do you agree?

src/vaults/BalanceSheet.sol
```diff
@@ -97,23 +99,36 @@ contract BalanceSheet is Auth, Recoverable, IBalanceSheet, IBalanceSheetGatewayH
         external
         authOrManager(poolId)
     {
-        AssetId assetId = poolManager.assetToId(asset, tokenId);
-        _noteDeposit(poolId, scId, assetId, asset, tokenId, msg.sender, amount);
-        _executeDeposit(poolId, asset, tokenId, msg.sender, amount);
+        noteDeposit(poolId, scId, asset, tokenId, msg.sender, amount);
```
hieronx, line 102: With the simplifications in the rest, I would say the internal methods are rather misdirection, and it is simpler to read to inline everything.

src/vaults/BalanceSheet.sol
```diff
@@ -178,42 +214,27 @@ contract BalanceSheet is Auth, Recoverable, IBalanceSheet, IBalanceSheetGatewayH
     // View methods
     //----------------------------------------------------------------------------------------------
 
+    /// @inheritdoc IBalanceSheet
+    function escrow(PoolId poolId) public view returns (IPoolEscrow) {
```
hieronx, line 218: This seems useful for balance sheet managers, to know e.g. where to transfer after noteDeposit

src/vaults/token/ShareToken.sol
```diff
@@ -42,7 +38,7 @@ contract ShareToken is ERC20, IShareToken {
     //----------------------------------------------------------------------------------------------
 
     /// @inheritdoc IShareToken
-    function file(bytes32 what, address data) external authOrHook {
+    function file(bytes32 what, address data) external auth {
```
hieronx, line 41: Fixes https://github.com/xmxanuel/centrifuge-protocol-v3-may-review/pull/14#discussion_r2079922517

src/common/Gateway.sol
```diff
@@ -70,7 +70,8 @@ contract Gateway is Auth, Recoverable, IGateway {
         root = root_;
         gasService = gasService_;
 
-        setRefundAddress(GLOBAL_POT, IRecoverable(address(this)));
+        subsidy[GLOBAL_POT].refund = IRecoverable(address(this));
```
hieronx, line 73: Fixes Cantina's comment.

lemunozm: Thanks for fixing this!

src/misc/ERC20.sol
lemunozm: Curious, why `_balanceOf()` is not public in `ShareToken.sol`?

src/vaults/BalanceSheet.sol
```diff
@@ -126,27 +141,48 @@ contract BalanceSheet is Auth, Recoverable, IBalanceSheet, IBalanceSheetGatewayH
...
-    function issue(PoolId poolId, ShareClassId scId, address to, uint128 shares) external authOrManager(poolId) {
-        _issue(poolId, scId, to, shares);
+    function issue(PoolId poolId, ShareClassId scId, address to, uint128 shares) public authOrManager(poolId) {
+        emit Issue(poolId, scId, to, _pricePoolPerShare(poolId, scId), shares);
+
+        if (queueEnabled[poolId][scId]) {
+            queuedShares[poolId][scId].increase += shares;
+        } else {
+            sender.sendUpdateShares(poolId, scId, shares, true);
+        }
```
lemunozm, line 167: Am I correctly assuming that all methods in BalanceSheet will always be subsidized? If not, this adds some complexity to the estimation that we should take care of.

src/vaults/token/ShareToken.sol
lemunozm: 😲 Does this come from LP?

lemunozm (approved):
Amazing simplification! Spetially the BalanceSheet much more easier to follow

src/vaults/BalanceSheet.sol
hieronx: We can implement balance sheet managers that can pay directly

src/misc/ERC20.sol
hieronx: Since users can use `balanceOf()` for the same purpose

src/vaults/BalanceSheet.sol
lemunozm: Then will be difficult to estimate the required gas 🤔

hieronx: True, although that is really an issue  for a higher layer. Maybe using subsidized funds for this makes more sense.

lemunozm: Yeah, I think it should. Another issue with integrators and let the pay directly is that they need to call to `gateway.payTransaction()` which is Auth. So right now, only contracts handled by us can be paid in the same transaction

src/common/MessageDispatcher.sol
mustermeiszer: Yes

src/vaults/BalanceSheet.sol
mustermeiszer: Good point! Then lets remove the `from` from the `revoke` too, like for deposits and assume the manager is the holder. Wdyt?

hieronx: Good point!

hieronx: This was actually a bit tricky, since it meant we need to first transfer shares to `asyncRequestManager` in `revokedShares` before they can be revoked. Which means `asyncRequestManager` needs to be endorsed.

lemunozm (approved):
LGTM!

PR #361, opened 2025-05-09
Allow holding updates before initialization

hieronx:


mustermeiszer (approved):
I like the change. Creates a lot more flexibility on the vaults side. 

NIT
* Wondering whether we should saturate `holding.assetAmountValue` at `0` instead of creating an underflow error?

src/hub/Holdings.sol
```diff
@@ -147,19 +149,20 @@ contract Holdings is Auth, IHoldings {
     // View methods
     //----------------------------------------------------------------------------------------------
 
+    /// @inheritdoc IHoldings
+    function initialized(PoolId poolId, ShareClassId scId, AssetId assetId) external view returns (bool) {
```
lemunozm, line 153: (Post-morten) NIT. I think predicates are easier to read as with `is` prefix, i.e: `isInitialized()`. At least, this is what I used for other methods in the codebase.

hieronx: Fair! Let me add that to another PR

PR #362, opened 2025-05-10
Remove vault managers as wards on share tokens

hieronx:


src/vaults/PoolManager.sol
```diff
@@ -369,29 +369,33 @@ contract PoolManager is
...
+        ShareClassDetails storage shareClass = _shareClass(poolId, scId);
 
         AssetIdKey memory assetIdKey = _idToAsset[assetId];
 
         IBaseRequestManager manager = vault.manager();
         manager.addVault(poolId, scId, vault, assetIdKey.asset, assetId);
-
         _vaultDetails[vault].isLinked = true;
 
+        IAuth(address(shareClass.shareToken)).rely(address(vault));
```
hieronx, line 380: Unless I miss something, I believe it is a lot cleaner to set this up here in the pool manager, as this removes the vault managers as wards from the share tokens entirely.

src/vaults/PoolManager.sol
```diff
@@ -404,13 +408,10 @@ contract PoolManager is
         ShareClassDetails storage shareClass = _shareClass(poolId, scId);
         require(vaultFactory[factory], InvalidFactory());
 
-        // Rely investment manager on vault so it can mint tokens
```
hieronx, line 407: Incorrect comment now

lemunozm (approved):
Another cool simplification!

PR #364, opened 2025-05-12
feat: Automated multi-chain deployment

mustermeiszer:


.gitignore
```diff
@@ -1,7 +1,7 @@
 cache/
 out/
 broadcast/
-.env
```
lemunozm, line 4: I think we should still keep `.env` to avoid pushing some private keys by error

script/CommonDeployer.s.sol
```diff
@@ -35,6 +35,7 @@ abstract contract CommonDeployer is Script, JsonRegistry {
     Gateway public gateway;
     MessageProcessor public messageProcessor;
     MessageDispatcher public messageDispatcher;
+    uint16 public localCentrifugeId;
```
lemunozm, line 38: Maybe we can initialize this by passing it through the constructor?

script/CommonDeployer.s.sol
```diff
@@ -60,9 +61,9 @@ abstract contract CommonDeployer is Script, JsonRegistry {
         messageProcessor = new MessageProcessor(root, tokenRecoverer, deployer);
 
         gasService = new GasService(maxBatchSize, messageGasLimit);
-        gateway = new Gateway(centrifugeId, root, gasService, deployer);
+        gateway = new Gateway(localCentrifugeId, root, gasService, deployer);
```
lemunozm, line 64: We're assuming here that another previous code will initialize `localCentrifugeId` before calling `deployCommon()`, creating an "invisible" side effect. Why not pass the `localCentrifugeId` as a parameter to express that relation?

script/FullDeployer.s.sol
```diff
@@ -8,9 +8,21 @@ import {HubDeployer} from "script/HubDeployer.s.sol";
...
+        removeFullDeployerAccess(msg.sender);
+        saveDeploymentOutput();
+
+        vm.stopBroadcast();
+    }
+
+
+    function deployFull(ISafe adminSafe_, address deployer, bool isTests) public {
+        deployHub(adminSafe_, deployer, isTests);
+        deployVaults(adminSafe_, deployer, isTests);
```
lemunozm, line 25: EndToEnd tests will call deployFull too, and we'll also need to initialize `localCentrifugeId`. We can move line 12 here.

Although I think passing it through a parameter express better that this method needs a centrifugeId

script/adapters/Axelar.s.sol
```diff
@@ -8,21 +8,12 @@ import {FullDeployer} from "script/FullDeployer.s.sol";
 
 // Script to deploy CP and CP with an AxelarScript Adapter.
 contract AxelarDeployer is FullDeployer {
-    function run() public {
-        uint16 localCentrifugeId = uint16(vm.envUint("CENTRIFUGE_ID"));
+    function deploy() public {
```
lemunozm, line 11: CR. I think we can not inherit from `FullDeployer`.

`wire` is assuming someone has initializate the contracts, but from this script perspective, everything is 0. 

I think we need code wire here, or pass the correct `gateway` and `root` addresses by env

deployments/deployMulti.sh
```diff
@@ -0,0 +1,132 @@
...
+
+################################################################################
+# 5 ▸ Deployment loop (iterate only over networks)
+################################################################################
+for NET in "${NETWORKS[@]}"; do
+  IFS='|' read -r RPCURL CHAINID <<< "$(network_params "$NET")"
+
+  echo "🚀 Deploying adapters [${ADAPTERS[*]}] to [$NET]  (test=$TEST broadcast=$BROADCAST)"
+
+  forge script "script/adapters/${DEPLOYER}.s.sol:${DEPLOYER}" \
```
lemunozm: CR. This will ending up as `script/adapters/FullDeployer"` but `FullDeployer` is not in `adapters`.

I think you need to deploy first the full deployer, and later the adapters, isn't?

lemunozm (commented):
I like the super clean deploy script!

Some CRs bellow

script/CommonDeployer.s.sol
mustermeiszer: Don't we do that all over the place?

lemunozm: Not for external values that come from env values.

PR #365, opened 2025-05-12
fix vault claim-redeem payment

lemunozm:
Calling `vault.withdraw` ends up calling `BalanceSheet.withdraw`, which can send a message. 

Not sure how we can track easily this. Seems like a future addition can send a new message without being able to link with the external API to perform the correct payment.


hieronx:
I would say this should either use subsidized funds, or queueing should be enabled and then payment is not needed.

lemunozm:
How can we ensure queuing is enabled before the investor call to `claimRedeem()`?

> I would say this should either use subsidized funds, or queueing should be enabled and then payment is not needed.

This reasoning should be applied to any other vault method where we currently use `_pay()`? I do not really know right now which method should:
- Be paid in the transaction
- Be subsidized
- Be queuing

lemunozm:
i.e. why `claimRedeem()` should be subsidized but `cancelRedeemRequest()` not?

hieronx:
> i.e. why claimRedeem() should be subsidized but cancelRedeemRequest() not?

Difference is that it's subsidizing the pool update (balance sheet call), not the investor specific request. If the pool manager wants these to not be subsidized, they should enable queueing.

lemunozm:
Then I think we can close it. Although I am still thinking about how we can have these statements strongly defined. Right now is very indirect 🤔

lemunozm:
> I would say this should either use subsidized funds, or queueing should be enabled and then payment is not needed

@hieronx, I was thinking more on this, that collides with another issue I have in other PR. 

What if the FM, forgets to call `setQueue` on the pool or just disable it? I think `VaultRouter` method should never use subsidized pool by default. For me it makes sense to add `pay()` here to protect agains that. Note, that it doesn't mean the user will end up paying for the message. If `setQueue()` is true, then the user will not pay anything, which is the normal behavior. 

To be in accordance with other methods, `VaultRouter.deposit()` is also calling a BalanceSheet message under the hood, but the payment is on top of the caller, as expected.

hieronx:
Fair enough, I see your point @lemunozm !

wischli (approved):

PR #366, opened 2025-05-13
Ref/remove gateway estimation

lemunozm:
### Problem

Gateway estimation is complex because:
1. Each adapter has different overhead.
2. The estimation cost depends on the message size.
3. The estimation depends on the target `centrifugeId`
4. We do not know from the UX which messages will be dispatched for an action. i.e. depending on the state, some action can send one or several messages.
5. Everything can be batched.

### Solution
- Having a `gateway.estimate(message)` is useless given points 4 and 5.
- Using `gateway.estimate(action)` as in https://github.com/centrifuge/protocol-v3/pull/357 also doesn't really help due 4.

The only way to estimate correctly is to simulate the tx/method in the node and check the payment to the adapter. The SDK should offer this.

For contract integration, we need to assume the contract has some extra funds and it overpays to ensure all actions always succeed, expecting to be refunded asynchronously


hieronx (approved):

wischli (approved):
Tenderly to the rescue!

lemunozm:
I will merge this. I will rescue the gateway estimation process in case we finally need it for the transfer shares

PR #368, opened 2025-05-13
add source centrifuge chain utility to messages

lemunozm:
Checks extracted from MessageProcessor

### DIFF
--- a/src/common/MessageProcessor.sol
+++ b/src/common/MessageProcessor.sol
@@ -7,7 +7,7 @@ import {Auth} from "src/misc/Auth.sol";
 import {D18, d18} from "src/misc/types/D18.sol";
 import {IRecoverable} from "src/misc/interfaces/IRecoverable.sol";
 
-import {MessageType, MessageLib, MessageDirection} from "src/common/libraries/MessageLib.sol";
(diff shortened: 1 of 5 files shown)


src/common/MessageProcessor.sol
```diff
@@ -68,11 +68,9 @@ contract MessageProcessor is Auth, IMessageProcessor {
     /// @inheritdoc IMessageHandler
     function handle(uint16 centrifugeId, bytes calldata message) external auth {
         MessageType kind = message.messageType();
+        uint16 sourceCentrifugeId = message.messageSourceCentrifugeId();
 
-        if (message.messageDirection() == MessageDirection.HubToVaults) {
-            // Hub => Vaults messages should only be able to interact with pools from that Hub chain
-            require(message.messagePoolId().centrifugeId() == centrifugeId, InvalidSourceChain());
-        }
+        require(sourceCentrifugeId == 0 || sourceCentrifugeId == centrifugeId, InvalidSourceChain());
```
lemunozm, line 70: It means: if there is some centrifugeId attached to the message as source one, then check against the adapter one

hieronx (approved):

PR #369, opened 2025-05-14
refactor: merged fulfilled messages

wischli:
Removes `FulfilledCancel{Deposit, Redeem}Request` by merging the cancelled amount into the corresponding `Fulfilled{Deposit, Redeem}Request` messages

As a result, we are back at having only 32 messages.


src/common/libraries/MessageLib.sol
```diff
@@ -108,21 +106,16 @@ library MessageLib {
...
         (73  << uint8(MessageType.TriggerIssueShares) * 8) +
-        (25  << uint8(MessageType.TriggerSubmitQueuedShares) * 8);
-
-    // forgefmt: disable-next-item
-    uint256 constant MESSAGE_LENGTHS_2 =
-        (41 << (uint8(MessageType.TriggerSubmitQueuedAssets) - 32) * 8) +
-        (26 << (uint8(MessageType.SetQueue) - 32) * 8);
+        (25  << uint8(MessageType.TriggerSubmitQueuedShares) * 8) +
+        (41 << uint8(MessageType.TriggerSubmitQueuedAssets) * 8) +
+        (26 << uint8(MessageType.SetQueue) * 8);
```
wischli, line 118: Leaving a breadcrumb here for when we exceed 32 messages again in the future

src/common/interfaces/IGatewayHandlers.sol
```diff
@@ -158,102 +158,41 @@ interface IRequestManagerGatewayHandler {
...
-    ) external;
-
-    /// @notice Fulfills deposit request cancellation after successful epoch execution on Hub.
-    ///         The amount of assets that can be claimed by the user is locked in the escrow contract.
-    ///         Updates claimableCancelDepositRequest bookkeeping value. The cancellation order execution can be
-    ///         partial.
-    /// @dev    The assets in the escrow are reserved for the user and are transferred to the user during
-    ///         claimCancelDepositRequest calls.
-    ///         `fulfillment` represents the decrease in `pendingDepositRequest`.
-    ///         This is a separate parameter from `assets` since there can be some precision loss when calculating this,
```
wischli, line 181: Artifact of v2 Centrifuge Chain implementation that luckily doesn't hold true for v3 (see [Slack thread](https://kflabs.slack.com/archives/C07PG2EUR9C/p1736937649220279?thread_ts=1736933248.761209&cid=C07PG2EUR9C)). With v3, we have always returned the exact same value here because we "swap" only during claiming at the price which was set during the approval of the corresponding epoch. So `fulfillment` and `assets` are always the same.

src/vaults/AsyncRequestManager.sol
```diff
@@ -202,22 +202,34 @@ contract AsyncRequestManager is BaseRequestManager, IAsyncRequestManager {
...
+        uint128 fulfilledShares,
+        uint128 cancelledAssets
     ) public auth {
         IAsyncVault vault_ = IAsyncVault(address(vault[poolId][scId][assetId]));
-
         AsyncInvestmentState storage state = investments[vault_][user];
+
         require(state.pendingDepositRequest != 0, NoPendingRequest());
+        if (cancelledAssets > 0) {
+            require(state.pendingCancelDepositRequest == true, NoPendingRequest());
```
wischli, line 214: Not sure if we want to keep this check at all. In practice, this should never throw. Curious about your opinion.

src/vaults/AsyncRequestManager.sol
```diff
@@ -226,64 +238,35 @@ contract AsyncRequestManager is BaseRequestManager, IAsyncRequestManager {
...
-    }
-
-    /// @inheritdoc IRequestManagerGatewayHandler
-    function fulfillCancelDepositRequest(
-        PoolId poolId,
-        ShareClassId scId,
-        address user,
-        AssetId assetId,
-        uint128 assets,
-        uint128 fulfillment
```
wischli, line 256: See [above](https://github.com/centrifuge/protocol-v3/pull/369#discussion_r2088533463) for my explanation why both parameters are represented by the same value in v3

src/vaults/AsyncRequestManager.sol
hieronx: No reason to not have it imo

hieronx: It could throw in case of message reordering.

src/common/interfaces/IGatewayHandlers.sol
hieronx: Really nice improvement actually!

hieronx (approved):
Turned out to be a great cleanup!

src/vaults/AsyncRequestManager.sol
wischli: > It could throw in case of message reordering.

I don't see how that's possible for this particular case because `state.pendingCancelDepositRequest == false` means there cannot be any pending cancellation request for the user anywhere in the system. Once this value is set to `true`, there will be exactly one fulfilled message with non-zero cancelled amount.

Independent of that, I agree there is now downside in keeping it.

src/vaults/AsyncRequestManager.sol
```diff
@@ -226,64 +235,36 @@ contract AsyncRequestManager is BaseRequestManager, IAsyncRequestManager {
...
+        uint128 fulfilledShares,
+        uint128 cancelledShares
     ) public auth {
         IAsyncRedeemVault vault_ = IAsyncRedeemVault(address(vault[poolId][scId][assetId]));
 
         AsyncInvestmentState storage state = investments[vault_][user];
         require(state.pendingRedeemRequest != 0, NoPendingRequest());
 
+        if (cancelledShares > 0) {
+            require(state.pendingCancelRedeemRequest == true, NoPendingRequest());
```
lemunozm: Ultra NIT.
```suggestion
            require(state.pendingCancelRedeemRequest, NoPendingRequest());
```

lemunozm (approved):
Such a good simplification! 💯 Definitely having less messages helps in the understanding

src/common/libraries/MessageLib.sol
hieronx: https://github.com/centrifuge/protocol-v3/pull/345/files will make it go to #33 btw

PR #370, opened 2025-05-14
Fix: start/end transactional payment

lemunozm:
Currently, only one message can be sent inside a transactional payment, because the fuel is reset after sending the first message. This PR modifies this behavior to allow multiple messages

### DIFF
--- a/src/hub/Hub.sol
+++ b/src/hub/Hub.sol
@@ -22,6 +22,7 @@ import {IHubRegistry} from "src/hub/interfaces/IHubRegistry.sol";
 import {IShareClassManager} from "src/hub/interfaces/IShareClassManager.sol";
 import {IHoldings, HoldingAccount} from "src/hub/interfaces/IHoldings.sol";
 import {IHub, AccountType} from "src/hub/interfaces/IHub.sol";
+import {IHubHelpers} from "src/hub/interfaces/IHubHelpers.sol";
 
 /// @title  Hub
 /// @notice Central pool management contract, that brings together all functions in one place.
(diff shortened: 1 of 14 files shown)


lemunozm:
Failing due hub size

lemunozm:
Note for myself. Missing endTransactionPayment UTs

src/hub/Hub.sol
```diff
@@ -54,6 +54,12 @@ contract Hub is Multicall, Auth, Recoverable, IHub, IHubGatewayHandler, IHubGuar
         gateway = gateway_;
     }
 
+    modifier payTransaction() {
```
wischli, line 61: FYI: Quick neutral summary of deployed byte code size of the Hub across our open PRs:
* `spearbit-fixreview`: 24309
* Hub Helper PR: 21183 -> 3126 decrease
* This PR: 25187 -> 878 increase

src/hub/Hub.sol
```diff
@@ -78,13 +84,14 @@ contract Hub is Multicall, Auth, Recoverable, IHub, IHubGatewayHandler, IHubGuar
         bool wasBatching = gateway.isBatching();
         if (!wasBatching) {
             gateway.startBatching();
-            gateway.payTransaction{value: msg.value}(msg.sender);
+            gateway.startTransactionPayment{value: msg.value}(msg.sender);
```
wischli, line 91: Sorry for the dumb question but can you quickly explain to me how the payment behaves for batches?

wischli (approved):
Great improvement!

src/hub/Hub.sol
lemunozm: In the same exact way. The only difference, is that the gasLimit used for the adapter estimation is a multiple of messages in the batch. Because having a batch would mean a lot of computation in the other side.

PR #372, opened 2025-05-14
MultiAdapter

lemunozm:
The base idea is to abstract the gateway to the fact that there can be different adapters. The gateway will only act with one adapter that will compose the others.

MultiAdapter:
- Compose adapters offering just one adapter to the gateway to deal with
- Knows about payloads and proofs
- The recovery system is attached to it because it belongs to the proof system (the `Guardian` should point now to `MultiAdapter`)

Gateway:
- Knows about payments, batches, failing messages and repayments.
- Just deal with one single adapter.

PROS:
- Easier to understand
- Easier to migrate, we can migrate just one of them


src/common/Gateway.sol
```diff
@@ -286,78 +151,35 @@ contract Gateway is Auth, Recoverable, IGateway {
...
-                data.gasCost[i] = adapters_[i].estimate(
-                    centrifugeId,
-                    i == PRIMARY_ADAPTER_ID - 1 ? batch : data.batchHash.serializeMessageProof(),
-                    batchGasLimit
-                );
-
-                total += data.gasCost[i];
-            }
+        bytes32 batchHash = keccak256(batch);
+        uint256 cost = adapter.estimate(centrifugeId, batch, batchGasLimit);
```
lemunozm: It's not easy to see in the diff, but finally the send method is easy to read 😭

src/common/Gateway.sol
```diff
@@ -132,71 +82,12 @@ contract Gateway is Auth, Recoverable, IGateway {
...
-        require(adapter.id != 0, InvalidAdapter());
-
-        // Verify adapter and parse message hash
-        bytes32 batchHash;
-        bool isMessageProof = payload.toUint8(0) == MessageProofLib.MESSAGE_PROOF_ID;
-        if (isMessageProof) {
-            require(adapter.id != PRIMARY_ADAPTER_ID, NonProofAdapter());
+    /// @inheritdoc IMessageHandler
+    function handle(uint16 centrifugeId, bytes memory batch) public pauseable {
+        require(address(adapter) == msg.sender, InvalidAdapter());
```
hieronx: I would make this method just `auth`, and remove the hardcoded `adapter` check

src/common/adapters/MultiAdapter.sol
```diff
@@ -0,0 +1,226 @@
...
+    // Outgoing
+    //----------------------------------------------------------------------------------------------
+
+    /// @inheritdoc IAdapter
+    function send(uint16 centrifugeId, bytes calldata payload, uint256 gasLimit, address refund)
+        external
+        payable
+        returns (bytes32)
+    {
+        require(msg.sender == address(gateway), NotGateway());
```
hieronx: Likewise, I would make this method just `auth`, and remove the hardcoded `gateway` check

src/common/adapters/MultiAdapter.sol
```diff
@@ -0,0 +1,226 @@
...
+    uint8 public constant PRIMARY_ADAPTER_ID = 1;
+    uint256 public constant RECOVERY_CHALLENGE_PERIOD = 7 days;
+
+    uint16 public immutable localCentrifugeId;
+    IMessageHandler public immutable gateway;
+
+    mapping(uint16 centrifugeId => IAdapter[]) public adapters;
+    mapping(uint16 centrifugeId => mapping(IAdapter adapter => Adapter)) internal _activeAdapters;
+    mapping(uint16 centrifugeId => mapping(bytes32 batchHash => Inbound)) public inbound;
+    mapping(uint16 centrifugeId => mapping(IAdapter adapter => mapping(bytes32 painbound => uint256 timestamp))) public
```
wischli: Nit: I don't understand the choice for using the word `painbound`. Could you explain? The previous `payloadHash` makes more sense to me.

src/common/Gateway.sol
```diff
@@ -29,23 +27,16 @@ import {MessageProofLib} from "src/common/libraries/MessageProofLib.sol";
...
-    uint256 public constant RECOVERY_CHALLENGE_PERIOD = 7 days;
     bytes32 public constant BATCH_LOCATORS_SLOT = bytes32(uint256(keccak256("Centrifuge/batch-locators")) - 1);
 
-    uint16 public immutable localCentrifugeId;
-
     // Dependencies
     IRoot public immutable root;
     IGasService public gasService;
     IMessageProcessor public processor;
+    IAdapter public adapter;
```
wischli, line 39: Seems to me like we are missing `file` support for the adapter as its not immutable?

src/common/Gateway.sol
```diff
@@ -286,78 +151,35 @@ contract Gateway is Auth, Recoverable, IGateway {
...
             } else {
-                emit SendProof(
-                    centrifugeId,
-                    data.payloadId,
-                    data.batchHash,
-                    adapters_[j],
-                    adapterData
-                );
+                underpaid[centrifugeId][batchHash].counter++;
+                underpaid[centrifugeId][batchHash].gasLimit = batchGasLimit;
```
wischli: Nit: AFAICT, we can save gas with
```suggestion
                Underpaid storage underpaid_ = underpaid[centrifugeId][batchHash];
                underpaid_.counter++;
                underpaid_.gasLimit = batchGasLimit;
```

src/common/adapters/MultiAdapter.sol
```diff
@@ -0,0 +1,226 @@
...
+            bytes32 adapterData = adapters_[i].send{value: cost}(centrifugeId, adapterPayload, gasLimit, refund);
+
+            if (i == PRIMARY_ADAPTER_ID - 1) {
+                emit SendBatch(centrifugeId, payloadId, payload, adapters_[i], adapterData, refund);
+            } else {
+                emit SendProof(centrifugeId, payloadId, hash, adapters_[i], adapterData);
+            }
+        }
+
+        return bytes32(0);
```
wischli, line 204: Nit Q: Can we drop this return as it always returns the same empty zero bytes32?

src/common/adapters/MultiAdapter.sol
```diff
@@ -0,0 +1,226 @@
...
+    }
+
+    /// @inheritdoc IAdapter
+    function estimate(uint16 centrifugeId, bytes calldata payload, uint256 gasLimit)
+        public
+        view
+        returns (uint256 total)
+    {
+        IAdapter[] memory adapters_ = adapters[centrifugeId];
```
wischli, line 213: Curious why we estimate for all and not just the active adapters.

wischli (commented):
Incredible work @lemunozm :rocket: When I review your Gateway & Adapter code, I always feel like a noob.

src/common/adapters/MultiAdapter.sol
lemunozm: Maybe the name here is a bit confusing, and we should rename this. `adapters` and `activeAdapters` are the same. One is a list to be able to iterate and the other is a map to be able to check against

src/common/adapters/MultiAdapter.sol
lemunozm: The problem here is that it implements a method defined in `IAdapter` that returns that value.

@hieronx could you extend a bit more about `adapterData`? Do we really need it? It's only used in one event, and only wormhole supports that.

src/common/Gateway.sol
lemunozm: Correct! Thanks for pointing this out, now we can do that. `send()` was pretty unreadable because we reached 16 stack variables long ago.

src/common/adapters/MultiAdapter.sol
lemunozm: It was a typo! Will fix this

src/common/adapters/MultiAdapter.sol
lemunozm: I've mimicked here the Adapter behavior in Axelar or Wormhole, which uses the same check.

Happy to change it, is there any reason you're thinking about this change?

lemunozm:
> I always feel like a noob.

Me too in my own code! 😆 was pretty difficult to fully understand previously, a lot of concepts

src/common/adapters/MultiAdapter.sol
hieronx: > could you extend a bit more about adapterData? Do we really need it? It's only used in one event, and only wormhole supports that.

It will be really useful for linking gateway events to wormhole events though. I would expect more adapters to support this, but even if not, I would say it's already worth it for wormhole.

src/common/adapters/MultiAdapter.sol
lemunozm: @hieronx what do you think about renaming `activeAdapters` to `adapterDetails`?

src/common/adapters/MultiAdapter.sol
hieronx: The reason it's hardcoded in the adapters is just gas efficiency (`gateway` is `immutable`) + it is very easy to swap adapters if we ever needed to swap the gateway.

In `MultiAdapter`, I would rather say it is not unlikely that we only want to migrate either `Gateway` or `MultiAdapter`, so it would be good to have that flexibility.

src/common/adapters/MultiAdapter.sol
lemunozm: Done

src/common/Gateway.sol
lemunozm: Done

src/common/Gateway.sol
lemunozm: Done

src/common/adapters/MultiAdapter.sol
lemunozm: I will leave it as it is by now, then 👍🏻

src/common/adapters/MultiAdapter.sol
hieronx: Works for me!

src/common/adapters/MultiAdapter.sol
lemunozm: Done

src/common/Gateway.sol
lemunozm: Done

src/common/adapters/MultiAdapter.sol
lemunozm: Done

src/common/adapters/AxelarAdapter.sol
```diff
@@ -19,17 +19,17 @@ import {IMessageHandler} from "src/common/interfaces/IMessageHandler.sol";
...
-    IMessageHandler public immutable gateway;
+    IMessageHandler public immutable startpoint;
     IAxelarGateway public immutable axelarGateway;
     IAxelarGasService public immutable axelarGasService;
 
     mapping(string axelarId => AxelarSource) public sources;
     mapping(uint16 centrifugeId => AxelarDestination) public destinations;
 
-    constructor(IMessageHandler gateway_, address axelarGateway_, address axelarGasService_, address deployer)
+    constructor(IMessageHandler startpoint_, address axelarGateway_, address axelarGasService_, address deployer)
```
hieronx: Could we call it `entrypoint`? `startpoint` is not an oft used English term afaik.

src/common/adapters/MultiAdapter.sol
```diff
@@ -0,0 +1,232 @@
...
+            cost = adapters_[i].estimate(centrifugeId, proof, gasLimit);
+            adapterData = adapters_[i].send{value: cost}(centrifugeId, proof, gasLimit, refund);
+            emit SendProof(centrifugeId, payloadId, payloadHash, adapters_[i], adapterData);
+        }
+
+        return bytes32(0);
+    }
+
+    /// @inheritdoc IAdapter
+    function estimate(uint16 centrifugeId, bytes calldata payload, uint256 gasLimit)
```
hieronx, line 208: Why re-add this?

lemunozm: The `Gateway` needs to estimate the `MultiAdapter` as if it were another adapter.

wischli (approved):
+1 on the `startpoint` nit - apart from that cleaner than a mountain lake 🎉

src/common/adapters/MultiAdapter.sol
hieronx: Right makes sense!

src/common/Guardian.sol
```diff
@@ -48,6 +48,7 @@ contract Guardian is IGuardian {
         if (what == "safe") safe = ISafe(data);
         else if (what == "sender") sender = IRootMessageSender(data);
         else if (what == "hub") hub = IHubGuardianActions(data);
+        else if (what == "multiAdapter") multiAdapter = IMultiAdapter(data);
```
lemunozm, line 51: This line is new, but I understand we want it.

src/common/adapters/MultiAdapter.sol
```diff
@@ -0,0 +1,239 @@
...
+import {ArrayLib} from "src/misc/libraries/ArrayLib.sol";
+import {MathLib} from "src/misc/libraries/MathLib.sol";
+
+import {IMessageHandler} from "src/common/interfaces/IMessageHandler.sol";
+import {IAdapter} from "src/common/interfaces/IAdapter.sol";
+import {MessageProofLib} from "src/common/libraries/MessageProofLib.sol";
+import {IMultiAdapter, MAX_ADAPTER_COUNT} from "src/common/interfaces/adapters/IMultiAdapter.sol";
+
+/// @title  Axelar Adapter
+/// @notice Routing contract that integrates with an Axelar Gateway
```
hieronx: Comment to be updated

hieronx: And title

src/common/adapters/MultiAdapter.sol
```diff
@@ -0,0 +1,239 @@
...
+
+        emit File(what, centrifugeId, addresses);
+    }
+
+    //----------------------------------------------------------------------------------------------
+    // Incoming
+    //----------------------------------------------------------------------------------------------
+
+    /// @inheritdoc IMessageHandler
+    function handle(uint16 centrifugeId, bytes calldata payload) external {
```
hieronx, line 91: This should be `pauseable`

src/common/adapters/MultiAdapter.sol
```diff
@@ -0,0 +1,239 @@
...
+            adapterData = adapters_[i].send{value: cost}(centrifugeId, proof, gasLimit, refund);
+            emit SendProof(centrifugeId, payloadId, payloadHash, adapters_[i], adapterData);
+        }
+
+        return bytes32(0);
+    }
+
+    /// @inheritdoc IAdapter
+    function estimate(uint16 centrifugeId, bytes calldata payload, uint256 gasLimit)
+        public
```
hieronx: This can be `external`

src/common/adapters/MultiAdapter.sol
lemunozm: Ooops!

src/common/adapters/MultiAdapter.sol
lemunozm: Correct. I apply this to Axelar and Wormhole adapters

src/common/Gateway.sol
```diff
@@ -483,26 +307,5 @@ contract Gateway is Auth, Recoverable, IGateway {
     function _outboundBatchSlot(uint16 centrifugeId, PoolId poolId) internal pure returns (bytes32) {
```
hieronx, line 311: Just for oganization purposes, could we add
```

    // Helpers
```

below `endBatching()`?

And potentially, also change `Outgoing` to `Outgoing handlers`, and add
```

    // Outgoing payments
```

above `subsidizePool`

lemunozm: Agree with the helpers. Regarding the outgoing payments, not sure given it's everything very related. For example `repay()`/`addUpnpaindMessage()` will go into "outgoing handler" or "outgoing payments"? 🤔

hieronx: Yeah fair

src/common/Gateway.sol
```diff
@@ -285,86 +149,46 @@ contract Gateway is Auth, Recoverable, IGateway {
...
-                    subsidy[poolId].value -= uint96(total);
-                } else {
-                    underpaid[centrifugeId][data.batchHash].counter++;
-                    underpaid[centrifugeId][data.batchHash].gasLimit = batchGasLimit;
-                    emit UnderpaidBatch(centrifugeId, batch);
-                    return false;
-                }
-            }
-        }
+                Underpaid storage underpaid_ = underpaid[centrifugeId][batchHash];
```
hieronx: I think we could call `addUnpaidMessage()` here

src/common/Gateway.sol
```diff
@@ -285,86 +149,46 @@ contract Gateway is Auth, Recoverable, IGateway {
...
+            centrifugeId,
+            batch,
+            batchGasLimit,
+            transactionRefund != address(0) ? transactionRefund : address(subsidy[poolId].refund)
+        );
+
         return true;
     }
 
     function addUnpaidMessage(uint16 centrifugeId, bytes memory message) external auth {
```
hieronx, line 182:
```suggestion
    /// @inheritdoc IGateway
    function addUnpaidMessage(uint16 centrifugeId, bytes memory message) external auth {
```

src/common/Gateway.sol
lemunozm: Super great recommendation!

lemunozm: Ah, I now remember I couldn't not. The gasLimit computation differs.

lemunozm: I did this anyway: [9ed8b25](https://github.com/centrifuge/protocol-v3/pull/372/commits/9ed8b25d93d16d4ab51215e0b88f1fe94c9b0eb0)

hieronx (approved):

PR #373, opened 2025-05-14
Add `HubHelpers`

hieronx:


src/hub/Hub.sol
```diff
@@ -666,9 +652,9 @@ contract Hub is Multicall, Auth, Recoverable, IHub, IHubGatewayHandler, IHubGuar
     /// @dev Ensure the sender is authorized
     function _auth() internal auth {}
 
-    /// @dev Ensure the method can be used without reentrancy issues, and the sender is a pool admin
+    /// @dev Ensure the method can be used without reentrancy issues, and the sender is a pool admin or a ward
     function _isManager(PoolId poolId) internal protected {
-        require(hubRegistry.manager(poolId, msg.sender), IHub.NotManager());
+        require(hubRegistry.manager(poolId, msg.sender) || wards[msg.sender] == 1, IHub.NotManager());
```
lemunozm, line 657: Why is this needed?

lemunozm (dismissed):

src/hub/Hub.sol
```diff
@@ -458,16 +422,18 @@ contract Hub is Multicall, Auth, Recoverable, IHub, IHubGatewayHandler, IHubGuar
...
-        accounts[0] = HoldingAccount(assetAccount, uint8(AccountType.Asset));
-        accounts[1] = HoldingAccount(equityAccount, uint8(AccountType.Equity));
-        accounts[2] = HoldingAccount(gainAccount, uint8(AccountType.Gain));
-        accounts[3] = HoldingAccount(lossAccount, uint8(AccountType.Loss));
+        {
+            HoldingAccount[] memory accounts = new HoldingAccount[](4);
+            accounts[0] = HoldingAccount(assetAccount, uint8(AccountType.Asset));
+            accounts[1] = HoldingAccount(equityAccount, uint8(AccountType.Equity));
+            accounts[2] = HoldingAccount(gainAccount, uint8(AccountType.Gain));
+            accounts[3] = HoldingAccount(lossAccount, uint8(AccountType.Loss));
```
wischli: Just a thought: Why don't we move the array initialization also to the hub helpers? IMO, this would align better with how we use the helper for other calls where the lower level logic is handled by the helper.

script/HubDeployer.s.sol
```diff
@@ -39,7 +41,8 @@ contract HubDeployer is CommonDeployer {
         accounting = new Accounting(deployer);
         holdings = new Holdings(hubRegistry, deployer);
         shareClassManager = new ShareClassManager(hubRegistry, deployer);
-        hub = new Hub(shareClassManager, hubRegistry, accounting, holdings, gateway, deployer);
+        hubHelpers = new HubHelpers(holdings, accounting, hubRegistry, shareClassManager, deployer);
+        hub = new Hub(shareClassManager, hubRegistry, accounting, hubHelpers, holdings, gateway, deployer);
```
wischli: Nit: I'd prefer to have the same order of contracts both for the helper and the hub.

src/hub/HubHelpers.sol
```diff
@@ -0,0 +1,183 @@
...
+import {IHoldings} from "src/hub/interfaces/IHoldings.sol";
+import {IHub, AccountType} from "src/hub/interfaces/IHub.sol";
+
+contract HubHelpers is Auth, IHubHelpers {
+    using MathLib for uint256;
+
+    IHoldings public immutable holdings;
+    IAccounting public immutable accounting;
+    IHubRegistry public immutable hubRegistry;
+    IShareClassManager public immutable shareClassManager;
```
wischli, line 27: Maybe CR: Curious why for the helper these are set to `immutable` whereas for the hub that's not the case. Seems erroneous to me because they should in sync regarding mutability?

hieronx: Making them `immutable` in the` Hub` increases the bytecode size :(

src/hub/Hub.sol
lemunozm: Same for liabilities

hieronx: Would you move just these 5 lines to `HubHelpers`, or the `initialize` call as well?

It seems to me we don't gain too much since you still need to pass all these params

wischli: > Would you move just these 5 lines to `HubHelpers`, or the `initialize` call as well?
> 
> It seems to me we don't gain too much since you still need to pass all these params

Good point. I think I'd prefer moving the initialization to the helper then as well but no strong opinion - let's merge as is.

wischli (approved):
Merge train here we go!

PR #376, opened 2025-05-15
Fix `sliceZeroPadded`

hieronx:
Fixes Cantina [#1](https://cantina.xyz/code/3b61c77c-4818-484c-ac0d-2addde3adf2f/findings/1) and [#6](https://cantina.xyz/code/3b61c77c-4818-484c-ac0d-2addde3adf2f/findings/6)


src/misc/libraries/BytesLib.sol
```diff
@@ -71,9 +70,17 @@ library BytesLib {
         return tempBytes;
     }
 
-    function slice(bytes memory _bytes, uint256 _start, uint256 _length) internal pure returns (bytes memory) {
-        require(_bytes.length >= _start + _length, SliceOutOfBounds());
-        return sliceZeroPadded(_bytes, _start, _length);
+    function sliceZeroPadded(bytes memory _bytes, uint256 _start, uint256 _length)
```
lemunozm, line 73: Where is this intended to be used?

lemunozm (approved):
Thanks!

BTW, the cantina links in the PR description are not the correct ones.

wischli:
> BTW, the cantina links in the PR description are not the correct ones.

Just added the links! The references where automatically set by Github because the links were missing.

wischli (approved):

src/misc/libraries/BytesLib.sol
hieronx: It's used for serialization of the token name in `NotifyShareClass`

lemunozm: Ah right! the diff has confused me, I thought it was a new method 😄

PR #377, opened 2025-05-15
Rename vaults to spokes, decouple vault managers and implementations

hieronx:


src/spokes/PoolManager.sol
```diff
@@ -66,8 +64,6 @@ contract PoolManager is
     ITokenFactory public tokenFactory;
     IVaultMessageSender public sender;
     IPoolEscrowFactory public poolEscrowFactory;
-    IAsyncRequestManager public asyncRequestManager;
```
hieronx, line 69: Only non-renaming in this PR is the removal of these 2 variables.

wischli: Great catch, thanks for the cleanup!

wischli (approved):
Spoke-tacular renaming!

PR #378, opened 2025-05-16
Feat: Legacay Migration Needs - V1

mustermeiszer:
Adds
* `LegacyVaultAdapterFactory` so that legacy vaults can be add through the normal flow during a migration
* Cuts out linking of share token so that we can run it during a spell 

Migration process of share tokens and vaults see [notion](https://www.notion.so/Protocol-Feature-Board-19d2eac24e1780ae9be7fb5be82cf430?p=1f52eac24e178099a1aced305d3da058&pm=s)

### DIFF
--- a/src/vaults/BalanceSheet.sol
+++ b/src/vaults/BalanceSheet.sol
@@ -214,6 +214,7 @@ contract BalanceSheet is Auth, Recoverable, IBalanceSheet, IBalanceSheetGatewayH
         TransientStorageLib.tstore(keccak256(abi.encode("pricePoolPerShareIsSet", poolId, scId)), true);
     }
     /// @inheritdoc IBalanceSheet
+
     function resetPricePoolPerShare(PoolId poolId, ShareClassId scId) external authOrManager(poolId) {
         TransientStorageLib.tstore(keccak256(abi.encode("pricePoolPerShareIsSet", poolId, scId)), false);
     }
--- a/src/vaults/PoolManager.sol
+++ b/src/vaults/PoolManager.sol
@@ -224,9 +224,8 @@ contract PoolManager is
             shareToken_.file("hook", hook);
         }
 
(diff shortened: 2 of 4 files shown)


hieronx:
> LegacyVaultAdapterFactory so that legacy vaults can be add through the normal flow during a migration

I'm not sure this is needed. Since we only do this once, we can just either deploy manually, or if it fits in the gas limit, deploy automatically in the spell.

mustermeiszer:
> I'm not sure this is needed. Since we only do this once, we can just either deploy manually, or if it fits in the gas limit, deploy automatically in the spell.

I agree, but then we need additional logic that mirrors or cuts out the relevant storage writes in `spokes.deployVault`

hieronx:
> I agree, but then we need additional logic that mirrors or cuts out the relevant storage writes in spokes.deployVault

Yes we should do that in a spell.

mustermeiszer:
But just for this one change we would also introduce another method that's auth callable instead of a factory? 

```
    /// @inheritdoc IPoolManager
    function deployVault(PoolId poolId, ShareClassId scId, AssetId assetId, IVaultFactory factory)
        public
        auth
        returns (IBaseVault)
    {

        // Deploy vault
        IBaseVault vault = IVaultFactory(factory).newVault(
            poolId, scId, assetIdKey.asset, assetIdKey.tokenId, shareClass.shareToken, new address[](0)
        );

        _vaultDetails[vault] = VaultDetails(assetId, assetIdKey.asset, assetIdKey.tokenId, false);
        emit DeployVault(poolId, scId, assetIdKey.asset, assetIdKey.tokenId, factory, vault, vault.vaultKind());

        return vault;
    }
 ```
 
 Would need to cur out the event and the storage write for this

hieronx:
@mustermeiszer Ah true.. I thought we could use `linkVault` but that is not the case.

PR #379, opened 2025-05-16
Add `MerkleProofManager`

hieronx:
The manager stores a policy, which is the root hash of a merkle tree. Each merkle tree leaf is a combination of:

- A *decoder*, which is a tiny smart contract used to support the integration
- A target contract address
- The function to call on the smart contract
- The address inputs for the function call
- Whether any native gas tokens should be passed when calling the function

To support a new protocol, a new decoder needs to be implemented. The goal is to return the address inputs from the function signature, so these can be filtered. This is a very simple smart contract to develop, e.g. to support deposit requests into an ERC-7540, the following is all that’s needed:

```solidity
function requestDeposit(uint256, address controller, address owner) external view virtual returns (bytes memory addressesFound) {
    addressesFound = abi.encodePacked(controller, owner);
}
```

Through this decoder + the leafs in the merkle tree, calls in the manager can be limited securely to the policy. Effectively this means a strategist, which is the caller of the manager, can be constrained in their allocation options, making it impossible to maliciously take out money.

**Credits to Boring Vault by Se7en-Seas for the design**


src/managers/MerkleProofManager.sol
```diff
@@ -0,0 +1,137 @@
...
+
+    /// @dev Convert hash of a merkle tree leaf
+    function _computeHash(PolicyLeaf memory leaf) internal pure returns (bytes32) {
+        return keccak256(abi.encodePacked(leaf.decoder, leaf.target, leaf.valueNonZero, leaf.selector, leaf.addresses));
+    }
+
+    /// @dev Execute a static call to a contract
+    function _staticCall(address target, bytes memory data) internal view returns (bytes memory) {
+        (bool success, bytes memory returnData) = target.staticcall(data);
+        require(success, CallFailed());
```
lemunozm, line 101: We should we add WrappedError here too I guess

src/managers/MerkleProofManager.sol
```diff
@@ -0,0 +1,137 @@
...
+
+        return returnData;
+    }
+
+    /// @dev Execute a call with value to a contract
+    function _callWithValue(address target, bytes memory data, uint256 value) internal returns (bytes memory) {
+        require(address(this).balance >= value, InsufficientBalance());
```
lemunozm, line 108: Is this not checked automatically by solidity?

src/managers/MerkleProofManager.sol
```diff
@@ -0,0 +1,137 @@
...
+
+    constructor(address spoke_, IBalanceSheet balanceSheet_) {
+        spoke = spoke_;
+        balanceSheet = balanceSheet_;
+    }
+
+    /// @inheritdoc IMerkleProofManagerFactory
+    function newManager(PoolId poolId) external returns (IMerkleProofManager) {
+        MerkleProofManager manager =
+            new MerkleProofManager{salt: bytes32(uint256(poolId.raw()))}(poolId, spoke, balanceSheet);
```
lemunozm: What if a malicious actor creates this manager before the correct manager one? I think the correct would never can create its manager.

I think this can not be permissionless if it's per pool, or we should add some `salt` as a parameter.

src/managers/decoders/VaultDecoder.sol
```diff
@@ -0,0 +1,112 @@
...
+pragma solidity 0.8.28;
+
+import {BaseDecoder} from "src/managers/decoders/BaseDecoder.sol";
+
+contract VaultDecoder is BaseDecoder {
+    // @desc deposit into the ERC4626 vault
+    // @tag receiver:address:the address of the receiver of the vault tokens
+    function deposit(uint256, address receiver) external view virtual returns (bytes memory addressesFound) {
+        addressesFound = abi.encodePacked(receiver);
+    }
```
lemunozm, line 11: Why is the `amount` not encoded? Sure it's related to the merkle tree idea but I'm not able to follow right now

src/managers/MerkleProofManager.sol
```diff
@@ -0,0 +1,137 @@
...
+import {IMerkleProofManagerFactory} from "src/managers/interfaces/IMerkleProofManagerFactory.sol";
+
+/// @title  Merkle Proof Manager
+/// @author Inspired by Boring Vaults from Se7en-Seas
+contract MerkleProofManager is IMerkleProofManager, IUpdateContract {
+    using CastLib for *;
+
+    PoolId public immutable poolId;
+    address public immutable spoke;
+    IBalanceSheet public immutable balanceSheet;
```
lemunozm: Where `balanceSheet` will be used in the contract?

src/managers/MerkleProofManager.sol
hieronx: There is a salt right? This is the point, it is deployed once permisisonlessly per pool, but then there is no special permissions, it's just controlled by `UpdateContract` from `Spoke`

src/managers/decoders/VaultDecoder.sol
hieronx: The idea is that in the policy merkle tree, you would store that you can deposit into a vault, with receiver e.g. only the merkle proof manager. If you also include the amount here, you could only deposit exactly that amount, which doesn't work. You want to allow any amount, as long as it's with the right address inputs.

src/managers/MerkleProofManager.sol
hieronx: Good point! It's unused.

src/managers/MerkleProofManager.sol
hieronx: I don't think so. WIll try to add a test for this to be sure

lemunozm: I think the method is reverted when it has not enough gas to pay

src/managers/MerkleProofManager.sol
lemunozm: Ok, now I see your point.

src/managers/decoders/VaultDecoder.sol
lemunozm: Ok... I understand.

What if, for example, I want to add a limited amount? Can I do it with this pattern?Should I add another method in the decoder to handle that?

hieronx: That doesn't work with this pattern unfortunately. What you can do is implement a simple contract on top, that adds just the amount validation, and then forwards to the merkle proof manager.

hieronx: I don't think limiting amount in any case would work though, since you could just repeat the transaction n times.

But you can add other numeric validation, like slippage if you are interacting with a DEX

src/managers/MerkleProofManager.sol
```diff
@@ -0,0 +1,131 @@
...
+        poolId = poolId_;
+        spoke = spoke_;
+    }
+
+    //----------------------------------------------------------------------------------------------
+    // Owner actions
+    //----------------------------------------------------------------------------------------------
+
+    /// @inheritdoc IUpdateContract
+    function update(PoolId poolId_, ShareClassId, /* scId */ bytes calldata payload) external {
```
wischli, line 36: Shouldn't this call be protected by `auth` which is missing?

src/managers/MerkleProofManager.sol
```diff
@@ -0,0 +1,131 @@
...
+    //----------------------------------------------------------------------------------------------
+    // Strategist actions
+    //----------------------------------------------------------------------------------------------
+
+    /// @inheritdoc IMerkleProofManager
+    function execute(Call[] calldata calls) external {
+        bytes32 strategistPolicy = policy[msg.sender];
+        require(strategistPolicy != bytes32(0), NotAStrategist());
+
+        for (uint256 i; i < calls.length; ++i) {
```
wischli: Curious why `++i` instead of `i++`

src/managers/MerkleProofManager.sol
```diff
@@ -0,0 +1,131 @@
...
+            target: call.target,
+            selector: bytes4(call.targetData),
+            addresses: addresses,
+            valueNonZero: call.value > 0
+        });
+    }
+
+    /// @dev Convert hash of a merkle tree leaf
+    function _computeHash(PolicyLeaf memory leaf) internal pure returns (bytes32) {
+        return keccak256(abi.encodePacked(leaf.decoder, leaf.target, leaf.valueNonZero, leaf.selector, leaf.addresses));
```
wischli, line 95: Any reason to not follow the order in `PolicyLeaf`?

```suggestion
        return keccak256(abi.encodePacked(leaf.decoder, leaf.target, leaf.selector, leaf.addresses,  leaf.valueNonZero));
```

src/managers/MerkleProofManager.sol
wischli: Very sure this is automatically handled by the EVM. From a quick search, could only find a reference to `transfer` in the docs ([ref](https://docs.soliditylang.org/en/v0.8.20/types.html#members-of-addresses:~:text=The%20transfer%20function%20fails%20if%20the%20balance%20of%20the%20current%20contract%20is%20not%20large%20enough%20or%20if%20the%20Ether%20transfer%20is%20rejected%20by%20the%20receiving%20account.%20The%20transfer%20function%20reverts%20on%20failure.))

src/managers/decoders/BaseDecoder.sol
```diff
@@ -0,0 +1,39 @@
...
+pragma solidity 0.8.28;
+
+import {PoolId} from "src/common/types/PoolId.sol";
+import {ShareClassId} from "src/common/types/ShareClassId.sol";
+
+contract BaseDecoder {
+    error FunctionNotImplemented(bytes _calldata);
+
+    // @desc The spender address to approve
+    // @tag spender:address
```
wischli: Nit: Not following our natspec. Same for the other decoders

src/managers/decoders/BaseDecoder.sol
```diff
@@ -0,0 +1,39 @@
...
+    }
+
+    // @desc deposit into the balance sheet
+    function deposit(PoolId, ShareClassId, address asset, uint256, uint128)
+        external
+        view
+        virtual
+        returns (bytes memory addressesFound)
+    {
+        addressesFound = abi.encodePacked(asset);
```
wischli: Can you elaborate on why we only filter for asset addresses but not the share class id?

test/managers/integration/MerkleProofManager.t.sol
```diff
@@ -0,0 +1,511 @@
...
+
+        vm.prank(address(spoke));
+        manager.update(
+            POOL_A,
+            defaultTypedShareClassId,
+            UpdateContractMessageLib.UpdateContractPolicy({
+                who: address(this).toBytes32(),
+                what: tree[tree.length - 1][0]
+            }).serialize()
+        );
```
wischli: Nit: This can be moved to a helper function as its used 4 times throughout the tests. Would make it much easier to follow.

test/managers/integration/MerkleProofManager.t.sol
```diff
@@ -0,0 +1,511 @@
...
+
+        bytes32[][] memory tree = MerkleTreeLib.generateMerkleTree(_computeHashes(leafs));
+
+        vm.prank(address(spoke));
+        manager.update(
+            POOL_A,
+            defaultTypedShareClassId,
+            UpdateContractMessageLib.UpdateContractPolicy({
+                who: address(this).toBytes32(),
+                what: tree[tree.length - 1][0]
```
wischli: Can you elaborate on why this particular part of the tree is sufficient?

src/managers/interfaces/IMerkleProofManager.sol
```diff
@@ -0,0 +1,42 @@
+// SPDX-License-Identifier: GPL-2.0-or-later
+pragma solidity >=0.5.0;
+
+struct Call {
+    address decoder;
```
wischli, line 5: Can you explain to me how it would work if I had to update the decoder of a MerkleProofManager? Say my current decoder needs to be extended by a new call. IIUC, the policy would have to be updated via `UpdateContract` because the new decoder leads to different merkle tree root hash. Therefore, old proofs would be invalid. I have a hard time understanding the impact from a product perspective.


How would that work and would I be able to seamlessly use that new decoder with the same MerkleProofManager or require some spell?

wischli (commented):
A bunch of questions from my side. Code looks good as far as I can understand with my current knowledge!

src/managers/MerkleProofManager.sol
hieronx: It is checking `msg.sender == spoke` so that is effectively protecting it

src/managers/MerkleProofManager.sol
hieronx: Slightly more gas efficient I believe for some arcane reason. Might not be true anymore with latest Solidity version though

src/managers/decoders/BaseDecoder.sol
hieronx: Yeah was matching the existing decoder design here. I agree would make sense to update

src/managers/decoders/BaseDecoder.sol
hieronx: Good point, should add that here!

test/managers/integration/MerkleProofManager.t.sol
hieronx: It is the root hash

wischli: Then my nit would be to move this to a helper function to remove the opaqueness

hieronx: Good point!

src/managers/interfaces/IMerkleProofManager.sol
hieronx: The main point to understand is: there can be many decoders, not just 1. Since each call and policy leaf points to a specific decoder. So you can have decoder A, with a series of leafs, and then you can also have decoder B, and add leafs pointing to that one.

src/managers/MerkleProofManager.sol
hieronx: Just wanted to match the original design. Will update the `PolicyLeaf` struct to match.

src/managers/interfaces/IMerkleProofManager.sol
wischli: Sorry, my question was poorly phrased due to the fuzziness of understanding the practical application in detail. Let me rephrase: Say we need to update an existing Decoder for vault interactions because we extend that ABI by at least a new call that needs to be covered by the decoder. What would be the series of actions we that need to be done, especially offchain? 

AFAIU, on the protocol level it suffices to update the policy for the affected address(es). This would invalidate old proofs for that address. For Apps, would it suffice to just generate the new proofs based on the new policy (signalled by Indexer) such that the policy update is entirely abstracted away from the user?

wischli (approved):
Thanks for applying my nits and updating the `BaseDecoder`! Apart from a follow-up test refactor, let's re-evaluate our open discussion about the necessity to (not) check call value sufficiency ([ref](https://github.com/centrifuge/protocol-v3/pull/379#discussion_r2113479172)).

PR #380, opened 2025-05-19
Remove ERC20 wrapping feature

hieronx:


lemunozm (approved):
This has always been a knowledge leak for me. Great to see this simpler now!

PR #381, opened 2025-05-19
Modify index from MessageProofId

lemunozm:
Minor change, but I think it's useful for:

- Have more space to keep everything in `MESSAGE_LENGTH_1`
- Conceptually, the message proof is not related to the rest of the messages, message proof is related to the adapters, not to our business logic. MessageProof is like an IP message, and the rest of the messages are like TCP messages, one level higher.
  - This distinction is more relevant once we merge the `MultiAdapter` contract


src/common/libraries/MessageLib.sol
```diff
@@ -10,8 +10,6 @@ import {AssetId} from "src/common/types/AssetId.sol";
 enum MessageType {
```
hieronx, line 11: Should we add a comment here, should never exceed 254 messages, to not re-use the message proof ID accidentally? Very unlikely but just to have documented.

lemunozm: There is a test to avoid that, but I agree in adding it

hieronx (approved):

PR #385, opened 2025-05-20
Holding & share snapshots

hieronx:
Fundamentally we realized that the accounting layer on the Hub side previously could not properly function, because any time an asset update was included on the hub side but a share update was not yet, then no hub manager could properly calculate the price per share, since the ratio would be off. This is a problem we needed to solve since it makes any kind of hub automation impossible, which is the point of the centralized accounting layer.

How we fixed it:

**Balance sheet**
- Stores `queuedAssetCounter` + `nonce` in the share queue storage in the balance sheet
  - This allows knowing the number of to be updated asset holdings.
- Increases `nonce` on every holding update messages to the hub.
    - This ensures message ordering just for holding and share update messages, which is required to not have asset/share mismatches.
- Moves queued shares storage to a delta, to allow fitting the full struct in 1 storage slot.
- Passes `isSnapshot` and `nonce` on every update message.
- Enforce queuing by default, and only allow disabling this when it's local (spoke <> hub on same chain). If not, someone could block the system by doing tiny sync deposits and not paying for the asset/share update messages, which would then block other messages because of the nonce.

**Hub**
- Store issuance of a share class per network
- Store whether the current state is a snapshot or not
- Block this update if the nonce is out of order, to enforce message ordering
- Add `ISnapshotHook`, that is called when the snapshot is in sync.

**Manager**
Based on these components, a higher-level hub manager should be able to update the NAV/share per network, whenever `Holdings.isSnapshot == true`

We also found another issue while going over all of this: previously, the total issuance was actually increased twice on every async request, because it was immediately increased in `scm.issueShares`, and then again in the callback loop `Hub.notifyDeposit => asyncRequestManager.issuedShares => balanceSheet.issue => hub.increaseShareClassIssuance`. This is now also fixed.


src/spokes/interfaces/IBalanceSheet.sol
```diff
@@ -13,11 +13,18 @@ import {ISpoke} from "src/spokes/interfaces/ISpoke.sol";
...
 import {IPoolEscrowProvider} from "src/spokes/interfaces/factories/IPoolEscrowFactory.sol";
 
-struct QueueAmount {
-    // Issuances of shares / deposits of assets
-    uint128 increase;
-    // Revocations of shares / withdraws of assets
-    uint128 decrease;
+struct ShareQueueAmount {
+    // Net queued shares
+    uint128 delta;
```
hieronx: Storing this as delta now to be able to pack the `queuedAssetCounter` into the same storage slot.

src/spokes/BalanceSheet.sol
```diff
@@ -160,9 +181,19 @@ contract BalanceSheet is Auth, Recoverable, IBalanceSheet, IBalanceSheetGatewayH
...
+            if (queuedShares[poolId][scId].isPositive || queuedShares[poolId][scId].delta == 0) {
+                queuedShares[poolId][scId].delta += shares;
+                queuedShares[poolId][scId].isPositive = true;
+            } else if (queuedShares[poolId][scId].delta > shares) {
+                queuedShares[poolId][scId].delta -= shares;
+                queuedShares[poolId][scId].isPositive = false;
+            } else {
+                queuedShares[poolId][scId].delta = shares - queuedShares[poolId][scId].delta;
+                queuedShares[poolId][scId].isPositive = true;
+            }
```
lemunozm: I think the logic is perfect!

Maybe we can factorize the maths into a `DeltaMathLib.sol` where we define a `Delta` struct with `add/sub` associated method? That way, we reduce the complexity at this contract level for the readers. Easy to test from the lib too.

test/hub/integration/Cases.t.sol
```diff
@@ -182,7 +182,7 @@ contract TestCases is BaseTest {
         uint128 poolDecimals = (10 ** hubRegistry.decimals(USD.raw())).toUint128();
         uint128 assetDecimals = (10 ** hubRegistry.decimals(USDC_C2.raw())).toUint128();
 
-        cv.updateHoldingAmount(poolId, scId, USDC_C2, 1000 * assetDecimals, D18.wrap(1e18), true);
+        cv.updateHoldingAmount(poolId, scId, USDC_C2, 1000 * assetDecimals, D18.wrap(1e18), true, true);
```
lemunozm: NIT. (I can do this in a future PR, and also for other booleans). What about defining a `IS_SNAPSHOT` bool constant?

lemunozm (commented):
As long as I understand the logic, looks perfect!

src/spokes/BalanceSheet.sol
hieronx: Good idea!

lemunozm:
BTW, `_submitQueuedShares()` and `_submitQueuedAssets()` are only used from one method. Maybe we can remove them and merged with the caller in this PR too?

hieronx:
> BTW, _submitQueuedShares() and _submitQueuedAssets() are only used from one method. Maybe we can remove them and merged with the caller in this PR too?

Agreed!

src/spokes/BalanceSheet.sol
hieronx: This is actually a bit challenging. We have 2 other properties on `ShareQueueAmount` struct, which are not generic. So we cannot just use a `Delta` struct.

We could add methods to convert to/from `ShareQueueAmount` and `Delta`, and then do the computations on `Delta`. But that seems like more overhead.

lemunozm: This should work, isn't?:

```solidity
struct Delta {

}

library DeltaLib {
    function add(Delta delta, ...)
}


struct ShareQueueAmount {
    Delta delta
    // Number of queued asset IDs for this share class
    uint32 queuedAssetCounter;
    // Nonce for share + asset messages to the hub
    uint88 nonce;
}
```

hieronx: Hm interesting. I wonder if it would still be able pack everything in 1 slot 🤔

src/hub/ShareClassManager.sol
```diff
@@ -340,24 +337,19 @@ contract ShareClassManager is Auth, IShareClassManager {
...
+        auth
+    {
         require(exists(poolId, scId_), ShareClassNotFound());
-        require(metrics[scId_].totalIssuance >= amount, DecreaseMoreThanIssued());
+        require(isIssuance || issuance[scId_][centrifugeId] >= amount, DecreaseMoreThanIssued());
 
-        uint128 newIssuance = metrics[scId_].totalIssuance - amount;
+        uint128 newIssuance = isIssuance ? metrics[scId_].totalIssuance + amount : metrics[scId_].totalIssuance - amount;
         metrics[scId_].totalIssuance = newIssuance;
+        issuance[scId_][centrifugeId] = newIssuance;
```
lemunozm: I think we have duplicated both storage. They track the same exact information

src/misc/libraries/BytesLib.sol
```diff
@@ -138,6 +138,17 @@ library BytesLib {
         return tempUint;
     }
 
+    function toUint88(bytes memory _bytes, uint256 _start) internal pure returns (uint88) {
+        require(_bytes.length >= _start + 9, SliceOutOfBounds());
```
lemunozm: CR: I think it should be 11 bytes.
```suggestion
        require(_bytes.length >= _start + 11, SliceOutOfBounds());
```
Also curious about why using `uint88` also

src/spoke/BalanceSheet.sol
```diff
@@ -244,47 +255,83 @@ contract BalanceSheet is Auth, Recoverable, IBalanceSheet, IBalanceSheetGatewayH
...
+        PoolId poolId,
+        ShareClassId scId,
+        AssetId assetId,
+        uint128 amount,
+        bool isDeposit,
+        D18 pricePoolPerAsset
+    ) internal {
+        ShareQueueAmount storage shareQueue = queuedShares[poolId][scId];
+        AssetQueueAmount storage assetQueue = queuedAssets[poolId][scId][assetId];
+        if (queueEnabled[poolId][scId]) {
```
lemunozm: IIUC, if we decide to always queue, we can remove the else branch, isn`t?

lemunozm (commented):
I think is not too much complex, but maybe I'm biased.

src/hub/ShareClassManager.sol
hieronx: Wdym? `metrics.totalIssuance` is across all networks, `issuance` is per network (`centrifugeId`)

src/misc/libraries/BytesLib.sol
hieronx: That is the maximum size that fits into 32 byte slot for `ShareQueueAmount` (which is `uint128 + bool + uint32 + uint88 = 16 + 1 + 4 + 11`)

src/spoke/BalanceSheet.sol
hieronx: We can still disable queuing on local chains though

hieronx: I agree it would be nice to simplify, but it seems valuable enough to keep non-queuing behaviour just for local chains.

lemunozm: WDYM by local chain?

hieronx: Spoke and hub on the same chain

src/hub/ShareClassManager.sol
lemunozm: Then I think `issuance[scId_][centrifugeId] = newIssuance` is wrong because `newIssuance` is computed from the total issuance accross all networks, but alter used only in one network. It means, it's populated with the values from other networks.

Should not be:

```solidity
uint128 newIssuance = isIssuance ? metrics[scId_].totalIssuance + amount : metrics[scId_].totalIssuance - amount;
metrics[scId_].totalIssuance = newIssuance;
uint128 newIssuancePerNetwork = isIssuance ? issuance[scId_][centrifugeId] + amount : issuance[scId_][centrifugeId]- amount;
issuance[scId_][centrifugeId] = newIssuancePerNetwork;
```

src/misc/libraries/BytesLib.sol
lemunozm: I think `uint64` is enough for `nonce`, it's the same we use for timestamps. It's impossible to overflow it in case you want to simplify this a bit, or reserve some bytes for other additions

src/spoke/BalanceSheet.sol
lemunozm: Maybe we can just check that instead of having a `queueEnabled` storage. I mean, only unqueue if `sender.localCentrifugeId == poolId.centrifugeId()`

lemunozm: At least that way we can test all paths out of the box with the current ITs and we can remove SetQueue message

src/hub/ShareClassManager.sol
wischli: Agreeing with @lemunozm.

src/spoke/BalanceSheet.sol
```diff
@@ -158,10 +157,22 @@ contract BalanceSheet is Auth, Recoverable, IBalanceSheet, IBalanceSheetGatewayH
     function revoke(PoolId poolId, ShareClassId scId, uint128 shares) external authOrManager(poolId) {
         emit Revoke(poolId, scId, msg.sender, _pricePoolPerShare(poolId, scId), shares);
 
+        ShareQueueAmount storage shareQueue = queuedShares[poolId][scId];
         if (queueEnabled[poolId][scId]) {
-            queuedShares[poolId][scId].decrease += shares;
+            if (!shareQueue.isPositive) {
+                shareQueue.delta += shares;
+                shareQueue.isPositive = false;
```
wischli: Nit: Can be removed since it doesn't change state
```suggestion
```

src/hub/ShareClassManager.sol
hieronx: You are absolutely right!

src/spoke/BalanceSheet.sol
hieronx: Yeah that is also the way I would go. Will wait for @mustermeiszer's feedback on this before implementing.

src/spoke/BalanceSheet.sol
```diff
@@ -244,47 +255,83 @@ contract BalanceSheet is Auth, Recoverable, IBalanceSheet, IBalanceSheetGatewayH
 
     /// @inheritdoc IBalanceSheetGatewayHandler
     function submitQueuedAssets(PoolId poolId, ShareClassId scId, AssetId assetId) external authOrManager(poolId) {
-        _submitQueuedAssets(poolId, scId, assetId);
+        ShareQueueAmount storage shareQueue = queuedShares[poolId][scId];
+        AssetQueueAmount storage assetQueue = queuedAssets[poolId][scId][assetId];
+
+        if (assetQueue.deposits > 0 || assetQueue.withdrawals > 0) {
+            shareQueue.queuedAssetCounter--;
```
wischli: Can you explain why we reduce the counter if something is queued for this `(scId, assetId)` and not increment it? From the docs, it appears to me that any queue amount greater 0 should increment and opposite for `== 0` but the implementation is opposite.
```solidity
struct ShareQueueAmount {
    // Net queued shares
    uint128 delta;
    // Whether the net queued shares lead to an issuance or revocation
    bool isPositive;
    // Number of queued asset IDs for this share class
    uint32 queuedAssetCounter;
    // Nonce for share + asset messages to the hub
    uint88 nonce;
}
```

wischli (commented):
As always, super solid work! A question from my side.

src/spokes/BalanceSheet.sol
wischli: AFAICT, this should still fit because we haven't changed field sizes and nested structs are laid out sequentially as if there were flattened

src/misc/libraries/BytesLib.sol
hieronx: Fair! WIll update

src/spokes/BalanceSheet.sol
hieronx: It doesn't seem to work that way unfortunately: https://github.com/centrifuge/protocol-v3/pull/401 You can see the gas snapshot `deposit_withQueue` in `SyncDepositVault.json` has increased by ~21k which is about the cost of 1 cold storage write.

lemunozm: Ouch! I'm super angry with the solidity compiler right now 😠

lemunozm: I'm searching and it seems like the solidity compile should do that work. I do not understand why you have such gas increment 🤔

src/spoke/BalanceSheet.sol
hieronx: This method is submitting the asset updates, meaning dequeuing, meaning decrementing the queued asset counter.

hieronx: I've rewritten this to a bit clearer, using an early return instead.

src/spokes/BalanceSheet.sol
hieronx: I would suggest to merge this now, and consider optimizing this in a followup PR, if possible. Wdyt?

lemunozm (approved):
LGTM!! Although I would appreciate also a @wischli review too before merging it. Great work here!

src/spoke/BalanceSheet.sol
mustermeiszer: I am not convinced this is a good idea. For the already mentioned reason to restrict the core to enforce that - mostly as I am not sure we can call submit directly afterwards all the time from higher level contracts. 

But I also see the benefits. So lets go that route.

mustermeiszer (approved):
Complex but I think a good solution to ensure composability on the hub side!

src/spoke/BalanceSheet.sol
hieronx: > Maybe we can just check that instead of having a queueEnabled storage. I mean, only unqueue if sender.localCentrifugeId == poolId.centrifugeId()

Actually no we shouldn't do that. On local chains, you'd still want the option to enable queuing, since without queuing the gas cost of a sync deposit is much higher (almost 3x). This is because of all the accounting updates.

So imo there should still be a `SetQueue` message, but the `Hub` should only allow triggering it to `localCentrifugeId`, and queuing should default to `true`.

src/common/interfaces/ISnapshotHook.sol
```diff
@@ -0,0 +1,20 @@
+// SPDX-License-Identifier: BUSL-1.1
+pragma solidity 0.8.28;
+
+import {PoolId} from "src/common/types/PoolId.sol";
+import {ShareClassId} from "src/common/types/ShareClassId.sol";
+
+/// @notice Hook interface that is called whenever the accounting system of the Hub, for a given share token on
```
hieronx, line 7: Would appreciate feedback if this comment is clear. Think this will be really critical for builders on top of the hub to understand.

lemunozm:
I think this line can now be removed: https://github.com/centrifuge/protocol-v3/blob/spearbit-fixreview/test/integration/EndToEnd.t.sol#L351 and the test should pass

lemunozm (approved):

src/common/MessageDispatcher.sol
```diff
@@ -451,15 +451,8 @@ contract MessageDispatcher is Auth, IMessageDispatcher {
...
-            balanceSheet.setQueue(poolId, scId, enabled);
-        } else {
-            gateway.send(
-                centrifugeId,
-                MessageLib.SetQueue({poolId: poolId.raw(), scId: scId.raw(), enabled: enabled}).serialize()
-            );
-        }
+    function sendSetQueue(PoolId poolId, ShareClassId scId, bool enabled) external auth {
+        balanceSheet.setQueue(poolId, scId, enabled);
     }
```
lemunozm, line 456: The message can be removed then, right?

src/common/interfaces/ISnapshotHook.sol
lemunozm: I was able to fully understand, but I could be biased.

Question. Why call this `onSnapshot` instead of `onSync`?

lemunozm (approved):
LGTM

lemunozm (approved):

src/common/interfaces/ISnapshotHook.sol
wischli: > Question. Why call this `onSnapshot` instead of `onSync`?

I had the same question. The latter is much clearer to me. Is it to distinguish from sync and async vaults?

wischli (commented):
Post mortem: Much clearer now, appreciate improving the explicitness!

PR #390, opened 2025-05-20
Pass tokenWards directly to the factory once.

lemunozm:
Minor simplification to avoid `Spoke` to know about the `BalanceSheet` existence.

Also, from a configuration PoV, I think it makes sense to pass the wards just once in the deployment phase instead of in every token creation

### DIFF
--- a/src/spokes/factories/TokenFactory.sol
+++ b/src/spokes/factories/TokenFactory.sol
@@ -13,19 +13,25 @@ import {IShareToken} from "src/spokes/interfaces/IShareToken.sol";
 ///         based on the pool id and share class id.
 contract TokenFactory is Auth, ITokenFactory {
     address public immutable root;
+    address[] public tokenWards;
 
     constructor(address _root, address deployer) Auth(deployer) {
         root = _root;
     }
 
+    function file(bytes32 what, address[] memory data) external auth {
+        if (what == "wards") tokenWards = data;
+        else revert FileUnrecognizedParam();
+
+        emit File(what, data);
+    }
+
     /// @inheritdoc ITokenFactory
-    function newToken(
-        string memory name,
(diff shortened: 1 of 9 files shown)


test/spokes/unit/factories/TokenFactory.t.sol
```diff
@@ -92,7 +92,7 @@ contract FactoryTest is Test {
     }
 
     function testShareShouldBeDeterministic(
-        address asyncRequestManager,
+        address balanceSheet,
```
lemunozm: Just a rename, it should be `balanceSheet` not `asyncRequestManager`

src/spokes/factories/TokenFactory.sol
```diff
@@ -13,19 +13,25 @@ import {IShareToken} from "src/spokes/interfaces/IShareToken.sol";
 ///         based on the pool id and share class id.
 contract TokenFactory is Auth, ITokenFactory {
     address public immutable root;
+    address public tokenWards;
```
hieronx: This should be an array right?

lemunozm: Yeah!

test/spokes/unit/factories/TokenFactory.t.sol
```diff
@@ -68,32 +69,7 @@ contract FactoryTest is Test {
...
-                                )
-                            )
-                        )
-                    )
-                )
-            )
-        );
-        TokenFactory tokenFactory = new TokenFactory{salt: salt}(root, address(this));
-        assertEq(address(tokenFactory), predictedAddress);
-    }
```
lemunozm, line 92: I've removed this because here we're testing the test.

hieronx (approved):

PR #242, opened 2025-04-10
feat: Shareclass manager refactor

mustermeiszer:

PR #392, opened 2025-05-20
ref: improve pricing precision

wischli:
Refactors calculations of `PricingLib` to reduce rounding from 2+ to 1 for call. As a result, this improves the precision of our calculations, which you can see for instance in changes for the ShareClassManager tests. The downside is that the math is now less readable IMO.

> Directly multiplying two D18 prices first and then performing the conversion afterward results in significant precision loss due to downscaling by 1e18, truncating the intermediate precision. Doing this step prematurely leads to a precision loss in subsequent mathematical operations, which further magnifies the issue. Instead, squeezing the entire calculation in the most minimal math operation (i.e. a single MathLib.mulDiv instead of multiple ones under the hood) with one consistent direction (instead of rounding down for intermediate mulDiv and then applying the desired rounding for the final computation) maintains the best possible accuracy, especially for edge cases involving very large or very small numbers.
> 
> In https://github.com/centrifuge/protocol-v3/pull/312, we already reduced the number of math operations by at least one for most calculations. However, after experimenting with different implementations, I noticed that the changes proposed in this PR offer greater accuracy due to the reasons mentioned above. I tried improving readability by abstracting away the Math.mulDiv into D18 functions. However, IMO this did not meaningfully enhance readability, so I decided to keep the explicit approach instead.

This PR also adds more PricingLib tests to cover edge cases and validate rounding.


test/hub/unit/ShareClassManager.t.sol
```diff
@@ -1694,11 +1694,7 @@ contract ShareClassManagerRoundingEdgeCasesRedeem is ShareClassManagerBaseTest {
             shareClass.claimRedeem(poolId, scId, INVESTOR_B, OTHER_STABLE);
 
         assertEq(claimedA, claimedB, "Claimed asset amount should be equal");
-        assertEq(
-            claimedA + claimedB + 1 * MAX_OTHER_STABLE_PRECISION,
-            assetPayout,
-            "System should have 1 max asset precision amount surplus"
-        );
+        assertEq(claimedA + claimedB + 1, assetPayout, "System should have 1 amount surplus");
```
wischli, line 1697: Proof of improved precision. We used to have this precision before #242.

src/common/libraries/PricingLib.sol
```diff
@@ -177,12 +182,36 @@ library PricingLib {
...
+
+    /// @dev Converts an amount using decimals and two prices.
+    ///
+    ///      NOTE: More precise than custom math with one price and convertWith{Reciprocal}Price for the other
+    function convertWithPrices(
+        uint256 baseAmount,
+        uint8 baseDecimals,
+        uint8 quoteDecimals,
+        D18 priceNumerator,
+        D18 priceDenominator,
```
lemunozm, line 203: Conceptually, I feel a bit lost here. 

Does it means that you're applying the composition of both prices? If it's, why not just provide a single price that takes already that composition?

AFAIK, if I have price `a` and price `b`, and both prices have the same decimals (18 decimals because they are `D18`). Could not the caller just do:

```
convertWithPrice(baseAmount, baseDecimals, quoteDecimals, priceNumerator * priceDenominator)
```

lemunozm (approved):
Despite my curiosity from above the PR looks great!!

src/common/libraries/PricingLib.sol
wischli: I totally understand how you feel because I have been there as well. Conceptually, you are right - both calculations should lead to the same result whereas mine is less readable. However, in practice numerical precision differs between the two approaches.

Directly multiplying two `D18` prices first and then performing the conversion afterward results in significant precision loss due to downscaling by `1e18`, truncating the intermediate precision. Doing this step prematurely leads to a precision loss in subsequent mathematical operations, which further magnifies the issue. Instead, squeezing the entire calculation in the most minimal math operation (i.e. a single `MathLib.mulDiv` instead of multiple ones under the hood) with one consistent direction (instead of rounding down for intermediate `mulDiv` and then applying the desired rounding for the final computation) maintains the best possible accuracy, especially for edge cases involving very large or very small numbers.

In #312, we already reduced the number of math operations by at least one for most calculations. However, after experimenting with different implementations, I noticed that the changes proposed in this PR offer greater accuracy due to the reasons mentioned above. I tried improving readability by abstracting away the `Math.mulDiv` into `D18` functions. However, IMO this did not meaningfully enhance readability, so I decided to keep the explicit approach instead.

test/misc/unit/libraries/PricingLib.t.sol
```diff
@@ -139,54 +437,131 @@ contract PurePricingLibTest is Test {
...
+            MathLib.Rounding.Up
         );
 
         uint256 result = PricingLib.shareToAssetAmount(
-            shareAmount, POOL_DECIMALS, assetDecimals, pricePoolPerAsset, pricePoolPerShare, ROUNDING_DOWN
+            shareAmount, POOL_DECIMALS, assetDecimals, pricePoolPerAsset, pricePoolPerShare, MathLib.Rounding.Down
         );
 
-        assertEq(result, expectedAssetAmount, "shareToAssetAmount failed");
+        assertGe(result, underestimate, "shareToAssetAmount should be at least as large as underestimate");
```
wischli, line 518: @lemunozm One example that proves `priceA.reciprocal(priceB.mul(amount))` (represented by `underestimate`) is less accurate than `Math.mulDiv(x, y, z)` (represented by `expected*`) because it is always less or equal than the actual value, proving that it suffered from truncation.

 I have followed this test template for all lib functions.

src/common/libraries/PricingLib.sol
lemunozm: Ok! Now I understand! Thanks so much for this deep explanation 🙌🏻

lemunozm (approved):
🚀

PR #393, opened 2025-05-20
Add token & vault update methods

hieronx:


lemunozm (approved):
LGTM!

src/spoke/Spoke.sol
```diff
@@ -185,20 +185,20 @@ contract Spoke is Auth, Recoverable, ReentrancyProtection, ISpoke, IUpdateContra
...
-        }
-
-        pool.shareClasses[scId].shareToken = shareToken_;
+        if (hook != address(0)) shareToken_.file("hook", hook);
+        linkToken(poolId, scId, shareToken_);
+    }
 
+    /// @inheritdoc ISpoke
+    function linkToken(PoolId poolId, ShareClassId scId, IShareToken shareToken_) public auth {
+        pools[poolId].shareClasses[scId].shareToken = shareToken_;
```
mustermeiszer, line 201: QUESTION

Why did you remove the requires? That here would allow to override share tokens essentially still. No blocker just wanted to highlight that this is a really invasive method now.

mustermeiszer (approved):

src/spoke/Spoke.sol
wischli: I'd also prefer moving the check of L188 here as this method is `public auth`. Curious about your train of thought.

hieronx: > QUESTION

> Why did you remove the requires? That here would allow to override share tokens essentially still. No blocker just wanted to highlight that this is a really invasive method now.

The requires are still in `addShareClass` rather than `linkToken`

hieronx: > I'd also prefer moving the check of L188 here as this method is public auth. Curious about your train of thought.

That would block updating the share token, which seems like a option that we want to keep for migration purposes.

wischli (approved):

src/spoke/Spoke.sol
```diff
@@ -384,18 +384,29 @@ contract Spoke is Auth, Recoverable, ReentrancyProtection, ISpoke, IUpdateContra
         ShareClassDetails storage shareClass = _shareClass(poolId, scId);
         require(vaultFactory[factory], InvalidFactory());
 
-        // Deploy vault
         AssetIdKey memory assetIdKey = _idToAsset[assetId];
         IBaseVault vault = IVaultFactory(factory).newVault(
             poolId, scId, assetIdKey.asset, assetIdKey.tokenId, shareClass.shareToken, new address[](0)
```
lemunozm, line 389: Curious @hieronx, do you know why we're passing here an empty array of wards? Is it dead code or do we expect some wards in the future?

hieronx: Generally it allows us to add wards in the future without swapping the vault factory. Seems potentially useful

PR #395, opened 2025-05-21
Ref: UpdateContract processed in MessageProcessor

lemunozm:
Right now all `UpdateContract` messages follows this flow:

```
MessageProcessor -> Spoke -> Other contract inside Spoke side 
```

We can release `Spoke` to that duty and make the `MessageProcessor` in charge of dispatching to the correct `target`.


hieronx:
@lemunozm  I don't think we should do this. Main reason: while in our system we use `auth`, for future integrations of `IUpdateContract`, I would expect these to check that they are valid by checking `msg.sender == spoke`. That is not a good idea with the `MessageProcessor` since that contract is very likely to be swapped.

PR #396, opened 2025-05-21
Fix: TransferShares issue

lemunozm:
If `CV1` is in `chainA`, `CV2` is in `ChainB`, but `Hub` is also in `ChainA`, the `InitiateTrasferShares` message will be sent through the Gateway when it must be bypassed. The gateway will never be configured to itself, making it fail.

Other minor changes:
- `centrifugeId` as first element in the methods.
- Using a modifier to avoid any future bugs in the transaction payment, forgetting to end the payment.


hieronx (approved):

wischli (approved):
Reminder for myself to finish the hop `TransferAssets` test(s) I started last week ASAP.

PR #397, opened 2025-05-21
Ref: Remove vaults from sync managers

lemunozm:


lemunozm:
Closed due future full sync vaults could use that storage

PR #398, opened 2025-05-22
fix: blocked requests edge case

wischli:
Fixes Cantina Finding 9.

Unfortunately, it is not sufficient to reduce pending requests just by the rounded up payment instead of the rounded down one because tests have proven that
1. difference between rounded down and rounded up payments is at most one atom
2. Dust of pending requests can be arbitrarily large depending on used prices and navPerShare if there is at least one other pending deposit with relatively large amounts that eats up >= 99% of the approved amount

Therefore, this PR adds a permissioned forced cancellation ([ref Slack thread](https://kflabs.slack.com/archives/C07PG2EUR9C/p1747923670959439?thread_ts=1747923645.670269&cid=C07PG2EUR9C)): It behaves the same as `cancel{Deposit, Redeem}Request` with one minor difference:
* Precondition: Processed cancellation request for `(scId, depositAssetId, investor)` in any epoch <= the current one
   * Important Note: This applies independently of the epoch ID in which the prior cancellation was processed. I.e., if a user cancellation is processed at epoch 1, force cancellation remains enabled indefinitely ~~until it is executed~~.
* Force cancellation enabling flag is ~~reset only after force cancellation~~ never reset

As a result, either instantly resolves cancellation (if pending amount can be mutated) or queues cancellation.


src/hub/ShareClassManager.sol
```diff
@@ -814,4 +851,13 @@ contract ShareClassManager is Auth, IShareClassManager {
             queued.isCancelling
         );
     }
+
+    /// @dev A user cannot mutate their pending amount at all times because it affects the total pending amount. It is
+    /// restricted to the following three conditions:
```
hieronx: NIT: indentation of 2nd line should line up with the A of the first line

src/hub/interfaces/IShareClassManager.sol
```diff
@@ -222,6 +223,17 @@ interface IShareClassManager {
         external
         returns (uint128 cancelledAssetAmount);
 
+    /// @notice Force cancels a pending deposit request.
+    ///
```
hieronx:
```suggestion
    /// @notice Force cancels a pending deposit request.
    ///                Only allowed if the user has cancelled a request at least once before. This is to protect against
    ///                cancelling a request of a smart contract user that does not support the cancellation interface,
    ///                and would thus be unable to claim the assets.
    ///
```

src/hub/interfaces/IShareClassManager.sol
```diff
@@ -245,6 +257,18 @@ interface IShareClassManager {
         external
         returns (uint128 cancelledShareAmount);
 
+    /// @notice Force cancels a pending redeem request.
+    ///
```
hieronx:
```suggestion
    /// @notice Force cancels a pending redeem request.
    ///                Only allowed if the user has cancelled a request at least once before. This is to protect against
    ///                cancelling a request of a smart contract user that does not support the cancellation interface,
    ///                and would thus be unable to claim the shares.
    ///
```

src/hub/Hub.sol
```diff
@@ -380,6 +380,39 @@ contract Hub is Multicall, Auth, Recoverable, IHub, IHubGatewayHandler, IHubGuar
...
+        payTransaction
+    {
+        _isManager(poolId);
+
+        uint128 cancelledAssetAmount =
+            shareClassManager.forceCancelDepositRequest(poolId, scId, investor, depositAssetId);
+
+        // Cancellation might have been queued such that it will be executed in the future during claiming
+        if (cancelledAssetAmount > 0) {
+            sender.sendFulfilledDepositRequest(poolId, scId, depositAssetId, investor, 0, 0, cancelledAssetAmount);
```
lemunozm, line 420: Nice to have one message for fulfilling and cancelling. "One message to rule them all"

lemunozm (approved):
Super clean and super tested!!

src/hub/interfaces/IShareClassManager.sol
```diff
@@ -222,6 +223,20 @@ interface IShareClassManager {
         external
         returns (uint128 cancelledAssetAmount);
 
+    /// @notice Force cancels a pending deposit request.
```
hieronx, line 226: Something off here, seems the indentation is shown differently in the Github suggestion preview than in the actual code :(

src/hub/ShareClassManager.sol
wischli: [45472fa](https://github.com/centrifuge/protocol-v3/pull/398/commits/45472fad831031c490085f329e6890d5b0089892)

src/hub/ShareClassManager.sol
```diff
@@ -386,7 +426,6 @@ contract ShareClassManager is Auth, IShareClassManager {
         // NOTE: Due to precision loss, the sum of claimable user amounts is leq than the amount of minted share class
         // tokens corresponding to the approved share amount (instead of equality). I.e., it is possible for an epoch to
         // have an excess of a share class tokens which cannot be claimed by anyone.
-        // This excess is at most n-1 share tokens for an epoch with n claimable users.
```
hieronx, line 389: Any reason this was removed?

lemunozm (approved):

src/hub/ShareClassManager.sol
```diff
@@ -814,4 +847,13 @@ contract ShareClassManager is Auth, IShareClassManager {
             queued.isCancelling
         );
     }
+
+    /// @dev A user cannot mutate their pending amount at all times because it affects the total pending amount. It is
+    ///     restricted to the following three conditions:
+    ///         1. It's the first epoch (currentEpoch <= 1), which implies userOrder.lastUpdate == 0
+    ///         2. User has no pending amount (userOrder.pending == 0)
+    ///         3. User's last update is not behind the current epoch (userOrder.lastUpdate >= currentEpoch)
+    function _canMutatePending(UserOrder memory userOrder, uint32 currentEpoch) private pure returns (bool) {
```
hieronx: Ultra-nit: we haven't formalized this in the style guide, but a convention I've followed and I think we do in the rest of the c odebase as well, is use `internal` by default, and only use `private` if there is a specific reason the method should not be in scope of parent contracts.

src/hub/ShareClassManager.sol
wischli: Yeah thanks for reminding me: My tests in [cd3631a](https://github.com/centrifuge/protocol-v3/pull/398/commits/cd3631ab3a24c170fde6f1cd02787f88d178f6f5) unfortunately proved this statement was incorrect. At this stage, I was not able to define a definite upper bound on the share token excess for an arbitrary price or respectively navPerShare.

src/hub/interfaces/IShareClassManager.sol
wischli: [c33f195](https://github.com/centrifuge/protocol-v3/pull/398/commits/c33f1951f6b2850d268495d9e7c455221144c311)

src/hub/ShareClassManager.sol
wischli: [53fb3f9](https://github.com/centrifuge/protocol-v3/pull/398/commits/53fb3f92b3971c624b78874bdcddd0bffd2935ed)

wischli: There is a bunch of remaining private methods. These all perform write operations and I don't see any reason why they should be exposed to parent contracts. Do you think we should make all of them internal?

hieronx: I would make them internal since there is no inheritance here. But really no strong opinion on this, feel free to leave as is.

wischli: [3dec0f1](https://github.com/centrifuge/protocol-v3/pull/398/commits/3dec0f187c8ee0d5ec32aae815a2558d8a95de47)

hieronx (approved):

PR #401, opened 2025-05-26
DONT MERGE: use separate struct for `Delta`

hieronx:


src/spoke/BalanceSheet.sol
```diff
@@ -133,15 +133,15 @@ contract BalanceSheet is Auth, Recoverable, IBalanceSheet, IBalanceSheetGatewayH
...
+                shareQueue.delta.amount += shares;
+                shareQueue.delta.isPositive = true;
+            } else if (shareQueue.delta.amount > shares) {
+                shareQueue.delta.amount -= shares;
+                shareQueue.delta.isPositive = false;
             } else {
-                shareQueue.delta = shares - shareQueue.delta;
-                shareQueue.isPositive = true;
+                shareQueue.delta.amount = shares - shareQueue.delta.amount;
+                shareQueue.delta.isPositive = true;
```
lemunozm, line 144: Avoid this comment if it's premature.

Should not all of this be moved to `DeltaLib.add(Delta delta, uint128 amount)`

hieronx: Yes but if the gas cost increases by 21k, I don't think we should do this change. That's why I put up this PR

lemunozm: Gas increment is less than 21, so it should not be due to the read. Maybe now the optimizer is not able to optimize other thing?

PR #404, opened 2025-05-26
fix: total redeem property

wischli:
Fixes `test_property_total_pending_redeem_geq_sum_pending_user_redeem_20` on base branch

The error roots in the original invariant property defined by myself which stated
> The total pending {deposit, redeem} amount `pending{Deposit, Redeem}[..]` is always geq the sum of pending user {deposit, redeem} amounts `{deposit, redeem}Request[..]`

However, this statement is correct if you switch `geq` with `leq` (or respectively left and right side of the equation):
> The **sum of pending user {deposit, redeem} amounts `{deposit, redeem}Request[..]`** is always geq the **total pending {deposit, redeem} amount `pending{Deposit, Redeem}[..]`**

I couldn't find the corresponding property for deposits, so I just fixed the redeem code and properties table

### DIFF
--- a/test/hub/fuzzing/recon-hub/Properties.sol
+++ b/test/hub/fuzzing/recon-hub/Properties.sol
@@ -51,10 +51,10 @@ abstract contract Properties is BeforeAfter, Asserts {
         }
     }
 
-    /// @dev Property: The total pending redeem amount pendingRedeem[..] is always >= the sum of pending user redeem amounts redeemRequest[..]
+    /// @dev Property: The the sum of pending user redeem amounts redeemRequest[..] is always >= total pending redeem amount pendingRedeem[..]
     /// @dev Property: The total pending redeem amount pendingRedeem[..] is always >= the approved redeem amount epochRedeemAmounts[..].approvedShareAmount
     // TODO: come back to this to check if accounting for case is correct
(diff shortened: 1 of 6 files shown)


test/hub/fuzzing/recon-hub/Properties.sol
```diff
@@ -51,10 +51,39 @@ abstract contract Properties is BeforeAfter, Asserts {
         }
     }
 
-    /// @dev Property: The total pending redeem amount pendingRedeem[..] is always >= the sum of pending user redeem amounts redeemRequest[..]
+    /// @dev Property: The sum of pending user deposit amounts depositRequest[..] is always >= total pending deposit amount pendingDeposit[..]
+    /// @dev Property: The total pending deposit amount pendingDeposit[..] is always >= the approved deposit amount epochInvestAmounts[..].approvedAssetAmount
+    function property_sum_pending_user_deposit_geq_total_pending_deposit() public {
```
nican0r, line 56: > I couldn't find the corresponding property for deposits, so I just fixed the redeem code and properties table

added the corresponding property for deposits here, for some reason I had defined it as an inlined instead of global property so I made it global for consistency with the redeem one

nican0r (approved):

PR #405, opened 2025-05-26
File reorganization (spoke and vaults modules)

lemunozm:
### New layout

```
src/spoke
├── vaults
│  ├── factories
│  │  ├── AsyncVaultFactory.sol
│  │  └── SyncDepositVaultFactory.sol
│  ├── interfaces
│  │  ├── IAsyncVault.sol
│  │  ├── IBaseRequestManager.sol
│  │  ├── IBaseVault.sol
│  │  ├── IVaultManagers.sol
│  │  └── IVaultRouter.sol
│  ├── legacy
│  │  ├── interfaces
│  │  │  ├── IInvestmentManager.sol
│  │  │  ├── ILegacyVault.sol
│  │  │  └── ILegacyVaultAdapter.sol
│  │  └── LegacyVaultAdapter.sol
│  ├── AsyncRequestManager.sol
│  ├── AsyncVault.sol
│  ├── BaseRequestManager.sol
│  ├── BaseVaults.sol
│  ├── SyncDepositVault.sol
│  ├── SyncRequestManager.sol
│  └── VaultRouter.sol
├── factories
│  ├── interfaces
│  │  ├── IPoolEscrowFactory.sol
│  │  ├── ITokenFactory.sol
│  │  └── IVaultFactory.sol
│  ├── PoolEscrowFactory.sol
│  └── TokenFactory.sol
├── interfaces
│  ├── IBalanceSheet.sol
│  ├── IEscrow.sol
│  ├── IShareToken.sol
│  ├── ISpoke.sol
│  └── IUpdateContract.sol
├── BalanceSheet.sol
├── Escrow.sol
├── ShareToken.sol
└── Spoke.sol
```

### Changes
- Interface files along with the submodule name.
- I divided some vault files to leave the `spoke` module agnostic from `vaults`, the vaults one. It means that `vaults` module is built on top of `spoke` without bidirectionality, which I think is a great feature for integrators.
  - Wonder if now `vaults` must be in a higher level like `hooks` or `managers`.
  - Note: `BaseVault` and `BaseRequestManager` are placed in `spoke` because they are the entities `spoke` use to call `vault` stuff.

**NOTE: Despite line changes, this file only move code from one file to another**


hieronx:
> Note: BaseVault and BaseRequestManager are placed in spoke because they are the entities spoke use to call vault stuff.

What part of `BaseVault` is actually used by `spoke`? I believe it's only `vaultKind()`. I think it's certainly plausible that vault implementations in the future won't use most of this.

Likewise for `BaseRequestManager`, we could have a vault for ERC6909 assets that has none of the current ERC4626 methods like `convertToShares`

I would prefer to keep it as is.

lemunozm:
> What part of BaseVault is actually used by spoke? I believe it's only vaultKind(). I think it's certainly plausible that vault implementations in the future won't use most of this.

Also `manager()`.

> Likewise for BaseRequestManager, we could have a vault for ERC6909 assets that has none of the current ERC4626 methods like convertToShares

For the `BaseRequestManager()`. We use at least `addVault()` and `removeVault()`.


The thing is that such dependency exists. Should we add a more generic interface to abstract the common part that in all scenarios `Spoke` needs to use?

hieronx:
> The thing is that such dependency exists. Should we add a more generic interface to abstract the common part that in all scenarios Spoke needs to use?

I think this would make sense yes!

lemunozm:
Nice! Will do a PR on top of this to make it easier for the review 👍🏻.

Also interested in your thoughts @wischli given you mostly developed that part. Any name?

I think it could be the simpler and generic ones: `IVault` and `IVaultManager`

hieronx (approved):

PR #407, opened 2025-05-27
Ref: max price age messages as independent messages

lemunozm:


src/common/libraries/MessageLib.sol
```diff
@@ -11,12 +11,12 @@ import {AssetId} from "src/common/types/AssetId.sol";
...
     /// @dev Placeholder for null message type
     _Invalid,
-    // -- Root messages
+    // -- Pool independent messages
     ScheduleUpgrade,
     CancelUpgrade,
     RecoverTokens,
-    // -- Pool manager messages
     RegisterAsset,
+    // -- Pool dependent messages
```
lemunozm: I think we can reserve some values here for the case in the future we need more pool-independent messages. WDTY?

We now anyways need to have two MESSAGE_LENGTH sections, so I think it's feasible

src/spoke/Spoke.sol
```diff
@@ -263,11 +263,7 @@ contract Spoke is Auth, Recoverable, ReentrancyProtection, ISpoke, IUpdateContra
 
     /// @inheritdoc ISpokeGatewayHandler
     function updateContract(PoolId poolId, ShareClassId scId, address target, bytes memory update_) public auth {
```
hieronx: Not shadowed anymore:
```suggestion
    function updateContract(PoolId poolId, ShareClassId scId, address target, bytes memory update) public auth {
```

src/common/libraries/MessageLib.sol
hieronx: Yeah that seems like a good idea.

lemunozm: [b7f3577](https://github.com/centrifuge/protocol-v3/pull/407/commits/b7f35770f0a79253abb5ff349f2faa1e9861b562)

src/spoke/Spoke.sol
lemunozm: Such an eye!

lemunozm:
Last suggestion I would add to this PR: 
- move update restriction submessages (only the submessages) to its own `UpdateRestrictionMessageLib` inside hooks. I think we should move this out from the core `MessageLib`.
- Also move update contract submessages to its own `UpdateContractMessageLib`, not sure where, I think this should still be in `common`.

Do you agree @hieronx?

hieronx (approved):

lemunozm:
Will add in a follow up PR

PR #347, opened 2025-05-05
ref: vault & investment manager

wischli:
* Extends #343 by removing `IVaultManager` entirely and being more explicit on used vault interfaces used by the corresponding investment manager contract
* Renames `Investment` to `Request` except for Legacy v2
* Moves `vaultKind` from `IBaseRequestManager` to `IBaseVault` and removes return of request manager

Follow-up food for thought: Make `asyncRedeemManager` and `asyncManager` internal for async vaults.

PR #409, opened 2025-05-27
Ref: generic vault <-> spoke interface

lemunozm:
### Problem

We do not want to never migrate `Spoke`. Nevertheless, `Spoke` knows about `IBaseVault` and `IBaseRequestManager`, which are vaults prepared to be `ERC20` vaults and those interfaces do not support another kind of vaults as ERC6909 vaults.

### Proposed solution

Add two interfaces that are the bare minimum needed for `Spoke` to handle vaults in the most agnostic way: `IVault` and `IVaultManager`. From this on, `spoke` module doesn't know anything about `vaults` module, which can help vault integrators

### Updated `spoke` diagram:

NOTE: All vaults interactions are done over `IVault*` interfaces, so now the `spoke` diagram becomes much more readable:

image

### DIFF
--- a/src/spoke/ShareToken.sol
+++ b/src/spoke/ShareToken.sol
@@ -51,6 +51,13 @@ contract ShareToken is ERC20, IShareToken {
 
     /// @inheritdoc IShareToken
     function updateVault(address asset, address vault_) external auth {
+        updateVault(asset, 0, vault_);
+    }
+
+    /// @inheritdoc IShareToken
+    function updateVault(address asset, uint256 tokenId, address vault_) public auth {
+        require(tokenId == 0, "unsupported");
+
         vault[asset] = vault_;
         emit VaultUpdate(asset, vault_);
     }
--- a/src/spoke/ShareToken.sol
+++ b/src/spoke/ShareToken.sol
@@ -51,15 +51,6 @@ contract ShareToken is ERC20, IShareToken {
 
     /// @inheritdoc IShareToken
     function updateVault(address asset, address vault_) external auth {
-        updateVault(asset, 0, vault_);
-    }
-
-    /// @inheritdoc IShareToken
-    function updateVault(address asset, uint256 tokenId, address vault_) public auth {
-        if (tokenId != 0) {
-            revert();
-        }
-
         vault[asset] = vault_;
         emit VaultUpdate(asset, vault_);
     }
--- a/src/spoke/ShareToken.sol
+++ b/src/spoke/ShareToken.sol
@@ -56,7 +56,9 @@ contract ShareToken is ERC20, IShareToken {
 
     /// @inheritdoc IShareToken
     function updateVault(address asset, uint256 tokenId, address vault_) public auth {
-        require(tokenId == 0, "unsupported");
+        if (tokenId != 0) {
+            revert();
+        }
 
         vault[asset] = vault_;
         emit VaultUpdate(asset, vault_);
--- a/src/spoke/vaults/BaseVaults.sol
+++ b/src/spoke/vaults/BaseVaults.sol
@@ -22,13 +22,15 @@ import {IAsyncRedeemManager} from "src/spoke/vaults/interfaces/IVaultManagers.so
 import {ISyncDepositManager} from "src/spoke/vaults/interfaces/IVaultManagers.sol";
 import {IBaseRequestManager} from "src/spoke/vaults/interfaces/IBaseRequestManager.sol";
 import {IShareToken} from "src/spoke/interfaces/IShareToken.sol";
+import {IVault} from "src/spoke/interfaces/IVaultManager.sol";
+import {IVaultManager} from "src/spoke/interfaces/IVaultManager.sol";
 
 abstract contract BaseVault is Auth, Recoverable, IBaseVault {
     /// @dev Requests for Centrifuge pool are non-fungible and all have ID = 0
     uint256 internal constant REQUEST_ID = 0;
 
     IRoot public immutable root;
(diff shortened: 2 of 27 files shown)


src/spoke/ShareToken.sol
```diff
@@ -51,6 +51,13 @@ contract ShareToken is ERC20, IShareToken {
 
     /// @inheritdoc IShareToken
     function updateVault(address asset, address vault_) external auth {
+        updateVault(asset, 0, vault_);
+    }
+
+    /// @inheritdoc IShareToken
+    function updateVault(address asset, uint256 tokenId, address vault_) public auth {
```
hieronx: I think this should be stored in a new vault mapping, with the token ID, rather than override the main one

src/spoke/vaults/BaseVaults.sol
```diff
@@ -22,13 +22,15 @@ import {IAsyncRedeemManager} from "src/spoke/vaults/interfaces/IVaultManagers.so
...
+import {IVault} from "src/spoke/interfaces/IVaultManager.sol";
+import {IVaultManager} from "src/spoke/interfaces/IVaultManager.sol";
 
 abstract contract BaseVault is Auth, Recoverable, IBaseVault {
     /// @dev Requests for Centrifuge pool are non-fungible and all have ID = 0
     uint256 internal constant REQUEST_ID = 0;
 
     IRoot public immutable root;
-    IBaseRequestManager public manager;
+    IBaseRequestManager public baseManager;
```
hieronx, line 33: Why this renaming?

lemunozm: I need to return a `manager() -> IVaultManager`, but it uses a child in the inheritance: IBaseRequestManager. So I need to distinguish the storage from the getter

src/spoke/ShareToken.sol
lemunozm: Not sure if I follow. The current implementation is only for ERC20, so we should panic if it is not used as it was before, isnt'?

hieronx: Ah now I see what you mean. 

Thinking about this more now, I actually don't think we really need the vault lookup in the share token. It is anyways not something anyone will rely on. Can we just only call `shareToken.updateVault` from `Spoke` if `tokenId == 0`?

hieronx: The downside of adding here is that the share token is absolutely unmigrateable, so it should stay as simple as possible.

src/spoke/vaults/SyncRequestManager.sol
```diff
@@ -88,41 +91,44 @@ contract SyncRequestManager is BaseRequestManager, ISyncRequestManager {
...
-        override(BaseRequestManager, IBaseRequestManager)
+        override(BaseRequestManager, IVaultManager)
         auth
     {
-        super.addVault(poolId, scId, vault_, asset_, assetId);
+        super.addVault(poolId, scId, assetId, vault_, asset_, tokenId);
 
-        (, uint256 tokenId) = spoke.idToAsset(assetId);
-        setMaxReserve(poolId, scId, asset_, tokenId, type(uint128).max);
+        (, uint256 tokenId_) = spoke.idToAsset(assetId);
```
wischli, line 102: Seems redundant due to the input `tokenId`?

wischli (commented):
Its quite funny to see that we revert the IVaultManager removal of #347 but I agree with the separation. Same concerns on the ShareToken changes as @hieronx

src/spoke/ShareToken.sol
lemunozm: > I actually don't think we really need the vault lookup in the share token. It is anyways not something anyone will rely on

AFAIK, we need it because `IERC7575Share` will require having that in the shareToken

> Can we just only call shareToken.updateVault from Spoke if tokenId == 0?

Yeah, I think that's another option, but future usages for ERC6909 will never be able to use that. If you're fine with it, I'm fine too.

> The downside of adding here is that the share token is absolutely unmigrateable, so it should stay as simple as possible.

The current behavior is exactly as it was, so nothing changes in that regard. I understand we will use other `ShareToken` impl if using ERC6909 assets in the vaults. This change at least is for avoid migrating the `Spoke`

src/spoke/vaults/SyncRequestManager.sol
lemunozm: I think it should be there to check the `assetId.tokenId` corresponds with the `vault.tokenId`. The same check that happens right now in `BaseRequestManager.addVault`

lemunozm:
> Its quite funny to see that we revert the IVaultManager removal

I know, and I was also laughing about this 😆.

For me, my major issue before #347 was that they were not associated with the inheritance. But after understanding all the tricky things regarding this, I could imagine the complexities you have been dealing with. Happy we finally found something in between!

src/spoke/ShareToken.sol
hieronx: > AFAIK, we need it because IERC7575Share will require having that in the shareToken

I meant, we have the vault lookup, but only for vaults with ERC20 assets.

> Yeah, I think that's another option, but future usages for ERC6909 will never be able to use that. If you're fine with it, I'm fine too.

Yes I think we should go for that

> The current behavior is exactly as it was, so nothing changes in that regard. I understand we will use other ShareToken impl if using ERC6909 assets in the vaults. This change at least is for avoid migrating the Spoke

No I don't think we'd use another ShareToken implementation for ERC6909 assets in vaults.

hieronx (approved):

PR #415, opened 2025-05-28
test: cross-chain shares transfer

wischli:
* Adds E2e setup for three chains `B <> A <> C` (currently only A <> B supported)
* Adds test to register asset `C -> A`
* Adds test to configure pools on `{B, C} -> A`
* Adds test to transfer shares `B -> A -> C` as B and C are not linked directly, only indirectly via hub on A

### DIFF
--- a/script/CommonDeployer.s.sol
+++ b/script/CommonDeployer.s.sol
@@ -26,8 +26,6 @@ abstract contract CommonDeployer is Script, JsonRegistry {
     uint128 constant FALLBACK_MSG_COST = uint128(0.02 ether); // in Weight
     uint128 constant FALLBACK_MAX_BATCH_SIZE = uint128(10_000_000 ether); // 10M in Weight
 
-    IAdapter[] adapters;
-
     ISafe public adminSafe;
     Root public root;
     TokenRecoverer public tokenRecoverer;
@@ -113,18 +111,11 @@ abstract contract CommonDeployer is Script, JsonRegistry {
     }
 
     function wire(uint16 centrifugeId, IAdapter adapter, address deployer) public {
-        bool isDuplicate = false;
-        for (uint256 i = 0; i < adapters.length; i++) {
-            if (address(adapters[i]) == address(adapter)) {
-                isDuplicate = true;
-                break;
-            }
-        }
-        if (!isDuplicate) {
-            adapters.push(adapter);
-            IAuth(address(adapter)).rely(address(root));
-            IAuth(address(adapter)).deny(deployer);
-        }
+        IAuth(address(adapter)).rely(address(root));
+        IAuth(address(adapter)).deny(deployer);
+
+        IAdapter[] memory adapters = new IAdapter[](1);
+        adapters[0] = adapter;
 
         multiAdapter.file("adapters", centrifugeId, adapters);
     }
--- a/script/CommonDeployer.s.sol
+++ b/script/CommonDeployer.s.sol
@@ -113,10 +113,20 @@ abstract contract CommonDeployer is Script, JsonRegistry {
     }
 
     function wire(uint16 centrifugeId, IAdapter adapter, address deployer) public {
-        adapters.push(adapter);
+        bool isDuplicate = false;
+        for (uint256 i = 0; i < adapters.length; i++) {
+            if (address(adapters[i]) == address(adapter)) {
+                isDuplicate = true;
+                break;
+            }
+        }
+        if (!isDuplicate) {
+            adapters.push(adapter);
+            IAuth(address(adapter)).rely(address(root));
+            IAuth(address(adapter)).deny(deployer);
+        }
+
         multiAdapter.file("adapters", centrifugeId, adapters);
-        IAuth(address(adapter)).rely(address(root));
-        IAuth(address(adapter)).deny(deployer);
     }
 
     function removeCommonDeployerAccess(address deployer) public {
--- a/src/spoke/Spoke.sol
+++ b/src/spoke/Spoke.sol
@@ -166,7 +166,7 @@ contract Spoke is Auth, Recoverable, ReentrancyProtection, ISpoke, ISpokeGateway
         pool.createdAt = block.timestamp;
 
         IPoolEscrow escrow = poolEscrowFactory.newEscrow(poolId);
-        gateway.setRefundAddress(PoolId.wrap(poolId.raw()), escrow);
+        gateway.setRefundAddress(poolId, escrow);
 
         emit AddPool(poolId);
     }
--- a/test/integration/EndToEnd.t.sol
+++ b/test/integration/EndToEnd.t.sol
@@ -267,31 +267,29 @@ contract EndToEndUtils is EndToEndDeployment {
         vm.stopPrank();
     }
 
-    function _configurePool(CSpoke memory spoke) internal {
-        _configureAsset(spoke);
+    function _configurePool(CSpoke memory s_) internal {
+        _configureAsset(s_);
 
         if (!h.hubRegistry.exists(POOL_A)) {
             _createPool();
         }
 
         vm.startPrank(FM);
-        h.hub.notifyPool{value: GAS}(POOL_A, spoke.centrifugeId);
-        h.hub.notifyShareClass{value: GAS}(
-            POOL_A, SC_1, spoke.centrifugeId, spoke.redemptionRestrictionsHook.toBytes32()
-        );
+        h.hub.notifyPool{value: GAS}(POOL_A, s_.centrifugeId);
+        h.hub.notifyShareClass{value: GAS}(POOL_A, SC_1, s_.centrifugeId, s_.redemptionRestrictionsHook.toBytes32());
 
         h.hub.initializeHolding(
-            POOL_A, SC_1, spoke.usdcId, h.identityValuation, ASSET_ACCOUNT, EQUITY_ACCOUNT, GAIN_ACCOUNT, LOSS_ACCOUNT
+            POOL_A, SC_1, s_.usdcId, h.identityValuation, ASSET_ACCOUNT, EQUITY_ACCOUNT, GAIN_ACCOUNT, LOSS_ACCOUNT
         );
-        h.hub.updateBalanceSheetManager{value: GAS}(spoke.centrifugeId, POOL_A, BSM.toBytes32(), true);
+        h.hub.updateBalanceSheetManager{value: GAS}(s_.centrifugeId, POOL_A, BSM.toBytes32(), true);
 
         h.hub.updateSharePrice(POOL_A, SC_1, IDENTITY_PRICE);
-        h.hub.notifySharePrice{value: GAS}(POOL_A, SC_1, spoke.centrifugeId);
-        h.hub.notifyAssetPrice{value: GAS}(POOL_A, SC_1, spoke.usdcId);
+        h.hub.notifySharePrice{value: GAS}(POOL_A, SC_1, s_.centrifugeId);
+        h.hub.notifyAssetPrice{value: GAS}(POOL_A, SC_1, s_.usdcId);
         vm.stopPrank();
 
         vm.startPrank(BSM);
-        spoke.gateway.subsidizePool{value: DEFAULT_SUBSIDY}(POOL_A);
+        s_.gateway.subsidizePool{value: DEFAULT_SUBSIDY}(POOL_A);
         vm.stopPrank();
     }
 }
--- a/test/integration/EndToEnd.t.sol
+++ b/test/integration/EndToEnd.t.sol
@@ -253,7 +253,6 @@ contract EndToEndUtils is EndToEndDeployment {
     function _createPool() internal {
         vm.startPrank(address(h.guardian.safe()));
         h.guardian.createPool(POOL_A, FM, USD_ID);
-        vm.stopPrank();
 
         vm.startPrank(FM);
         h.hub.setPoolMetadata(POOL_A, bytes("Testing pool"));
@@ -286,7 +285,6 @@ contract EndToEndUtils is EndToEndDeployment {
         h.hub.updateSharePrice(POOL_A, SC_1, IDENTITY_PRICE);
         h.hub.notifySharePrice{value: GAS}(POOL_A, SC_1, s_.centrifugeId);
         h.hub.notifyAssetPrice{value: GAS}(POOL_A, SC_1, s_.usdcId);
-        vm.stopPrank();
 
         vm.startPrank(BSM);
         s_.gateway.subsidizePool{value: DEFAULT_SUBSIDY}(POOL_A);
--- a/test/integration/EndToEnd.t.sol
+++ b/test/integration/EndToEnd.t.sol
@@ -102,6 +102,7 @@ contract EndToEndDeployment is Test {
         address redemptionRestrictionsHook;
         // Others
         ERC20 usdc;
+        AssetId usdcId;
     }
 
     ISafe immutable safeAdminA = ISafe(makeAddr("SafeAdminA"));
@@ -117,7 +118,6 @@ contract EndToEndDeployment is Test {
     address immutable BSM = makeAddr("BSM");
     address immutable INVESTOR_A = makeAddr("INVESTOR_A");
     address immutable ANY = makeAddr("ANY");
-    address immutable REFUND = makeAddr("REFUND");
 
     uint128 constant INVESTOR_A_USDC_AMOUNT = 1_000_000e6; // Measured in USDC: 1M of USDC
 
@@ -129,13 +129,12 @@ contract EndToEndDeployment is Test {
     AssetId USD_ID;
     PoolId POOL_A;
     ShareClassId SC_1;
-    AssetId USDC_ID;
 
     FullDeployer deployA = new FullDeployer();
     FullDeployer deployB = new FullDeployer();
 
-    LocalAdapter adapterA;
-    LocalAdapter adapterB;
+    LocalAdapter adapterAToB;
+    LocalAdapter adapterBToA;
 
     CHub h;
     CSpoke s;
@@ -146,12 +145,12 @@ contract EndToEndDeployment is Test {
     function setUp() public virtual {
         vm.setEnv(MESSAGE_COST_ENV, vm.toString(GAS));
 
-        adapterA = _deployChain(deployA, CENTRIFUGE_ID_A, CENTRIFUGE_ID_B, safeAdminA);
-        adapterB = _deployChain(deployB, CENTRIFUGE_ID_B, CENTRIFUGE_ID_A, safeAdminB);
+        adapterAToB = _deployChain(deployA, CENTRIFUGE_ID_A, CENTRIFUGE_ID_B, safeAdminA);
+        adapterBToA = _deployChain(deployB, CENTRIFUGE_ID_B, CENTRIFUGE_ID_A, safeAdminB);
 
         // We connect both deploys through the adapters
-        adapterA.setEndpoint(adapterB);
-        adapterB.setEndpoint(adapterA);
+        adapterAToB.setEndpoint(adapterBToA);
+        adapterBToA.setEndpoint(adapterAToB);
 
         // Initialize accounts
         vm.deal(FM, 1 ether);
@@ -180,8 +179,8 @@ contract EndToEndDeployment is Test {
         POOL_A = h.hubRegistry.poolId(CENTRIFUGE_ID_A, 1);
         SC_1 = h.shareClassManager.previewNextShareClassId(POOL_A);
 
-        vm.label(address(adapterA), "AdapterA");
-        vm.label(address(adapterB), "AdapterB");
+        vm.label(address(adapterAToB), "AdapterAToB");
+        vm.label(address(adapterBToA), "AdapterBToA");
         vm.label(address(h.hub), "Hub");
     }
 
@@ -194,37 +193,41 @@ contract EndToEndDeployment is Test {
         adapter = new LocalAdapter(localCentrifugeId, deploy.multiAdapter(), address(deploy));
         deploy.wire(remoteCentrifugeId, adapter, address(deploy));
 
-        // TODO(wischli): Revert when 3-chain-setup works
+        // TODO(later): Re-enable if wire is moved to Guardian
+        //             (ref: https://github.com/centrifuge/protocol-v3/pull/415#discussion_r2121671364)
         // deploy.removeFullDeployerAccess(address(deploy));
     }
 
-    function _setSpoke(bool sameChain) internal {
-        if (s.centrifugeId != 0) return; // Already set
+    function _setSpoke(FullDeployer deploy, uint16 centrifugeId, CSpoke storage spoke) internal {
+        if (spoke.centrifugeId != 0) return; // Already set
 
-        FullDeployer deploy = (sameChain) ? deployA : deployB;
-        uint16 centrifugeId = (sameChain) ? CENTRIFUGE_ID_A : CENTRIFUGE_ID_B;
-        s = CSpoke({
-            centrifugeId: centrifugeId,
-            root: deploy.root(),
-            guardian: deploy.guardian(),
-            gateway: deploy.gateway(),
-            balanceSheet: deploy.balanceSheet(),
-            spoke: deploy.spoke(),
-            router: deploy.vaultRouter(),
-            fullRestrictionsHook: deploy.fullRestrictionsHook(),
-            redemptionRestrictionsHook: deploy.redemptionRestrictionsHook(),
-            asyncVaultFactory: address(deploy.asyncVaultFactory()).toBytes32(),
-            syncDepositVaultFactory: address(deploy.syncDepositVaultFactory()).toBytes32(),
-            asyncRequestManager: deploy.asyncRequestManager(),
-            syncRequestManager: deploy.syncRequestManager(),
-            usdc: new ERC20(6)
-        });
+        spoke.centrifugeId = centrifugeId;
+        spoke.root = deploy.root();
+        spoke.guardian = deploy.guardian();
+        spoke.gateway = deploy.gateway();
+        spoke.balanceSheet = deploy.balanceSheet();
+        spoke.spoke = deploy.spoke();
+        spoke.router = deploy.vaultRouter();
+        spoke.fullRestrictionsHook = deploy.fullRestrictionsHook();
+        spoke.redemptionRestrictionsHook = deploy.redemptionRestrictionsHook();
+        spoke.asyncVaultFactory = address(deploy.asyncVaultFactory()).toBytes32();
+        spoke.syncDepositVaultFactory = address(deploy.syncDepositVaultFactory()).toBytes32();
+        spoke.asyncRequestManager = deploy.asyncRequestManager();
+        spoke.syncRequestManager = deploy.syncRequestManager();
+        spoke.usdc = new ERC20(6);
+        spoke.usdcId = newAssetId(centrifugeId, 1);
 
         // Initialize default values
-        s.usdc.file("name", "USD Coin");
-        s.usdc.file("symbol", "USDC");
+        spoke.usdc.file("name", "USD Coin");
+        spoke.usdc.file("symbol", "USDC");
+    }
 
-        USDC_ID = newAssetId(centrifugeId, 1);
+    function _setSpoke(bool sameChain) internal {
+        if (sameChain) {
+            _setSpoke(deployA, CENTRIFUGE_ID_A, s);
+        } else {
+            _setSpoke(deployB, CENTRIFUGE_ID_B, s);
+        }
     }
 }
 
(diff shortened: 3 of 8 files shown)


test/integration/ThreeChainEndToEnd.t.sol
```diff
@@ -0,0 +1,84 @@
...
+import "forge-std/Test.sol";
+
+import "test/integration/EndToEnd.t.sol";
+
+import {IAdapter} from "src/common/interfaces/IAdapter.sol";
+
+/// @title  Three Chain End-to-End Test
+/// @notice Extends the dual-chain setup to include a third chain (C) which acts as an additional spoke
+///         Hub is on Chain A, with spokes on Chains B and C
+contract ThreeChainEndToEndDeployment is EndToEndDeployment {
```
lemunozm: You've thought exactly what I had in mind with this inheritance. Awesome!

test/integration/ThreeChainEndToEnd.t.sol
```diff
@@ -0,0 +1,91 @@
...
+            address(deployA)
+        );
+
+        // Connect Chain A to Chain C (spoke 2)
+        vm.startPrank(address(deployA.root()));
+        deployA.wire(CENTRIFUGE_ID_C, adapterAToC, address(adapterAToC));
+        vm.stopPrank();
+
+        adapterC.setEndpoint(adapterAToC);
+        adapterAToC.setEndpoint(adapterC);
```
wischli: @lemunozm Here's my setup for the adapter from A to C that leads to the `NonPayloadAdapter` error

script/CommonDeployer.s.sol
```diff
@@ -26,8 +26,6 @@ abstract contract CommonDeployer is Script, JsonRegistry {
     uint128 constant FALLBACK_MSG_COST = uint128(0.02 ether); // in Weight
     uint128 constant FALLBACK_MAX_BATCH_SIZE = uint128(10_000_000 ether); // 10M in Weight
 
-    IAdapter[] adapters;
```
wischli, line 29: Curious about your opinion to drop this storage @lemunozm as I couldn't really see its purpose. If we want to keep it, it needs to be a map IMO or be deleted at the end of `wire` which makes it redundant.

test/integration/ThreeChainEndToEnd.t.sol
```diff
@@ -0,0 +1,213 @@
...
+///         Hub is on Chain A, with spokes on Chains B and C
+contract ThreeChainEndToEndDeployment is EndToEndUtils {
+    using CastLib for *;
+
+    uint16 constant CENTRIFUGE_ID_C = 7;
+    ISafe immutable safeAdminC = ISafe(makeAddr("SafeAdminC"));
+
+    FullDeployer deployC = new FullDeployer();
+    LocalAdapter adapterC;
+    LocalAdapter adapterAToC;
```
wischli, line 40: Maybe I should also add the following to be more explicit?
```suggestion
    LocalAdapter adapterAToC;
    LocalAdapter adapterAToB = adapterA;
```

test/integration/EndToEnd.t.sol
```diff
@@ -173,6 +179,10 @@ contract EndToEndDeployment is Test {
         USD_ID = deployA.USD_ID();
         POOL_A = h.hubRegistry.poolId(CENTRIFUGE_ID_A, 1);
         SC_1 = h.shareClassManager.previewNextShareClassId(POOL_A);
+
+        vm.label(address(adapterA), "AdapterA");
+        vm.label(address(adapterB), "AdapterB");
+        vm.label(address(h.hub), "Hub");
```
lemunozm: Note for myself to do this for all addresses

test/integration/EndToEnd.t.sol
```diff
@@ -184,7 +194,8 @@ contract EndToEndDeployment is Test {
         adapter = new LocalAdapter(localCentrifugeId, deploy.multiAdapter(), address(deploy));
         deploy.wire(remoteCentrifugeId, adapter, address(deploy));
 
-        deploy.removeFullDeployerAccess(address(deploy));
+        // TODO(wischli): Revert when 3-chain-setup works
+        // deploy.removeFullDeployerAccess(address(deploy));
```
lemunozm: Missing todo?

test/integration/ThreeChainEndToEnd.t.sol
```diff
@@ -0,0 +1,213 @@
...
+        uint128 amount = 1000 * 1e18;
+
+        testConfigurePoolWithThreeChains();
+
+        // B + C: Deploy vaults on both chains
+        vm.startPrank(FM);
+        AssetId USDC_ID_B = newAssetId(sB.centrifugeId, 1);
+        h.hub.updateVault{value: GAS}(POOL_A, SC_1, USDC_ID_B, sB.asyncVaultFactory, VaultUpdateKind.DeployAndLink);
+        AssetId USDC_ID_C = newAssetId(sC.centrifugeId, 1);
+        h.hub.updateVault{value: GAS}(POOL_A, SC_1, USDC_ID_C, sC.asyncVaultFactory, VaultUpdateKind.DeployAndLink);
```
lemunozm: Do we need to deploy vaults?

test/integration/ThreeChainEndToEnd.t.sol
```diff
@@ -0,0 +1,213 @@
...
+            poolId: PoolId.unwrap(POOL_A),
+            scId: ShareClassId.unwrap(SC_1),
+            receiver: INVESTOR_A.toBytes32(),
+            amount: amount
+        }).serialize();
+        vm.expectEmit(true, false, false, false);
+        emit IMultiAdapter.HandlePayload(h.centrifugeId, bytes32(""), bytes(""), adapterC);
+        vm.expectEmit();
+        emit IERC20.Transfer(address(0), INVESTOR_A, amount);
+        emit ISpoke.ExecuteTransferShares(POOL_A, SC_1, INVESTOR_A, amount);
```
lemunozm: From EndToEnd perspective, I think we do not need to check these events. As long as the amount reaches the destination, it works. Dispatching the event is already tested in the UTs

test/integration/ThreeChainEndToEnd.t.sol
```diff
@@ -0,0 +1,213 @@
...
+        vm.stopPrank();
+        assertEq(shareTokenB.balanceOf(INVESTOR_A), amount, "Investor should have minted shares on chain B");
+
+        // B: Initiate transfer of shares
+        vm.expectEmit();
+        emit ISpoke.TransferShares(sC.centrifugeId, POOL_A, SC_1, INVESTOR_A, INVESTOR_A.toBytes32(), amount);
+        emit IHub.ForwardTransferShares(sC.centrifugeId, POOL_A, SC_1, INVESTOR_A.toBytes32(), amount);
+        vm.expectEmit(true, false, false, false);
+        emit IGateway.UnderpaidBatch(sC.centrifugeId, bytes(""));
+        vm.prank(INVESTOR_A);
```
lemunozm: Same here

test/integration/ThreeChainEndToEnd.t.sol
```diff
@@ -0,0 +1,213 @@
...
+
+        AssetId USDC_ID_B = newAssetId(sB.centrifugeId, 1);
+        AssetId USDC_ID_C = newAssetId(sC.centrifugeId, 1);
+
+        h.hub.initializeHolding(
+            POOL_A, SC_1, USDC_ID_B, h.identityValuation, ASSET_ACCOUNT, EQUITY_ACCOUNT, GAIN_ACCOUNT, LOSS_ACCOUNT
+        );
+        h.hub.initializeHolding(
+            POOL_A, SC_1, USDC_ID_C, h.identityValuation, ASSET_ACCOUNT, EQUITY_ACCOUNT, GAIN_ACCOUNT, LOSS_ACCOUNT
+        );
```
lemunozm: Do we need to configure the holdings for these tests? I think we can get rid of that part here.

test/integration/ThreeChainEndToEnd.t.sol
```diff
@@ -0,0 +1,213 @@
...
+
+        h.hub.updateBalanceSheetManager{value: GAS}(sB.centrifugeId, POOL_A, BSM.toBytes32(), true);
+        h.hub.updateBalanceSheetManager{value: GAS}(sC.centrifugeId, POOL_A, BSM.toBytes32(), true);
```
lemunozm: Same here (?)

lemunozm (commented):
Looks super great! Thanks for dealing with the adapter stuff!

Just some comments in order to minimize the test without loosing coverage. My point of view regarding this, is that we should keep ITs as minimum as possible to easy debugging/refactoring/understanding. There is nothing more obscure and annoying than having an error in an IT and trying to debug it 🥲

script/CommonDeployer.s.sol
lemunozm: Fully agree in removing this!

test/integration/ThreeChainEndToEnd.t.sol
lemunozm: I like it!

test/integration/ThreeChainEndToEnd.t.sol
wischli: What I wanted to achieve with this is to ensure all events reflecting the hop are dispatched properly because theoretically, `shareTokenC` could also exist on chain B such that just asserting amounts is not fully reliable IMO.

test/integration/ThreeChainEndToEnd.t.sol
wischli: I feel like we want to keep those lines because this test configures a pool as done in the source 2-chain e2e counterpart. It's true it's not necessary for the share transfer.

test/integration/ThreeChainEndToEnd.t.sol
lemunozm: But `shareTokenC` comes from `sC.spoke.shareToken(POOL_A, SC_1)`,  ins't? So it only exists in the "emmulated" chainC.

test/integration/EndToEnd.t.sol
wischli: Ah thanks for catching that! Now I'm running into expected issues at `deployA.wire(CENTRIFUGE_ID_C, adapterAToC, address(adapterAToC))` even when pranking the root or guardian of A. I was under the impression that should always work? Else, how would we add new adapters?

test/integration/ThreeChainEndToEnd.t.sol
lemunozm: If you want to have that test as a generic config for future testing with three chains, then ok 👍🏻

test/integration/EndToEnd.t.sol
lemunozm: That's something still WIP right now. If you need to wire other adapters, you need to not remove the deployer access. We're still discussing if `wire` should be moved to `Guardian`. 

For now, I think it's ok removing that line or leaving it commented as it is!

test/integration/ThreeChainEndToEnd.t.sol
lemunozm: My concern is, that if someone modifies something i.e. in the `Gateway/multiAdapter`, this test will be broken. And that person, without knowing about this topic, will need to understand the whole test and configuration to fix this. It's like we're adding a side effect of knowledge in the test.

I talk a bit from the frustration because I lost a lot of time with ITs for this reason. Especially with the Gateway side effects.

For me, if the funds reach the destination, the Gateway (or everything in between) have worked as expected in this regard without being explicit about that.

lemunozm:
@wischli, can I ask you a favor? Could you rename `transferShares` to `crossTransferShares()`? Is it something we talked here: https://kflabs.slack.com/archives/C04HF8SHRRC/p1748861001053769?thread_ts=1748855971.469459&cid=C04HF8SHRRC

test/integration/ThreeChainEndToEnd.t.sol
```diff
@@ -0,0 +1,213 @@
...
+        vm.stopPrank();
+
+        vm.startPrank(BSM);
+        sB.gateway.subsidizePool{value: DEFAULT_SUBSIDY}(POOL_A);
+        sC.gateway.subsidizePool{value: DEFAULT_SUBSIDY}(POOL_A);
+    }
+
+    /// @notice Test transferring shares between Chain B and Chain C via Hub A
+    /// forge-config: default.isolate = true
+    function testTransferSharesWithHop() public {
```
lemunozm: Is there any possibility to run this test with the boolean `bool sameChain`?  (same pattern as `EndToEnd.sol`)

test/integration/ThreeChainEndToEnd.t.sol
lemunozm: Not sure if we can reuse the `testConfigurePool()` method from EndToEnd and only configure here the part of spoke2 (?)

test/integration/ThreeChainEndToEnd.t.sol
wischli: Totally understand your point. Mine is kind of the opposite: At this point in time (i.e. shortly before launch), I see e2e tests as an assurance of a consistent full workflow. If a code change leads to an update of e2e tests, this signals a breaking change to consumers (i.e. the SDK, Apps and indexers). Otherwise, this could fall under our radar. Our event emission coverage is not the greatest and particularly for cross chain share transfers, we won't ever have unit tests covering all the three associated events IMO because that would be OOS for unit tests.

Before the audit, I wouldn't have discussed a second with you regarding this topic due to expected breaking changes. At this point however, we can assume minimal changes to the existing codebase and therefore I strongly argue for meaningful e2e tests that go beyond just asserting start and end amounts.

lemunozm: > At this point in time (i.e. shortly before launch), I see e2e tests as an assurance of a consistent full workflow

Yeah, I also understand you. I have a very theoretical PoV regarding tests.

> we won't ever have unit tests covering all the three associated events IMO because that would be OOS for unit tests

Yeah, that's other point to work for me. I feel we move a lot of things from UT to IT.

> At this point however, we can assume minimal changes to the existing codebase and therefore I strongly argue for meaningful e2e tests that go beyond just asserting start and end amounts.

The same we said about liquidity-pools 😛 . Kidding. I understand your point. No strong opinion. Let's see how everything evolves!

test/integration/ThreeChainEndToEnd.t.sol
wischli: Good catch, we don't need to!

test/integration/ThreeChainEndToEnd.t.sol
wischli: If it's fine with you, I would like to keep all event assertions for this particular test except for the ERC20 transfer, which I removed in cedc36443c6c6c8d063dcc3295fc49e33f96b05f.

test/integration/ThreeChainEndToEnd.t.sol
wischli: Great idea! Refactored the entire setup in [139e9e5](https://github.com/centrifuge/protocol-v3/pull/415/commits/139e9e565fc51f5fa5cda37c7c98362186b67977). Curious about your thoughts!

test/integration/ThreeChainEndToEnd.t.sol
wischli: Eventually I decided to keep it as part of `configurePool` which is now part of `EndToEndUtils` and generic for a given spoke chain.

test/integration/ThreeChainEndToEnd.t.sol
wischli: IMO, that doesn't make sense because
* If `A == B`, then we essentially test a transfer from different chains `A` to `B` 
* `A == B == C` is infeasible because we must only have one spoke per chain by design

So for me, the `ThreeChainEndToEnd` assumes chains A, B and C to be different. Tests with less than 3 actual chains should be included elsewhere. WDYT?

lemunozm: Mmm... the scenario I had in mind was that A and B are actually the same chain. That case has not been tested IIUC. So A (spoke) sends locally to B, and the B send through the gateway to to C.

To achieve that, we should rename ThreeChainEndToEnd to ThreeSidesEndToEnd or something like that. Conceptually, at least in EndToEnd, we're testing "sides"

test/integration/EndToEnd.t.sol
```diff
@@ -233,87 +236,112 @@ contract EndToEndUtils is EndToEndDeployment {
...
     }
 
-    /// forge-config: default.isolate = true
-    function testConfigurePool(bool sameChain) public {
-        testConfigureAsset(sameChain);
-
+    function _createPool() internal {
         vm.startPrank(address(h.guardian.safe()));
         h.guardian.createPool(POOL_A, FM, USD_ID);
+        vm.stopPrank();
```
lemunozm: AFAIK. There is no need to `stopPrank` if another `startPrank` is used.

test/integration/EndToEnd.t.sol
```diff
@@ -233,87 +236,112 @@ contract EndToEndUtils is EndToEndDeployment {
...
 
         h.hub.createAccount(POOL_A, ASSET_ACCOUNT, true);
         h.hub.createAccount(POOL_A, EQUITY_ACCOUNT, false);
         h.hub.createAccount(POOL_A, LOSS_ACCOUNT, false);
         h.hub.createAccount(POOL_A, GAIN_ACCOUNT, false);
+
+        vm.stopPrank();
+    }
+
+    function _configurePool(CSpoke memory spoke) internal {
```
lemunozm: Could we call it `s` to distinguish it from `spoke` contract? In order to normalize names

test/integration/ThreeChainEndToEnd.t.sol
```diff
@@ -90,81 +63,24 @@ contract ThreeChainEndToEndUseCases is ThreeChainEndToEndDeployment {
...
     }
 
     /// @notice Test transferring shares between Chain B and Chain C via Hub A
     /// forge-config: default.isolate = true
-    function testTransferSharesWithHop() public {
-        uint128 amount = 1000 * 1e18;
-
-        testConfigurePoolWithThreeChains();
+    function testCrossChainTransferShares(uint128 amount) public {
+        vm.assume(amount != 0);
```
lemunozm: NIT. Maybe for debugging, is easier not to fuzzing? I think we do not gain anything fuzzing here because the transfer is independent from the amount

test/integration/ThreeChainEndToEnd.t.sol
lemunozm: I like it! Left some comments above

test/integration/EndToEnd.t.sol
wischli: That's correct. I added `stopPrank` because in my `_configurePool` there was a `startPrank` that was not stopped in the closure of the call leading to an error that I wasn't able to quickly debug. Moreover, I prefer being explicit and there's no real overhead IMO.

test/integration/EndToEnd.t.sol
wischli: Then I will go with `s_` because `s` already exists in `EndToEnd` as the only spoke.

test/integration/ThreeChainEndToEnd.t.sol
wischli: Agree

test/integration/ThreeChainEndToEnd.t.sol
wischli: Done in [9354886](https://github.com/centrifuge/protocol-v3/pull/415/commits/93548868a78081776dfed242d35195700d051fb0)

wischli: On a note: I noticed Foundry runs those tests with a single bool param more than twice because the fuzzer is designed to test a function multiple times with different inputs where the number of times it runs a fuzzed test is primarily controlled by the `fuzz.runs` setting in our toml.

At some point (low prio IMO) we should move all such tests `test*` to internal `_test*` and manually create `test*_{True, False}` tests

test/integration/EndToEnd.t.sol
wischli: [2e11eee](https://github.com/centrifuge/protocol-v3/pull/415/commits/2e11eeeaf020dae24e8b7a6e4cdbc256060f5952)

lemunozm: Correct, thanks!

lemunozm:
I was also thinking that we should test A <-> B == C. Do you think that's easy given the current setup? I do not want to block this PR for that.

test/integration/ThreeChainEndToEnd.t.sol
lemunozm: Interesting. Thanks for notify that!

test/integration/EndToEnd.t.sol
lemunozm: I started initially adding `stopPrank` that, and felt that the readability of the tests decreased given the extra noise (Probably I'm being NIT here). Maybe can we only add `stopPrank` at the end of the methods as a middle point between you and me?

test/integration/ThreeChainEndToEnd.t.sol
```diff
@@ -0,0 +1,141 @@
...
+
+        assertEq(shareTokenB.balanceOf(INVESTOR_A), 0, "Shares should be burned on chain B");
+
+        // C: Transfer expected to be pending on A due to message being unpaid
+        IShareToken shareTokenC = IShareToken(sC.spoke.shareToken(POOL_A, SC_1));
+        assertEq(shareTokenC.balanceOf(INVESTOR_A), 0, "Share transfer not executed due to unpaid message");
+
+        // A: Before calling repay, set a refund address for the pool
+        vm.prank(address(h.root));
+        h.gateway.setRefundAddress(POOL_A, IRecoverable(h.gateway));
```
lemunozm: IIUC, this is not needed for this test

test/integration/EndToEnd.t.sol
wischli: I have a different opinion: For me `stopPrank` completes a closure. Moreover, it appears to me that in most cases, we need to close a started prank manually anyways because subsequent `startPrank`s don't seem to be that common. So IMO it is more consistent to keep that behavior: When you start pranking, you should stop it as well.

lemunozm: > because subsequent startPranks don't seem to be that common

What do you mean by that?

wischli: > What do you mean by that?

I was referring to code paths like the following in `ThreeChainEndToEnd.t.sol`. However, I take back my comment and will follow your recommendation.

```solidity

        // Connect Chain A to Chain C (spoke 2)
        adapterAToC = new LocalAdapter(CENTRIFUGE_ID_A, deployA.multiAdapter(), address(deployA));
        vm.startPrank(address(h.guardian.safe()));
        deployA.wire(CENTRIFUGE_ID_C, adapterAToC, address(adapterAToC));
        vm.stopPrank();

        adapterCToA.setEndpoint(adapterAToC);
        adapterAToC.setEndpoint(adapterCToA);


        // B: Mint shares
        vm.startPrank(address(sB.root));
        IShareToken shareTokenB = IShareToken(sB.spoke.shareToken(POOL_A, SC_1));
        shareTokenB.mint(INVESTOR_A, amount);
        vm.stopPrank();
        assertEq(shareTokenB.balanceOf(INVESTOR_A), amount, "Investor should have minted shares on chain B");
```

wischli: [2b3bf65](https://github.com/centrifuge/protocol-v3/pull/415/commits/2b3bf650a108aa90c85f347e4eac732da82f34ae)

test/integration/ThreeChainEndToEnd.t.sol
wischli: Full support for all directions as discussed on Slack fixed in: [707bf41](https://github.com/centrifuge/protocol-v3/pull/415/commits/707bf4139dfc5f4810340b768c655a5fecb767af)

test/integration/ThreeChainEndToEnd.t.sol
wischli: Without setting it, transferring from hub chain and with intermediary hub chain fails with `RefundAddressNotSet`

wischli: But you are right, it should be set as part of notifying the pool. I can also see it in the logs 😕

wischli: Apparently, the issue is that in those cases the subsidy is set for the spoke gateway which is a different address than the hub gateway. We need to call `h.gateway.repay` though as the message is pending as underpaid there.

test/integration/ThreeChainEndToEnd.t.sol
```diff
@@ -53,64 +63,40 @@ contract ThreeChainEndToEndDeployment is EndToEndUtils {
...
 
     /// @notice Test transferring shares between Chain B and Chain C via Hub A
-    /// forge-config: default.isolate = true
-    function testCrossChainTransferSharesSameChain() public {
-        _testCrossChainTransferShares(true);
-    }
-
-    /// @notice Test transferring shares between Chain B and Chain C via Hub A
-    function _testCrossChainTransferShares(bool sameChain) internal {
+    function _testCrossChainTransferShares(CrossChainDirection direction) internal {
```
lemunozm, line 105: Is the fuzzer not working with enums? Is it related to the Boolean issue you found?

lemunozm (approved):
Let's merge it!

test/integration/ThreeChainEndToEnd.t.sol
```diff
@@ -0,0 +1,202 @@
...
+            poolId: PoolId.unwrap(POOL_A),
+            scId: ShareClassId.unwrap(SC_1),
+            receiver: INVESTOR_A.toBytes32(),
+            amount: amount
+        }).serialize();
+        vm.expectEmit(true, false, false, false);
+        emit IMultiAdapter.HandlePayload(h.centrifugeId, bytes32(""), bytes(""), adapterCToA);
+        vm.expectEmit();
+        emit ISpoke.ExecuteTransferShares(POOL_A, SC_1, INVESTOR_A, amount);
+        h.gateway.repay{value: DEFAULT_SUBSIDY}(sC.centrifugeId, message);
```
lemunozm: BTW: this must be `GAS`

test/integration/EndToEnd.t.sol
```diff
@@ -222,87 +236,108 @@ contract EndToEndUtils is EndToEndDeployment {
...
+
+    function _configurePool(CSpoke memory s_) internal {
+        _configureAsset(s_);
+
+        if (!h.hubRegistry.exists(POOL_A)) {
+            _createPool();
+        }
+
+        vm.startPrank(FM);
+        h.hub.notifyPool{value: GAS}(POOL_A, s_.centrifugeId);
```
wischli, line 277: @lemunozm Reference of always notifying pool independent of the cross chain setup.

test/integration/ThreeChainEndToEnd.t.sol
wischli: The param type is irrelevant for the fuzzer. AFAIU, as long as there is at least one fuzzed variable, foundry fuzzes the test with the specified runs.

test/integration/ThreeChainEndToEnd.t.sol
```diff
@@ -0,0 +1,202 @@
...
+    }
+
+    /// @notice Test transferring shares between Chain B and Chain C via Hub A
+    function _testCrossChainTransferShares(CrossChainDirection direction) internal {
+        uint128 amount = 1e18;
+
+        _testConfigurePool(direction);
+
+        // B: Mint shares
+        vm.startPrank(address(sB.root));
```
lemunozm, line 111: I've used in other sections `vm.startPrank(DEPLOYER);` to indicate that we're somehow adding a line that can not be done in production. I would leave `vm.startPrank(address(sB.root));` for cases where actually we need root because we mimic some real use case in that way. WDYT?

test/integration/ThreeChainEndToEnd.t.sol
```diff
@@ -0,0 +1,202 @@
...
+    CSpoke sC;
+    CSpoke sB;
+
+    function setUp() public override {
+        // Call the original setUp to set up chains A and B
+        super.setUp();
+
+        // Deploy the third chain (C)
+        adapterCToA = _deployChain(deployC, CENTRIFUGE_ID_C, CENTRIFUGE_ID_A, safeAdminC);
+        _setSpoke(deployC, CENTRIFUGE_ID_C, sC);
```
lemunozm: This line is not needed, right? The spoke must be set under `setSpokes`

lemunozm: Actually I think that's the issue your're having. We set sC to deployC, and setSpoke will avoid to reset that later when configure the chains

test/integration/ThreeChainEndToEnd.t.sol
lemunozm: Ah so it performs a lot of duplications, right?

Wonder why we have specific tests for each variant instead of fuzzing the enum

wischli: > Ah so it performs a lot of duplications, right?

Yeah it runs for the configured fuzz_runs even if it exhausted all possible values.

> Wonder why we have specific tests for each variant instead of fuzzing the enum

This way, each test will run once and debugging is much smoother!

test/integration/ThreeChainEndToEnd.t.sol
wischli: Ah thanks for that catch! I thought I removed it so I didn't check again.

wischli: [02e12fd](https://github.com/centrifuge/protocol-v3/pull/415/commits/02e12fdc8b916497f127fc278558ea9511b64204)

test/integration/ThreeChainEndToEnd.t.sol
wischli: Unfortunately, the deployer is not authorized to mint shares here. We could restrict it to `sB.spoke` but not sure if we win anything by that.

src/spoke/Spoke.sol
```diff
@@ -166,7 +166,7 @@ contract Spoke is Auth, Recoverable, ReentrancyProtection, ISpoke, ISpokeGateway
         pool.createdAt = block.timestamp;
 
         IPoolEscrow escrow = poolEscrowFactory.newEscrow(poolId);
-        gateway.setRefundAddress(PoolId.wrap(poolId.raw()), escrow);
+        gateway.setRefundAddress(poolId, escrow);
```
lemunozm, line 169: My eyes can now enjoy the beautiful view

test/integration/ThreeChainEndToEnd.t.sol
lemunozm: Ahh that's right

lemunozm (approved):
Finally everything solved! Let's merge this. Thanks for fighting with the setup

PR #416, opened 2025-05-28
fix: deployer wire

wischli:
Ref: https://kflabs.slack.com/archives/C07PG2EUR9C/p1748446241285339?thread_ts=1748443446.176869&cid=C07PG2EUR9C


lemunozm (approved):
I wrote a comment in the slack thread regarding change this to a list of adapters. Anyways this is the correct solution for the current behavior so I think we can merge it.

Thanks for this!

wischli:
Closing because provided solution has flaws

PR #418, opened 2025-05-29
Tests: BalanceSheet UTs

lemunozm:
- Removed escrow erc20 variation authTransfer. Required for pointing to the correct selector from tests. Didn't find any other workaround
- Fix BalanceSheet issues: https://kflabs.slack.com/archives/C07PG2EUR9C/p1749467185844379

### DIFF
--- a/src/spoke/BalanceSheet.sol
+++ b/src/spoke/BalanceSheet.sol
@@ -201,12 +201,12 @@ contract BalanceSheet is Auth, Recoverable, IBalanceSheet, IBalanceSheetGatewayH
     /// @inheritdoc IBalanceSheet
     function submitQueuedAssets(PoolId poolId, ShareClassId scId, AssetId assetId) external authOrManager(poolId) {
         AssetQueueAmount storage assetQueue = queuedAssets[poolId][scId][assetId];
-        if (assetQueue.deposits == 0 && assetQueue.withdrawals == 0) return;
-
         ShareQueueAmount storage shareQueue = queuedShares[poolId][scId];
         D18 pricePoolPerAsset = _pricePoolPerAsset(poolId, scId, assetId);
 
-        bool isSnapshot = shareQueue.delta == 0 && shareQueue.queuedAssetCounter == 1;
+        uint32 assetCounter = (assetQueue.deposits != 0 || assetQueue.withdrawals != 0) ? 1 : 0;
+        bool isSnapshot = shareQueue.delta == 0 && shareQueue.queuedAssetCounter == assetCounter;
+
         emit SubmitQueuedAssets(
             poolId,
             scId,
@@ -233,13 +233,12 @@ contract BalanceSheet is Auth, Recoverable, IBalanceSheet, IBalanceSheetGatewayH
         assetQueue.deposits = 0;
         assetQueue.withdrawals = 0;
         shareQueue.nonce++;
-        shareQueue.queuedAssetCounter--;
+        shareQueue.queuedAssetCounter -= assetCounter;
     }
 
     /// @inheritdoc IBalanceSheet
     function submitQueuedShares(PoolId poolId, ShareClassId scId) external authOrManager(poolId) {
         ShareQueueAmount storage shareQueue = queuedShares[poolId][scId];
-        if (shareQueue.delta == 0) return;
 
         bool isSnapshot = queuedShares[poolId][scId].queuedAssetCounter == 0;
         emit SubmitQueuedShares(poolId, scId, shareQueue.delta, shareQueue.isPositive, isSnapshot, shareQueue.nonce);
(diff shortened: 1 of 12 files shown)


src/spoke/BalanceSheet.sol
```diff
@@ -1,6 +1,8 @@
 // SPDX-License-Identifier: BUSL-1.1
 pragma solidity 0.8.28;
 
+import "forge-std/Test.sol";
```
wischli: Debug remnant which needs to be removed 😄

test/spoke/unit/BalanceSheet.t.sol
```diff
@@ -0,0 +1,591 @@
...
+}
+
+contract BalanceSheetTestDeposit is BalanceSheetTest {
+    function testErrNotAuthorized() public {
+        vm.prank(ANY);
+        vm.expectRevert(IAuth.NotAuthorized.selector);
+        balanceSheet.deposit(POOL_A, SC_1, erc20, 0, AMOUNT);
+    }
+
+    function testDepositERC20(bool managerOrAuth) public {
```
wischli, line 247: Q: Wouldn't we want to ensure the escrow received `AMOUNT` as well?

test/spoke/unit/BalanceSheet.t.sol
```diff
@@ -0,0 +1,591 @@
...
+
+        vm.startPrank(AUTH);
+        balanceSheet.noteDeposit(POOL_A, SC_1, erc20, 0, AMOUNT);
+        balanceSheet.noteDeposit(POOL_A, SC_1, erc20, 0, AMOUNT);
+
+        (,, uint32 queuedAssetCounter,) = balanceSheet.queuedShares(POOL_A, SC_1);
+        assertEq(queuedAssetCounter, 1);
+
+        (uint128 deposits,) = balanceSheet.queuedAssets(POOL_A, SC_1, ASSET_20);
+        assertEq(deposits, AMOUNT * 2);
```
wischli, line 236: Q: Why are we not ensuring these as part of testing `deposit`?

test/spoke/unit/BalanceSheet.t.sol
```diff
@@ -0,0 +1,591 @@
...
+        _mockShareBurn(AMOUNT);
+        _mockShareAuthTransferFrom(AUTH, address(balanceSheet), AMOUNT);
+        _mockSendUpdateShares(AMOUNT, !IS_ISSUANCE, IS_SNAPSHOT, 0);
+
+        vm.startPrank(AUTH);
+        balanceSheet.revoke(POOL_A, SC_1, AMOUNT);
+        balanceSheet.submitQueuedShares(POOL_A, SC_1);
+    }
+
+    function testSubmitQueuedSharesAfterUpdateAssets() public {
```
wischli, line 535: Feels like we are missing assertions for this test or should it only ensure that the call order does not revert?

wischli (approved):
Thanks for the catch, the fix and the great UT refactor! A few questions from my side but definitely not blocking!

test/spoke/unit/BalanceSheet.t.sol
lemunozm: It is checked in the next line: 248

src/spoke/BalanceSheet.sol
lemunozm: Thanks for this warn!

test/spoke/unit/BalanceSheet.t.sol
lemunozm: I've omitted because `deposit()` is composed with `noteDeposit()`. So we would be testing the same code twice. It seems like it's `noteDeposit()` duty to set those variables correctly.

test/spoke/unit/BalanceSheet.t.sol
lemunozm: All _mock() calls already ensure that all params internally are the ones that are expected. The storage changes are tested in `testSubmitQueuedSharesWithDeltaPositive()`

src/spoke/BalanceSheet.sol
lemunozm: Fixed: [21b5820](https://github.com/centrifuge/protocol-v3/pull/418/commits/21b582096b0b4749d5ac60d76fe54c03fe063368)

src/spoke/Escrow.sol
```diff
@@ -30,11 +30,6 @@ contract Escrow is Auth, IEscrow {
 
         emit AuthTransferTo(asset, tokenId, receiver, amount);
     }
-
-    /// @inheritdoc IEscrow
```
hieronx, line 34: Why did we remove this? I think it's a helpful simplified API

lemunozm: > - Removed escrow erc20 variation authTransfer. Required for pointing to the correct selector from tests. Didn't find any other workaround

If you know a workaround, I can revert this

PR #420, opened 2025-05-30
Fix default gas limit

hieronx:


gpmayorga (approved):

script/CommonDeployer.s.sol
```diff
@@ -23,7 +23,7 @@ string constant MAX_BATCH_SIZE_ENV = "MAX_BATCH_SIZE";
 abstract contract CommonDeployer is Script, JsonRegistry {
     uint256 constant DELAY = 48 hours;
     bytes32 immutable SALT;
-    uint128 constant FALLBACK_MSG_COST = uint128(0.02 ether); // in Weight
```
lemunozm, line 26: I think we should not rely on this value and we should pass `MESSAGE_COST_ENV` env with the correct value. The fallback is only used for tests, actually I think to avoid confusions we should extract that from here (something pending to do by me).

PR #423, opened 2025-06-01
Fix adapter wiring

hieronx:


deployments/deploy.sh
```diff
@@ -8,23 +8,23 @@ cd "$SCRIPT_DIR" || exit
 ROOT_DIR="$(cd "$SCRIPT_DIR/.." && pwd)"
 
 # Usage: ./deploy.sh <network> <step> [forge_args...]
-# Example: ./deploy.sh sepolia fulldeploy
-# Example: ./deploy.sh base-sepolia adapters --priority-gas-price 2
-# Example: ./deploy.sh sepolia testdata --nonce 4765
+# Example: ./deploy.sh sepolia deploy:full
```
hieronx, line 11: Nit: I think this is a bit cleaner, feel free to change/ignore if you disagree.

deployments/deploy.sh
```diff
@@ -147,9 +147,10 @@ case "$STEP" in
         exit 1
     fi
 
-    run_forge_script "TestData"
+    run_forge_script "Adapters"
```
hieronx, line 150: It wasn't actually deploying adapters before.

env/base-sepolia.json
```diff
@@ -2,25 +2,20 @@
   "network": {
     "chainId": 84532,
     "environment": "testnet",
-    "centrifugeId": 2
+    "centrifugeId": 2,
+    "connectsTo": [
```
hieronx, line 6: For now it's just setting it up with the assumption that the adapters on chain A are always the same as the adapters on chain B, for every connection. That won't be true in reality, but let's consider this a next step.

env/base-sepolia.json
```diff
@@ -2,25 +2,20 @@
...
   "adapters": {
     "wormhole": {
-      "deploy": true,
-      "relayer": "0x7B1bD7a6b4E61c2a123AC6BC2cbfC614437D0470",
-      "chain-id": 10004,
-      "connectsTo": [
-        "sepolia"
-      ]
+      "wormholeId": 10004,
+      "relayer": "0x93BAD53DDfB6132b0aC8E37f6029163E63372cEE"
```
hieronx, line 13: Relayer was off, should be the one from https://wormhole.com/docs/build/reference/contract-addresses/#__tabbed_3_2

env/base-sepolia.json
```diff
@@ -2,25 +2,20 @@
...
     },
     "axelar": {
-      "deploy": false,
-      "chain-id": "axelar-testnet-lisbon-3",
+      "axelarId": "base-sepolia",
```
hieronx, line 16: Axelar ID was off, should be the one on https://testnet.axelarscan.io/resources/chains

script/WireAdapters.s.sol
```diff
@@ -65,19 +66,25 @@ contract WireAdapters is Script {
...
             // Wire WormholeAdapter
             if (wormholeAddr != address(0)) {
-                uint16 wormholeId2 = uint16(vm.parseJsonUint(config2, "$.adapters.wormhole.chain-id"));
+                address wormholeAddr2 = vm.parseJsonAddress(config2, "$.contracts.wormholeAdapter");
+                uint16 wormholeId2 = uint16(vm.parseJsonUint(config2, "$.adapters.wormhole.wormholeId"));
+
                 WormholeAdapter wormholeAdapter = WormholeAdapter(wormholeAddr);
-                wormholeAdapter.file("sources", wormholeId2, centrifugeId2, wormholeAddr);
-                wormholeAdapter.file("destinations", centrifugeId2, wormholeId2, wormholeAddr);
+                wormholeAdapter.file("sources", centrifugeId2, wormholeId2, wormholeAddr2);
```
hieronx: Main issue was that the last param here should be of the remote chain, not local chain

env/base-sepolia.json
gpmayorga: I found theese:
https://docs.axelar.dev/resources/contract-addresses/testnet/
It seems to be the same except I took chain-id literal (top of the page) and didn't consider chain-name at all
Might be worth having an env.example.json for reference of where to find this values

env/base-sepolia.json
gpmayorga: Agree, it is easier and clenaer. But will this hold true for now even in production at least for the time being?

deployments/deploy.sh
gpmayorga: Yup, why not, I think it reads better this way

deployments/deploy.sh
gpmayorga: I agree, I was trying to be as close as possible to the original deployment flow

env/base-sepolia.json
hieronx: For now I think so yes! We are also going to change how adapters are wired (per pool rather than on protocol level, so it will be setup quite differently then), so I would leave it like this until then.

env/base-sepolia.json
hieronx: That is confusing yeah! I like the example file you added, great idea!

deployments/deploy.sh
```diff
@@ -8,23 +8,23 @@ cd "$SCRIPT_DIR" || exit
...
     echo "Network options: sepolia, base-sepolia, etc. (must match env/<network>.json)"
     echo "Step options:"
-    echo "  fulldeploy    - Deploy everything (hub, spoke, adapters, wiring)"
-    echo "  adapters      - Deploy only adapters"
-    echo "  adapterwiring - Wire adapters to hub/spoke"
-    echo "  testdata      - Deploy test data"
+    echo "  deploy:full      - Deploy everything (hub, spoke, adapters)"
+    echo "  deploy:adapters  - Deploy only adapters"
+    echo "  wire:adapters    - Wire adapters to hub/spoke"
+    echo "  test            - Deploy test data"
```
lemunozm: Ultra NIT.
```suggestion
    echo "  test             - Deploy test data"
```

lemunozm (commented):
Looks really good!

PR #425, opened 2025-06-02
Add missing events

hieronx:


src/spoke/BalanceSheet.sol
```diff
@@ -90,6 +90,7 @@ contract BalanceSheet is Auth, Recoverable, IBalanceSheet, IBalanceSheetGatewayH
         } else {
             IERC6909(asset).transferFrom(msg.sender, escrow_, tokenId, amount);
         }
+        emit Deposit(poolId, scId, asset, tokenId, amount);
```
lemunozm, line 93: Each time we call a deposit, we will emit `NoteDeposit` and `Deposit`. If we're fine with that, I'm also fine

lemunozm (approved):
LGTM

src/spoke/BalanceSheet.sol
hieronx: I think this is fine, we definitely should have events for both.

src/spoke/BalanceSheet.sol
```diff
@@ -238,6 +251,7 @@ contract BalanceSheet is Auth, Recoverable, IBalanceSheet, IBalanceSheetGatewayH
         require(!root.endorsed(from), CannotTransferFromEndorsedContract());
         IShareToken token = IShareToken(spoke.shareToken(poolId, scId));
         token.authTransferFrom(from, from, to, amount);
+        emit TransferSharesFrom(poolId, scId, from, to, amount);
```
wischli, line 254: This will also lead to two emitted events:
1. `ERC20.Transfer(from, to, amount)`
2. `BalanceSheet.TransferSharesFrom(..)`

Suppose that's totally fine as the latter includes the pool id and scid.

src/spoke/Spoke.sol
```diff
@@ -142,16 +142,16 @@ contract Spoke is Auth, Recoverable, ReentrancyProtection, ISpoke, ISpokeGateway
...
             _assetCounter++;
             assetId = newAssetId(sender.localCentrifugeId(), _assetCounter);
 
             _idToAsset[assetId] = AssetIdKey(asset, tokenId);
             _assetToId[asset][tokenId] = assetId;
-
-            emit RegisterAsset(assetId, asset, tokenId, name, symbol, decimals);
         }
 
+        emit RegisterAsset(assetId, asset, tokenId, name, symbol, decimals, isInitialization);
```
wischli, line 154: Just to be sure: Do we not rather want to have a separate event for updating a previously registered asset?

wischli (approved):

src/spoke/BalanceSheet.sol
hieronx: Yeah I was 50/50 on whether this one was really needed. Always a tradeoff

src/spoke/Spoke.sol
hieronx: I think boolean is likely easier as the indexer can just add a handler for 1 event

PR #427, opened 2025-06-02
EndToEnd: fuzzing prices

lemunozm:
- [x] Perform all tests fuzzing prices
- [x] Support for infallible reciprocals
- [ ] Test with zero prices

### DIFF
--- a/test/integration/EndToEnd.t.sol
+++ b/test/integration/EndToEnd.t.sol
@@ -140,6 +140,7 @@ contract EndToEndDeployment is Test {
     CHub h;
     CSpoke s;
 
+    D18 immutable ZERO_PRICE = d18(0);
     D18 immutable IDENTITY_PRICE = d18(1, 1);
     D18 immutable TEN_PERCENT = d18(1, 10);
     D18 immutable MAX_PRICE = d18(1e24);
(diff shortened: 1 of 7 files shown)


test/integration/EndToEnd.t.sol
```diff
@@ -136,13 +142,24 @@ contract EndToEndDeployment is Test {
...
         vm.setEnv(MESSAGE_COST_ENV, vm.toString(GAS));
 
+        // Active waiting to ensure the env value is set in the OS
+        console.log(vm.envUint(MESSAGE_COST_ENV));
+
         LocalAdapter adapterA = _deployChain(deployA, CENTRIFUGE_ID_A, CENTRIFUGE_ID_B, safeAdminA);
         LocalAdapter adapterB = _deployChain(deployB, CENTRIFUGE_ID_B, CENTRIFUGE_ID_A, safeAdminB);
 
+        console.log(deployB.gasService().gasLimit(0, ""));
+
```
lemunozm, line 163: I've waited a long time dealing with the concurrence issue in EndToEnd, and it doesn't make sense to me. I thought it was due some OS behavior but seems not.

The following sometimes print the correct gas value `10` in both console lines.
But sometimes (more frecuente as more test we have) print: `10` and `500000000`. Not idea from where that `500000000` comes. Makes no sense to me... It's not even the fallback value, that is `0.02 eth`

lemunozm: `gasService.gasLimit()` just returns whatever it was configured in the `GasService` constructor. It's `GAS` value, which is `10`. Or worst case, the fallback value if setEnv was not reached by the OS before the test read that value in `CommonDeployer`.

lemunozm:
Integration tests: https://github.com/centrifuge/protocol-v3/pull/433/files
Zero pricing in another PR

PR #428, opened 2025-06-02
Ref: Normalize d18.raw

lemunozm:
We have in `D18` both `.raw()` and `.inner()` that are exactly the same. I removed `inner()` and I left `raw()` as is the pattern used for other custom defined types


wischli (approved):
Great normalization I also had on my mind!

PR #432, opened 2025-06-04
ref: normalize PricingLib return types to `uint128`

wischli:
* Normalizes return types of PricingLib functions to `uint128` as we are using solely that throughout our codebase
* Apart from a larger amount of required changes in the corresponding test function, there were no other changes necessary
   * Proves this change effectively does nothing apart from moving to uint128 conversion check layer from consumers of the lib to itself
   * Improves PricingLib fuzzed bounds (unfortunately more complex than with `uint256` return types)
* Ref creep 1: Minor precision improvement in SCM by merging two split calculations into one each

### DIFF
--- a/src/hub/ShareClassManager.sol
+++ b/src/hub/ShareClassManager.sol
@@ -174,7 +174,7 @@ contract ShareClassManager is Auth, IShareClassManager {
 
         approvedPoolAmount = PricingLib.convertWithPrice(
             approvedAssetAmount, hubRegistry.decimals(depositAssetId), hubRegistry.decimals(poolId), pricePoolPerAsset
-        ).toUint128();
+        );
 
         // Update epoch data
         EpochInvestAmounts storage epochAmounts = epochInvestAmounts[scId_][depositAssetId][nowDepositEpochId];
@@ -297,13 +297,14 @@ contract ShareClassManager is Auth, IShareClassManager {
         // NOTE: shares and pool currency have the same decimals - no conversion needed!
         payoutPoolAmount = navPoolPerShare.mulUint128(epochAmounts.approvedShareAmount, MathLib.Rounding.Down);
 
-        payoutAssetAmount = PricingLib.poolToAssetAmount(
-            payoutPoolAmount,
+        payoutAssetAmount = PricingLib.shareToAssetAmount(
+            epochAmounts.approvedShareAmount,
             hubRegistry.decimals(poolId),
             hubRegistry.decimals(payoutAssetId),
             epochAmounts.pricePoolPerAsset,
+            epochAmounts.navPoolPerShare,
             MathLib.Rounding.Down
-        ).toUint128();
+        );
         revokedShareAmount = epochAmounts.approvedShareAmount;
 
         epochAmounts.payoutAssetAmount = payoutAssetAmount;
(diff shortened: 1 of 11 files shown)


src/hub/ShareClassManager.sol
```diff
@@ -297,13 +297,14 @@ contract ShareClassManager is Auth, IShareClassManager {
...
-            payoutPoolAmount,
+        payoutAssetAmount = PricingLib.shareToAssetAmount(
+            epochAmounts.approvedShareAmount,
             hubRegistry.decimals(poolId),
             hubRegistry.decimals(payoutAssetId),
             epochAmounts.pricePoolPerAsset,
+            epochAmounts.navPoolPerShare,
             MathLib.Rounding.Down
-        ).toUint128();
+        );
```
wischli, line 307: Same calculation but in 1 step instead of 2 which is more precise. Forgot to update this in last pricing improvements.

src/hub/ShareClassManager.sol
```diff
@@ -423,14 +424,14 @@ contract ShareClassManager is Auth, IShareClassManager {
...
                 paymentAssetAmount,
                 hubRegistry.decimals(depositAssetId),
                 hubRegistry.decimals(poolId),
-                epochAmounts.pricePoolPerAsset
+                epochAmounts.pricePoolPerAsset,
+                epochAmounts.navPoolPerShare,
+                MathLib.Rounding.Down
             );
-            payoutShareAmount =
-                epochAmounts.navPoolPerShare.reciprocalMulUint256(paymentPoolAmount, MathLib.Rounding.Down).toUint128();
```
wischli, line 433: Same as above

lemunozm (approved):
Really thanks for this change!!

Just a question, `AsyncRequestManager` and `SyncRequestManager` still uses some uint256 for shares and assets. Should we also change that?

lemunozm:
A request that maybe can be added quickly to this PR 🙏🏻 :

`assetToSharesAmount` and `shareToAssetAmount` doesn't follow the same order in their params, which for me has lead me to confusion in some usages, and maybe for future API users too. Could we normalize that? Right now it is as follows:

```solidity
    function assetToShareAmount(
        uint256 assetAmount,
        uint8 assetDecimals,
        uint8 shareDecimals,
        D18 pricePoolPerAsset,
        D18 pricePoolPerShare,
        MathLib.Rounding rounding
    ) internal pure returns (uint128 shareAmount) {
        return
            convertWithPrices(assetAmount, assetDecimals, shareDecimals, pricePoolPerAsset, pricePoolPerShare, rounding);
    }

    /// @dev Converts share amount to asset asset amount.
    ///
    ///      NOTE: Pool and share denomination are always equal by design
    function shareToAssetAmount(
        uint256 shareAmount,
        uint8 shareDecimals,
        uint8 assetDecimals,
        D18 pricePoolPerAsset,
        D18 pricePoolPerShare,
        MathLib.Rounding rounding
    ) internal pure returns (uint128 assetAmount) {
        return
            convertWithPrices(shareAmount, shareDecimals, assetDecimals, pricePoolPerShare, pricePoolPerAsset, rounding);
    }

```

Decimals do not follow the same order in both methods. I would either:
- Use the same order for both methods
- Interchange the order of prices in shareToAssetAmount, to have the params of that method in opposite directions (maybe this makes more sense, given de conversion direction)

wischli:
> A request that maybe can be added quickly to this PR 🙏🏻 :

I understand your point but IMO this is not a change worth our time right now due to the upcoming audits and higher priority on extending e2e and invariant tests. Moreover, this will increase the changes diff outside of tests by orders of magnitude which has to be reviewed by the auditors.

Is that acceptable for you @lemunozm? If at all, I would change the order of prices.

lemunozm:
I agree to doing this in a future PR. But I think we should change this before the deployment. I wasted quite a lot of time realizing what was wrong in my case, and I could imagine some future integrators too.

PR #433, opened 2025-06-05
Tests: add more end-to-end tests

lemunozm:
### Current state of End to End testing

I think we already have the most common flows:

- pool configuration flow
- basic fund management flow
- async deposit flow 
- sync deposit flow
- async deposit flow + fund management
- sync deposit flow + fund management
- async deposit flow + async redeem flow
- sync deposit flow + async redeem flow
- async deposit flow + async redeem flow + fund management
- sync deposit flow + async redeem flow + fund management
- async deposit cancellation flow
- sync deposit + async redeem cancelation flow
- async deposit + async redeem cancelation flow

All of them tested with local and remote (emulated) chains through the gateway and different prices for assets and shares.

### Next steps. 

Same flows but mixing more stuff:

- Play with different investors
- Play with different approval amounts
- Play with different prices for deposit vs redeem

Add more checks to the end of existing tests

### Production changes
- Added `PricingLib.assetToPool`, useful for integrators.
- Fixed 3 issues from BSM: 
  - Manuel issue regarding snapshots: https://kflabs.slack.com/archives/C04HF8SHRRC/p1748945986778699
  - nonce incremented in all messages: https://kflabs.slack.com/archives/C07PG2EUR9C/p1749055334179079
  - uUsent when deposits == withdrawals: https://kflabs.slack.com/archives/C07PG2EUR9C/p1749110746190999

### DIFF
--- a/src/common/libraries/PricingLib.sol
+++ b/src/common/libraries/PricingLib.sol
@@ -251,6 +251,17 @@ library PricingLib {
         return convertWithReciprocalPrice(poolAmount, poolDecimals, assetDecimals, pricePoolPerAsset, rounding);
     }
 
+    /// @dev Converts asset amount to pool amount.
+    function assetToPoolAmount(
+        uint256 assetAmount,
+        uint8 assetDecimals,
+        uint8 poolDecimals,
+        D18 pricePoolPerAsset,
+        MathLib.Rounding rounding
+    ) internal pure returns (uint256 poolAmount) {
+        return convertWithPrice(assetAmount, assetDecimals, poolDecimals, pricePoolPerAsset, rounding);
+    }
+
     /// @dev Returns the asset price per share denominated in ASSET_UNIT/SHARE_UNIT
     ///
     ///      NOTE: Should never be used for calculating amounts due to precision loss. Instead, please refer to
--- a/src/common/libraries/PricingLib.sol
+++ b/src/common/libraries/PricingLib.sol
@@ -253,12 +253,12 @@ library PricingLib {
 
     /// @dev Converts asset amount to pool amount.
     function assetToPoolAmount(
-        uint256 assetAmount,
+        uint128 assetAmount,
         uint8 assetDecimals,
         uint8 poolDecimals,
         D18 pricePoolPerAsset,
         MathLib.Rounding rounding
-    ) internal pure returns (uint256 poolAmount) {
+    ) internal pure returns (uint128 poolAmount) {
         return convertWithPrice(assetAmount, assetDecimals, poolDecimals, pricePoolPerAsset, rounding);
     }
 
--- a/src/spoke/BalanceSheet.sol
+++ b/src/spoke/BalanceSheet.sol
@@ -204,7 +204,7 @@ contract BalanceSheet is Auth, Recoverable, IBalanceSheet, IBalanceSheetGatewayH
         ShareQueueAmount storage shareQueue = queuedShares[poolId][scId];
         D18 pricePoolPerAsset = _pricePoolPerAsset(poolId, scId, assetId);
 
-        bool isSnapshot = shareQueue.delta == 0 && shareQueue.queuedAssetCounter == 0;
+        bool isSnapshot = shareQueue.delta == 0 && shareQueue.queuedAssetCounter == 1;
         emit SubmitQueuedAssets(
             poolId,
             scId,
@@ -215,29 +215,18 @@ contract BalanceSheet is Auth, Recoverable, IBalanceSheet, IBalanceSheetGatewayH
             isSnapshot,
             shareQueue.nonce
         );
-        if (assetQueue.deposits > assetQueue.withdrawals) {
-            sender.sendUpdateHoldingAmount(
-                poolId,
-                scId,
-                assetId,
-                assetQueue.deposits - assetQueue.withdrawals,
-                pricePoolPerAsset,
-                true,
-                isSnapshot,
-                shareQueue.nonce
-            );
-        } else if (assetQueue.withdrawals > assetQueue.deposits) {
-            sender.sendUpdateHoldingAmount(
-                poolId,
-                scId,
-                assetId,
-                assetQueue.withdrawals - assetQueue.deposits,
-                pricePoolPerAsset,
-                false,
-                isSnapshot,
-                shareQueue.nonce
-            );
-        }
+        sender.sendUpdateHoldingAmount(
+            poolId,
+            scId,
+            assetId,
+            (assetQueue.deposits >= assetQueue.withdrawals)
+                ? assetQueue.deposits - assetQueue.withdrawals
+                : assetQueue.withdrawals - assetQueue.deposits,
+            pricePoolPerAsset,
+            assetQueue.deposits >= assetQueue.withdrawals,
+            isSnapshot,
+            shareQueue.nonce
+        );
 
         assetQueue.deposits = 0;
         assetQueue.withdrawals = 0;
--- a/test/integration/EndToEnd.t.sol
+++ b/test/integration/EndToEnd.t.sol
@@ -275,6 +275,12 @@ contract EndToEndUtils is EndToEndDeployment {
             poolAmount, POOL_DECIMALS, USDC_DECIMALS, ASSET_PRICE, MathLib.Rounding.Down
         ).toUint128();
     }
+
+    function checkAccountValue(AccountId accountId, uint128 value, bool isPositive) public view {
+        (bool accountIsPositive, uint128 accountValue) = h.accounting.accountValue(POOL_A, accountId);
+        assertEq(accountValue, value);
+        assertEq(accountIsPositive, isPositive);
+    }
 }
 
 /// Common and generic flows ready to be used in different tests
@@ -419,11 +425,7 @@ contract EndToEndUseCases is EndToEndFlows {
 
     /// forge-config: default.isolate = true
     function testAsyncRedeem(bool sameChain, bool afterAsyncDeposit) public {
-        if (afterAsyncDeposit) {
-            testAsyncDeposit(sameChain);
-        } else {
-            testSyncDeposit(sameChain);
-        }
+        (afterAsyncDeposit) ? testAsyncDeposit(sameChain) : testSyncDeposit(sameChain);
 
         vm.startPrank(FM);
         h.hub.updateRestriction{value: GAS}(POOL_A, SC_1, s.centrifugeId, _updateRestrictionMemberMsg(INVESTOR_A));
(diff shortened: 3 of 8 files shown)


src/common/libraries/PricingLib.sol
```diff
@@ -251,6 +251,17 @@ library PricingLib {
...
+    /// @dev Converts asset amount to pool amount.
+    function assetToPoolAmount(
+        uint128 assetAmount,
+        uint8 assetDecimals,
+        uint8 poolDecimals,
+        D18 pricePoolPerAsset,
+        MathLib.Rounding rounding
+    ) internal pure returns (uint128 poolAmount) {
+        return convertWithPrice(assetAmount, assetDecimals, poolDecimals, pricePoolPerAsset, rounding);
+    }
```
lemunozm, line 263: @wischli I tried to add a test for this following the same pattern as the opposite one, but I din't feel able to do it in a reasonable amount of time. Seems like you went really deep in the maths in those tests 🙌🏻 

Could be a 5-min task for you to do it? I would really appreciate it

wischli: I will take a look but it will probably take longer than 5 minutes. FWIW, fixing the PricingLib tests for #432 took multiple hours and was really really cumbersome.

src/spoke/BalanceSheet.sol
```diff
@@ -199,29 +199,18 @@ contract BalanceSheet is Auth, Recoverable, IBalanceSheet, IBalanceSheetGatewayH
...
-        }
+        sender.sendUpdateHoldingAmount(
+            poolId,
+            scId,
+            assetId,
+            (assetQueue.deposits >= assetQueue.withdrawals)
+                ? assetQueue.deposits - assetQueue.withdrawals
+                : assetQueue.withdrawals - assetQueue.deposits,
+            pricePoolPerAsset,
+            assetQueue.deposits >= assetQueue.withdrawals,
```
wischli, line 226: Q: Previously we didn't send an update for `==`, now we do. Is this intended?

test/integration/EndToEnd.t.sol
```diff
@@ -139,8 +147,17 @@ contract EndToEndDeployment is Test {
...
     D18 immutable IDENTITY_PRICE = d18(1, 1);
     D18 immutable TEN_PERCENT = d18(1, 10);
+    D18 immutable ASSET_PRICE = d18(1, 2);
+    D18 immutable SHARE_PRICE = d18(4, 1);
+
+    uint8 constant USDC_DECIMALS = 6;
+    uint8 constant POOL_DECIMALS = 18;
+    uint8 constant SHARE_DECIMALS = POOL_DECIMALS;
+
+    uint256 constant NO_USED = 0;
```
wischli: Nit: Can we make the naming more explicit? Seems like it refers to an amount

```suggestion
    uint256 constant ZERO_AMOUNT = 0;
```

test/integration/EndToEnd.t.sol
```diff
@@ -243,11 +296,12 @@ contract EndToEndUtils is EndToEndDeployment {
...
-    function _configureAsset(CSpoke memory spoke) internal {
-        spoke.usdc.mint(INVESTOR_A, INVESTOR_A_USDC_AMOUNT);
-        spoke.spoke.registerAsset{value: GAS}(h.centrifugeId, address(spoke.usdc), 0);
+    function _configureAsset(CSpoke memory s_) internal {
+        vm.startPrank(DEPLOYER);
+        s_.usdc.mint(INVESTOR_A, USDC_AMOUNT_1);
 
-        assertEq(h.hubRegistry.decimals(spoke.usdcId), 6, "expected decimals");
+        vm.startPrank(ANY);
+        s_.spoke.registerAsset{value: GAS}(h.centrifugeId, address(s_.usdc), 0);
```
wischli, line 304: Nit: Please add stop the prank as part of function closure. Otherwise this can lead to errors which are not easy to debug

```suggestion
        s_.spoke.registerAsset{value: GAS}(h.centrifugeId, address(s_.usdc), 0);
        vm.stopPrank();
```

test/integration/EndToEnd.t.sol
```diff
@@ -278,121 +332,164 @@ contract EndToEndUtils is EndToEndDeployment {
...
+
+        /// We subsidize the hub using the local spoke deployment
+        if (s.centrifugeId != h.centrifugeId) {
+            vm.startPrank(FM);
+            h.hub.notifyPool{value: GAS}(POOL_A, h.centrifugeId);
+            h.hub.updateBalanceSheetManager{value: GAS}(h.centrifugeId, POOL_A, BSM.toBytes32(), true);
+
+            vm.startPrank(BSM);
+            h.gateway.subsidizePool{value: DEFAULT_SUBSIDY}(POOL_A);
+        }
```
wischli, line 358: This can be removed because we execute those steps already in the above `_configurePool(s)`. Just tested it on your branch without this code path and all tests succeed.

test/integration/EndToEnd.t.sol
```diff
@@ -278,121 +332,164 @@ contract EndToEndUtils is EndToEndDeployment {
...
+    }
+
+    function _configurePrices(D18 assetPrice, D18 sharePrice) internal {
+        h.valuation.setPrice(s.usdcId, USD_ID, assetPrice);
+
+        vm.startPrank(FM);
+        h.hub.updateSharePrice(POOL_A, SC_1, sharePrice);
+        h.hub.notifySharePrice{value: GAS}(POOL_A, SC_1, s.centrifugeId);
+        h.hub.notifyAssetPrice{value: GAS}(POOL_A, SC_1, s.usdcId);
```
wischli, line 367: Nit: Please add stop the prank as part of function closure. Otherwise this can lead to errors which are not easy to debug

```suggestion
        h.hub.notifyAssetPrice{value: GAS}(POOL_A, SC_1, s.usdcId);
        vm.stopPrank();
```

src/common/libraries/PricingLib.sol
```diff
@@ -251,6 +251,17 @@ library PricingLib {
         return convertWithReciprocalPrice(poolAmount, poolDecimals, assetDecimals, pricePoolPerAsset, rounding);
     }
 
+    /// @dev Converts asset amount to pool amount.
```
hieronx, line 254: If this function is only used in tests, I don't think it should be in the src contract. It should be a test helper then. Otherwise it is exposed to audits etc.

src/spoke/BalanceSheet.sol
```diff
@@ -199,29 +199,18 @@ contract BalanceSheet is Auth, Recoverable, IBalanceSheet, IBalanceSheetGatewayH
...
-                scId,
-                assetId,
-                assetQueue.withdrawals - assetQueue.deposits,
-                pricePoolPerAsset,
-                false,
-                isSnapshot,
-                shareQueue.nonce
-            );
-        }
+        sender.sendUpdateHoldingAmount(
```
hieronx, line 218: Actually a lot cleaner to not have the if else statement, good idea

src/common/libraries/PricingLib.sol
lemunozm: My thought is that this can be quite interesting for integrators. So if we have this audited, they can rely on it. Soon or later they will need that conversion. Especially when dealing with accounting.

WDYT?

wischli (approved):
Thanks for giving us more confidence about the code and covering most happy paths! A few nits from my side.

src/common/libraries/PricingLib.sol
hieronx: That is a fair point!

src/spoke/BalanceSheet.sol
hieronx: See this Slack thread: https://kflabs.slack.com/archives/C07PG2EUR9C/p1749110746190999

test/integration/EndToEnd.t.sol
lemunozm: I named NO_USED, because that constant is used in parameters that are not really used. So you can set the value you want. Just to notify the reader. 

Maybe another name like... `ANY_VALUE` represents more of what I want to transmit?

test/integration/EndToEnd.t.sol
lemunozm: I think I've removed by error after some refactors but I think those issues are fixed now because the `mint()` part of that method is now under `vm.startPrank(DEPLOYER);`

I feel like it's scary to use `stopPrank()` in the sense that the caller becomes `DEPLOYER` again, which means that you can call anything without blockers. But from IT you want to ensure the correct actor before each call. Actually the error you experienced I think is the correct error because you were calling a method without using the correct actor

test/integration/EndToEnd.t.sol
lemunozm: Are you sure you tested with the updated branch? It fails to me.

Note that this step is for the local chain attached to the hub when it's configured with `!sameChain`

src/common/libraries/PricingLib.sol
lemunozm: OMG! We can leave this for post-deployment actually. It's been tested already under EndToEnd and will be audited

test/integration/EndToEnd.t.sol
lemunozm: For me should be a pattern the fact of calling `startPrank()` before any API call to ensure the correct actor and never rely on what is set by default

test/integration/EndToEnd.t.sol
wischli: I like that or `PLACEHOLDER` to indicate its irrelevant.

lemunozm: Good one!

test/integration/EndToEnd.t.sol
wischli: Damn you are right, I ran the tests in my cloned repo which I used for testing Guillermos deployments.

test/integration/EndToEnd.t.sol
wischli: I see and understand that. I still think we should we should stop pranks when a closure ends because we don't follow your pattern anywhere else in the codebase so far. But no strong opinion!

lemunozm: I agree it doesn't follow the pattern right now. This is something I was thinking about in the offsite to avoid by default have the deployer in the tests. In the incoming BalanceSheet UTs I already apply that.

hieronx (approved):

test/integration/EndToEnd.t.sol
lemunozm: What is unneeded, I just realize right now, is the `updateBalanceSheetManager` call. Will remove in a future PR

PR #434, opened 2025-06-05
ref: switch price param order in shareToAssetAmount

wischli:
Follow up from #432 requested by @lemunozm ([ref](https://github.com/centrifuge/protocol-v3/pull/432#issuecomment-2941707773)) which turned out to be smaller than expected, thus not postponed


lemunozm (approved):
Thank you so much for this!! I thought there were more consumers

PR #435, opened 2025-06-05
Add `reserve/unreserve` methods to `BalanceSheet`

hieronx:


src/spoke/factories/PoolEscrowFactory.sol
```diff
@@ -37,9 +33,7 @@ contract PoolEscrowFactory is Auth, IPoolEscrowFactory {
 
         escrow_.rely(root);
         escrow_.rely(gateway);
-        escrow_.rely(spoke);
```
hieronx, line 40: Wasn't used at all

src/vaults/AsyncRequestManager.sol
```diff
@@ -42,7 +41,6 @@ contract AsyncRequestManager is BaseRequestManager, IAsyncRequestManager {
     using MathLib for uint256;
 
     ISpokeMessageSender public sender;
-    IBalanceSheet public balanceSheet;
```
hieronx, line 45: Now added to `BaseRequestManager` since both contracts use this anyways

src/vaults/AsyncRequestManager.sol
```diff
@@ -58,7 +56,6 @@ contract AsyncRequestManager is BaseRequestManager, IAsyncRequestManager {
         if (what == "sender") sender = ISpokeMessageSender(data);
         else if (what == "spoke") spoke = ISpoke(data);
         else if (what == "balanceSheet") balanceSheet = IBalanceSheet(data);
-        else if (what == "poolEscrowProvider") poolEscrowProvider = IPoolEscrowProvider(data);
```
hieronx, line 61: I think it's cleaner to just rely on `balanceSheet.escrow()`

src/spoke/BalanceSheet.sol
```diff
@@ -128,6 +128,22 @@ contract BalanceSheet is Auth, Recoverable, IBalanceSheet, IBalanceSheetGatewayH
         escrow_.authTransferTo(asset, tokenId, receiver, amount);
     }
 
+    /// @inheritdoc IBalanceSheet
+    function reserveIncrease(PoolId poolId, ShareClassId scId, address asset, uint256 tokenId, uint128 amount)
+        public
+        authOrManager(poolId)
+    {
+        escrow(poolId).reserveIncrease(scId, asset, tokenId, amount);
```
lemunozm: Q: Isn't the name flipped? 🤔

lemunozm (approved):
Much more cleaner actually!

src/spoke/BalanceSheet.sol
hieronx: Yeah I thought this too, just matched the escrow naming. Not sure if @mustermeiszer had a specific reason for that.

mustermeiszer: No strong opinion!

hieronx: @lemunozm I would actually rename to `reserve` and `unreserve`, also in the escrow. Wdyt ?

src/vaults/BaseRequestManager.sol
```diff
@@ -118,7 +118,7 @@ abstract contract BaseRequestManager is Auth, Recoverable, IBaseRequestManager {
 
     /// @inheritdoc IBaseRequestManager
     function poolEscrow(PoolId poolId) public view returns (IPoolEscrow) {
-        return poolEscrowProvider.escrow(poolId);
+        return balanceSheet.escrow(poolId);
```
wischli, line 121: Curious why you prefer this as this adds ~700 gas

wischli (approved):

src/vaults/BaseRequestManager.sol
lemunozm: I think pool escrow provider is no longer a ward of the request manager

lemunozm (approved):

src/vaults/BaseRequestManager.sol
hieronx: > Curious why you prefer this as this adds ~700 gas

How do you get to this value?

It adds an external call but since we are already doing other external calls to the balance sheet in most methods anyways, it shouldn't add any gas cost I believe.

Reason I prefer it is just being able to rely fully on the balance sheet from the request managers perspective.

src/spoke/BalanceSheet.sol
```diff
@@ -128,6 +128,22 @@ contract BalanceSheet is Auth, Recoverable, IBalanceSheet, IBalanceSheetGatewayH
...
+    }
+
+    /// @inheritdoc IBalanceSheet
+    function unreserve(PoolId poolId, ShareClassId scId, address asset, uint256 tokenId, uint128 amount)
+        public
+        authOrManager(poolId)
+    {
+        escrow(poolId).unreserve(scId, asset, tokenId, amount);
+    }
+
```
mustermeiszer, line 146: @hieronx why do we need those? We had this logic already beforehand and agreed on not needing it.

hieronx: This is now used by the `asyncRequestManager`?

lemunozm: To extend, it's a way of avoiding the asyncRequestManager ward on escrows

PR #436, opened 2025-06-05
tests: add assetToPool

wischli:
Ref: https://github.com/centrifuge/protocol-v3/pull/433#discussion_r2128617825

In hindsight, it really annoys me that I didn't think of inheriting the test contracts for `assetToShare`, `shareToAsset` and `poolToAsset` before which just utilize either `convertWithPrice`, `convertWithPrices` or `convertWithReciprocalPrice` under the hood 🙈


lemunozm (approved):
Thank you so much!!!

PR #439, opened 2025-06-08
Decouple request managers

hieronx:
* Add new `RequestMessageLib`  and `RequestCallbackMessageLib`. These are sub-messages, that are forwarded by `Spoke` contract on the spoke side
* Add `setRequestManager(poolId, scId, manager)`  message and storage in `Spoke`
  * `mapping(poolId => mapping (scId => manager)) requestManager`
  * This is the entity that can trigger `Request` messages spoke => hub, and that will receive `RequestCallback` messages hub => spoke.
* Remove all top level messages, and the corresponding dispatchers and processors. And remove `vaultFactories` lookup.
* Add process logic for `RequestCallback` in `asyncRequestManager`

Slighlty simplified explanation of changes:

### Before

```mermaid
graph LR
    subgraph "HUB"
        HUB1["Hardcoded messaging<br/>• requestDeposit()<br/>• cancelDepositRequest()<br/>• requestRedeem()<br/>• cancelRedeemRequest()<br/>• ..."]
    end

    subgraph "VAULTS"
        ARM1["AsyncRequestManager<br/>• Fixed implementation<br/>• Cannot be swapped easily<br/>• Ward permissions required"]
    end

    HUB1 <==>|"Basically Request Messages<br/>RequestCallback Messages"| ARM1
```

### After

```mermaid
graph LR
    subgraph "HUB"
        HUB2["Generic messaging<br/>• request()<br/>• requestCallback()"]
    end

    subgraph "SPOKE"
        RR["Request Router<br/>• requestManager mapping<br/>• Dynamic routing<br/>• setRequestManager()<br/>• request(payload)<br/>• requestCallback(payload)<br/>• Configurable per (pool, scId)"]
    end

    subgraph "VAULTS"
        ARM2["AsyncRequestManager<br/>• Implements interface<br/>• Pluggable<br/>• No ward required"]
        CRM["CustomRequestManager<br/>• Custom logic<br/>• Pool-specific<br/>• Epoch-based"]
    end

    HUB2 <==>|"Request Messages<br/>RequestCallback Messages"| RR
    RR --> |"Route to configured request mgr"| ARM2
    RR --> |"Route to configured request mgr"| CRM
```

# Breaking Changes

This PR introduces breaking changes to request manager architecture. The SDK's ABI interactions, message parsing, and event handling will be affected.

### 1. **New Hub Methods for Request Manager Setup**
- `hub.setRequestManager(poolId, scId, assetId, manager)`
- `hub.updateBalanceSheetManager(spokeId, poolId, manager, authorized)`
- **SDK Impact**: Add these methods to Hub ABI
  - Example Async Vault: https://github.com/centrifuge/protocol-v3/blob/e5bb69cddbc85f44a743c08b37e8823dc9dc0d59/script/TestData.s.sol#L98-L99
  - Example Sync Vault: https://github.com/centrifuge/protocol-v3/blob/e5bb69cddbc85f44a743c08b37e8823dc9dc0d59/script/TestData.s.sol#L188-L190
- **Required Setup**: Setting the request manager happens in Hub.sol and requires both method calls for proper configuration

#### RequestMessage Wrapper (New)
- **Before**: Direct message types sent individually
  - `DepositRequest`
  - `RedeemRequest` 
  - `CancelDepositRequest`
  - `CancelRedeemRequest`
- **After**: All request messages wrapped in `RequestMessage` with nested message types
- **Structure**: `RequestMessage` contains message type + serialized nested message
- **SDK Impact**: Update request message parsing to handle wrapper + nested structure

#### RequestCallbackMessageLib (New)
- **Before**: Callback messages part of core message library
  - `ApprovedDeposits`
  - `IssuedShares`
  - `RevokedShares`
  - `FulfilledDepositRequest`
  - `FulfilledRedeemRequest`
- **After**: Separate `RequestCallbackMessageLib` with own message types
- **Structure**: Callback messages now have dedicated library with type enum + serialized data
- **SDK Impact**: Update callback message parsing to use new library structure

### New Events
- `RequestManagerSet` - Emitted when setting request managers via Hub
- `BalanceSheetManagerUpdated` - Emitted when updating balance sheet authorization

### Renamed/Modified Events
- **Request events**: Now emit wrapper message instead of direct message content
- **Callback events**: Now reference `RequestCallbackMessageLib` types instead of core message types
- **Message routing events**: New events for manager-based routing

### ABI Updates
1. **Hub contract**: Add `setRequestManager()` and `updateBalanceSheetManager()` methods
2. **Message libraries**: 
  - Add `RequestMessage` wrapper ABI
  - Add `RequestCallbackMessageLib` ABI
3. **Event definitions**: Update for new/renamed events

### Message Parsing Changes
1. **Request messages**: 
  - **Old**: Parse message type directly
  - **New**: Parse `RequestMessage` wrapper → extract type → parse nested message
2. **Callback messages**:
  - **Old**: Parse from core message library
  - **New**: Parse from `RequestCallbackMessageLib` with separate type system

### Transaction Flows
1. **Pool Setup**: Use Hub methods for request manager configuration and authorization:
2. **Message Handling**: Handle two-layer message structure (wrapper + nested)
3. **Event Processing**: Update event listeners for new event names and structures


src/common/libraries/RequestCallbackMessageLib.sol
```diff
@@ -0,0 +1,162 @@
...
+
+import {BytesLib} from "src/misc/libraries/BytesLib.sol";
+import {CastLib} from "src/misc/libraries/CastLib.sol";
+
+enum RequestCallbackType {
+    /// @dev Placeholder for null request callback type
+    Invalid,
+    ApprovedDeposits,
+    IssuedShares,
+    RevokedShares,
```
lemunozm, line 12: I think these messages relate to the `BalanceSheet` instead of vaults. I feel like they are ok where they are placed now

src/common/libraries/RequestCallbackMessageLib.sol
```diff
@@ -0,0 +1,162 @@
+// SPDX-License-Identifier: BUSL-1.1
+pragma solidity 0.8.28;
+
+import {BytesLib} from "src/misc/libraries/BytesLib.sol";
+import {CastLib} from "src/misc/libraries/CastLib.sol";
+
+enum RequestCallbackType {
```
lemunozm, line 7: NIT. Not sure if the "callback" name correctly defines all types of messages. Some of the messages are not necessarily callbacks. Maybe `RequestNotification` is s wide term? 🤔

src/hub/Hub.sol
```diff
@@ -122,8 +124,13 @@ contract Hub is Multicall, Auth, Recoverable, IHub, IHubGatewayHandler, IHubGuar
...
         if (totalPaymentAssetAmount > 0 || cancelledAssetAmount > 0) {
-            sender.sendFulfilledDepositRequest(
-                poolId, scId, assetId, investor, totalPaymentAssetAmount, totalPayoutShareAmount, cancelledAssetAmount
+            sender.sendRequestCallback(
+                poolId,
+                scId,
+                assetId,
+                RequestCallbackMessageLib.FulfilledDepositRequest(
+                    investor, totalPaymentAssetAmount, totalPayoutShareAmount, cancelledAssetAmount
+                ).serialize()
```
lemunozm, line 139: From an organization PoV feels weird that we create this submessage, but later the core knows about that. Then, why not leave the message in `MessageLib` as it was?

I think we do not need to move the messages as submessages. Just leave in the spoke side to be able to configure custom handlers for them. But the messages still belong to the core.

src/common/libraries/RequestCallbackMessageLib.sol
hieronx: The problem if they are not in a separate message lib, is that we can't easily handle the forwarding of the processing of these messages to another contract. Right now, we can simply forward all emssages in `RequestCallbackMessageLib` to `Spoke.requestCallback`.

src/common/libraries/RequestCallbackMessageLib.sol
hieronx: I think callback seems quite correct here since it's the return action of some request. It is imo more than a notification.

hieronx: In terms of them being core: I think you could argue we could build an implementation of `Spoke` on a new chain (e.g. Solana) without building vaults, and in that case we wouldn't need to implement the request and request callback messages. So it is decoupled from the core in that sense.

src/common/libraries/RequestCallbackMessageLib.sol
hieronx: Otherwise you need 4 outgoing and 5 incoming additional methods in Spoke, to forward all these to the corresponding request manager, I think that's a lot of extra code that would be nice to avoid.

src/hub/Hub.sol
hieronx: In terms of them being core: I think you could argue we could build an implementation of Spoke on a new chain (e.g. Solana) without building vaults, and in that case we wouldn't need to implement the request and request callback messages. So it is decoupled from the core in that sense.

src/common/libraries/RequestCallbackMessageLib.sol
hieronx: What I mean, otherwise `Spoke` would need something like
```solidity
function depositRequest(..) public {
    // Look up requestManager
    require(msg.sender == requestManager);
    sender.depositRequest(..);
}

function redeemRequest(..) public {
    // Look up requestManager
    require(msg.sender == requestManager);
    sender.redeemRequest(..);
}

function cancelDepositRequest(..) public {
    // Look up requestManager
    require(msg.sender == requestManager);
    sender.cancelDepositRequest(..);
}

function cancelRedeemRequest(..) public {
    // Look up requestManager
    require(msg.sender == requestManager);
    sender.cancelRedeemRequest(..);
}

function approvedDeposits(..) public auth {
    // Look up requestManager
    requestManager.approvedDeposits(..);
}

function issuedShares(..) public auth {
    // Look up requestManager
    requestManager.issuedShares(..);
}

function revokedShares(..) public auth {
    // Look up requestManager
    requestManager.revokedShares(..);
}

function fulfilledDepositRequest(..) public auth {
    // Look up requestManager
    requestManager.fulfilledDepositRequest(..);
}

function fulfilledRedeemRequest(..) public auth {
    // Look up requestManager
    requestManager.fulfilledRedeemRequest(..);
}
```

src/common/libraries/RequestCallbackMessageLib.sol
lemunozm: 🤔 mmm... but for example, in case of `hub.issueShares()`, it's not a callback right? it's just an action from hub -> spoke

src/common/libraries/RequestCallbackMessageLib.sol
lemunozm: But I do not understand in this case why `ApprovedDeposits`,  `issuedShares` and    `RevokedShares` need to support different future handlers. IIUC, those 3 messages should always be handled by the `BalanceSheet`.

Of course, this is different from the request/fulfilled messages, where I agree more in the separation

lemunozm: Why should the outgoing methods should be in `spoke`? Should not be in the request manager/vault?

Maybe it makes sense to have a new thin contract to handle those, like a `RequestRouter` contract or so.

src/hub/Hub.sol
lemunozm: I understand you PoV. Although IIUC they are still from the core, but they are just optional to implement. Not sure if having it as submessage helps in that regard 🤔

src/common/libraries/RequestCallbackMessageLib.sol
hieronx: > But I do not understand in this case why ApprovedDeposits, issuedShares and RevokedShares need to support different future handlers. IIUC, those 3 messages should always be handled by the BalanceSheet.

I don't think these should be in BalanceSheet. E.g. in ApprovedDeposits, we transfer from global escrow. That is async request specific logic, that balance sheet shouldn't know about.

hieronx: > Why should the outgoing methods should be in spoke? Should not be in the request manager/vault?
> Maybe it makes sense to have a new thin contract to handle those, like a RequestRouter contract or so.

Yes introducing a small routing contract in between would be an alternative solution. But I would feel the separate message libs is much cleaner then.

src/common/libraries/RequestCallbackMessageLib.sol
hieronx: It's a second order callback from requestDeposit

src/common/libraries/RequestCallbackMessageLib.sol
lemunozm: >  don't think these should be in BalanceSheet. E.g. in ApprovedDeposits, we transfer from global escrow. That is async request specific logic, that balance sheet shouldn't know about.

Correct! I thought for a moment they were processed in the `BalanceSheet` 👍🏻

src/common/libraries/RequestCallbackMessageLib.sol
lemunozm: Well, that's a good argument 😆

script/SpokeDeployer.s.sol
```diff
@@ -110,8 +110,6 @@ contract SpokeDeployer is CommonDeployer {
         gateway.rely(address(spoke));
 
         // Rely async requests manager
-        balanceSheet.rely(address(asyncRequestManager));
```
hieronx, line 109: This means a pool always needs to add the async request manager as a balance sheet manager. I think that is a good decision, making it fully up to the pool to choose the impl.

src/spoke/Spoke.sol
```diff
@@ -200,6 +200,13 @@ contract Spoke is Auth, Recoverable, ReentrancyProtection, ISpoke, ISpokeGateway
         emit AddShareClass(poolId, scId, shareToken_);
     }
 
+    /// @inheritdoc ISpokeGatewayHandler
+    function updateRequestManager(PoolId poolId, ShareClassId scId, AssetId assetId, address manager) public auth {
```
hieronx: We should likely also add `manager` to `NotifyPool` message, to set up on initialization

lemunozm: We can batch both messages then, right?

There are cases like subsidizing the hub through a local spoke where we do not need a request manager

src/spoke/Spoke.sol
```diff
@@ -304,6 +311,25 @@ contract Spoke is Auth, Recoverable, ReentrancyProtection, ISpoke, ISpokeGateway
     // Vault management
     //----------------------------------------------------------------------------------------------
 
+    /// @inheritdoc ISpoke
+    function request(PoolId poolId, ShareClassId scId, AssetId assetId, bytes memory payload) external {
+        ShareClassDetails storage shareClass = _shareClass(poolId, scId);
+        require(msg.sender == address(shareClass.manager[assetId]), NotAuthorized());
+
+        sender.sendRequest(poolId, scId, assetId, payload);
+        // emit Request(poolId, scId, payload);
```
wischli: Reminder for myself to implement

src/spoke/Spoke.sol
```diff
@@ -304,6 +311,25 @@ contract Spoke is Auth, Recoverable, ReentrancyProtection, ISpoke, ISpokeGateway
...
+    }
+
+    /// @inheritdoc ISpokeGatewayHandler
+    function requestCallback(PoolId poolId, ShareClassId scId, AssetId assetId, bytes memory payload) external auth {
+        ShareClassDetails storage shareClass = _shareClass(poolId, scId);
+        IRequestCallback manager = shareClass.manager[assetId];
+        require(address(manager) != address(0), InvalidRequestManager());
+
+        manager.callback(poolId, scId, assetId, payload);
+        // emit RequestCallback(poolId, scId, assetId, payload);
```
wischli: Reminder for myself to implement

src/vaults/AsyncRequestManager.sol
```diff
@@ -85,11 +84,19 @@ contract AsyncRequestManager is BaseRequestManager, IAsyncRequestManager {
         require(state.pendingCancelDepositRequest != true, CancellationIsPending());
 
         state.pendingDepositRequest += assets_;
-        sender.sendDepositRequest(poolId, scId, controller.toBytes32(), vaultDetails.assetId, assets_);
+        _sendDepositRequest(poolId, scId, vaultDetails.assetId, controller, assets_);
```
wischli: Nit: For cancel requests, we directly invoke `spoke.request(...)`. Since `_sendDepositRequest` is needed exactly once and doesn't do more than simply invoking `spoke.request(...)`, I'd prefer to do it here directly and save the marginal gas cost on execution, deployment and have more linear readability.

src/vaults/AsyncRequestManager.sol
```diff
@@ -115,13 +122,20 @@ contract AsyncRequestManager is BaseRequestManager, IAsyncRequestManager {
         require(state.pendingCancelRedeemRequest != true, CancellationIsPending());
 
         state.pendingRedeemRequest = state.pendingRedeemRequest + shares_;
-        sender.sendRedeemRequest(poolId, scId, controller.toBytes32(), vaultDetails.assetId, shares_);
-
+        _sendRedeemRequest(poolId, scId, vaultDetails.assetId, controller, shares_);
```
wischli: Nit: Same as above for deposit

src/vaults/AsyncRequestManager.sol
```diff
@@ -156,22 +174,65 @@ contract AsyncRequestManager is BaseRequestManager, IAsyncRequestManager {
...
+
+    /// @inheritdoc IAsyncRequestManager
     function approvedDeposits(
         PoolId poolId,
         ShareClassId scId,
         AssetId assetId,
         uint128 assetAmount,
         D18 pricePoolPerAsset
-    ) external auth {
+    ) public auth {
```
wischli, line 189: Can you elaborate on why these methods are required to be `public` now? Since we are not calling those methods within the contract, I fail to see the reason.

wischli (commented):
No objections from my side regarding the proposed path. Will take over fixing tests in a PR on top of this.

src/vaults/AsyncRequestManager.sol
hieronx: They are called from the `callback` method right?

src/spoke/Spoke.sol
hieronx: True!

src/vaults/AsyncRequestManager.sol
hieronx: I believe i moved this because of a stack too deep issue.

src/vaults/AsyncRequestManager.sol
wischli: Ah sorry, my brain searched with prefixed `.` which of course doesn't make sense when checking internal usage. You are right!

src/spoke/Spoke.sol
lemunozm: After rethinking and discussing in the sync, you're maybe right with your first thought. Just to avoid the pool manager to modify the requestManager in between of some messages and break the system

mustermeiszer (commented):
Questions I have: 

* Factories currentlty have hardcoded managers - do we want the factories to dynamically look-up the latest stored manager? 
* Already deployed vaults could have different managers then the latest one set - do we want vaults to dynamically look-up the latest manager or do we allow to orphan the vaults? 

@lemunozm had a good proposal

* We could implement a hook that is called everytime a request or requestCallback comes in

test/hub/integration/Cases.t.sol
```diff
@@ -92,33 +92,34 @@ contract TestCases is BaseTest {
...
-            m2.fulfilledShareAmount,
-            PricingLib.convertWithPrice(
-                APPROVED_INVESTOR_AMOUNT,
-                hubRegistry.decimals(USDC_C2),
-                hubRegistry.decimals(poolId),
-                NAV_PER_SHARE.reciprocal()
-            )
-        );
-        assertEq(m2.cancelledAssetAmount, INVESTOR_AMOUNT - APPROVED_INVESTOR_AMOUNT);
+        // MessageLib.ApprovedDeposits memory m0 = MessageLib.deserializeApprovedDeposits(cv.popMessage());
```
wischli: Reminder for myself to update these

src/spoke/Spoke.sol
wischli: Reminder for myself to add simple tests that ensure one time setup once implemented

src/vaults/AsyncRequestManager.sol
```diff
@@ -115,9 +108,14 @@ contract AsyncRequestManager is BaseRequestManager, IAsyncRequestManager {
...
-        sender.sendRedeemRequest(poolId, scId, controller.toBytes32(), vaultDetails.assetId, shares_);
+        spoke.request(
+            vault_.poolId(),
+            vault_.scId(),
+            spoke.vaultDetails(vault_).assetId,
+            RequestMessageLib.RedeemRequest(controller.toBytes32(), shares_).serialize()
+        );
 
-        _executeRedeemTransfer(poolId, scId, sender_, owner, address(globalEscrow), shares_);
+        _executeRedeemTransfer(vault_.poolId(), vault_.scId(), sender_, owner, address(globalEscrow), shares_);
```
wischli: Nit: Maybe we want to store all these multiply read storage values to variables?

src/common/libraries/RequestCallbackMessageLib.sol
```diff
@@ -0,0 +1,159 @@
...
+    using CastLib for *;
+
+    error UnknownRequestCallbackType();
+
+    function requestCallbackType(bytes memory message) internal pure returns (RequestCallbackType) {
+        return RequestCallbackType(message.toUint8(0));
+    }
+
+    //---------------------------------------
+    //    RequestCallback.ApprovedDeposits (submsg)
```
wischli, line 29: Ping @hieronx to check my change here 

```diff
-    //    Request.ApprovedDeposits (submsg)
+    //    RequestCallback.ApprovedDeposits (submsg)
```

test/spoke/integration/Deposit.t.sol
```diff
@@ -18,9 +19,19 @@ import {IAsyncVault} from "src/vaults/interfaces/IAsyncVault.sol";
 
 contract DepositTest is BaseTest {
     using MessageLib for *;
+    using RequestMessageLib for *;
     using CastLib for *;
     using MathLib for uint256;
 
+    function _checkCancelDepositRequestMessage(MockAdapter adapter, IBaseVault vault, uint128 assetId) internal {
```
wischli: Note: Quick fix to resolve stack too deep

wischli (approved):
All tests should be re-enabled and fixed now. Rest LGTM except for the post-merge TODOs with which I agree.

src/common/libraries/RequestCallbackMessageLib.sol
hieronx: Yeah looks good!

test/common/unit/libraries/RequestCallbackMessageLib.t.sol
```diff
@@ -0,0 +1,131 @@
...
+        assertEq(a.assetAmount, b.assetAmount);
+        assertEq(a.pricePoolPerAsset, b.pricePoolPerAsset);
+
+        assertEq(
+            uint8(RequestCallbackMessageLib.requestCallbackType(a.serialize())),
+            uint8(RequestCallbackType.ApprovedDeposits)
+        );
+
+        // 1 byte type + 16 bytes assetAmount + 16 bytes pricePoolPerAsset
+        assertEq(a.serialize().length, 1 + 16 + 16);
```
lemunozm, line 34: NIT NOTE. We do not really need to test the length of the submessages, because we don't require that explicit length anywhere. This is different to main messages where the length is required for batching.

What I want to mean, if tomorrow we modify the message, this line would fail without warning a real issue.

test/common/unit/libraries/RequestCallbackMessageLib.t.sol
```diff
@@ -0,0 +1,131 @@
...
+        RequestCallbackMessageLib.IssuedShares memory a =
+            RequestCallbackMessageLib.IssuedShares({shareAmount: shareAmount, pricePoolPerShare: pricePoolPerShare});
+        RequestCallbackMessageLib.IssuedShares memory b = a.serialize().deserializeIssuedShares();
+
+        assertEq(a.shareAmount, b.shareAmount);
+        assertEq(a.pricePoolPerShare, b.pricePoolPerShare);
+
+        assertEq(
+            uint8(RequestCallbackMessageLib.requestCallbackType(a.serialize())), uint8(RequestCallbackType.IssuedShares)
+        );
```
lemunozm, line 47: This is a good check that is missing in other messages. We should also add them to the rest of the messages

src/spoke/Spoke.sol
```diff
@@ -203,6 +195,14 @@ contract Spoke is Auth, Recoverable, ReentrancyProtection, ISpoke, ISpokeGateway
         emit AddShareClass(poolId, scId, shareToken_);
     }
 
+    /// @inheritdoc ISpokeGatewayHandler
+    function initializeRequestManager(PoolId poolId, ShareClassId scId, AssetId assetId, address manager) public auth {
+        ShareClassDetails storage shareClass = _shareClass(poolId, scId);
+        require(address(shareClass.manager[assetId]) == address(0), RequestManagerAlreadySet());
```
lemunozm: Wondering what happens if the FM just makes a mistake here and puts an incorrect manager. How can they revert that?

lemunozm (commented):
LGTM! Good job here @hieronx 

Just some minor comments

lemunozm (approved):

src/spoke/Spoke.sol
hieronx: Assumption is, they'd do this while setting up a share class. They can just set up a new share class then.

test/common/unit/libraries/RequestCallbackMessageLib.t.sol
hieronx: https://github.com/centrifuge/protocol-v3/issues/452

src/vaults/AsyncRequestManager.sol
hieronx: Yeah I was struggling to stay below stack too deep issues with that. Will check once more

src/vaults/AsyncRequestManager.sol
```diff
@@ -155,23 +129,64 @@ contract AsyncRequestManager is BaseRequestManager, IAsyncRequestManager {
...
+    function _sendRequest(IBaseVault vault_, bytes memory payload) internal {
+        spoke.request(vault_.poolId(), vault_.scId(), spoke.vaultDetails(vault_).assetId, payload);
     }
 
     //----------------------------------------------------------------------------------------------
     // Gateway handlers
     //----------------------------------------------------------------------------------------------
 
-    /// @inheritdoc IRequestManagerGatewayHandler
+    function callback(PoolId poolId, ShareClassId scId, AssetId assetId, bytes calldata payload) external {
```
hieronx: Just realized this misses `auth`

src/spoke/Spoke.sol
lemunozm: But this is per asset right?

Nothing prevent to:
1. Create the share class
2. Create a bunch of correct managers for different assets
3. After one year, create another requestManager for another asset and mess it up

src/vaults/AsyncRequestManager.sol
hieronx: Actually cleaner is just checking the `msg.sender == spoke`, similar to the on-offramp manager. Can then remove spoke as a ward on this contract.

src/spoke/Spoke.sol
hieronx: What would they mess up?

lemunozm: I mean, they will block the usage of that asset for that share class for the entire life of the pool, if they configure a wrong address

lemunozm: I feel quite reasonable, that someone copy an address that is not the correct one in the  `initilizeRequestManager`, over a share class that has been living quite a while.

lemunozm: Maybe, it's not difficult to achieve, to add some boolean to `shareClass.manager[assetId]` storage, that only becomes true after the first request/callback message.

That way, at least we know the manager is at least valid before freezing it forever

lemunozm: As pseudocode: 

```diff
diff --git a/src/spoke/Spoke.sol b/src/spoke/Spoke.sol
index c5d0d30b..0e5c1741 100644
--- a/src/spoke/Spoke.sol
+++ b/src/spoke/Spoke.sol
@@ -198,7 +198,7 @@ contract Spoke is Auth, Recoverable, ReentrancyProtection, ISpoke, ISpokeGateway
     /// @inheritdoc ISpokeGatewayHandler
     function initializeRequestManager(PoolId poolId, ShareClassId scId, AssetId assetId, address manager) public auth {
         ShareClassDetails storage shareClass = _shareClass(poolId, scId);
-        require(address(shareClass.manager[assetId]) == address(0), RequestManagerAlreadySet());
+        require(!shareClass.manager[assetId].isSet, RequestManagerAlreadySet());
         shareClass.manager[assetId] = IRequestManager(manager);
         emit InitializeRequestManager(poolId, scId, assetId, IRequestManager(manager));
     }
@@ -312,6 +312,8 @@ contract Spoke is Auth, Recoverable, ReentrancyProtection, ISpoke, ISpokeGateway
         ShareClassDetails storage shareClass = _shareClass(poolId, scId);
         require(msg.sender == address(shareClass.manager[assetId]), NotAuthorized());

+        shareClass.manager[AssetId].isSet = true;
+
         sender.sendRequest(poolId, scId, assetId, payload);
     }

@@ -321,6 +323,8 @@ contract Spoke is Auth, Recoverable, ReentrancyProtection, ISpoke, ISpokeGateway
         IRequestManager manager = shareClass.manager[assetId];
         require(address(manager) != address(0), InvalidRequestManager());

+        shareClass.manager[AssetId].isSet = true;
+
         manager.callback(poolId, scId, assetId, payload);
     }
```

(We could only set it to true if it was not true to save some gas)

hieronx: Actually I think we can do something better, we can check `shareClass.vaults[asset][tokenId].length`. If that is 0, we can update.

Then we should rename the message and function again, to `update`.

hieronx: Actually, that storage is unused right now 🤔  We can add a sepearate struct like you propose, but rather tracking whether there are any linked vaults. I think that's cleaner. Just pushed a proposal, let me know what you think!

script/adapters/Localhost.s.sol
```diff
@@ -156,6 +159,10 @@ contract LocalhostDeployer is FullDeployer {
         hub.notifyPool(poolId, centrifugeId);
         hub.notifyShareClass(poolId, scId, centrifugeId, bytes32(bytes20(redemptionRestrictionsHook)));
 
+        hub.setRequestManager(poolId, scId, assetId, address(asyncRequestManager).toBytes32());
+        hub.updateBalanceSheetManager(centrifugeId, poolId, address(asyncRequestManager).toBytes32(), true);
+        hub.updateBalanceSheetManager(centrifugeId, poolId, address(syncRequestManager).toBytes32(), true);
```
lemunozm, line 164: Why is the manager here the `asyncRequestManager`? Shouldn't there be some account?

script/adapters/Localhost.s.sol
```diff
@@ -69,6 +69,9 @@ contract LocalhostDeployer is FullDeployer {
         hub.notifyPool(poolId, centrifugeId);
         hub.notifyShareClass(poolId, scId, centrifugeId, bytes32(bytes20(redemptionRestrictionsHook)));
 
+        hub.setRequestManager(poolId, scId, assetId, address(asyncRequestManager).toBytes32());
+        hub.updateBalanceSheetManager(centrifugeId, poolId, address(asyncRequestManager).toBytes32(), true);
```
lemunozm, line 73: And here

lemunozm (approved):
Love your solution!! I think now it works without any side issue

src/spoke/Spoke.sol
lemunozm: Love it!! ❤️

script/adapters/Localhost.s.sol
hieronx: The request managers also need to be balance sheet managers, to note deposits etc

hieronx: Same change in end to end and spoke integration tests

lemunozm: Mmm... shouldn't that be done automatically then? When we register the `requestManager` or the vault, we automatically enable that requestManager as the manager

lemunozm: Seems like something the user should not know about. Like some internal linking. From UX it's weird to explain why this is needed

lemunozm: I think actually, when we link the vault, we can check if its vault.manager() is already a manager in the BalanceSheet, and if it's not, do it there. We will also save gas by configuring this there

lemunozm: Ahhh.... but the spoke doesn't know anything about `BalanceSheet`

hieronx: It could. But I don't think this should be done automatically, it can very easily be batched, and it is important to know that the async and sync managers give full access to the balance sheet.

wischli (approved):

src/hub/HubHelpers.sol
```diff
@@ -164,6 +172,48 @@ contract HubHelpers is Auth, IHubHelpers {
...
+                    poolId,
+                    scId,
+                    assetId,
+                    RequestCallbackMessageLib.FulfilledRedeemRequest(m.investor, 0, 0, cancelledShareAmount).serialize()
+                );
+            }
+        } else {
+            revert UnknownRequestType();
+        }
+    }
```
mustermeiszer, line 215: @hieronx why is this in the helpers? Looks like it should go into the core.

src/spoke/interfaces/ISpoke.sol
```diff
@@ -11,22 +11,27 @@ import {IShareToken} from "src/spoke/interfaces/IShareToken.sol";
...
 struct Pool {
     uint256 createdAt;
     mapping(ShareClassId scId => ShareClassDetails) shareClasses;
 }
 
+struct ShareClassAsset {
+    /// @dev Manager that can send requests, and handles the request callbacks.
+    IRequestManager manager;
+    /// @dev Number of linked vaults.
+    uint32 numVaults;
```
mustermeiszer, line 26: What is the reason we have this counter?

src/hub/HubHelpers.sol
hieronx: Doesn't fit in the bytecode size

src/spoke/interfaces/ISpoke.sol
hieronx: To block updating request manager if there is >0 linked vault.

src/hub/HubHelpers.sol
hieronx: In general I think in v3.1, we want to find a way to refactor `Hub` and `HubHelpers` somehow. Potentially just moving all incoming gateway actions.

PR #440, opened 2025-06-09
Feat: zero price support

lemunozm:
Support set `price = 0`, which means zero amounts at the end without revert due division-by-zero errors

To avoid any mathematical mistake silently, the bypass `n / 0 == 0` is done at the highest possible level when we can handle any issue depending on the usage.


hieronx:
What about `Holdings.increase/decrease` if `pricePoolPerAsset == 0`?

lemunozm:
The `amountValue` obtained from that increase/decrease will be 0.

src/hub/ShareClassManager.sol
```diff
@@ -250,14 +250,16 @@ contract ShareClassManager is Auth, IShareClassManager {
...
 
-        issuedShareAmount = PricingLib.assetToShareAmount(
-            epochAmounts.approvedAssetAmount,
-            hubRegistry.decimals(depositAssetId),
-            hubRegistry.decimals(poolId),
-            epochAmounts.pricePoolPerAsset,
-            navPoolPerShare,
-            MathLib.Rounding.Down
-        ).toUint128();
+        issuedShareAmount = (navPoolPerShare.raw() != 0)
```
hieronx: Wouldn't it be cleaner to add `isZero()` to `D18`?

hieronx: Maybe even `isNotZero()`

src/vaults/SyncRequestManager.sol
```diff
@@ -212,11 +212,12 @@ contract SyncRequestManager is BaseRequestManager, ISyncRequestManager {
...
-        Prices memory prices_ = prices(vault_.poolId(), vault_.scId(), vaultDetails.assetId);
 
-        return super._assetToShareAmount(
-            vault_, vaultDetails, assets, prices_.poolPerAsset, prices_.poolPerShare, MathLib.Rounding.Down
-        );
+        D18 poolPerShare = pricePoolPerShare(vault_.poolId(), vault_.scId());
+        D18 poolPerAsset = spoke.pricePoolPerAsset(vault_.poolId(), vault_.scId(), vaultDetails.assetId, true);
+
+        return
+            super._assetToShareAmount(vault_, vaultDetails, assets, poolPerAsset, poolPerShare, MathLib.Rounding.Down);
```
hieronx, line 220: Why doesn't `BaseRequestManager.convertToAssets` need an update?

Also, if you are not using `prices()` here anymore, it might make sense to also update `_shareToAssetAmount` and just remove `prices()`?

src/hub/ShareClassManager.sol
lemunozm: Yeah, could be, that's something I always think about, but I never have time. I would go with `isZero()` due to it being a positive sentence and usually found in most libraries

src/vaults/SyncRequestManager.sol
lemunozm: Could `prices()` be need it as a view method? If not I agree

> Why doesn't BaseRequestManager.convertToAssets need an update?

This was already handled here: https://github.com/centrifuge/protocol-v3/blob/post-spearbit/src/common/libraries/PricingLib.sol#L56

src/hub/ShareClassManager.sol
hieronx: Yeah just realized you will then end up doing `!price.isZero()` in most cases with your current pattern. In that case I'd argue `price.isNotZero()` is actually a bit more readable.

lemunozm: Yeah, we can add both then

src/vaults/SyncRequestManager.sol
hieronx: I think it's not really needed as a view method

lemunozm: Done: [46bff63](https://github.com/centrifuge/protocol-v3/pull/440/commits/46bff63d7caea266dc3cced7cb19e7eed04e31ba)

src/hub/ShareClassManager.sol
lemunozm: [1696187](https://github.com/centrifuge/protocol-v3/pull/440/commits/1696187c119aab664020f71d151d0217169b1907)

wischli (approved):
Overall LGTM but I have a nit that I would like to have resolved before we deploy v3.

src/vaults/SyncRequestManager.sol
lemunozm: Thanks so much for your research!

> My proposal is to remove defaulting to 0 for denominator zero prices entirely in the PricingLib and handling that on the higher level.

I saw them and I thought the same. I fully agree.

> #### On the other hand, these functions don't explicitly handle zero denominator prices:

> 1. convertWithPrice (both variants) - No explicit zero check, but price is typically used as numerator
>2. assetToShareAmount ([pure version L216+](https://github.com/centrifuge/protocol-v3/blob/733c06dc786afb9fb23eb0bac9204e773cb18528/src/common/libraries/PricingLib.sol#L216)) - Relies on `convertWithPrices` which has require protection
>3. shareToAssetAmount ([pure version L231+ ](https://github.com/centrifuge/protocol-v3/blob/733c06dc786afb9fb23eb0bac9204e773cb18528/src/common/libraries/PricingLib.sol#L231)) - Relies on` convertWithPrices` which has require protection
>4. poolToAssetAmount ([L244+](https://github.com/centrifuge/protocol-v3/blob/733c06dc786afb9fb23eb0bac9204e773cb18528/src/common/libraries/PricingLib.sol#L244)) - Relies on convertWithReciprocalPrice which has require protection
>6. assetToPoolAmount ([L255+](https://github.com/centrifuge/protocol-v3/blob/733c06dc786afb9fb23eb0bac9204e773cb18528/src/common/libraries/PricingLib.sol#L255)) - Relies on convertWithPrice, no explicit protection but price used as numerator
>7. priceAssetPerShare ([L 269+](https://github.com/centrifuge/protocol-v3/blob/733c06dc786afb9fb23eb0bac9204e773cb18528/src/common/libraries/PricingLib.sol#L269)) - Uses direct division operator with no protection

From my understanding, these would not need to handle zero denominator prices. It should be the caller who handle this. You agree with that IIUC, right?

wischli: > From my understanding, these would not need to handle zero denominator prices. It should be the caller who handle this. You agree with that IIUC, right?

Yes!

wischli:
@lemunozm When you have the time, could you check my changes in [b8efae6](https://github.com/centrifuge/protocol-v3/pull/440/commits/b8efae6a20cb4c8e3004e96c42d790a033be170d)? Thanks!

src/vaults/SyncRequestManager.sol
wischli: Done in [b8efae6](https://github.com/centrifuge/protocol-v3/pull/440/commits/b8efae6a20cb4c8e3004e96c42d790a033be170d)

src/vaults/AsyncRequestManager.sol
```diff
@@ -524,9 +524,11 @@ contract AsyncRequestManager is BaseRequestManager, IAsyncRequestManager {
         VaultDetails memory vaultDetails = spoke.vaultDetails(vault_);
         address shareToken = vault_.share();
 
-        return PricingLib.assetToShareAmount(
-            shareToken, vaultDetails.asset, vaultDetails.tokenId, assets, d18(priceAssetPerShare.toUint128()), rounding
-        );
+        return priceAssetPerShare == 0
```
lemunozm: What was the reason we're not using `D18` here?

src/vaults/BaseRequestManager.sol
```diff
@@ -134,15 +134,17 @@ abstract contract BaseRequestManager is Auth, Recoverable, IBaseRequestManager {
...
-        return PricingLib.assetToShareAmount(
-            vault_.share(),
-            vaultDetails.asset,
-            vaultDetails.tokenId,
-            assets.toUint128(),
-            pricePoolPerAsset,
-            pricePoolPerShare,
-            rounding
-        );
+        return pricePoolPerShare.raw() == 0
```
lemunozm: NIT. We already have `isZero()` and `isNotZero()` attached to `D18`

lemunozm (commented):
I thought we would need to modify something in EndToEnd, happy it's not need it!

Thanks for tackling this!

src/vaults/AsyncRequestManager.sol
wischli: Good point! Fixed in [5a8d733](https://github.com/centrifuge/protocol-v3/pull/440/commits/5a8d7331e1f788875003e3944032581791ed3df7)

src/vaults/BaseRequestManager.sol
wischli: Ah yes! Sorry, fixed in [5a8d733](https://github.com/centrifuge/protocol-v3/pull/440/commits/5a8d7331e1f788875003e3944032581791ed3df7)

src/vaults/AsyncRequestManager.sol
lemunozm: Thanks for doing this!

src/vaults/BaseRequestManager.sol
lemunozm: It was a recommendation from @hieronx actually. Thanks for this!

lemunozm:
I can not approve your latest changes because it's my PR but, I APPROVE

wischli (approved):
Approving our mutual work

PR #443, opened 2025-06-09
Post Spearbit review changes

hieronx:


src/vaults/AsyncRequestManager.sol
```diff
@@ -325,7 +325,7 @@ contract AsyncRequestManager is BaseRequestManager, IAsyncRequestManager {
         state.maxMint = state.maxMint > sharesUp ? state.maxMint - sharesUp : 0;
 
         if (sharesDown > 0) {
-            globalEscrow.authTransferTo(vault_.share(), receiver, sharesDown);
+            globalEscrow.authTransferTo(vault_.share(), 0, receiver, sharesDown);
```
xmxanuel, line 379: Why not using a constant for `0`?

src/spoke/BalanceSheet.sol
```diff
@@ -199,12 +199,12 @@ contract BalanceSheet is Auth, Recoverable, IBalanceSheet, IBalanceSheetGatewayH
     /// @inheritdoc IBalanceSheet
     function submitQueuedAssets(PoolId poolId, ShareClassId scId, AssetId assetId) external authOrManager(poolId) {
         AssetQueueAmount storage assetQueue = queuedAssets[poolId][scId][assetId];
-        if (assetQueue.deposits == 0 && assetQueue.withdrawals == 0) return;
```
xmxanuel, line 185: Why the removal of this early return? I understand the `assetCounter` helper variable usage as a result.

hieronx: Otherwise, we could not notify the hub that the snapshot value may have changed. i.e:
- Last hub snapshot is “synced”
- balanceSheet.issue()
- balanceSheet.submitQueuedAssets() //Nothing is sent here. Hub is not notified that the snapshot is no longer synced.

This is indeed a very border case, and probably the manager will use submitQueuedShares() instead. Nevertheless, from a user perspective, I would expect to trigger the snapshotHook for each submit* call

PR #444, opened 2025-06-10
Feat: extra gas limit

lemunozm:
- Add a new parameter to methods that requires extra and custom computation (i.e, by a hook) that we can not preview in the `GasService`

### DIFF
--- a/src/common/MessageDispatcher.sol
+++ b/src/common/MessageDispatcher.sol
@@ -261,13 +261,17 @@ contract MessageDispatcher is Auth, IMessageDispatcher {
     }
 
     /// @inheritdoc IHubMessageSender
-    function sendUpdateRestriction(uint16 centrifugeId, PoolId poolId, ShareClassId scId, bytes calldata payload)
-        external
-        auth
-    {
+    function sendUpdateRestriction(
+        uint16 centrifugeId,
+        PoolId poolId,
+        ShareClassId scId,
+        bytes calldata payload,
+        uint128 extraGasLimit
+    ) external auth {
         if (centrifugeId == localCentrifugeId) {
             spoke.updateRestriction(poolId, scId, payload);
         } else {
+            gateway.setExtraGasLimit(extraGasLimit);
             gateway.send(
                 centrifugeId,
                 MessageLib.UpdateRestriction({poolId: poolId.raw(), scId: scId.raw(), payload: payload}).serialize()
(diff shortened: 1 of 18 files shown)


src/spoke/BalanceSheet.sol
```diff
@@ -162,7 +162,7 @@ contract BalanceSheet is Auth, Recoverable, IBalanceSheet, IBalanceSheetGatewayH
             }
         } else {
             bool isSnapshot = shareQueue.queuedAssetCounter == 0;
-            sender.sendUpdateShares(poolId, scId, shares, true, isSnapshot, shareQueue.nonce);
+            sender.sendUpdateShares(poolId, scId, shares, true, isSnapshot, shareQueue.nonce, 0 /*TODO*/ );
```
lemunozm, line 184: Left TODOs in some methods in `BalanceSheet` in case we no longer need them after "always queuing"

lemunozm:
The vaults have some callbacks like `onDepositClaimable()`. Right now, we're filling this because we own the vaults. Nevertheless, with the vault separation we're taking right now, I think it could make sense to also provide extraGasLimit to the actions that ends up in the vaults, given in the future other vaults could compute anything inside of it. WDYT @hieronx?

src/common/MessageDispatcher.sol
```diff
@@ -281,11 +285,13 @@ contract MessageDispatcher is Auth, IMessageDispatcher {
         PoolId poolId,
         ShareClassId scId,
         bytes32 target,
-        bytes calldata payload
+        bytes calldata payload,
+        uint128 extraGasLimit
```
wischli, line 289: Q: AFAICT, right now there is no need to have a non-zero extra gas limit for `UpdateContract` messages right? So this adjustment is just future safety or a I missing something?

I basically want to understand why it is added for the changed cases here. The `sendUpdateRestriction` makes totally sense to me but for the others I need some more input.

src/spoke/BalanceSheet.sol
```diff
@@ -231,7 +233,7 @@ contract BalanceSheet is Auth, Multicall, Recoverable, IBalanceSheet, IBalanceSh
             assetQueue.deposits,
             assetQueue.withdrawals,
             pricePoolPerAsset,
-            isSnapshot,
+            shareQueue.delta == 0 && shareQueue.queuedAssetCounter == assetCounter, //isSnapshot
```
wischli, line 236: Q: Can you elaborate on this change? Seems like it slightly increases gas while also reducing readability

wischli: Q: Why do we not need extra gas for submitting queued assets but need it for queued shares?

wischli (commented):
Need to resolve my knowledge gap before approving, though I am sure all is perfect.

src/common/MessageDispatcher.sol
lemunozm: `UpdateContract` can potentially run any method from any contract, which makes it totally uncertain what the gas is for that computation

src/spoke/BalanceSheet.sol
lemunozm: > Can you elaborate on this change? Seems like it slightly increases gas while also reducing readability

Yes stack too deep issue 😢

> Q: Why do we not need extra gas for submitting queued assets but need it for queued shares?

If I'm right, both methods have `extraGasLimit`

wischli: > If I'm right, both methods have extraGasLimit

You're right, sorry! Looked in the wrong row where the event emission happens.

lemunozm: Yeah, I think these gas values don't need to be added to the events

wischli (approved):
Thanks for the explanations!

PR #447, opened 2025-06-11
chore: finish Decouple request managers #439

wischli:
Does finishing touches on base PR #439 
* Adds missing docs
* Fixes issues from broken tests 
* Adds missing tests

If we end up agreeing with #439, then the following TODOs need to be resolved

- [x] Add minor tests to `MessageLib.t.sol` for new generic messages
- [ ] Reduce size of `Hub`


hieronx:
>  Add minor tests to MessageLib.t.sol for new generic messages

These already exist: https://github.com/centrifuge/protocol-v3/blob/decouple-request-managers/test/common/unit/libraries/MessageLib.t.sol#L264

PR #448, opened 2025-06-11
Ref: Remove non-queueing support and always queueing

lemunozm:


test/spoke/integration/SyncDeposit.t.sol
```diff
@@ -176,21 +176,10 @@ contract SyncDepositTest is SyncDepositTestHelper {
...
-            vm.startSnapshotGas("SyncDepositVault", "deposit_withoutQueue");
+            vm.startSnapshotGas("SyncDepositVault", "deposit");
         }
         // _assertDepositEvents(syncVault, shares.toUint128(), pricePoolPerShare, pricePoolPerAsset);
-        syncVault.deposit(amount / 2, self);
-        if (snap) {
-            vm.stopSnapshotGas();
-        }
-
-        balanceSheet.setQueue(syncVault.poolId(), syncVault.scId(), true);
```
lemunozm, line 187: IIUC, the measure of this was wrong. We're already queuing by default, so the previous measurement was also taking into account that it was queuing

lemunozm: The snapshot value difference was that only the first `deposit()` from the test was increasing `sharesQueue.assetCounter`

wischli: Yeah seems like this wasn't adapted when we we switched to queuing by default.

wischli (approved):
Quick and clean!

PR #449, opened 2025-06-11
Ref: remove gateway ward from escrow

lemunozm:
Thread: https://kflabs.slack.com/archives/C04HF8SHRRC/p1749645728300619?thread_ts=1748869090.686359&cid=C04HF8SHRRC

The `gateway` ward from escrow is highly migratable, instead `tokenRecoverer` is an immutable contract that should never be migrated.

NOTE: I would like to create an EndToEnd with the whole refunding cycle, it is:

1. Sending an overpaid message
2. Get the PoolEscrow refunded
3. Send a second message to reuse those funds.

In this PR or in a follow one

### DIFF
--- a/src/common/TokenRecoverer.sol
+++ b/src/common/TokenRecoverer.sol
@@ -22,14 +22,21 @@ contract TokenRecoverer is Auth, ITokenRecoverer {
     {
         root.relyContract(address(target), address(this));
 
+        withdrawTokens(target, token, tokenId, to, amount);
+
+        root.denyContract(address(target), address(this));
+
+        emit RecoverTokens(target, token, tokenId, to, amount);
+    }
+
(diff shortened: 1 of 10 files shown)


src/common/TokenRecoverer.sol
```diff
@@ -22,14 +22,21 @@ contract TokenRecoverer is Auth, ITokenRecoverer {
     {
         root.relyContract(address(target), address(this));
 
+        withdrawTokens(target, token, tokenId, to, amount);
+
+        root.denyContract(address(target), address(this));
```
lemunozm, line 27: We could completely avoid the `tokenRecoverer` ward by using the root.relyContract, but I think that would dispatch root-level events that we want to avoid

lemunozm:
Finally decided not to go with this

PR #450, opened 2025-06-12
Test: Finish BalanceSheet UTs

lemunozm:
- Full coverage in `BalanceSheet`
- Remove previous `BalanceSheet` tests that have now been ported to the UTs. Now, between the current UTs and the End-to-End we should be able to test all possible paths & casuistics related to `BalanceSheet`


test/spoke/integration/BalanceSheet.t.sol
```diff
@@ -87,316 +31,4 @@ contract BalanceSheetTest is BaseTest {
         assertEq(balanceSheet.wards(address(messageDispatcher)), 1);
```
lemunozm, line 31: What we're testing here is the deployment scripts, not things related to the contract. I think we should move all these tests to a common test inside `test/script`. WDYT?

src/spoke/BalanceSheet.sol
```diff
@@ -270,8 +270,8 @@ contract BalanceSheet is Auth, Multicall, Recoverable, IBalanceSheet, IBalanceSh
         address to,
         uint256 amount
     ) external authOrManager(poolId) {
-        require(!root.endorsed(from), CannotTransferFromEndorsedContract());
-        IShareToken token = IShareToken(spoke.shareToken(poolId, scId));
+        require(!root.endorsed(from), CannotTransferFromEndorsedContract()); // mainly from the escrow
```
hieronx: i think this comment would below in the natspec (if needed at all)

lemunozm: Correct. I will remove this for now. Maybe it was just me

lemunozm: Done

hieronx (approved):

PR #451, opened 2025-06-12
Docs: update diagrams

lemunozm:
### Spoke diagram

image

### Common diagram

image

### Message diagram

image


docs/architecture/spoke.puml
```diff
@@ -0,0 +1,66 @@
...
+hide empty members
+
+package misc {
+    abstract Recoverable
+    class Escrow
+}
+
+package common {
+    class Gateway
+    class MessageDispatcher
```
wischli, line 6: MessageProcessor is missing here or did you omit it on purpose?

wischli (approved):

wischli:
Thanks for handling these diagrams and the code cleanups! 

If you have the time, could you also update the `messaging.puml`?

hieronx:
> Spoke diagram

NIT: I feel like hierarchically, `common` and `misc` should visually be on top rather than on the bottom of the diagram. Wdyt @lemunozm ?

lemunozm:
Having them on the bottom feels like they are the bare bones over what the other is built on (?)

No strong opinion, I can flip the diagram vertically if you want

docs/architecture/spoke.puml
lemunozm: I did on purpose, because if not, things get messy and theoretically, I'm only representing what the `spoke` is calling to, which is only `MessageDispatcher`

hieronx (approved):

PR #454, opened 2025-06-16
ref: import order

wischli:
* **Fixes ONLY import order** for all files based on the folllowing priority list

```python
IMPORT_PRIORITY = [
    "src/misc",
    "src/common", 
    "src/vaults",
    "src/hub",
    "src/spoke",
    "src/hooks",
    "src/managers",
    "script",
    "test",
    "forge-std"
]
```

* Adds `script/utils/fix_imports.py` which we could utilize in CI as check in the future
* Within an import block, sorts by ascending line length order

```diff
- import {AccountId} from "src/common/types/AccountId.sol";
- import {AssetId, newAssetId} from "src/common/types/AssetId.sol";
+ import {PoolId} from "src/common/types/PoolId.sol";
+ import {AccountId} from "src/common/types/AccountId.sol";
import {ISafe} from "src/common/interfaces/IGuardian.sol";
- import {PoolId} from "src/common/types/PoolId.sol";
- import {ShareClassId} from "src/common/types/ShareClassId.sol";
+ import {ShareClassId} from "src/common/types/ShareClassId.sol";
+ import {AssetId, newAssetId} from "src/common/types/AssetId.sol";
import {VaultUpdateKind} from "src/common/libraries/MessageLib.sol"
```

### DIFF
--- a/script/utils/fix_imports.py
+++ b/script/utils/fix_imports.py
@@ -1,0 +1,202 @@
+#!/usr/bin/env python3
+
+import os
+import re
+import subprocess
+from typing import List, Dict, Set
+
+# Priority order for imports
+IMPORT_PRIORITY = [
+    "src/misc",
+    "src/common", 
+    "src/vaults",
+    "src/hub",
+    "src/spoke",
+    "src/managers",
+    "src/hooks",
+    "script",
+    "test"
+]
+
+def get_all_solidity_files() -> List[str]:
+    """Get all Solidity files excluding lib directory"""
+    result = subprocess.run(
+        ["find", ".", "-name", "*.sol", "-not", "-path", "./lib/*"],
+        capture_output=True,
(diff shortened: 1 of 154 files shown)


script/utils/fix_imports.py
```diff
@@ -0,0 +1,202 @@
+#!/usr/bin/env python3
```
wischli, line 1: Happy to remove the script. Just thought it might be handy in the future. I can envision this to run in a `post-commit` CI action.

hieronx: I think it's handy to keep yes!

hieronx (approved):

lemunozm:
Looks really cool! Just one thing, should not be `vaults` below `spoke` given spoke is more generic/core than vaults?

PR #455, opened 2025-06-16
ci: add import checks

wischli:
* Adds `check-order` CLI to import script
* Adds out filtering of cached/out files
* Adds import order check and unused import variables each as separate step to CI

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -46,6 +46,32 @@ jobs:
         env:
           FOUNDRY_PROFILE: ci
 
+  import-order:
+    runs-on: ubuntu-latest
+    steps:
+      - uses: actions/checkout@v3
+
+      - name: Setup Python
+        uses: actions/setup-python@v4
+        with:
+          python-version: "3.x"
+
+      - name: Check import order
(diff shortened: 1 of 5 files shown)


wischli:
PoC import order failure: https://github.com/centrifuge/protocol-v3/actions/runs/15684521032/job/44184196056

.github/workflows/ci.yml
```diff
@@ -46,6 +46,32 @@ jobs:
...
+
+      - name: Setup Python
+        uses: actions/setup-python@v4
+        with:
+          python-version: "3.x"
+
+      - name: Check import order
+        run: python3 script/utils/fix_imports.py --check-order
+
+  unused-imports:
```
hieronx, line 62: Can you remove `check-unused-imports` (line 107 onwards)? :)

wischli: @hieronx Can you elaborate on why you believe this to not be helpful? IMO it's inconsistent to check and require ascending line lengths without checking for unused imports.

hieronx: It was just duplicated, i had already added `check-unused-imports` and you then also added `unused-imports`.

PR #459, opened 2025-06-19
Add `protected` modifier to `Hub.notifyDepositRedeem` methods

hieronx:


wischli (approved):

src/hub/Hub.sol
```diff
@@ -706,6 +710,9 @@ contract Hub is Multicall, Auth, Recoverable, IHub, IHubGatewayHandler, IHubGuar
     /// @dev Ensure the sender is authorized
     function _auth() internal auth {}
 
+    /// @dev Protect against reentrancy
+    function _protected() internal protected {}
+
     /// @dev Ensure the method can be used without reentrancy issues, and the sender is a pool admin
     function _isManager(PoolId poolId) internal protected {
```
wischli, line 717: Nit: The new `_protected` should also be applied here

wischli (approved):

src/hub/Hub.sol
hieronx: It doesn't really make a difference. Since it's only in the code once, and all `_isManager` calls don't copy-paste the modifier.

PR #460, opened 2025-06-20
Mainnet deployments

gpmayorga:
## WIP
CreateX factory
Ledger support
Catapulta.sh wrapper support


script/CommonDeployer.s.sol
```diff
@@ -143,4 +153,107 @@ abstract contract CommonDeployer is Script, JsonRegistry {
...
+    // Helper function to deploy basic contracts in a separate scope to avoid stack too deep errors
+    function _deployBasicContracts(address deployer, uint128 messageGasLimit, uint128 maxBatchSize) private {
+        // Root
+        bytes32 rootSalt = generateSalt("root");
+        bytes memory rootBytecode = abi.encodePacked(
+            type(Root).creationCode,
+            abi.encode(DELAY, deployer)
+        );
+        root = Root(create3(rootSalt, rootBytecode));
+        console.log("Root deployed at:", address(root));
```
lemunozm: We should only log this if `!isTests`, otherwise we would spam all ITs with a lot of log lines

script/HubDeployer.s.sol
```diff
@@ -119,4 +170,31 @@ contract HubDeployer is CommonDeployer {
...
+    }
+
+    // Helper function to deploy Hub contract in a separate scope to avoid stack too deep errors
+    // The Hub constructor requires 7 parameters which was causing compilation issues
+    function _deployHubContract(address deployer) private returns (Hub) {
+        bytes32 hubSalt = generateSalt("hub");
+        
+        // Store variables to reduce stack pressure
+        address gatewayAddr = address(gateway);
+        address hubHelpersAddr = address(hubHelpers);
```
lemunozm, line 176: Curious, what do you mean by stack pressure?

script/SpokeDeployer.s.sol
```diff
@@ -195,4 +196,133 @@ contract SpokeDeployer is CommonDeployer {
...
+
+    // Helper function to deploy escrows, factories, and managers in a separate scope to avoid stack too deep errors
+    function _deployEscrowsAndFactories(address deployer) private {
+        // RouterEscrow
+        bytes32 routerEscrowSalt = generateSalt("routerEscrow");
+        bytes memory routerEscrowBytecode = abi.encodePacked(
+            type(Escrow).creationCode,
+            abi.encode(deployer)
+        );
+        routerEscrow = Escrow(create3(routerEscrowSalt, routerEscrowBytecode));
```
lemunozm: Another option to solve the stack too deep issue without splitting in different functions (that maybe becomes more readable):
```suggestion
        routerEscrow = Escrow(create3(
            generateSalt("routerEscrow"), 
            abi.encodePacked(
                type(Escrow).creationCode,
                abi.encode(deployer)
            )
        ));
```

.gitmodules
```diff
@@ -4,3 +4,6 @@
 [submodule "lib/chimera"]
 	path = lib/chimera
 	url = https://github.com/Recon-Fuzz/chimera
+[submodule "lib/createx-forge"]
+	path = lib/createx-forge
+	url = https://github.com/radeksvarz/createx-forge
```
lemunozm, line 9: We should add the next line to `remappings.txt`:
```
createx-forge/=lib/createx-forge/
```

script/CommonDeployer.s.sol
lemunozm: Actually I think we do not need those lines since the contract addresses are already register in `_commonRegister()` method

.gitmodules
lemunozm: Fixed here: https://github.com/centrifuge/protocol-v3/pull/473

script/CommonDeployer.s.sol
hieronx: Removed now!

PR #464, opened 2025-06-23
Check pool id in merkle proof manager factory

hieronx:


wischli (approved):
Nit: Would be great to cover changes by tests though I understand for this one we don't really win anything apart from a slight better code coverage.

hieronx:
Yeah I'll note to add test coverage for the factory in a followup PR!

src/managers/MerkleProofManager.sol
```diff
@@ -115,15 +116,18 @@ contract MerkleProofManager is IMerkleProofManager, IUpdateContract {
...
 
-    constructor(address spoke_) {
+    constructor(ISpoke spoke_) {
         spoke = spoke_;
     }
 
     /// @inheritdoc IMerkleProofManagerFactory
     function newManager(PoolId poolId) external returns (IMerkleProofManager) {
-        MerkleProofManager manager = new MerkleProofManager{salt: bytes32(uint256(poolId.raw()))}(poolId, spoke);
+        require(spoke.isPoolActive(poolId), InvalidPoolId());
```
lemunozm, line 127: Curious, why the check? Could not ideally create the manager before having been notified by the hub that the pool is active?

hieronx: It came out of https://github.com/electisec-fellowship-block-7/centrifuge-v3-vaults-report/issues/5

I think it can be confusing if anyone can just create a manager for any ID. It might also be problematic for the indexer, as it would need to check whether a manager actually needs to be tracked

PR #466, opened 2025-06-23
Fix: normalize delta positive

lemunozm:
Internal thread: https://kflabs.slack.com/archives/C07PG2EUR9C/p1750674308484539?thread_ts=1750660523.181429&cid=C07PG2EUR9C
Related issue: https://github.com/electisec-fellowship-block-7/centrifuge-v3-vaults-report/issues/9


src/spoke/BalanceSheet.sol
```diff
@@ -169,7 +169,10 @@ contract BalanceSheet is Auth, Multicall, Recoverable, IBalanceSheet, IBalanceSh
         if (shareQueue.isPositive || shareQueue.delta == 0) {
             shareQueue.delta += shares;
             shareQueue.isPositive = true;
-        } else if (shareQueue.delta > shares) {
+            if (shareQueue.delta == 0) {
+                shareQueue.isPositive = false;
```
lemunozm, line 173: The neutral value should go along `isPositive = false` because that's the value contained in an uninitialized `shareQueue` storage.

src/spoke/BalanceSheet.sol
```diff
@@ -169,7 +169,10 @@ contract BalanceSheet is Auth, Multicall, Recoverable, IBalanceSheet, IBalanceSh
...
+            }
+        } else if (shareQueue.delta >= shares) {
             shareQueue.delta -= shares;
             shareQueue.isPositive = false;
         } else {
```
lemunozm, line 178: NOTE: change from `>` to `>=` on purpose to force the `delta = 0` in this branch where `isPositive` is set to `false`

src/spoke/BalanceSheet.sol
```diff
@@ -191,6 +194,9 @@ contract BalanceSheet is Auth, Multicall, Recoverable, IBalanceSheet, IBalanceSh
         } else if (shareQueue.delta > shares) {
             shareQueue.delta -= shares;
             shareQueue.isPositive = true;
+            if (shareQueue.delta == 0) {
+                shareQueue.isPositive = false;
+            }
```
wischli: This branch seems to be unreachable because the upper check is `shareQueue.delta > uint128(shares)` such that the delta can never be zero. I suppose you intended to move this `if` branch to the upper level `if else`?

lemunozm: Ah correct! Then that can be removed. Good catch!

wischli (approved):

PR #471, opened 2025-06-25
Test: Add UT to ensure price = 1.0 when no holdings

lemunozm:
It's simple, but I think it's enough to cover the case.

### DIFF
--- a/test/hub/unit/HubHelpers.t.sol
+++ b/test/hub/unit/HubHelpers.t.sol
@@ -2,6 +2,7 @@
 pragma solidity ^0.8.28;
 
 import {IAuth} from "src/misc/interfaces/IAuth.sol";
+import {d18} from "src/misc/types/D18.sol";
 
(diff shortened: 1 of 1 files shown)


test/hub/unit/HubHelpers.t.sol
```diff
@@ -2,6 +2,7 @@
 pragma solidity ^0.8.28;
 
 import {IAuth} from "src/misc/interfaces/IAuth.sol";
+import {d18} from "src/misc/types/D18.sol";
```
hieronx:
```suggestion
import {d18} from "src/misc/types/D18.sol";
import {IAuth} from "src/misc/interfaces/IAuth.sol";
```

hieronx (commented):
Very simple :)

wischli (approved):

hieronx (approved):

PR #472, opened 2025-06-25
Add `receive()` method to merkle proof manager

hieronx:


lemunozm (approved):

wischli (approved):
Nit: We should add a simple test like

```solidity
function testReceiveEther() public {
    (bool success,) = address(manager).call{value: 1 ether}("");
    assertTrue(success, "Failed to send Ether");

    assertEq(address(manager).balance, 1 ether);
}
```

hieronx:
> Nit: We should add a simple test like

Was planning to (why I hadn't requested reviews yet), but was fighting with a complex integration test using the actual merkle proof logic to transfer ETH. Agreed a unit is much more feasible here for now, so thanks for writing the sample!

wischli (approved):
Thanks for adding the test! Would have been nice to see it with the actual merkle tree logic but this should suffice 😄

PR #474, opened 2025-06-26
Fix: Extra validation checks to unlinked vaults

lemunozm:
Thread: https://kflabs.slack.com/archives/C07PG2EUR9C/p1750936510769599?thread_ts=1750934812.842919&cid=C07PG2EUR9C


wischli (approved):
Thanks for the quick fix! As a nit, we should add a test for this but can be done later.

lemunozm:
That's right!

PR #476, opened 2025-06-27
Swap update order in `BalanceSheet` for CEI-compatibility

hieronx:


src/common/interfaces/IGatewaySenders.sol
```diff
@@ -124,6 +124,13 @@ interface IHubMessageSender is ILocalCentrifugeId {
 
 /// @notice Interface for dispatch-only gateway
 interface ISpokeMessageSender is ILocalCentrifugeId {
+    struct UpdateData {
```
hieronx, line 127: Went back and forth on this, but overall felt this was the cleanest outcome:
- Struct is needed regardless, to avoid stack too deep issue in balance sheet
- These 4 params are shared between the 2 update methods, so it makes sense to re-use
- Putting it in `ISpokeMessageSender` makes the calls and events shorter, which overall makes the code easier to read IMO

hieronx: I didn't add `extraGasLimit` because it semantically seemed unrelated to the other params + isn't reset so no need to store in the struct.

src/common/interfaces/IGatewaySenders.sol
```diff
@@ -124,6 +124,13 @@ interface IHubMessageSender is ILocalCentrifugeId {
...
+        uint128 netAmount;
```
lemunozm, line 128: Curious about the `netAmount` name. What do you mean by "net"?

lemunozm (approved):
LGTM!!

src/common/interfaces/IGatewaySenders.sol
hieronx: "A net value is the resultant amount after accounting for the sum or difference of two or more variables". Wanted to make it clear that it is the sum of the deposits/withdrawals or issuances/revocations.

PR #478, opened 2025-06-29
Fix legacy transfer

hieronx:


src/vaults/AsyncRequestManager.sol
```diff
@@ -141,7 +144,11 @@ contract AsyncRequestManager is Auth, Recoverable, IAsyncRequestManager {
         state.pendingRedeemRequest = state.pendingRedeemRequest + shares_;
 
         _sendRequest(vault_, RequestMessageLib.RedeemRequest(controller.toBytes32(), shares_).serialize());
-        balanceSheet.transferSharesFrom(vault_.poolId(), vault_.scId(), sender_, owner, address(globalEscrow), shares_);
+        if (transfer) {
+            balanceSheet.transferSharesFrom(
+                vault_.poolId(), vault_.scId(), sender_, owner, address(globalEscrow), shares_
+            );
```
lemunozm, line 150: That would mean that the legacy vault will never dispatch the `BalanceSheet.TransferSharesFrom` event. Hope this will not lead to some indexing issue

lemunozm (approved):
LGTM! This was an amazing catch by the auditor

src/vaults/AsyncRequestManager.sol
hieronx: True! I think that's fine.

PR #481, opened 2025-06-30
Ref: split deployers

lemunozm:
### Changes

- [x] Split deployers
- [x] Add managers' deployment
- [x] Fix stack too deep issues

### New schema

image

### DIFF
--- a/script/VaultsDeployer.s.sol
+++ b/script/VaultsDeployer.s.sol
@@ -1,0 +1,154 @@
+// SPDX-License-Identifier: BUSL-1.1
+pragma solidity 0.8.28;
+
+import {Escrow} from "src/misc/Escrow.sol";
+import {IAuth} from "src/misc/interfaces/IAuth.sol";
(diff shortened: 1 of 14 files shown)


hieronx (approved):
🔥

script/VaultsDeployer.s.sol
```diff
@@ -0,0 +1,147 @@
...
+        globalEscrow.rely(address(root));
+        asyncVaultFactory.rely(address(root));
+        syncDepositVaultFactory.rely(address(root));
+
+        // Rely others
+        routerEscrow.rely(address(vaultRouter));
+        syncManager.rely(address(syncDepositVaultFactory));
+
+        // Rely VaultRouter
+        gateway.rely(address(vaultRouter));
```
lemunozm, line 125: FYI. This rely dependency seems off since vaults are a layer on top of spoke now. But not sure how to fix it without opening any security issues in the Gateway.

PR #482, opened 2025-06-30
Fix share amount in transfer checks of max methods

hieronx:


src/vaults/AsyncRequestManager.sol
```diff
@@ -509,20 +509,23 @@ contract AsyncRequestManager is Auth, Recoverable, IAsyncRequestManager {
...
+        if (!_canTransfer(vault_, user, address(0), shares)) return 0;
         assets = uint256(investments[vault_][user].maxWithdraw);
     }
 
     /// @inheritdoc IRedeemManager
     function maxRedeem(IBaseVault vault_, address user) public view returns (uint256 shares) {
-        if (!_canTransfer(vault_, user, address(0), 0)) return 0;
-        shares = uint256(_maxRedeem(vault_, user));
+        AsyncInvestmentState memory state = investments[vault_][user];
+        shares = uint256(_assetToShareAmount(vault_, state.maxWithdraw, state.redeemPrice, MathLib.Rounding.Down));
```
lemunozm: Why not call to `uint256(_maxRedeem(vault_, user));` as it was before?

src/vaults/AsyncRequestManager.sol
```diff
@@ -509,20 +509,23 @@ contract AsyncRequestManager is Auth, Recoverable, IAsyncRequestManager {
...
-        if (!_canTransfer(vault_, ESCROW_HOOK_ID, user, 0)) return 0;
         shares = uint256(investments[vault_][user].maxMint);
+        if (!_canTransfer(vault_, ESCROW_HOOK_ID, user, shares)) return 0;
     }
 
     /// @inheritdoc IRedeemManager
     function maxWithdraw(IBaseVault vault_, address user) public view returns (uint256 assets) {
-        if (!_canTransfer(vault_, user, address(0), 0)) return 0;
+        AsyncInvestmentState memory state = investments[vault_][user];
+        uint128 shares = _assetToShareAmount(vault_, state.maxWithdraw, state.redeemPrice, MathLib.Rounding.Down);
```
lemunozm: Same here

src/vaults/AsyncRequestManager.sol
hieronx: We need to have the `_canTransfer` check, that uses `shares` only in `maxRedeem`. `_maxRedeem` should not include the transfer check

lemunozm: But as long as I see these two lines are exactly `_maxRedeem()`, so this method implementation could be:

```solidity
shares = _maxRedeem(vault_, user);
if (!_canTransfer(vault_, user, address(0), shares)) return 0;
```

hieronx: True! Fixed.

lemunozm (approved):
🚀

PR #483, opened 2025-06-30
Deploy: allow to use a living root contract in deployment

lemunozm:


script/CommonDeployer.s.sol
```diff
@@ -140,13 +144,16 @@ abstract contract CommonDeployer is Script, JsonRegistry, CreateXScript {
...
     }
 
-    function _commonRegister() private {
-        register("root", address(root));
+    function _commonRegister(address inputRoot) private {
+        if (inputRoot == address(0)) {
+            register("root", address(root));
+            // Otherwise already present in load_vars.sh and not needed to be registered
+        }
         // register("adminSafe", address(adminSafe)); => Already present in load_vars.sh and not needed to be registered
```
lemunozm, line 162: I understand this would be true. I mean, if the already deployed root is used, then it will be present in `load_vars.sh` like `safeAdmin` is

lemunozm:
Discussed with @gpmayorga about move the envvars to enviromental file config (preferred in another PR)

gpmayorga (approved):

PR #485, opened 2025-07-01
Deploy: Use CBD instances

lemunozm:
### Description

image

PR #486, opened 2025-07-01
Deploy: Use CBD instances for batching actions

lemunozm:
The initial approach was this: https://github.com/centrifuge/protocol-v3/pull/485, where we batched everything: contract creation and rely/deny/file actions. Nevertheless batching contract creation creates bytecodes > 300KB.

This approach only batches the rely/deny/file actions. To achieve that, we have new entities with the form `*ActionBatcher` for each deployer. These entities now have the deployer access instead of the deployer itself, to be able to perform the auth actions over the contracts

image

#### Sub-feature: `postDeploy` method

The pipeline is the follows when calling `deployX()`:
1. In a tree, all deployers will execute `preDeployX()` in order (the same deploy impl as before)
1. After that, all deployers will execute `postDeployX()` in order.

The idea is that in the future, we can add more post-deploy actions to the deployers without worrying about wrong states or breaking test configurations.


script/FullDeployer.s.sol
```diff
@@ -51,16 +55,16 @@ contract FullDeployer is HubDeployer, ExtendedSpokeDeployer {
             version: vm.envOr("VERSION", bytes32(0))
         });
 
-        // Use the regular deployment functions - they now use CreateX internally
-        deployFull(input, msg.sender);
+        FullActionBatcher batcher = new FullActionBatcher();
```
lemunozm, line 68: The batcher is constructed using CREATE1 because it's a temporal contract just used for the permissions during the deployment phase.

Because all other contracts are created using CREATE3, the batcher address does not affect the other contract addresses

script/HubDeployer.s.sol
```diff
@@ -12,11 +12,86 @@ import {HubHelpers} from "src/hub/HubHelpers.sol";
...
+    Accounting accounting;
+    Holdings holdings;
+    ShareClassManager shareClassManager;
+    HubHelpers hubHelpers;
+    Hub hub;
+    IdentityValuation identityValuation;
+}
+
+contract HubActionBatcher is CommonActionBatcher, HubConstants {
+    function engageHub(HubReport memory report) public {
```
lemunozm: I tried to add this pattern: https://github.com/centrifuge/vaults-internal/blob/b8f64daddcec8c9a53143c98a80119b2b25d5399/test/spell/Spell.sol#L54

Nevertheless, it becomes a bit dirty because we need a `done` variable per `*ActionBatcher`. I really think, although those methods are public, they're fully protected because the batcher has a deployer access that will be removed in the final stage. So our `done` check is `wards == batcher` in this case.

Do you still have an strong opinion @hieronx?

hieronx: Why not implement the done variable in a modifier?

lemunozm: Fixed here: https://github.com/centrifuge/protocol-v3/pull/490

hieronx (approved):

PR #490, opened 2025-07-02
Ref: Add ContractUpdater contract

lemunozm:
### The problem

We're using `spoke` as source of truth for any contract that implements `IUpdateContract` interface. Nevertheless, if we need to migrate the spoke (probably in v3.1) that would mean we need to also migrate all contract that use that spoke as source of truth. This is especially annoying for the managers who:
- Are deployed by `pool` or worst, by `pool`/`scId` tuple
- Have an immutable reference to spoke just to detect the correct origin for `IUpdateContract` calls

### Proposed solution

A non-migratable `ContractUpdater` that receives the `UpdateContractMessage` from the `MessageDispatcher/Processor` and calls the `IUpdateContract` interface in the target contract. `ContractUpdater` serves as immutable source of truth for those messages, allowing spokes to be more easily migratable.

### Questions

- Should we also move `UpdateRestriction` to this contract?

**Blocked by https://github.com/centrifuge/protocol-v3/pull/486**


hieronx (approved):

hieronx (dismissed):

hieronx (approved):

src/managers/OnOfframpManager.sol
```diff
@@ -51,7 +50,7 @@ contract OnOfframpManager is IOnOfframpManager {
     function update(PoolId poolId_, ShareClassId scId_, bytes calldata payload) external {
         require(poolId == poolId_, InvalidPoolId());
         require(scId == scId_, InvalidShareClassId());
-        require(msg.sender == spoke, NotSpoke());
```
xmxanuel, line 54: nitpick: error type could be re-named from `NotSpoke`

src/managers/MerkleProofManager.sol
```diff
@@ -20,13 +20,13 @@ contract MerkleProofManager is IMerkleProofManager, IUpdateContract {
...
 
     PoolId public immutable poolId;
-    address public immutable spoke;
+    address public immutable contractUpdater;
 
     mapping(address strategist => bytes32 root) public policy;
 
     constructor(PoolId poolId_, address spoke_) {
         poolId = poolId_;
-        spoke = spoke_;
```
xmxanuel, line 29:
```diff
- contractUpdater = spoke_;
+ contractUpdater = contractUpdater_;
```

src/managers/OnOfframpManager.sol
hieronx: What would you think is a better error name?

lemunozm: Good catch!

PR #491, opened 2025-07-02
Feat: Add share hook gas support

lemunozm:
Thread: https://kflabs.slack.com/archives/C07PG2EUR9C/p1751445755299359

### Changes

- Rename from `transferShares` to `crossTransferShares` as discussed previously in some thread.
- Add `extraGasLimit` to:
  - `hub.issueShares()`
  - `hub.revokeShares()`
- Add `remoteExtraGasLimit` to:
  - `spoke.crossTransferShares`. In this case, the gas limit value needs to be in the message to be executed from the hub side.


lemunozm:
I'm exceeding the Hub contract size. Options I suggest:
- Remove the immutability from hubRegistry.
- Remove `file()` method. AFAIK, we don't really need a `file()` method here, we can just re-upload the Hub

src/spoke/Spoke.sol
```diff
@@ -87,12 +87,14 @@ contract Spoke is Auth, Recoverable, ReentrancyProtection, ISpoke, ISpokeGateway
     //----------------------------------------------------------------------------------------------
 
     /// @inheritdoc ISpoke
-    function transferShares(uint16 centrifugeId, PoolId poolId, ShareClassId scId, bytes32 receiver, uint128 amount)
-        external
-        payable
-        payTransaction
-        protected
-    {
+    function crossTransferShares(
```
hieronx: Nit, I would do
```suggestion
    function crosschainTransfer(
```

Matches similar naming in https://eips.ethereum.org/EIPS/eip-7802

src/spoke/Spoke.sol
```diff
@@ -104,7 +106,7 @@ contract Spoke is Auth, Recoverable, ReentrancyProtection, ISpoke, ISpokeGateway
         share.burn(address(this), amount);
 
         emit TransferShares(centrifugeId, poolId, scId, msg.sender, receiver, amount);
```
hieronx: Would also rename this then
```suggestion
        emit CrosschainTransfer(centrifugeId, poolId, scId, msg.sender, receiver, amount);
```

src/spoke/Spoke.sol
lemunozm: Should we avoid the "shares" name? Maybe it makes things more confusing?

lemunozm: I would leave the "Shares" name, it's true that the EIP-7802 doesn't say shares, but it is in the context of shares. From our level can be confused with assets

hieronx: Yeah fair, then would do `crosschainTransferShares`

src/spoke/Spoke.sol
lemunozm: I've renamed it to `InitiateTransferShares` because it matches with the other stage `ExecuteTransferShares`. Both events compose the cross-chain transfer.

Tell me if you disagree, and I will revert it

hieronx: Also fine!

src/hub/Hub.sol
```diff
@@ -36,8 +36,7 @@ contract Hub is Multicall, Auth, Recoverable, IHub, IHubGatewayHandler, IHubGuar
     using MathLib for uint256;
     using RequestCallbackMessageLib for *;
 
-    IHubRegistry public immutable hubRegistry;
-
+    IHubRegistry public hubRegistry;
     IGateway public gateway;
     IHoldings public holdings;
     IHubHelpers public hubHelpers;
```
hieronx:
```suggestion
    IGateway public gateway;
    IHoldings public holdings;
    IHubHelpers public hubHelpers;
    IHubRegistry public hubRegistry;
```

hieronx (commented):
Only 1 nit!

src/hub/Hub.sol
lemunozm: Done

hieronx (approved):
🔥

PR #492, opened 2025-07-02
Deploy: CBD actions protected

lemunozm:


script/FullDeployer.s.sol
```diff
@@ -58,15 +58,17 @@ contract FullDeployer is HubDeployer, ExtendedSpokeDeployer {
...
-        saveDeploymentOutput();
-
         bool isMainnet = keccak256(abi.encodePacked(environment)) == keccak256(abi.encodePacked("mainnet"));
         if (isMainnet) {
             removeFullDeployerAccess(batcher);
         } else {
             guardian.file("safe", address(adminSafe));
         }
 
+        batcher.lock();
```
lemunozm, line 68: Here we lock the batch and no more actions can be done, but it means another tx

hieronx (approved):

PR #493, opened 2025-07-03
Add freely transferable hook, fix managers

hieronx:


src/managers/MerkleProofManager.sol
```diff
@@ -24,9 +24,9 @@ contract MerkleProofManager is IMerkleProofManager, IUpdateContract {
 
     mapping(address strategist => bytes32 root) public policy;
 
-    constructor(PoolId poolId_, address spoke_) {
+    constructor(PoolId poolId_, address contractUpdater_) {
         poolId = poolId_;
-        contractUpdater = spoke_;
+        contractUpdater = contractUpdater_;
```
mustermeiszer, line 29: That change I don't understand.

src/hooks/FreelyTransferable.sol
```diff
@@ -0,0 +1,171 @@
...
+        returns (bool)
+    {
+        uint128 fromHookData = uint128(hookData.from);
+        if (fromHookData.getBit(FREEZE_BIT) == true && !root.endorsed(from)) {
+            // Source is frozen and not endorsed
+            return false;
+        }
+
+        uint128 toHookData = uint128(hookData.to);
+        if (toHookData.getBit(FREEZE_BIT) == true) {
```
mustermeiszer, line 80: QUESTION 

Should we have this here? IIRC, we said we don't need that from a legal point of view?

mustermeiszer (approved):

src/managers/MerkleProofManager.sol
hieronx: It was just a mistake in a previous PR, where the constructor param wasn't properly renamed

src/hooks/FreelyTransferable.sol
hieronx: I think having freeze feature is always good optionality

src/hooks/FreelyTransferable.sol
```diff
@@ -0,0 +1,171 @@
...
+        uint128 hookData = uint128(validUntil) << 64;
+        hookData.setBit(FREEZE_BIT, isFrozen(token, user));
+        IShareToken(token).setHookData(user, bytes16(hookData));
+
+        emit UpdateMember(token, user, validUntil);
+    }
+
+    /// @inheritdoc IMemberlist
+    function isMember(address token, address user) external view returns (bool isValid, uint64 validUntil) {
+        validUntil = abi.encodePacked(IShareToken(token).hookDataOf(user)).toUint64(0);
```
lemunozm, line 159: Not a blocker:

`validUntil` is obtained using both `>> 64` and with `toUint64(0)`. Maybe we can normalize or isolate that into a method/lib?

lemunozm (approved):
LGMT!

src/hooks/FreelyTransferable.sol
hieronx: I think the hooks in general should be cleaned up in v3.1

wischli: Added as generic item to the v3.1 board ([ref](https://www.notion.so/Clean-up-hooks-2302eac24e178008b016dddb9ee95f84?source=copy_link))

hieronx: I had already added `Implement BaseTransferHook` btw, just not in the v3.1 list

PR #495, opened 2025-07-03
Feat: GasService with fixed gas values

lemunozm:
- Rename `maxBatchSize` to `batchGasLimit`
- Remove `messageGasLimit` configuration variable.
- Static gas values per message values and getters for UX experience

### DIFF
--- a/script/FullDeployer.s.sol
+++ b/script/FullDeployer.s.sol
@@ -60,8 +60,8 @@ contract FullDeployer is HubDeployer, ExtendedSpokeDeployer {
             centrifugeId: centrifugeId,
             root: IRoot(vm.envAddress("ROOT")),
             adminSafe: ISafe(vm.envAddress("ADMIN")),
-            messageGasLimit: uint128(vm.envUint("MESSAGE_COST")),
-            maxBatchSize: uint128(vm.envUint("MAX_BATCH_SIZE")),
+            messageGasLimit: uint128(vm.envUint("MESSAGE_GAS_LIMIT")),
+            batchGasLimit: uint128(vm.envUint("BATCH_GAS_LIMIT")),
             version: vm.envOr("VERSION", bytes32(0))
         });
 
--- a/script/FullDeployer.s.sol
+++ b/script/FullDeployer.s.sol
@@ -60,7 +60,6 @@ contract FullDeployer is HubDeployer, ExtendedSpokeDeployer {
             centrifugeId: centrifugeId,
             root: IRoot(vm.envAddress("ROOT")),
             adminSafe: ISafe(vm.envAddress("ADMIN")),
-            messageGasLimit: uint128(vm.envUint("MESSAGE_GAS_LIMIT")),
             batchGasLimit: uint128(vm.envUint("BATCH_GAS_LIMIT")),
             version: vm.envOr("VERSION", bytes32(0))
         });
--- a/src/common/GasService.sol
+++ b/src/common/GasService.sol
@@ -127,7 +127,7 @@ contract GasService is IGasService {
         } else if (kind == MessageType.MaxSharePriceAge) {
             return maxSharePriceAge;
         } else {
-            return 0; // Unreachable
+            return type(uint128).max; // Unreachable
         }
     }
 }
--- a/src/common/GasService.sol
+++ b/src/common/GasService.sol
@@ -16,7 +16,6 @@ contract GasService is IGasService {
     uint128 public constant SMALL_COST = 100_000;
 
     uint128 internal immutable _batchGasLimit;
-    uint128 internal immutable _messageGasLimit;
 
     uint128 public immutable scheduleUpgrade;
     uint128 public immutable cancelUpgrade;
@@ -42,9 +41,8 @@ contract GasService is IGasService {
     uint128 public immutable maxAssetPriceAge;
     uint128 public immutable maxSharePriceAge;
 
-    constructor(uint128 batchGasLimit_, uint128 messageGasLimit_) {
+    constructor(uint128 batchGasLimit_) {
         _batchGasLimit = batchGasLimit_;
-        _messageGasLimit = messageGasLimit_;
 
         scheduleUpgrade = BASE_COST + SMALL_COST;
         cancelUpgrade = BASE_COST + SMALL_COST;
--- a/src/common/GasService.sol
+++ b/src/common/GasService.sol
@@ -27,7 +27,7 @@ contract GasService is IGasService {
     uint128 public immutable notifyPricePoolPerShare;
     uint128 public immutable notifyPricePoolPerAsset;
     uint128 public immutable notifyShareMetadata;
-    uint128 public immutable notifyShareHook;
+    uint128 public immutable updateShareHook;
     uint128 public immutable initiateTransferShares;
     uint128 public immutable executeTransferShares;
     uint128 public immutable updateRestriction;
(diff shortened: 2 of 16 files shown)


script/FullDeployer.s.sol
```diff
@@ -60,8 +60,8 @@ contract FullDeployer is HubDeployer, ExtendedSpokeDeployer {
             centrifugeId: centrifugeId,
             root: IRoot(vm.envAddress("ROOT")),
             adminSafe: ISafe(vm.envAddress("ADMIN")),
-            messageGasLimit: uint128(vm.envUint("MESSAGE_COST")),
-            maxBatchSize: uint128(vm.envUint("MAX_BATCH_SIZE")),
+            messageGasLimit: uint128(vm.envUint("MESSAGE_GAS_LIMIT")),
+            batchGasLimit: uint128(vm.envUint("BATCH_GAS_LIMIT")),
```
lemunozm: @gpmayorga relevant part for your

src/common/GasService.sol
```diff
@@ -2,26 +2,84 @@
...
+        } else if (kind == MessageType.UpdateHoldingAmount) {
+            return BASE_COST + 220866; // hub.updateHoldingAmount
+        } else if (kind == MessageType.UpdateShares) {
+            return BASE_COST + 49968; // hub.updateShares
+        } else if (kind == MessageType.MaxAssetPriceAge) {
+            return BASE_COST + SMALL_COST;
+        } else if (kind == MessageType.MaxSharePriceAge) {
+            return BASE_COST + SMALL_COST;
+        } else {
+            return _messageGasLimit;
```
lemunozm: Actually, we no longer need this value, and we can just remove it

src/common/GasService.sol
```diff
@@ -2,26 +2,84 @@
...
+        } else if (kind == MessageType.CancelUpgrade) {
+            return BASE_COST + SMALL_COST;
+        } else if (kind == MessageType.RecoverTokens) {
+            return BASE_COST + SMALL_COST;
+        } else if (kind == MessageType.RegisterAsset) {
+            return BASE_COST + 34329; // hub.registerAsset()
+        } else if (kind == MessageType.Request) {
+            return BASE_COST + 86084; // hub.request(requestDeposit)
+        } else if (kind == MessageType.NotifyPool) {
+            return BASE_COST + 38190; // spoke.notifyPool()
```
lemunozm: Ideally, all these values should have their own getter. That way, an integrator can do:

```solidity
hub.notifyPool{gas: gasService.notifyPool()}(..)
```

Which is great from a UX experience

lemunozm: Done here: [0507541](https://github.com/centrifuge/protocol-v3/pull/495/commits/0507541f25312d11de0014ff1e9972e45c4c47b1), assuming only EVM systems, but for now I think it's ok.

src/common/GasService.sol
lemunozm: Done: [12ec8e0](https://github.com/centrifuge/protocol-v3/pull/495/commits/12ec8e045299fbc6c857b0247d250fe4854f0285)

src/common/GasService.sol
```diff
@@ -2,26 +2,130 @@
...
 contract GasService is IGasService {
-    uint128 internal immutable _maxBatchSize;
-    uint128 internal immutable _messageGasLimit;
+    using MessageLib for *;
 
-    constructor(uint128 maxBatchSize_, uint128 messageGasLimit) {
-        _maxBatchSize = maxBatchSize_;
-        _messageGasLimit = messageGasLimit;
+    /// @dev Takes into account Adapter + Gateway processing + some mismatch happened regarding the reference
+    uint128 public constant BASE_COST = 200_000;
```
lemunozm, line 14: BTW, processing a message from the `Gateway` & `MultiAdapter` having proofs takes around `50_000` (checking quorum, adapters, read the message...). So we have `150_000` of reminder gas for adapter stuff and any possible extra cost that some code path could take, depending on the input values.

src/common/GasService.sol
```diff
@@ -2,26 +2,132 @@
...
+        } else if (kind == MessageType.UpdateHoldingAmount) {
+            return updateHoldingAmount;
+        } else if (kind == MessageType.UpdateShares) {
+            return updateShares;
+        } else if (kind == MessageType.MaxAssetPriceAge) {
+            return maxAssetPriceAge;
+        } else if (kind == MessageType.MaxSharePriceAge) {
+            return maxSharePriceAge;
+        } else {
+            return 0; // Unreachable
```
hieronx: Maybe we return `type(uint128).max` here, as a way to error out (won't be possible to be paid)

src/common/Gateway.sol
```diff
@@ -127,7 +127,7 @@ contract Gateway is Auth, Recoverable, IGateway {
 
         emit PrepareMessage(centrifugeId, poolId, message);
 
-        uint128 gasLimit = gasService.gasLimit(centrifugeId, message) + extraGasLimit;
+        uint128 gasLimit = gasService.messageGasLimit(centrifugeId, message) + extraGasLimit;
```
hieronx, line 130: Why did we rename this?

hieronx: I'd rather minimize changes and stick to current name

src/common/GasService.sol
lemunozm: Makes sense 👍🏻

src/common/Gateway.sol
lemunozm: I did that because if not, we have `gasLimit` and `batchGasLimit`, which are confusing. That way, we have two clear names for two different things

lemunozm: If you still have a super strong opinion, I can revert this, but I would say the change is minimal (?)

hieronx: I get it, seems okay then

hieronx (approved):

src/common/GasService.sol
```diff
@@ -27,7 +27,7 @@ contract GasService is IGasService {
     uint128 public immutable notifyPricePoolPerShare;
     uint128 public immutable notifyPricePoolPerAsset;
     uint128 public immutable notifyShareMetadata;
-    uint128 public immutable notifyShareHook;
+    uint128 public immutable updateShareHook;
```
lemunozm, line 30: Thanks for this naming catch!

src/common/GasService.sol
```diff
@@ -18,9 +18,57 @@ contract GasService is IGasService {
...
+        initiateTransferShares = BASE_COST + 52195;
+        executeTransferShares = BASE_COST + 70267;
+        updateRestriction = BASE_COST + 35992;
+        updateContract = BASE_COST + SMALL_COST;
+        requestCallback = BASE_COST + 186947; // approve deposit case
+        updateVault = BASE_COST + 2770342; // deploy vault case
+        setRequestManager = BASE_COST + 30039;
+        updateBalanceSheetManager = BASE_COST + 35241;
+        updateHoldingAmount = BASE_COST + 220866;
+        updateShares = BASE_COST + 49968;
```
wischli: Curious where these custom values come from. Could you point me somewhere @lemunozm?

lemunozm: They are manually gathered when executing EndToEnd tests, just in the function called by the MessageDispatcher/MessageProcessor.

The idea is to add some code to `LocalAdapter` (used in EndToEnd tests) and write the measured time there into some file, later some python script can create the `GasService` contract file from those values as a template.

wischli: Thanks for the explanation! IIUC, this PR already provided the core functionality for the v3.1 feature of gas estimates per message id ([ref](https://www.notion.so/Update-GasService-to-have-estimates-per-message-ID-2002eac24e1780859764e9984294bd3e?source=copy_link)), right?

lemunozm: Yeah, exactly!

PR #497, opened 2025-07-07
Add `Holdings.updateIsLiability`

hieronx:


src/hub/Holdings.sol
```diff
@@ -89,6 +89,17 @@ contract Holdings is Auth, IHoldings {
         emit UpdateValuation(poolId, scId, assetId, valuation_);
     }
 
+    /// @inheritdoc IHoldings
+    function updateIsLiability(PoolId poolId, ShareClassId scId, AssetId assetId, bool isLiability_) external auth {
+        Holding storage holding_ = holding[poolId][scId][assetId];
+        require(address(holding_.valuation) != address(0), HoldingNotFound());
+        require(holding_.assetAmount == 0, HoldingNotZero());
```
mustermeiszer: NIT

should we maybe also check `holding_.assetValue == 0` ?

mustermeiszer (commented):
Looks good already

src/hub/Holdings.sol
hieronx: Isn't that guaranteed to be true if `assetAmount == 0` @mustermeiszer ? Or in what case wouldn't it be?

mustermeiszer: Not sure tbh. It isn't if increase was 

-5 at price of 1
 
And decrease was 
- 5 at price of 0.5

hieronx: Isn't that a temporary invalid state then in a way, since once you'd call `Holdings.update`, `assetAmountValue` would become 0?

hieronx: I guess the point is, you'd want to ensure that `Holdings.update` is called before `Holdings.updateIsLiability`

src/hub/Hub.sol
```diff
@@ -602,6 +602,16 @@ contract Hub is Multicall, Auth, Recoverable, IHub, IHubGatewayHandler, IHubGuar
...
 
+    /// @inheritdoc IHub
+    function updateHoldingIsLiability(PoolId poolId, ShareClassId scId, AssetId assetId, bool isLiability)
+        external
+        payable
+    {
+        _isManager(poolId);
+
+        holdings.updateIsLiability(poolId, scId, assetId, isLiability);
+    }
```
lemunozm, line 612: Maybe using `tongleHoldingIsLibility(PoolId poolId, ShareClassId scId, AssetId assetId)` or similar, we can avoid passing one argument to the holding, saving the extra 18 bytes.

Otherwise, maybe removing some entries in the `file()` method is the easier way....

lemunozm (dismissed):
LGTM

src/hub/Hub.sol
hieronx: I think having a toggle is quite easy to mess up for integrators, unfortunately.

> Otherwise, maybe removing some entries in the file() method is the easier way....

I went for this instead! `accounting` seems least likely to need migrations.

foundry.toml
```diff
@@ -98,9 +98,5 @@ quote_style = "double"
 number_underscore = "preserve"
 wrap_comments = true
 ignore = [
-  "test/*.sol",
-  #Check if https://github.com/foundry-rs/foundry/issues/9088 is fixed to remove the following:
```
hieronx, line 102: This is now fixed

src/common/Gateway.sol
```diff
@@ -159,7 +159,8 @@ contract Gateway is Auth, Recoverable, IGateway {
         if (transactionRefund != address(0)) {
             require(cost <= fuel, NotEnoughTransactionGas());
             fuel -= cost;
-        } else { // Subsidized pool payment
+        } else {
```
hieronx, line 162: Caused by the re-enabling of `fmt`

src/hub/Accounting.sol
```diff
@@ -125,7 +125,7 @@ contract Accounting is Auth, IAccounting {
     //----------------------------------------------------------------------------------------------
 
     /// @inheritdoc IAccounting
-    function accountValue(PoolId poolId, AccountId account) external view returns (bool /* isPositive */, uint128) {
+    function accountValue(PoolId poolId, AccountId account) external view returns (bool, /* isPositive */ uint128) {
```
hieronx, line 128: Caused by the re-enabling of `fmt`

src/hub/Hub.sol
```diff
@@ -79,7 +79,6 @@ contract Hub is Multicall, Auth, Recoverable, IHub, IHubGatewayHandler, IHubGuar
         if (what == "sender") sender = IHubMessageSender(data);
         else if (what == "holdings") holdings = IHoldings(data);
         else if (what == "hubHelpers") hubHelpers = IHubHelpers(data);
-        else if (what == "accounting") accounting = IAccounting(data);
```
hieronx, line 82: To keep within the contract size limit

mustermeiszer (approved):

foundry.toml
lemunozm: That's great!

lemunozm (approved):

PR #500, opened 2025-07-08
Test: Improve EndToEnd coverage & accurate gas values

lemunozm:
- Test under EndToEnd all supported messages in the MessageLib
- Use real values for all messages
- Split `UpdateVault` gas value in three values depending on the `UpdateVaultKind` message value

### DIFF
--- a/src/common/GasService.sol
+++ b/src/common/GasService.sol
@@ -71,8 +71,8 @@ contract GasService is IGasService {
         updateBalanceSheetManager = BASE_COST + 35241;
         updateHoldingAmount = BASE_COST + 220866;
         updateShares = BASE_COST + 49968;
-        maxAssetPriceAge = BASE_COST + SMALL_COST;
-        maxSharePriceAge = BASE_COST + SMALL_COST;
+        maxAssetPriceAge = BASE_COST + 27260;
+        maxSharePriceAge = BASE_COST + 26032;
     }
 
     /// @inheritdoc IGasService
--- a/src/common/GasService.sol
+++ b/src/common/GasService.sol
@@ -2,7 +2,7 @@
 pragma solidity 0.8.28;
 
 import {IGasService} from "src/common/interfaces/IGasService.sol";
-import {MessageLib, MessageType} from "src/common/libraries/MessageLib.sol";
+import {MessageLib, MessageType, VaultUpdateKind} from "src/common/libraries/MessageLib.sol";
 
 /// @title  GasService
 /// @notice This is a utility contract used to determine the execution gas limit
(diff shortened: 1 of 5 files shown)


src/common/GasService.sol
```diff
@@ -46,29 +46,31 @@ contract GasService is IGasService {
...
         initiateTransferShares = BASE_COST + 52195;
         executeTransferShares = BASE_COST + 70267;
         updateRestriction = BASE_COST + 35992;
-        updateContract = BASE_COST + SMALL_COST;
+        updateContract = BASE_COST + 53345;
         requestCallback = BASE_COST + 186947; // approve deposit case
-        updateVault = BASE_COST + 2770342; // deploy vault case
+        updateVaultDeployAndLink = BASE_COST + 2770342;
+        updateVaultLink = BASE_COST + 100567;
+        updateVaultUnlink = BASE_COST + 20814;
```
lemunozm, line 67: Note `DeployAndLink` is 130 times higher than `Unlink` so I think this optimization makes sense for this message

lemunozm:
Coverage improved by 2% 🚀, getting close to our target!

hieronx (approved):

test/integration/EndToEnd.t.sol
```diff
@@ -580,6 +582,44 @@ contract EndToEndUseCases is EndToEndFlows {
...
+
+        _setSpoke(sameChain);
+
+        vm.startPrank(ANY);
+        h.gateway.subsidizePool{value: DEFAULT_SUBSIDY}(PoolId.wrap(0));
+
+        s.gateway.subsidizePool{value: VALUE}(PoolId.wrap(0));
+
+        vm.startPrank(address(SAFE_ADMIN_A));
+        h.guardian.recoverTokens(s.centrifugeId, address(s.gateway), ETH_ADDRESS, 0, RECEIVER, VALUE);
```
wischli, line 618: Nit: Shouldn't we perform this check also for `s.guardian`?

lemunozm: Here, what is happening is that we're sending a `RecoverTokens` message through the same chain or the other chain based on `sameChain` config. So using the `h.guardian`, we're already testing both casuistics.

If using `s.guardian`, we would always be bypassing the message, because given how the test is set, the recovery is always done in the spoke side.

lemunozm: Note the line `h.gateway.subsidizePool...` is there not to be recovered, just to be able to have the funds for sending that message

wischli: Thanks for the explanation! Makes sense to me.

PR #504, opened 2025-07-09
IT: Test the entire refund cycle

lemunozm:
Added an IT to test the refund cycle. It consists of:

1. Each message is sent with a higher gas value than it (this always happens).
2. The adapter, if sending in subsidize mode, will refund the excess of gas to the pool escrow associated to the message (using the PoolId).
3. When a next message does not have enough subsidized funds, it will take them first from the pool escrow to subsidize the message.


test/integration/EndToEnd.t.sol
```diff
@@ -648,6 +644,40 @@ contract EndToEndUseCases is EndToEndFlows {
...
+        h.hub.notifySharePrice{value: GAS}(POOL_A, SC_1, s.centrifugeId);
+        h.hub.setSnapshotHook(POOL_A, h.snapshotHook);
+
+        // Each message will return half of the gas wasted
+        adapterBToA.setRefundedValue(h.gasService.updateShares() / 2);
+
+        // We just subsidize for two message
+        vm.startPrank(ANY);
+        s.gateway.subsidizePool{value: h.gasService.updateShares() * 2}(POOL_A);
+
```
hieronx: I think this would make the test easier to follow:
```suggestion
        // We just subsidize for two message
        vm.startPrank(ANY);
        s.gateway.subsidizePool{value: h.gasService.updateShares() * 2}(POOL_A);
        assertEq(address(s.balanceSheet.escrow(POOL_A)).balance, 0);
```

Potentially also asserting the subsidized funds in the gateway. So you can see the amount of subsidized funds available going down, and the amount in the escrow going up.

Similar to test style like this: https://github.com/sparkdotfi/spark-alm-controller/blob/dev/test/mainnet-fork/Centrifuge.t.sol#L221

lemunozm: Makes sense your proposal 👍🏻

lemunozm: 77c3d88f35cf34bd290e3ac47ce290967c830e2f

hieronx (approved):

PR #509, opened 2025-07-10
Swap deposit & issue order in `SyncManager`

hieronx:


lemunozm (commented):
LGTM!

Is this change out from the last testnet deployment, right?

lemunozm (approved):

PR #510, opened 2025-07-11
Deploy: adapters as part of the deployer pipeline

lemunozm:
Missing:
- [x] Add deployment tests for adapters (already missed before)
- [x] Configure the FullDeployer (see below comment)
- [x] Remove previous, no longer used, adapter script

image


script/FullDeployer.s.sol
```diff
@@ -78,21 +94,33 @@ contract FullDeployer is HubDeployer, ExtendedSpokeDeployer {
...
+            wormhole: WormholeInput({
+                shouldDeploy: true, // TODO
+                relayer: address(0) // TODO
+            }),
+            axelar: AxelarInput({
+                shouldDeploy: true, // TODO
+                gateway: address(0), // TODO
+                gasService: address(0) // TODO
+            })
+        });
```
lemunozm: This must be populated with the adapter's config

lemunozm:
I'm unsure if the current inheritance is the correct one.

Should instead inherit `FullDeployer` with a `FullWithAdaptersDeployer`? That way, from testing point of view we can just deploy the whole system without adapters at all 🤔.

EDIT: I think we can go as it is right now

gpmayorga (changes requested):
I modified the python script to remove all Adapters logic. Anvil check should be green in order to merge this PR IMO as otherwise the deployment is failing even if the unit tests pass.

gpmayorga:
Don't forget to consider the `"deploy": [true,false]` on the configurations, some networks do not have wormhole and/or do not have Axelar and cannot be deployed.
```"adapters": {
        "wormhole": {
            "wormholeId": "https://wormhole.com/docs/build/reference/chain-ids/",
            "relayer": "https://wormhole.com/docs/build/reference/contract-addresses",
            "deploy": false
        },
        "axelar": {
            "axelarId": "Chain Name",
            "gateway": "0x...",
            "gasService": "0x...",
            "deploy": false
        }
    },```

lemunozm:
NOTE: I've did a python revert here [55d4696](https://github.com/centrifuge/protocol-v3/pull/510/commits/55d4696faa72ac34e3f9d2d692c7ce6c639044cd) to let it as it was before this change from here: [13b0b1f](https://github.com/centrifuge/protocol-v3/pull/510/commits/13b0b1ff103b74f584073ade352a6f4c8eb0fbed). If not the python scripts does not "compile". IIUC, we don't need that parameter there as it is. Would like a double-check from @gpmayorga but not sure I could before deployment.

From this change the Anvil CI passes

hieronx (approved):

wischli (approved):

PR #514, opened 2025-07-14
Update mainnet configs

hieronx:


env/plume.json
```diff
@@ -2,17 +2,18 @@
...
-        "batchGasLimit": "25000000",
-        "connectsTo": []
+        "safeAdmin": null
     },
     "adapters": {
         "wormhole": {
             "wormholeId": "55",
-            "relayer": null,
-            "deploy": false
+            "relayer": "0xaBf89de706B583424328B54dD05a8fC986750Da8",
```
wischli, line 14: Just to double check: This relayer was created two weeks ago according to the block explorer ([ref](https://explorer.plume.org/address/0xaBf89de706B583424328B54dD05a8fC986750Da8?tab=index)). Is that expected?

wischli (dismissed):
AFAICT, all good!

env/plume.json
```diff
@@ -2,17 +2,18 @@
...
-        "centrifugeId": 7,
+        "centrifugeId": 4,
+        "connectsTo": [
+            "ethereum"
+        ],
         "root": null,
-        "safeAdmin": null,
-        "batchGasLimit": "25000000",
-        "connectsTo": []
+        "safeAdmin": null
```
lemunozm: Why is this null?

env/ethereum.json
```diff
@@ -19,7 +21,7 @@
         "axelar": {
             "axelarId": "Ethereum",
             "gateway": "0x4F4495243837681061C4743b74B3eEdf548D56A5",
-            "gasService": "0x2d5d7d31F671F86C782533cc367F14136aCD5838",
+            "gasService": "0x2d5d7d31F671F86C782533cc367F14109a082712",
```
lemunozm, line 24: Curious about this change only in the last digits (?), was wrong before?

env/plume.json
lemunozm, line 14: ✅

lemunozm (dismissed):

env/plume.json
lemunozm: Seems it's from the official web: https://wormhole.com/docs/products/reference/contract-addresses/#wormhole-relayer for plume

lemunozm (dismissed):

env/plume.json
hieronx: Yeah that is expected!

env/plume.json
hieronx: To be deployed, it's now set!

env/ethereum.json
hieronx: Yeah it seems it was wrong before. See https://axelarscan.io/resources/chains as source of truth

lemunozm (dismissed):

lemunozm (approved):

PR #519, opened 2025-07-15
Fix: protect ActionBatcher

lemunozm:


script/CommonDeployer.s.sol
```diff
@@ -37,18 +37,28 @@ struct CommonReport {
 }
 
 contract CommonActionBatcher {
-    // @dev lock to ensure we forbid calling engage/revoke methods after deployment
-    bool isLock;
```
lemunozm, line 41: Lock is no longer needed, because with `address deployer` we can already control the locking mechanism

script/CommonDeployer.s.sol
```diff
@@ -37,18 +37,28 @@ struct CommonReport {
...
+    }
+
+    modifier onlyDeployer() {
+        require(msg.sender == deployer, NotDeployer());
+        _;
+    }
+
+    function setDeployer(address newDeployer) public onlyDeployer {
+        deployer = newDeployer;
     }
```
lemunozm, line 55: This is needed for testing, when the deployer of the ActionBatcher can not be the sender

script/CommonDeployer.s.sol
```diff
@@ -37,18 +37,28 @@ struct CommonReport {
...
-    bool isLock;
+    error NotDeployer();
 
-    modifier unlocked() {
-        if (!isLock) _;
+    address deployer;
+
+    constructor() {
+        deployer = msg.sender;
+    }
```
lemunozm, line 46: Forced to do this instead of passing a parameter due: https://github.com/ethereum/solidity/issues/4199

test/spoke/BaseTest.sol
```diff
@@ -115,10 +107,6 @@ contract BaseTest is ExtendedSpokeDeployer, Test, ExtendedSpokeActionBatcher {
         testAdapters.push(adapter2);
         testAdapters.push(adapter3);
 
-        // wire contracts
-        _wire(OTHER_CHAIN_ID, adapter1);
```
lemunozm, line 119: Was not need it because the adapters were already set below

script/CommonDeployer.s.sol
```diff
@@ -37,18 +37,28 @@ struct CommonReport {
 }
 
 contract CommonActionBatcher {
-    // @dev lock to ensure we forbid calling engage/revoke methods after deployment
-    bool isLock;
+    error NotDeployer();
 
-    modifier unlocked() {
-        if (!isLock) _;
+    address deployer;
```
hieronx, line 42: This can be `address immutable deployer` right?

snapshots/VaultRouter.json
```diff
@@ -1,7 +1,7 @@
 {
-  "claimDeposit": "1153430",
+  "claimDeposit": "1104497",
```
lemunozm, line 2: This comes from remove the `wire()` method in tests (although I don't fully understand why 🤔)

script/CommonDeployer.s.sol
```diff
@@ -37,18 +37,28 @@ struct CommonReport {
...
+        deployer = msg.sender;
+    }
+
+    modifier onlyDeployer() {
+        require(msg.sender == deployer, NotDeployer());
+        _;
+    }
+
+    function setDeployer(address newDeployer) public onlyDeployer {
+        deployer = newDeployer;
```
hieronx, line 54: Can we add `require(deployer != address(0), AlreadyLocked())`?

To make it impossible for deployer to ever unlock again.

script/CommonDeployer.s.sol
lemunozm: Not if we want to support the testing 😢  https://github.com/centrifuge/protocol-v3/pull/519#discussion_r2207001950

hieronx: Right yeah, never mind

script/CommonDeployer.s.sol
hieronx: This is fine, also for testing, right @lemunozm ?

lemunozm: This method is just for testing, but we need to keep this here. I was not able to find a workaround (https://github.com/centrifuge/protocol-v3/pull/519#discussion_r2207002736). Note the `lock()` method still exist and is what is called from the `FullDeployer.run()`

lemunozm: Note that the method is already `onlyDeployer`

hieronx: But that means our deployer key can still call this as well.

What is the problem for testing with not allowing it to be re-set, after it was set to `address(0)`?

script/CommonDeployer.s.sol
```diff
@@ -37,18 +37,28 @@ struct CommonReport {
...
-    // @dev lock to ensure we forbid calling engage/revoke methods after deployment
-    bool isLock;
+    error NotDeployer();
 
-    modifier unlocked() {
-        if (!isLock) _;
+    address deployer;
+
+    constructor() {
+        deployer = msg.sender;
```
lemunozm, line 45: FYI. Anvil CI fails because `msg.sender` is create3. Need to think how to fix this then...

script/CommonDeployer.s.sol
lemunozm: Once you deploy everything, we call to `lock()` which sets the deployer to `address(0)`. So nobody, even the originator `deployer` can call `setDeployer` again. The require you suggest will never be reached

snapshots/VaultRouter.json
lemunozm: Want to double-check this, in case we have some issue in the Gateway/Multiadapters. 

Seems somehow that adding adapters and overwritting them later, increase the computation of parsing messages later 🤔

hieronx: Seems more likely to be an issue with the way we are computing gas for this. Would ignore it.

hieronx (approved):

snapshots/VaultRouter.json
lemunozm: Found the issue, nothing to worry about IIUC. This code was computed previously: https://github.com/centrifuge/protocol-v3/blob/main/src/common/adapters/MultiAdapter.sol#L122-L126

But everything works as expected, we're just delete an already deleted storage

PR #521, opened 2025-07-15
Fix Alchemy IDs, disable Plume, add mainnet deployments

hieronx:


lemunozm (approved):
Not blocker: some json indentation has changed from 4 to 2

PR #522, opened 2025-07-15
Clean: remove legacy vault files

lemunozm:
IMO, we can remove these files from the (not yet created?) `v3` branch because they were not deployed


hieronx (approved):

wischli (approved):
Great pre release cleanup!

PR #526, opened 2025-07-16
Ref: Use relative imports & add integration tests for externals

lemunozm:
Base tests for integrators. Come in two flavours:
- `CentrifugeIntegrationTest`: The most generic way to start doing tests
- `CentrifugeIntegrationTestWithUtils`: An extension of the above with some customized utilities. If some integrator requires a specialized configuration, they should use `CentrifugeIntegrationTest` instead; if not, it can take advantage of the offered utilities (we will add more over time).

#### Extra feature:
Label the address by default automatically in all integration tests.

### DIFF
--- a/script/utils/JsonRegistry.s.sol
+++ b/script/utils/JsonRegistry.s.sol
@@ -7,6 +7,8 @@ import {console} from "forge-std/console.sol";
 contract JsonRegistry is Script {
     string deploymentOutput;
     uint256 registeredContracts = 0;
+    bool shouldLaberAddresses;
+    string addressLabelPrefix;
 
     function register(string memory name, address target) public {
         deploymentOutput = (registeredContracts == 0)
@@ -14,6 +16,15 @@ contract JsonRegistry is Script {
             : string(abi.encodePacked(deploymentOutput, ',\n    "', name, '": "', vm.toString(target), '"'));
 
         registeredContracts += 1;
+
+        if (shouldLaberAddresses) {
+            vm.label(address(target), string(abi.encodePacked(addressLabelPrefix, name)));
+        }
+    }
+
+    function labelAddresses(string memory prefix) public {
+        shouldLaberAddresses = true;
+        addressLabelPrefix = prefix;
     }
 
     function startDeploymentOutput() public {
--- a/test/integration/Integration.t.sol
+++ b/test/integration/Integration.t.sol
@@ -63,7 +63,7 @@ contract CentrifugeIntegrationTestWithUtils is CentrifugeIntegrationTest {
     ERC20 usdc;
     AssetId usdcId;
 
-    function setUp() public override {
+    function setUp() public virtual override {
         super.setUp();
 
         POOL_A = hubRegistry.poolId(LOCAL_CENTRIFUGE_ID, 1);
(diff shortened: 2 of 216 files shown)


lemunozm:
I think this could make sense to merge this branch into a `v3` branch, to support integrators tagged on v3

hieronx:
> I think this could make sense to merge this branch into a v3 branch, to support integrators tagged on v3

Yeah makes sense! Let's merge into v3 first then ,and then merge from v3 into main? So we don't merge the other commits on main into v3

lemunozm:
Yes, we should merge stuff like this into two stages. If v3 diverge somehow (i.e. readme stuff or so), then we should merge the PR in both branches individually

wischli:
> Yes, we should merge stuff like this into two stages. If v3 diverge somehow (i.e. readme stuff or so), then we should merge the PR in both branches individually

I have just pushed a branch `release-v3.0.0` for this purpose.

hieronx:
> I have just pushed a branch release-v3.0.0 for this purpose.

Had already created `v3` for the same purpose. Happy to move to yours though. Although the tag is on `v3`.

script/utils/JsonRegistry.s.sol
```diff
@@ -7,13 +7,24 @@ import {console} from "forge-std/console.sol";
...
         registeredContracts += 1;
+
+        if (shouldLaberAddresses) {
+            vm.label(address(target), string(abi.encodePacked(addressLabelPrefix, name)));
+        }
+    }
+
+    function labelAddresses(string memory prefix) public {
+        shouldLaberAddresses = true;
+        addressLabelPrefix = prefix;
```
wischli, line 27: Love it! So elegant and concise

test/integration/Integration.t.sol
```diff
@@ -0,0 +1,78 @@
...
+    ShareClassId SC_1;
+
+    function setUp() public override {
+        super.setUp();
+
+        POOL_A = hubRegistry.poolId(LOCAL_CENTRIFUGE_ID, 1);
+        SC_1 = shareClassManager.previewNextShareClassId(POOL_A);
+    }
+
+    function createPool() public {
```
wischli: Shouldn't we also expose the pool configuration including all necessary messages to the spoke side?

wischli:
> > I have just pushed a branch release-v3.0.0 for this purpose.
> 
> Had already created `v3` for the same purpose. Happy to move to yours though. Although the tag is on `v3`.

Sorry, I was looking for `release-*` branches. I'd prefer using semantic versioning at least for the release branches and having the prefix as this appears to be widely adopted in open-source development to the best of my knowledge.

test/integration/Integration.t.sol
lemunozm: Maybe... I don't know yet. It all depends on how the integrator wants (or needs) to configure the pool for its needs. Which accounts? holdings or liability? which hooks or managers?

lemunozm:
Add `registerUSDC()` here: [d6d0fbd](https://github.com/centrifuge/protocol-v3/pull/526/commits/d6d0fbd25e6e5ee39fde40638c9245e0ffb36fa7)

lemunozm:
I'm gonna cherry-pickicking the Jeroen commits.

hieronx (approved):
Can you just update the title?

PR #525, opened 2025-07-16
Ref: move adapters its own module

lemunozm:
Already discussed with Jeroen about moving adapters to its own module, allowing integrators to build their own (similar to hooks, managers, vaults...)

PR #528, opened 2025-07-16
Ref: guardian agnostic from the kind of adapters

lemunozm:
After https://github.com/centrifuge/protocol-v3/pull/525, the `Guardian` depends of the `adapters` module, which is not correct. 

This PR changes a bit how the `Guardian` wires, in order to be adapter agnostic, and also, to support new adapters in the future without requiring to migrate the `Guardian` each time.

#### Extra addition:
Increase a bit the adapter coverage


src/common/Guardian.sol
```diff
@@ -119,28 +116,13 @@ contract Guardian is IGuardian {
...
-    /// @inheritdoc IGuardian
-    function wireAxelarAdapter(
-        IAxelarAdapter localAdapter,
-        uint16 centrifugeId,
-        string calldata axelarId,
-        string calldata adapter
-    ) external onlySafe {
-        localAdapter.file("sources", axelarId, centrifugeId, adapter);
-        localAdapter.file("destinations", centrifugeId, axelarId, adapter);
+    function wireAdapter(IConfigurableAdapter adapter, bytes memory encodedParams) external onlySafe {
```
hieronx, line 124: This will be much harder to call manually. Since we need to get the encoding right, and the ABI decoding on the Gnosis UI won't work.

hieronx: I think we should keep it as is.

lemunozm: And how can we fix this, then?

Suppose tomorrow we want to add a new adapter, with its own parameters, should I migrate all Guardians in all networks? We need a way to do this generically. 

Note that these calls are only for admin management. With adapters-per-pool, the pool admin should call directly to the `adapter.file()` with the static signature

hieronx: > Suppose tomorrow we want to add a new adapter, with its own parameters, should I migrate all Guardians in all networks? We need a way to do this generically.

I was thinking this is what's needed yes.

But definitely not ideal.

Let me think about it a bit more.

src/adapters/AxelarAdapter.sol
```diff
@@ -41,22 +41,27 @@ contract AxelarAdapter is Auth, IAxelarAdapter {
...
-        auth
-    {
+    function file(bytes32 what, uint16 centrifugeId, string memory axelarId, string memory destination) public auth {
         if (what == "destinations") destinations[centrifugeId] = AxelarDestination(axelarId, destination);
         else revert FileUnrecognizedParam();
         emit File(what, centrifugeId, axelarId, destination);
     }
 
+    /// @inheritdoc IConfigurableAdapter
+    function wire(bytes memory encodedParams) external auth {
```
wischli, line 58: Since this addition introduces a breaking change, it appears to me we need to update the Guardian and all adapters on each network once we deploy v3.1, right? Otherwise, v3 adapters don't work with the v3.1 Guardian wiring protocol.

src/adapters/AxelarAdapter.sol
```diff
@@ -41,22 +41,27 @@ contract AxelarAdapter is Auth, IAxelarAdapter {
...
+        (uint16 centrifugeId, string memory axelarId, string memory adapter) =
+            abi.decode(encodedParams, (uint16, string, string));
```
wischli, line 60: Did you put a thought into ensuring the provided `string axelarId` actually is in the correct format on the protocol level?

src/common/Guardian.sol
wischli: I have the same reluctance, especially because verifying the validity of the data for all the safe calls will be difficult going forward with this. For my limited PoV, this feels like a premature abstraction that outweighs the decreased UX as new types of adapters aren't yet in the pipeline. But definitely no strong opinion!

src/common/interfaces/IGuardian.sol
```diff
@@ -77,32 +74,15 @@ interface IGuardian {
...
-    /// @param centrifugeId The remote chain's chain ID
-    /// @param axelarId The remote chain's Axelar ID
-    /// @param adapter The remote chain's Axelar adapter address
-    function wireAxelarAdapter(
-        IAxelarAdapter localAdapter,
-        uint16 centrifugeId,
-        string calldata axelarId,
-        string calldata adapter
-    ) external;
+    function setAdapters(uint16 centrifugeId, IAdapter[] calldata adapters) external;
```
wischli, line 82: Definitely in favor of this renaming!

wischli (commented):
Code LGTM but I feel like the decreased UX outweighs the added scalability

src/common/Guardian.sol
lemunozm: The same way we have to use each time we need to use a multicall, right? I mean, we need to encode params of the calls. Feel like this issue happens everywhere.

BTW. Note that also conceptually, right now we have the `common` module pointing to the `adapters` module, which should never happen because `adapters` is an integration module.

src/adapters/AxelarAdapter.sol
lemunozm: Correct, agree that's not ideal

hieronx: I just realized two things:
1. Swapping adapters wil be quite hard ,as any in-flight messages will need to be recovered.
2. We anyways will need to swap adapters as `entrypoint` is `immutable`, and we will be swapping `entrypoint` aka `multiAdapter` for the per-pool adapter setup

lemunozm: Correct on both 😅.

Hopefully, pausing the gateway for some time (30 min or so) before the update help with 1, and only very few messages need to be recovered. Also happy that our recovery method is already in the Gateway in the failed message storage, instead of fully rely on the adapters.

Maybe we should modify the immutable endpoint to a mutable one, for future upgrades after v3.1, WDYT?

src/adapters/AxelarAdapter.sol
lemunozm: What do you mean by " on the protocol level?" ?

lemunozm:
No longer need it after #566

PR #530, opened 2025-07-16
Tests: increase coverage in MessageLib and MessageProofLib

lemunozm:


lemunozm:
Not sure why the lines still appear in the coverage report when they shouldn't 🤔, will double-check that

wischli (dismissed):

lemunozm:
Fixed, although I don't know why the % has not increased. At least the lines no longer appear are not covered

wischli (approved):
Thanks for continuously extending our test coverage!

PR #531, opened 2025-07-16
Allow protocol-v3 to be used as dependency of other projects

lemunozm:
### The problem:

Right now, we can not be used as a dependency. To determine where a struct/enum is, solidity uses the path from the import, so different paths to the same file are actually different things.

A dependency can import i.e. `PoolId` as: `"centrifuge-v3/src/common/types/PoolId.sol"`, but later our own repo, internally also import `PoolId` as `"src/common/types/PoolId.sol"`. It means different paths, and consequently different types.

### Solution:
1. Use relative paths, like `"./types/PoolId.sol"`, nevertheless, this means, AFAICT, a manual refactor over 200 files. Unless we create or find a script to do that. Also, future refactors could be a pain if using relative paths. Uniswap uses this BTW

2. Use absolute paths prefixed with a name that our customer can use in the remapping. Our customer needs to do: `centrifuge-v3/=lib/protocol-v3/src`, that way, both paths are the same, not matter from where you import the file.

I've chosen 2, because it is the fastest way to do this, and both ways are supposed to be ok. (chatgtp recommends option 2 as more stable, BTW, but not sure how reliable it is). I'm not against 1, if we found a way to do it and maintain the repo structure

### DIFF
--- a/remappings.txt
+++ b/remappings.txt
@@ -2,3 +2,5 @@ forge-std/=lib/forge-std/src/
 @chimera/=lib/chimera/src/
 createx-forge/=lib/createx-forge/
 centrifuge-v3/src/=src/
+centrifuge-v3/test/=test/
+centrifuge-v3/script/=script/
--- a/remappings.txt
+++ b/remappings.txt
@@ -1,3 +1,4 @@
 forge-std/=lib/forge-std/src/
 @chimera/=lib/chimera/src/
 createx-forge/=lib/createx-forge/
+centrifuge-v3/src/=src/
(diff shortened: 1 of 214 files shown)


hieronx:
> centrifuge-v3/=lib/protocol-v3/src

FYI this doesn't work because integrators need to use `test/integration/...`, so that is inaccessible if the remapping points to `src`

lemunozm:
Closed in favor of: https://github.com/centrifuge/protocol-v3/pull/532

lemunozm:
NOTE: we should cherry pick the additions to integration tests
- [ ] Done

PR #533, opened 2025-07-21
Add new chain

hieronx:


script/FullDeployer.s.sol
```diff
@@ -56,6 +71,43 @@ contract FullDeployer is HubDeployer, ExtendedSpokeDeployer, AdaptersDeployer {
...
+        require(address(syncDepositVaultFactory) == 0x3568184784E8ACCaacF51A7F710a3DE0144E4f29);
+        require(address(vaultRouter) == 0xdbCcee499563D4AC2D3788DeD3acb14FB92B175D);
+        require(address(freezeOnlyHook) == 0xBb7ABFB0E62dfb36e02CeeCDA59ADFD71f50c88e);
+        require(address(fullRestrictionsHook) == 0xa2C98F0F76Da0C97039688CA6280d082942d0b48);
+        require(address(freelyTransferableHook) == 0xbce8C1f411484C28a64f7A6e3fA63C56b6f3dDDE);
+        require(address(redemptionRestrictionsHook) == 0xf0C36EFD5F6465D18B9679ee1407a3FC9A2955dD);
+        require(address(onOfframpManagerFactory) == 0xcb084F79e8AE54e1373130F4F7119214FCe972a9);
+        require(address(merkleProofManagerFactory) == 0xaBd3cDc17C15a9E7771876cE24aB10A8E722781d);
+        require(address(vaultDecoder) == 0x72B188c37bD8Eb002d0D9c63CCd77F2Ff71d272e);
+        require(address(circleDecoder) == 0x6fce63E718fED6E20bAa8179e313C24cbF2EDa24);
```
lemunozm, line 108: Curious, this works as automatic verification?

lemunozm (approved):

script/FullDeployer.s.sol
hieronx: It is just a sanity check, to make sure the deterministic deployments fully work. It works because the deployment script always simulates the entire deployment, so the deployment will be stopped at the start if any of these requires are false.

lemunozm: Makes sense! Good check

script/utils/verify.sh
```diff
@@ -0,0 +1,56 @@
...
+fi
+
+contract_address=$1
+
+echo "vault: $contract_address"
+if ! cast call $contract_address 'share()(address)' --rpc-url $RPC_URL &> /dev/null; then
+    echo "Error: Must pass a vault address."
+    exit 1
+fi
+poolId=$(cast call $contract_address 'poolId()(uint64)' --rpc-url $RPC_URL | awk '{print $1}')
```
lemunozm, line 39: For reference (and for me), could you explain this syntax `poolId()(uint64)`? 🙏🏻

lemunozm (approved):
LGTM!

script/utils/verify.sh
hieronx: It is a view function with no parameters, and with `uint64` return type

env/bnb-smart-chain.json
```diff
@@ -0,0 +1,69 @@
+{
+  "network": {
+    "chainId": 56,
+    "environment": "mainnet",
+    "centrifugeId": 6,
+    "connectsTo": [
+      "ethereum"
```
mustermeiszer: @gpmayorga and @hieronx shouldn't his be all others?

script/utils/verify.sh
```diff
@@ -0,0 +1,56 @@
...
+    echo "This script verifies the vault contract and its associated tranche and restriction manager contracts."
+    echo
+    echo "Usage: $0 contract_address"
+    echo
+    echo "Arguments:"
+    echo "  contract_address      The address of the vault to verify"
+    echo
+    echo "Required Environment Variables:"
+    echo "  RPC_URL               The RPC URL"
+    echo "  ETHERSCAN_KEY         The Etherscan API key"
```
wischli: Nit: Can we change this to `ETHERSCAN_API_KEY` which we already use in our other scripts?

hieronx: Good point! Fixed

wischli (approved):
Thanks for the fixes!

PR #534, opened 2025-07-22
Refactor hooks

hieronx:


src/hooks/FullRestrictions.sol
```diff
@@ -1,169 +1,35 @@
...
-    function supportsInterface(bytes4 interfaceId) external pure override returns (bool) {
-        return interfaceId == type(ITransferHook).interfaceId || interfaceId == type(IERC165).interfaceId;
+        if (isSourceOrTargetFrozen(from, to, hookData)) return false;
+
+        if (isDepositRequest(from, to)) return isTargetMember(to, hookData);
+        if (isDepositFulfillment(from, to)) return true;
+        if (isDepositClaim(from, to)) return isTargetMember(to, hookData);
+        if (isRedeemRequest(from, to)) return isSourceMember(from, hookData);
+        if (isRedeemFulfillment(from, to)) return true;
+        if (isRedeemClaim(from, to)) return isSourceMember(from, hookData);
```
wischli: This line should be the root for the broken test. AFAICT, in v3 we don't check if the source is actually a member, i.e. with the new API we do
```diff
-        if (isRedeemClaim(from, to)) return isSourceMember(from, hookData);
+        if (isRedeemClaim(from, to)) return true;
````

because we only check the freeze bit on the `fromHookData`: https://github.com/centrifuge/protocol-v3/blob/6f60024a3d49d83a2efc1611756e6ab229c333fc/src/hooks/FullRestrictions.sol#L72-L93

**Review by wischli — COMMENTED** [review]
Great improvement! I proposed an enum-based variant in #537 which can happily be dropped for this one.

File: script/HooksDeployer.s.sol [hunk 2]
```diff
@@ -61,28 +61,34 @@ contract HooksDeployer is SpokeDeployer {
         freezeOnlyHook = FreezeOnly(
             create3(
                 generateSalt("freezeOnlyHook"),
-                abi.encodePacked(type(FreezeOnly).creationCode, abi.encode(address(root), batcher))
+                abi.encodePacked(type(FreezeOnly).creationCode, abi.encode(address(root), address(spoke), batcher))
```
lemunozm: IIRC, don't we want to avoid the spoke dependencies and instead use balanceSheet as source of truth?

src/hooks/BaseHook.sol
```diff
@@ -0,0 +1,184 @@
...
+
+    /// @inheritdoc ITransferHook
+    function onERC20AuthTransfer(
+        address, /* sender */
+        address, /* from */
+        address, /* to */
+        uint256, /* value */
+        HookData calldata /* hookData */
+    ) external pure virtual returns (bytes4) {
+        return ITransferHook.onERC20AuthTransfer.selector;
```
lemunozm: Curious, should we still keep the return value? IIUC, this is only for [compatibility with V2](https://github.com/centrifuge/protocol-v3/blob/main/src/common/interfaces/ITransferHook.sol#L45). Maybe this can also be simplified

lemunozm (commented):
For sure much much more simplified!

lemunozm:
Could you modify the `generateSalt` method of the hooks as `"*-2"`?

src/hooks/BaseHook.sol
hieronx: Yeah just for compatibility

lemunozm (dismissed):
LGTM!

Last double-check: https://github.com/centrifuge/protocol-v3/pull/534#discussion_r2224426473

script/HooksDeployer.s.sol
lemunozm: @hieronx did you think about this?

hieronx: It would just be too expensive to get the spoke from the balance sheet every call.

Hooks are not too difficult to migrate since they are stateless

But I think we should just not update these hooks until v3.1, and then spoke should not change anymore either

lemunozm: Makes sense!

lemunozm (approved):
LGTM! Love it!

src/common/interfaces/ITransferHook.sol
```diff
@@ -22,12 +22,15 @@ address constant ESCROW_HOOK_ID = address(uint160(0x1CF60));
 /// @notice Hook interface to customize share token behaviour
 /// @dev    To detect specific system actions:
 ///           Deposit request:                  address(0)      -> address(user)
-///           Deposit request fulfillment:       address(0)      -> ESCROW_HOOK_ID
-///           Deposit claim:                    ESCROW_HOOK_ID  -> address(user)
+///           Deposit request fulfillment:       address(0)      -> Endorsed
+///           Deposit or cancel redeem claim:   Endorsed        -> address(user)
```
lemunozm, line 26: These are no longer endorsed I think

src/hooks/BaseTransferHook.sol
```diff
@@ -0,0 +1,194 @@
...
+    function isDepositFulfillment(address from, address to) public view returns (bool) {
+        return from == address(0) && to == depositTarget;
+    }
+
+    function isDepositClaim(address from, address to) public view returns (bool) {
+        return from == depositTarget && to != address(0);
+    }
+
+    function isRedeemRequest(address, address to) public pure returns (bool) {
+        return to == ESCROW_HOOK_ID;
```
wischli, line 93: Nit: Shouldn't we also check that `from != address(0)` based on the `ITransferHook` description just to be sync between description and implementation?

src/hooks/FullRestrictions.sol
```diff
@@ -1,169 +1,41 @@
...
-    function supportsInterface(bytes4 interfaceId) external pure override returns (bool) {
-        return interfaceId == type(ITransferHook).interfaceId || interfaceId == type(IERC165).interfaceId;
+        if (isSourceOrTargetFrozen(from, to, hookData)) return false;
+
+        if (isDepositRequest(from, to)) return isTargetMember(to, hookData);
+        if (isDepositFulfillment(from, to)) return true;
+        if (isDepositClaim(from, to)) return isTargetMember(to, hookData);
+        if (isRedeemRequest(from, to)) return isSourceMember(from, hookData);
+        if (isRedeemFulfillment(from, to)) return true;
+        if (isRedeemClaim(from, to)) return true;
```
wischli, line 35: Just to be sure: We always allow offramping shares (claiming redemption) but onramping (claiming deposit) requires membership as reflected here right?

src/hooks/RedemptionRestrictions.sol
```diff
@@ -1,171 +1,36 @@
...
-            return true;
-        }
-
-        if (to == ESCROW_HOOK_ID && fromHookData >> 64 < block.timestamp) {
-            // Destination is escrow, so it's a redemption request, and the user is not a member
-            return false;
-        }
+        if (isSourceOrTargetFrozen(from, to, hookData)) return false;
+        if (isRedeemRequest(from, to)) return isSourceMember(from, hookData);
+        if (isRedeemClaim(from, to)) return true;
```
wischli, line 31: Nit: could remove that and extend comment below to slightly improve gas

src/common/interfaces/ITransferHook.sol
hieronx: They still are, just not the way they are checked in `BaseTransferHook`. I'm not sure how to best document this here, seeing as it's not necessarily always `asyncRequestManager`/`globalEscrow`

wischli: AFAIU `Endorsed` here refers to either the `GlobalEscrow` (deposit request fulfillment, deposit claim) or `AsyncRequestManager` (redeem request fulfilment). Is that assessment correct @hieronx ?

wischli: Nit: This is definitely more readable than before but without the `BaseTransferHook` implementation, it's still quite opaque, at least to me. Can we make this even clearer or is the intention to keep it this generic? Asking because if setting up a hook with `depositTarget == redeemSource` (based on unifying `Endorsed`), both conditions `isDepositClaim(from, to)` and `isRedeemFulfillment(from, to)` would be true.

wischli (commented):
Overall looks great! Got a question regarding a potential stricter contructor requirement.

src/hooks/BaseTransferHook.sol
hieronx: It is a bit more gas efficient this way. Would keep it as is.

src/hooks/FullRestrictions.sol
hieronx: That is right!

hieronx: Not always btw, freezing would still block claiming a redemption.

src/common/interfaces/ITransferHook.sol
hieronx: How do you think we can make this even clearer? Would be happy to but I don't see a good solution for it.

src/hooks/BaseTransferHook.sol
```diff
@@ -0,0 +1,194 @@
...
+    constructor(
+        address root_,
+        address redeemSource_,
+        address depositTarget_,
+        address crosschainSource_,
+        address deployer
+    ) Auth(deployer) {
+        root = IRoot(root_);
+        redeemSource = redeemSource_;
+        depositTarget = depositTarget_;
```
wischli, line 45: Somehow my review comment disappeared: Wondering if we should restrict `redeemSource != depositTarget` based on my [above comment](https://github.com/centrifuge/protocol-v3/pull/534#discussion_r2239265356) of opening up two conditions if they match which doesn't seem to be logically desired.

hieronx: Yeah makes sense, will add

PR #535, opened 2025-07-22
Fix: gas service estimation for addPool

lemunozm:
The gas value of addPool was taken from an EndToEnd in the same chain where the Escrow creation is not taken into account. The correct maximum estimation should be taken from a test using different chains

### DIFF
--- a/script/CommonDeployer.s.sol
+++ b/script/CommonDeployer.s.sol
@@ -161,7 +161,7 @@ abstract contract CommonDeployer is Script, JsonRegistry, CreateXScript {
 
         gasService = GasService(
             create3(
-                generateSalt("gasService"),
+                generateSalt("gasService-2"),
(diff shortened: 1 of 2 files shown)


lemunozm:
Right now we're using a `VERSION` value for all contracts to compute the create3 address. Nevertheless, if we only want to redeploy one, we should either:
1. configure a version per contract in the env/json.
2. hardcode the new version here: https://github.com/centrifuge/protocol-v3/blob/main/script/CommonDeployer.s.sol#L164
  - In this case, this would be: `"gasService-2"` or so.
  
Do you agree in following option 2? which is simpler? (for me, it makes sense to hardcode it)
cc @hieronx @wischli

hieronx:
Yeah I would agree option 2 makes sense.

hieronx:
Yeah I would agree option 2 makes sense.

hieronx (approved):

hieronx (approved):

PR #536, opened 2025-07-22
Import changes from release-v3.0.0

hieronx:


wischli (dismissed):

wischli:
I think in the future we should merge from a `release` branch fork such that we can rebase the source branch to the target one which we don't want here as `main` includes changes that shouldn't be included in `release-v3.0.0`. I almost pressed that button 😅

lemunozm:
Yeah, also not sure if Github allow to open two PRs from the same origin & destination branches, in case tomorrow we need to merge again `release-v3.0.0` into `main`

lemunozm (approved):

PR #537, opened 2025-07-22
Alternative to refactored hooks

wischli:
Enum based alternative to the great hook refactoring proposed in #534 just for comparison as it reduces the lines of code slightly. For me, it provides better readability but I have no strong opinion! Feel free to close this PR and go forward with the first proposal.

The main difference is that all types of transfers are represented by an enum and the transfer type determined by a new call `checkTransferPolicy` which is implemented by all hooks instead of the custom `checkERC20Transfer` implementations. The `BaseHook` then always checks first if source or member are frozen as this check is performed for all types of hooks and then executes the corresponding `checkTransferPolicy`.

NOTE: The broken test is inherited from #534


src/hooks/FullRestrictions.sol
```diff
@@ -12,24 +12,32 @@ import {BaseHook} from "src/hooks/BaseHook.sol";
...
+                || transferType == TransferType.LocalTransfer
+        ) {
+            return isTargetMember(to, hookData);
+        }
 
-        // Else, it's a transfer
-        return isTargetMember(to, hookData);
+        if (transferType == TransferType.RedeemRequest || transferType == TransferType.RedeemClaim) {
+            return isSourceMember(from, hookData);
+        }
```
lemunozm, line 38: Just syntactically and visually, I would prefer the line-by-line case we had (independently of using enum or not)

wischli:
Superseded by #547

PR #538, opened 2025-07-23
Test: full coverage for spoke

lemunozm:
Add intensive UTs for all possible cases (and some combinations) in Spoke.

Also added some "fixes/improvements" in Spoke. Commented below.

### Next steps in other PRs:
- Use D18 for prices instead of plain `uint128`. Note the abi is the same so not a breaking change.
- Modify storages to avoid the internal struct mapping we have


src/spoke/Spoke.sol
```diff
@@ -205,11 +205,14 @@ contract Spoke is Auth, Recoverable, ReentrancyProtection, ISpoke, ISpokeGateway
...
+    function setRequestManager(PoolId poolId, ShareClassId scId, AssetId assetId, IRequestManager manager)
+        public
+        auth
+    {
         ShareClassDetails storage shareClass = _shareClass(poolId, scId);
         require(shareClass.asset[assetId].numVaults == 0, MoreThanZeroLinkedVaults());
-        shareClass.asset[assetId].manager = IRequestManager(manager);
-        emit SetRequestManager(poolId, scId, assetId, IRequestManager(manager));
+        shareClass.asset[assetId].manager = manager;
+        emit SetRequestManager(poolId, scId, assetId, manager);
```
lemunozm, line 215: Ref: `address` by `IRequestManager`. 

Note the ABI is the same, so we don't need to redeploy `MessageDispatcher` nor `MessageProcessor`

src/spoke/Spoke.sol
```diff
@@ -257,14 +260,16 @@ contract Spoke is Auth, Recoverable, ReentrancyProtection, ISpoke, ISpokeGateway
...
     function updatePricePoolPerShare(PoolId poolId, ShareClassId scId, uint128 price, uint64 computedAt) public auth {
         ShareClassDetails storage shareClass = _shareClass(poolId, scId);
+        Price storage poolPerShare = shareClass.pricePoolPerShare;
         require(computedAt >= shareClass.pricePoolPerShare.computedAt, CannotSetOlderPrice());
 
-        // Disable expiration of the price
-        if (shareClass.pricePoolPerShare.computedAt == 0) {
-            shareClass.pricePoolPerShare.maxAge = type(uint64).max;
+        // Disable expiration of the price if never initialized
+        if (poolPerShare.computedAt == 0 && poolPerShare.maxAge == 0) {
```
lemunozm, line 267: FIx issue: https://kflabs.slack.com/archives/C07PG2EUR9C/p1753179660610469

src/spoke/Spoke.sol
```diff
@@ -281,8 +286,8 @@ contract Spoke is Auth, Recoverable, ReentrancyProtection, ISpoke, ISpokeGateway
         Price storage poolPerAsset = shareClass.pricePoolPerAsset[asset][tokenId];
         require(computedAt >= poolPerAsset.computedAt, CannotSetOlderPrice());
 
-        // Disable expiration of the price
-        if (poolPerAsset.computedAt == 0) {
+        // Disable expiration of the price if never initialized
+        if (poolPerAsset.computedAt == 0 && poolPerAsset.maxAge == 0) {
```
lemunozm, line 290: FIx issue: https://kflabs.slack.com/archives/C07PG2EUR9C/p1753179660610469

src/spoke/Spoke.sol
```diff
@@ -352,23 +356,11 @@ contract Spoke is Auth, Recoverable, ReentrancyProtection, ISpoke, ISpokeGateway
         ShareClassDetails storage shareClass = _shareClass(poolId, scId);
         (address asset, uint256 tokenId) = idToAsset(assetId);
         IVault vault = factory.newVault(poolId, scId, asset, tokenId, shareClass.shareToken, new address[](0));
-        registerVault(poolId, scId, assetId, asset, tokenId, factory, vault);
 
-        return vault;
-    }
-
-    /// @inheritdoc ISpoke
-    function registerVault(
```
lemunozm, line 361: No longer need it because we don't need to attach v2 vaults. 

I can revert if we want to maintain this

src/spoke/Spoke.sol
```diff
@@ -428,9 +420,7 @@ contract Spoke is Auth, Recoverable, ReentrancyProtection, ISpoke, ISpokeGateway
 
     /// @inheritdoc ISpoke
     function shareToken(PoolId poolId, ShareClassId scId) public view returns (IShareToken) {
-        IShareToken token = pools[poolId].shareClasses[scId].shareToken;
-        require(address(token) != address(0), UnknownToken());
-        return token;
+        return _shareClass(poolId, scId).shareToken;
     }
```
lemunozm, line 424: This change is idempotent, we now reuse code and the same error. 

Note that we had `UnknownToken` and `ShareTokenDoesNotExist` errors for the same purpose. Kept only `ShareTokenDoesNotExist`

src/spoke/Spoke.sol
```diff
@@ -462,7 +452,9 @@ contract Spoke is Auth, Recoverable, ReentrancyProtection, ISpoke, ISpokeGateway
         view
         returns (D18 price)
     {
-        (Price memory poolPerAsset,) = _pricesPoolPer(poolId, scId, assetId, false);
+        ShareClassDetails storage shareClass = _shareClass(poolId, scId);
```
lemunozm, line 455: This is a minor bug fix:

Previously, a caller to `pricePoolPerAsset()` with `checkValidity = true` could receive an `InvalidPrice` error if the share price is invalid although the asset one is valid, when they shouldn't.

This happened because `_pricesPoolPer` performs validations for both prices when only the asset price is used

lemunozm (commented):
Commented the `src` changes below:

lemunozm:
Ready to review, although I still need to remove old tests from `integration/Spoke.t.sol` that fully overlap between these new tests and the EndToEnd tests.

The coverage tool complains about some lines that are already covered. Not sure why or how to debug this better.

hieronx (dismissed):

lemunozm:
I've cleaned overlapping tests. Ready to merge now

hieronx (approved):

PR #543, opened 2025-07-28
Test: improve test organization

lemunozm:
- Move all integration tests to `integration/` folders (some of these were previously under `unit/`)
- Remove `MockHook` and `MockSafe` no longer used.
- Move vault tests to `test/vaults/` folder.
- Remove some no longer used or already cover test


test/spoke/unit/factories/TokenFactory.t.sol
```diff
@@ -6,69 +6,17 @@ import {IAuth} from "../../../../src/misc/interfaces/IAuth.sol";
...
-    function setUp() public {
-        if (vm.envOr("FORK_TESTS", false)) {
-            mainnetFork = vm.createFork(vm.rpcUrl("ethereum-mainnet"));
-            polygonFork = vm.createFork(vm.rpcUrl("polygon-mainnet"));
-        }
-
-        root = address(new Root(48 hours, address(this)));
-    }
-
-    function testTokenFactoryIsDeterministicAcrossChains(bytes16 scId) public {
```
lemunozm, line 35: IIUC, this test is no longer representative for us. Factories are deployed with `create3`, and the Token factory creates tokens with create2, fully independent of which chain it is, only parameters are `salt` and `decimals`

hieronx (approved):

PR #544, opened 2025-07-29
Ref: normalize spoke price usage to D18

lemunozm:
Some parts of `Spoke` still uses `uint128` for handling prices. A remaining from a previous refactor in spoke to change all `uint128` to `D18`. Now all prices used are handled as `D18`.

## Breaking change
* ABI-compatible with integrators
* Breaking change for indexer


src/spoke/interfaces/ISpoke.sol
```diff
@@ -86,10 +86,10 @@ interface ISpoke {
         ShareClassId indexed scId,
         address indexed asset,
         uint256 tokenId,
-        uint256 price,
```
hieronx, line 89: Isn't this an ABI-incompatible change since `D18=uint128` rather than `uint256`?

lemunozm: abi-compatible in the public calls for integrators, I wanted to meant. So hooks or managers do not need to migrate themself to adapt to a new api

lemunozm: But true that it will require changes in the indexer, good eye!

hieronx (commented):
LGTM but would wait for @wischli 's approval

wischli (approved):
Thanks for cleaning up the missed D18 price refactor code path(s) 🎉

PR #545, opened 2025-07-29
Minor hook fixes

hieronx:


src/hooks/BaseTransferHook.sol
```diff
@@ -40,6 +42,8 @@ abstract contract BaseTransferHook is Auth, IMemberlist, IFreezable, ITransferHo
         address crosschainSource_,
         address deployer
     ) Auth(deployer) {
+        require(redeemSource_ != depositTarget_ && depositTarget_ != crosschainSource_, InvalidInputs());
```
wischli: Do we also want to restrict redeemSource != crosschainSource?

wischli (dismissed):
Thanks for applying my nits and concern!

src/hooks/BaseTransferHook.sol
hieronx: Yeah we should true

wischli (approved):

PR #548, opened 2025-07-29
Ref: simplify spoke storage for asset prices

lemunozm:
Remove one storage from Spoke. 

Asset prices can be stored by assetId reusing the current mapping for that purpose. We simplify the storage and save some gas too (5K).


wischli (dismissed):
Happy to see that nesting being a bit flattened in a more reasonable way. Not looking forward to the spell though 😅

src/spoke/Spoke.sol
```diff
@@ -465,7 +464,16 @@ contract Spoke is Auth, Recoverable, ReentrancyProtection, ISpoke, ISpokeGateway
         view
         returns (D18 pricePoolPerAsset_, D18 pricePoolPerShare_)
     {
-        (Price memory poolPerAsset, Price memory poolPerShare) = _pricesPoolPer(poolId, scId, assetId, checkValidity);
+        ShareClassDetails storage shareClass = _shareClass(poolId, scId);
+
+        Price memory poolPerAsset = shareClass.asset[assetId].pricePoolPerAsset;
+        Price memory poolPerShare = shareClass.pricePoolPerShare;
+
+        if (checkValidity) {
```
hieronx: What about
```suggestion
        require(!checkValidity || (poolPerAsset.isValid() && poolPerShare.isValid()), InvalidPrice());
```

hieronx: I think this is as readable

lemunozm: I like it!

hieronx: Can you change this in `pricePoolPerShare` and `pricePoolPerAsset` as well?

lemunozm: Done

hieronx (approved):

PR #555, opened 2025-08-04
Ref: Spoke: Extract inner storage to top level contract

lemunozm:
### Changes
- Move shareClass and asset inner storages to top level
- Rename `pools` to `pool`
- Rename `shareClasses` to `shareClass`
- Remove hook check (discussed with Jeroen)
- Re-added registerVault for migrations
- Add a check to enforce requestManager existence before deploy a non-sync vault


src/spoke/Spoke.sol
```diff
@@ -150,8 +151,7 @@ contract Spoke is Auth, Recoverable, ReentrancyProtection, ISpoke, ISpokeGateway
 
     /// @inheritdoc ISpoke
     function request(PoolId poolId, ShareClassId scId, AssetId assetId, bytes memory payload) external {
-        ShareClassDetails storage shareClass = _shareClass(poolId, scId);
```
lemunozm, line 153: The share class check is no longer needed here, because it's checked when the request manager is configured

src/spoke/Spoke.sol
```diff
@@ -313,8 +311,7 @@ contract Spoke is Auth, Recoverable, ReentrancyProtection, ISpoke, ISpokeGateway
 
     /// @inheritdoc ISpokeGatewayHandler
     function requestCallback(PoolId poolId, ShareClassId scId, AssetId assetId, bytes memory payload) external auth {
-        ShareClassDetails storage shareClass = _shareClass(poolId, scId);
```
lemunozm, line 316: Same here

src/spoke/Spoke.sol
```diff
@@ -300,10 +300,8 @@ contract Spoke is Auth, Recoverable, ReentrancyProtection, ISpoke, ISpokeGateway
     }
 
     function setMaxAssetPriceAge(PoolId poolId, ShareClassId scId, AssetId assetId, uint64 maxPriceAge) external auth {
-        ShareClassDetails storage shareClass = _shareClass(poolId, scId);
```
lemunozm, line 303: The share class check is not need it in this case, and in case it is, it should be done at `Hub` level

src/spoke/Spoke.sol
```diff
@@ -279,8 +280,7 @@ contract Spoke is Auth, Recoverable, ReentrancyProtection, ISpoke, ISpokeGateway
         auth
     {
         (address asset, uint256 tokenId) = idToAsset(assetId);
-        ShareClassDetails storage shareClass = _shareClass(poolId, scId);
```
lemunozm, line 282: The share class check is not need it in this case, and in case it is, it should be done at `Hub` level

src/spoke/Spoke.sol
```diff
@@ -488,8 +484,7 @@ contract Spoke is Auth, Recoverable, ReentrancyProtection, ISpoke, ISpokeGateway
         view
         returns (uint64 computedAt, uint64 maxAge, uint64 validUntil)
     {
-        ShareClassDetails storage shareClass = _shareClass(poolId, scId);
```
lemunozm, line 491: The share class check is no longer need it. No consumers of `markersPricePoolPerAsset` use this check as an implicit ward

src/spoke/Spoke.sol
```diff
@@ -447,8 +444,7 @@ contract Spoke is Auth, Recoverable, ReentrancyProtection, ISpoke, ISpokeGateway
         view
         returns (D18 price)
     {
-        ShareClassDetails storage shareClass = _shareClass(poolId, scId);
```
lemunozm, line 450: This is the most controversial check removal. This method is used in `SyncManager` and `BalanceSheet`. Removing the share class check is not a problem for the `SyncManager` usage because the usage there is with price validity. `BalanceSheet` does not use validity, so theoretically, this check validates the existence of the share class implicitly. Nevertheless:

- The entire contract never does that validation in most places.
- Things like overridePrice + submitQueued bypasses the check.

So I think the correct solution is to avoid the check here and be explicit in `BalanceSheet`. Anyways, not having this check is not an issue [ref](https://kflabs.slack.com/archives/C07PG2EUR9C/p1754295432002099)

src/spoke/Spoke.sol
```diff
@@ -150,8 +151,7 @@ contract Spoke is Auth, Recoverable, ReentrancyProtection, ISpoke, ISpokeGateway
 
     /// @inheritdoc ISpoke
     function request(PoolId poolId, ShareClassId scId, AssetId assetId, bytes memory payload) external {
-        ShareClassDetails storage shareClass = _shareClass(poolId, scId);
-        require(msg.sender == address(shareClass.asset[assetId].manager), NotAuthorized());
+        require(msg.sender == address(assetInfo[poolId][scId][assetId].manager), NotAuthorized());
```
wischli: Nit: We could consider using a new `InvalidRequestManager()` error for better clarity and explicit manager existence check before authorization
```suggestion
        IRequestManager manager = assetInfo[poolId][scId][assetId].manager;
        require(address(manager) != address(0), InvalidRequestManager());
        require(msg.sender == address(manager), NotAuthorized());
```

src/spoke/Spoke.sol
```diff
@@ -350,34 +344,53 @@ contract Spoke is Auth, Recoverable, ReentrancyProtection, ISpoke, ISpokeGateway
...
         (address asset, uint256 tokenId) = idToAsset(assetId);
-        IVault vault = factory.newVault(poolId, scId, asset, tokenId, shareClass.shareToken, new address[](0));
+        IVault vault = factory.newVault(poolId, scId, asset, tokenId, shareClass_.shareToken, new address[](0));
 
-        _vaultDetails[vault] = VaultDetails(assetId, asset, tokenId, false);
-        emit DeployVault(poolId, scId, asset, tokenId, factory, vault, vault.vaultKind());
+        require(
+            vault.vaultKind() == VaultKind.Sync || address(assetInfo[poolId][scId][assetId].manager) != address(0),
+            InvalidRequestManager()
+        );
```
wischli, line 356: 🚀  Nice addition

wischli (dismissed):
Super clean changes. Really appreciate the context on the controversial ones which helped me a lot to verify the changes don't break anything!

src/spoke/Spoke.sol
hieronx: @lemunozm I'm thinking more about this one. `vaultKind` is quite opinionated to our current vault options. It is very possible there will be future vault implementations that don't adhere to our current enum values.

I wonder if we should remove this.

There could be future vault implementations that are not synchronous, but still don't use our request manager logic, e.g. because they handle requests locally (on the spoke chain).

lemunozm: > There could be future vault implementations that are not synchronous, but still don't use our request manager logic, e.g. because they handle requests locally (on the spoke chain).

In that case could not they use a `requestManager` as indicator of that? Maybe some hardcoded value: `0xaaaaaaaaaaaa` or so. Meaning: it's asynchronous and but handled manually.

I added this check from some Frederik feedback in the configuration, that wanted to avoid deploying and linking a vault without configuring the requestManager first.

This check goes also in line with this on top refactor: https://github.com/centrifuge/protocol-v3/pull/563 where the spoke handles the vaults indexed by the manager, so we need anyways some "manager" address to index the vault correctly. I think using some harcoded addresses can work for the scenario you propose

hieronx: > In that case could not they use a requestManager as indicator of that? Maybe some hardcoded value: 0xaaaaaaaaaaaa or so. Meaning: it's asynchronous and but handled manually.

I guess that could be done, although I don't fully see the point.

But still leaves the question of whether we should keep vaultKind as is, despite it being quite specific to current vaults impl.

lemunozm: > I guess that could be done, although I don't fully see the point.

The point is that we will store the vault in spoke by: `vault[poolId][scId][assetId][manager] = vault_`, so we need to attach some "manager" to it, somehow.

> But still leaves the question of whether we should keep vaultKind as is, despite it being quite specific to current vaults impl.

Maybe we can leave this as an open point to resolve later? You could be right that maybe the `vaultKind` no longer applies to this new schema. But we should at least distinguish between vaults that require a `requestManager` and vaults that don't, IIUC.

hieronx: > The point is that we will store the vault in spoke by: vault[poolId][scId][assetId][manager] = vault_, so we need to attach some "manager" to it, somehow.

Yeah okay I see it!

> Maybe we can leave this as an open point to resolve later? You could be right that maybe the vaultKind no longer applies to this new schema. But we should at least distinguish between vaults that require a requestManager and vaults that don't, IIUC.

Definitely fine to revisit later.

src/spoke/Spoke.sol
lemunozm: Agree with your NIT!

src/spoke/Spoke.sol
wischli: Definitely agree the `vaultKind` reliance dependence would be good to get rid of given the context and main goal of no further spoke migration if possible.

src/spoke/Spoke.sol
lemunozm: [131b342](https://github.com/centrifuge/protocol-v3/pull/555/commits/131b342b848df3574b04cbd788fd37ba1b2178ae)

wischli (approved):

PR #556, opened 2025-08-04
Deploy: register tokenRecoverer

lemunozm:


script/FullDeployer.s.sol
```diff
@@ -26,12 +26,8 @@ contract FullDeployer is ExtendedHubDeployer, ExtendedSpokeDeployer, AdaptersDep
     function deployFull(CommonInput memory commonInput, AdaptersInput memory adaptersInput, FullActionBatcher batcher)
         public
     {
-        if (commonInput.version == keccak256(abi.encodePacked(("3")))) _verifyAdmin(commonInput);
-
         _preDeployFull(commonInput, adaptersInput, batcher);
         _postDeployFull(batcher);
-
-        if (commonInput.version == keccak256(abi.encodePacked(("3")))) _verifyMainnetAddresses();
```
lemunozm, line 37: Just moved to `run()` because theoretically, those changes only concern to real deployments, not to deployments used in EndToEnd or integration tests

script/CommonDeployer.s.sol
```diff
@@ -225,14 +225,14 @@ abstract contract CommonDeployer is Script, JsonRegistry, CreateXScript {
         batcher.engageCommon(_commonReport());
 
         register("root", address(root));
-        // register("adminSafe", address(adminSafe)); => Already present in load_vars.sh and not needed to be registered
```
lemunozm, line 228: I think this line led to confusion. We don't need to register this because we're not deploying this contract. We only must call `register()` to contracts we deploy

wischli (commented):
Thanks for the quick fixes! I'd prefer the deployment jsons to not be reformatted to keep the diff lower though as a nit.

Also not sure if we rather want to merge the v3.0.1 fixes into `release-v3.0.1` and then backport to `main` when done. Otherwise, we have to cherry-pick everything eventually into `release-v3.0.1`.

lemunozm:
> I'd prefer the deployment jsons to not be reformatted to keep the diff lower though as a nit.

Yeah, sorry for that, actually, the json was in 4-space tab and I convert it to 2-space tab, which is the expected output when running the deployers. So actually, it was wrong before (maybe by me too).

Only real changes is adding one line for the `tokenRecoverer`

lemunozm:
Rebased to release-v3.0.1

lemunozm:
Tell me if you want me to backport this later. I use `git rebase --onto` which works pretty nice in these contexts

wischli (approved):

lemunozm:
Here in main: https://github.com/centrifuge/protocol-v3/pull/559

PR #557, opened 2025-08-04
tests: base transfer hook 100% coverage

wischli:
Cherry-picks `BaseTransferHook` tests from #547 and refactors them to non-enum transfer type support


lemunozm (approved):
Awesome! Thanks for these tests!

PR #560, opened 2025-08-05
feat: vault factory create2 salts

wischli:
Adds `create2` vault salts to both vault factories which ensures deterministic vault addresses for the same `(poolId, shareClassId, assetAddress)` triplet across different chains

### DIFF
--- a/src/vaults/factories/AsyncVaultFactory.sol
+++ b/src/vaults/factories/AsyncVaultFactory.sol
@@ -35,7 +35,7 @@ contract AsyncVaultFactory is Auth, IVaultFactory {
     ) public auth returns (IVault) {
         require(tokenId == 0, UnsupportedTokenId());
 
-        bytes32 salt = _generateSalt(poolId, scId, asset);
+        bytes32 salt = keccak256(abi.encode(poolId, scId, asset));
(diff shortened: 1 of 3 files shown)


src/vaults/factories/AsyncVaultFactory.sol
```diff
@@ -49,4 +51,12 @@ contract AsyncVaultFactory is Auth, IVaultFactory {
...
         return vault;
     }
+
+    /// @notice Generate deterministic salt for CREATE2 deployment
+    /// @dev Uses keccak256 hash of encoded poolId, scId, and asset address to ensure
+    ///      deterministic vault addresses across all chains. Same inputs will always
+    ///      produce the same salt, enabling predictable cross-chain vault addresses.
+    function _generateSalt(PoolId poolId, ShareClassId scId, address asset) internal pure returns (bytes32 salt) {
+        salt = keccak256(abi.encode(poolId, scId, asset));
+    }
```
lemunozm: Do we really need the function? Just to reduce the number of lines of the fix for the audits

lemunozm (dismissed):

lemunozm (approved):

PR #562, opened 2025-08-05
tests: Add fork validation

wischli:
* Adds fork validation checks of `rely` and `file` calls from the various deployer scripts
* Also checks wired adapters
* Extends constants of `IntegrationTests`
  * Decided to keep constant addresses instead of dynamically reading from JSON due to performance and readability

Subsequent PR Plans
* Support all main networks instead of just Ethereum
* In a follow-up PR, we can add checks for other dynamic filings and configurations such as `setRequestManager` or `updateBalanceSheetManager` results


test/integration/utils/VMLabeling.sol
```diff
@@ -0,0 +1,59 @@
...
+
+import "forge-std/Test.sol";
+
+/// @title VMLabeling
+/// @notice Abstract contract that provides VM labeling for all integration test contracts
+abstract contract VMLabeling is Test {
+    /// @notice Sets up VM labels for all contracts to enable automatic name resolution in test outputs
+    function _setupVMLabels() internal {
+        // Core system contracts
+        vm.label(IntegrationConstants.ROOT, "Root");
```
lemunozm, line 14: I think they can initiate them as an immutable `makeAddr()` to avoid this list of `vm.label`

test/integration/EndToEnd.t.sol
```diff
@@ -172,6 +173,8 @@ contract EndToEndDeployment is Test {
     //----------------------------------------------------------------------------------------------
 
     function setUp() public virtual {
+        _setupVMLabels();
+
         adapterAToB = _deployChain(deployA, CENTRIFUGE_ID_A, CENTRIFUGE_ID_B, SAFE_ADMIN_A);
         adapterBToA = _deployChain(deployB, CENTRIFUGE_ID_B, CENTRIFUGE_ID_A, SAFE_ADMIN_B);
```
lemunozm: I think here you don't need to call setupVMLabels because they are already set by `JsonRegistry`: https://github.com/centrifuge/protocol-v3/blob/main/script/utils/JsonRegistry.s.sol#L21

wischli: I definitely had to label them because none of them were labelled. AFAICT, the above codepath is only executed when parsing from JSON which we are not doing here in order to have a full homogeneous static list of constants in contrast to a mixed list of dynamically JSON parsed and hardcoded constants.

test/integration/utils/VMLabeling.sol
wischli: Can you elaborate? I cannot declare them as a non-const variable with `makeAddr` in the `IntegrationConstants` library unfortunately, doesn't compile.

test/integration/EndToEnd.t.sol
wischli: I'm curious: Would you prefer moving the contents of `IntegrationConstants` to JSON files? I suppose we can add most there and then dynamically read from there. This would improve readability for separating network-dependant addresses and constants. But for me, it's was much more helpful to have the addresses hardcoded for spell and test writing. The downside is of course that this replicates a lot of entries from our JSONs.

test/integration/utils/VMLabeling.sol
lemunozm: Something like that can not be achievable in the library? https://github.com/centrifuge/protocol-v3/blob/main/test/adapters/Deployment.t.sol#L19

Maybe the fact of being a library disallow the immutable initialization

test/integration/EndToEnd.t.sol
lemunozm: > I definitely had to label them because none of them were labelled

Ah, that's because you're using the EndToEnd infrastructure for the fork test, I guess. Normal EndToEnd tests are already labeled. Maybe to clarify that, can `_setupVMLabels()` be called from the Fork test `setUp()`?

No strong opinion right now about the json files. Let's keep it as it is right now!

test/integration/utils/VMLabeling.sol
wischli: Sadly, libraries can't inherit from `Test` so they don't have access to `makeAddr`. Also, even if you could use `makeAddr("0x123...")` with a real address, the debugger would show the generated address instead of the var name, so you'd still need `vm.label()` calls for readable debugging.

test/integration/EndToEnd.t.sol
wischli: > Ah, that's because you're using the EndToEnd infrastructure for the fork test, I guess. Normal EndToEnd tests are already labeled. Maybe to clarify that, can _setupVMLabels() be called from the Fork test setUp()?

Great point! Agree it can be moved down in the inheritance chain because normal e2e integration tests don't need this, only fork tests. Will apply!

test/integration/utils/VMLabeling.sol
lemunozm: That's correct 👍🏻

test/integration/EndToEnd.t.sol
wischli: > No strong opinion right now about the json files. Let's keep it as it is right now!

I think unfortunately this is not feasible and creates worse UX. With the help of Claude Code, I planned it out with a simple prototype. There were some wins but overall appears to be just added unnecessary complexity just for the sake of fork tests. For instance, compilation time is expected to increase quite significantly. Also, if we want to move all vault addresses to the jsons and use them in code, some of these will be optional (like deJTRSY JTRSY denominated vault only on Eth). In Rust, this would be fine but not in Solidity where these optional values are just a pain to deal with.

test/integration/fork/ForkTestInvestments.sol
```diff
@@ -0,0 +1,227 @@
...
+/// @notice Fork tests for async investment flows on Ethereum mainnet
+contract ForkTestAsyncInvestments is ForkTestBase, VMLabeling {
+    // TODO(later): After v2 disable, switch to JAAA
+    IBaseVault constant VAULT = IBaseVault(IntegrationConstants.ETH_DEJAA_USDC_VAULT);
+
+    uint128 constant depositAmount = IntegrationConstants.DEFAULT_USDC_AMOUNT;
+
+    function setUp() public virtual override {
+        super.setUp();
+        _setupVMLabels();
```
wischli, line 53: @lemunozm Moved from `EndToEndDeployments` here as desired!

wischli:
@lemunozm If you agree, let's merge the current state. I will then open a follow-up PR which adds support for true multi chain validation of our 6 chains.

lemunozm:
Awesome!! Thanks!

lemunozm (approved):
Amazing job!

PR #563, opened 2025-08-05
Ref: vaults registered in spoke

lemunozm:
Spoke now handles the asyncRequestManager storage to handle vaults. IVaultManager is no longer used.

Blocked by https://github.com/centrifuge/protocol-v3/pull/555


src/spoke/Spoke.sol
```diff
@@ -323,17 +328,17 @@ contract Spoke is Auth, Recoverable, ReentrancyProtection, ISpoke, ISpokeGateway
         VaultUpdateKind kind
     ) external auth {
         if (kind == VaultUpdateKind.DeployAndLink) {
-            IVault vault = deployVault(poolId, scId, assetId, IVaultFactory(vaultOrFactory));
-            linkVault(poolId, scId, assetId, vault);
+            IVault vault_ = deployVault(poolId, scId, assetId, IVaultFactory(vaultOrFactory));
```
lemunozm, line 331: note most lines changes are just renaming local variable `vault` to `vault_`

src/spoke/Spoke.sol
```diff
@@ -367,56 +372,62 @@ contract Spoke is Auth, Recoverable, ReentrancyProtection, ISpoke, ISpokeGateway
...
-        manager.addVault(poolId, scId, assetId, vault, asset, tokenId);
-
+        IRequestManager manager = assetInfo[poolId][scId][assetId].manager;
+        vault[poolId][scId][assetId][manager] = vault_;
         assetInfo[poolId][scId][assetId].numVaults++;
         vaultDetails_.isLinked = true;
 
+        if (manager == REQUEST_MANAGER_V3_0) {
+            IVaultManager(address(manager)).addVault(poolId, scId, assetId, vault_, asset, tokenId);
+        }
```
lemunozm, line 398: this

src/spoke/Spoke.sol
```diff
@@ -367,56 +372,62 @@ contract Spoke is Auth, Recoverable, ReentrancyProtection, ISpoke, ISpokeGateway
...
-        manager.removeVault(poolId, scId, assetId, vault, asset, tokenId);
-
+        IRequestManager manager = assetInfo[poolId][scId][assetId].manager;
+        delete vault[poolId][scId][assetId][manager];
         assetInfo[poolId][scId][assetId].numVaults--;
         vaultDetails_.isLinked = false;
 
+        if (manager == REQUEST_MANAGER_V3_0) {
+            IVaultManager(address(manager)).removeVault(poolId, scId, assetId, vault_, asset, tokenId);
+        }
```
lemunozm, line 424: and this

lemunozm (commented):
These are the only real changes in Spoke

lemunozm:
Ready to review. 

I need to fix some issues with the Fork tests and import orders, but all the remaining tests pass

src/spoke/interfaces/ISpoke.sol
```diff
@@ -328,4 +328,11 @@ interface ISpoke {
         external
         view
         returns (uint64 computedAt, uint64 maxAge, uint64 validUntil);
+
+    /// @notice Returns the address of the vault for a given pool, share class asset and requestManager
+    /// @param manager the request manager associated to the vault, if 0, then it correspond to a full sync vault.
+    function vault(PoolId poolId, ShareClassId scId, AssetId assetId, IRequestManager manager)
```
wischli, line 334: I suppose you left out the remaining params to reduce noise?

src/spoke/Spoke.sol
```diff
@@ -31,6 +31,8 @@ import {ISpokeGatewayHandler} from "../common/interfaces/IGatewayHandlers.sol";
 import {VaultUpdateKind, MessageLib} from "../common/libraries/MessageLib.sol";
 import {IPoolEscrowFactory} from "../common/factories/interfaces/IPoolEscrowFactory.sol";
 
+IRequestManager constant REQUEST_MANAGER_V3_0 = IRequestManager(0x58d57896EBbF000c293327ADf33689D0a7Fd3d9A);
```
wischli: CR: The live AsyncRequestManager from v3.0.1 is `0xF06f89a1b6C601235729A689595571B7455dD433` ([ref](https://docs.centrifuge.io/developer/protocol/deployments/#mainnet))

src/spoke/Spoke.sol
wischli: Can you elaborate on why we hardcode the manager? Just for backwards compatibility? 

The latter seems sound to me as the `vault()` lookup is changed from `AsyncRequestManager` to `Spoke` in this PR. It is still needed from the `AsyncRequestManager` whenever a request is fulfilled which is either available in >v3.0.1 via Spoke or in =v3.0.1 via AsyncRequestManager.

src/spoke/Spoke.sol
```diff
@@ -53,6 +55,9 @@ contract Spoke is Auth, Recoverable, ReentrancyProtection, ISpoke, ISpokeGateway
     mapping(PoolId => Pool) public pool;
     mapping(PoolId => mapping(ShareClassId scId => ShareClassDetails)) public shareClass;
     mapping(PoolId => mapping(ShareClassId => mapping(AssetId => ShareClassAsset))) public assetInfo;
+    mapping(
+        PoolId poolId => mapping(ShareClassId scId => mapping(AssetId assetId => mapping(IRequestManager => IVault)))
+    ) public vault;
```
wischli, line 58: Just for the sake of completeness, I was looking at a more gas efficient variant due to the 4 keccak hashes that are necessary to read and write a `vault` entry. However, TLDR is that only ~190 gas is saved on read and write with worse readability. So not proposing that, just sharing as informational:

```solidity
      mapping(bytes32 => IVault) public vault;

      function _vaultKey(PoolId poolId, ShareClassId scId, AssetId assetId, IRequestManager manager)
          private pure returns (bytes32) {
          return keccak256(abi.encode(poolId, scId, assetId, manager));
      }

      function getVault(PoolId poolId, ShareClassId scId, AssetId assetId, IRequestManager manager)
          external view returns (IVault) {
          return vault[_vaultKey(poolId, scId, assetId, manager)];
      }

      function _setVault(PoolId poolId, ShareClassId scId, AssetId assetId, IRequestManager manager, IVault vault_)
          internal {
          vault[_vaultKey(poolId, scId, assetId, manager)] = vault_;
      }
```

| Operation | Option 1 (Nested) | Option 2 (Single) | Difference |
|-----------|------------------|-------------------|------------|
| **Cold Read** | ~2,520 gas | ~2,330 gas | **190 gas saved** |
| **Warm Read** | ~520 gas | ~330 gas | **190 gas saved** |
| **Write (0→x)** | ~20,420 gas | ~20,230 gas | **190 gas saved** |
| **Write (x→y)** | ~3,320 gas | ~3,130 gas | **190 gas saved** |
| **Write (x→0)** | +1,620 gas* | +1,430 gas* | **190 gas saved** |

**Option 1 costs:**
- 4× keccak256: ~420 gas (4 × ~105 each)
- 1× SLOAD/SSTORE: variable

**Option 2 costs:**
- `abi.encode`: ~130 gas
- 2× keccak256: ~100 gas (explicit) + ~100 gas (internal)
- 1× SLOAD/SSTORE: variable

src/spoke/interfaces/ISpoke.sol
lemunozm: Yes, but I can add them too. No strong opinion.

src/spoke/Spoke.sol
lemunozm: Correct! Thanks for pointing this out

src/spoke/Spoke.sol
lemunozm: Thanks so much for this research. That's something I was interested on. 

I think, somehow, the compiler under the hood does some `keccak256(abi.encode(poolId, scId, assetId, manager))` (maybe with `encodePacked` instead and that's the 190 gas diff?) each time they need to access some value with several nested mappings

src/spoke/Spoke.sol
lemunozm: > Just for backwards compatibility?

Yes. In case we need to link/unlink a vault that is handled by the old asyncRequestManager, we need to call addVault/removeVault. New manager will read the spoke storage directly

lemunozm:
NOTE: Currently the fork tests fail, because the fork tests are trying to check the existence of v3.0 vaults using the new `Spoke`

wischli:
> NOTE: Currently the fork tests fail, because the fork tests are trying to check the existence of v3.0 vaults using the new `Spoke`

Are you sure about that? The fork tests should use the configured contracts such that this PR should have no impact on them.

lemunozm:
Yeah, because, IIUC:

- The vaults no are no longer stored in `AsyncRequestManager`. so `vaultByKind()` no longer exists.
- Spoke now stores the vaults, and can be retrived by `spoke.vault()`
- vaults from v3.0 are not registered in spoke, so they can not be get using `spoke.vault()`.

I think, ideally when we upload v3.1, we should unlink/link all vaults to be able to store them in `spoke.vaults`. In the meantime we need some kind of distincion on how to retrive the vaults in the EndToEnd utilities/flows

wischli:
> I think, ideally when we upload v3.1, we should unlink/link all vaults to be able to store them in `spoke.vaults`. In the meantime we need some kind of distincion on how to retrive the vaults in the EndToEnd utilities/flows

Agree this should be the migration path. 

For the failing tests, we should apply a fix in this PR. Since fork tests should simulate the live behaviour whereas e2e tests should simulate the latest state, the issue is that fork tests reuse e2e infra. Looking deeper, I think we should mutate `_ensureAsyncVaultExists` and the corresponding lines in `_asyncRedeemFlow` which do a similar check to support both versions of the system. Will provide some inline comments.

test/integration/EndToEnd.t.sol
```diff
@@ -565,7 +565,7 @@ contract EndToEndFlows is EndToEndUtils {
         vm.startPrank(poolManager);
 
         // Check if vault already exists (for fork tests)
-        address existingVault = address(spoke.asyncRequestManager.vaultByAssetId(poolId, shareClassId, assetId));
+        address existingVault = address(spoke.spoke.vault(poolId, shareClassId, assetId, spoke.asyncRequestManager));
```
wischli:
```suggestion
        address existingVault = _getVaultWithFallback(spoke, poolId, shareClassId, assetId);
```

test/integration/EndToEnd.t.sol
```diff
@@ -798,7 +799,7 @@ contract EndToEndFlows is EndToEndUtils {
         // Resolve vault - use existing if provided, otherwise get from manager
         IAsyncRedeemVault vault = existingVault != address(0)
             ? IAsyncRedeemVault(existingVault)
-            : IAsyncRedeemVault(address(spoke.asyncRequestManager.vaultByAssetId(poolId, shareClassId, assetId)));
+            : IAsyncRedeemVault(address(spoke.spoke.vault(poolId, shareClassId, assetId, spoke.asyncRequestManager)));
```
wischli: Draft proposal of supporting legacy v3 and new

```suggestion
            : IAsyncRedeemVault(_getVaultWithFallback(spoke, poolId, shareClassId, assetId));
```

Helper 
```solidity
  function _getVaultWithFallback(
      CSpoke memory spoke,
      PoolId poolId,
      ShareClassId shareClassId,
      AssetId assetId
  ) internal view returns (address vaultAddr) {
      // Try new system first
      vaultAddr = address(spoke.spoke.vault(poolId, shareClassId, assetId, 
  spoke.asyncRequestManager));
      
      // Fallback to legacy V3.0 lookup if not found in new system
      if (vaultAddr == address(0) && spoke.asyncRequestManager == REQUEST_MANAGER_V3_0) {
          vaultAddr = address(IVaultManager(address(spoke.asyncRequestManager)).vaultByAssetId(poolId,
   shareClassId, assetId));
      }

      return vaultAddr;
  }
```

wischli (commented):
Draft proposal for unblocking e2e and fork tests

lemunozm:
Thanks for the ideas! Will look into it tomorrow. BTW, somehow the fork test are still running under `test` job. Weird because seems the CI command is the correct one.

test/integration/EndToEnd.t.sol
lemunozm: I think I need to try-catch the first call, given the forked spoke doesn't have the `vault()` method but this is the idea!

.github/workflows/ci.yml
```diff
@@ -21,7 +21,7 @@ jobs:
           version: nightly
 
       - name: Run tests
-        run: forge test --no-match-path "test/spoke/fuzzing/**/*.sol,test/integration/fork/**/*.sol" --deny-warnings
+        run: forge test --no-match-path "test/integration/fork/**/*.sol" --deny-warnings
```
wischli, line 24: This should have worked but fine with merging. I'll fix it when we finally merge the invariant tests 😅

wischli (approved):
Loving the simplistic e2e & fork test fix handler. Hope we don't need many more in the future which would justify decoupling both as discussed.

.github/workflows/ci.yml
lemunozm: Yeah, somehow the test CI job was also executing the fork ones, I tried if that helped (but not checked TBH)

PR #566, opened 2025-08-06
Ref: wire adapters without using the Guardian

lemunozm:
A simplified solution to avoid the Guardian to wire adapters.

Each adapter now has its own `wire()` method and a ward on `safeAdmin`. This allows:
- To have the Guardian without any adapter knowledge.
- To avoid maintaining and upgrading a new `AdapterManager` to handle the wiring process.

Also, it prepares some foundations for adapters-per-pool where the pool admin should be able to handle the adapter directly to wire it.


script/WireAdapters.s.sol
```diff
@@ -66,48 +66,30 @@ contract WireAdapters is Script {
...
-                    );
-                }
+                IWormholeAdapter(localWormholeAddr).wire(
+                    remoteCentrifugeId,
+                    uint16(vm.parseJsonUint(remoteConfig, "$.adapters.wormhole.wormholeId")),
+                    vm.parseJsonAddress(remoteConfig, "$.contracts.wormholeAdapter")
+                );
+
+                console.log("Wired WormholeAdapter from", localNetwork, "to", remoteNetwork);
             }
```
lemunozm, line 91: IIUC, line 75 is useless because if `localWormholeAddr != 0` then it has configuration. So the `catch` section is unreachable

src/adapters/AxelarAdapter.sol
```diff
@@ -41,22 +41,28 @@ contract AxelarAdapter is Auth, IAxelarAdapter {
...
     /// @inheritdoc IAxelarAdapter
     function file(bytes32 what, uint16 centrifugeId, string calldata axelarId, string calldata destination)
-        external
+        public
         auth
     {
         if (what == "destinations") destinations[centrifugeId] = AxelarDestination(axelarId, destination);
         else revert FileUnrecognizedParam();
         emit File(what, centrifugeId, axelarId, destination);
     }
```
lemunozm: @hieronx I think these `file` methods are no longer needed because the configuration can always be done through `wire()`. That would remove some complexity in dealing and understanding these file methods. If you agree I can remove it in this PR

wischli (dismissed):
Codewise all LGTM. Since we planned on re-deploying adapters and the Guardian, no blocker from my side.

IMO, we can remove/internalize the `{Axelar Wormhole}Adapter.file` methods due to the new `wire` ones.

src/adapters/AxelarAdapter.sol
lemunozm: Removed here: [c2ae254](https://github.com/centrifuge/protocol-v3/pull/566/commits/c2ae25444c043bfc2a6d7110543f92569989fe63), which can be reverted easily if disaggre. A lot of lines removed!

wischli (approved):

PR #570, opened 2025-08-12
Feat: adapters per pool

lemunozm:
Allows the pool manager to configure its optional adapters. Initial spec: https://www.notion.so/Adapters-per-pool-24d2eac24e1780e8bdd3ddf0f98cf3a4?source=copy_link

Current solution: "best effort": New adapters are configured when the message is received. Any message received from old adapters will be lost and needs to be recovered.

Next solution: https://kflabs.slack.com/archives/C07PG2EUR9C/p1757322269874999?thread_ts=1757319528.336239&cid=C07PG2EUR9C using the recoverer as manager to disable/enable sending messages. The pool manager can ensure no messages on the air before configuring new adapters => https://github.com/centrifuge/protocol-v3/pull/601


src/hub/Hub.sol
```diff
@@ -646,6 +647,29 @@ contract Hub is Multicall, Auth, Recoverable, IHub, IHubGatewayHandler, IHubGuar
...
+        // Should we send a message to recover message from remote chains?
+        multiAdapter.disputeRecovery(centrifugeId, PoolId.wrap(0), adapter, payloadHash);
+    }
+
+    /// @inheritdoc IHub
+    function setPoolAdapters(uint16 centrifugeId, PoolId poolId, IAdapter[] memory adapters) external {
+        _isManager(poolId);
+
+        sender.sendSetPoolAdapters(centrifugeId, poolId, adapters);
+    }
```
lemunozm: Follows the same pattern as the `Guardian` for these stuff, but using isManager as authorization for the pool actions

src/common/MessageDispatcher.sol
```diff
@@ -522,4 +523,17 @@ contract MessageDispatcher is Auth, IMessageDispatcher {
...
     }
+
+    /// @inheritdoc IHubMessageSender
+    function sendSetPoolAdapters(uint16 centrifugeId, PoolId poolId, IAdapter[] memory adapters) external {
+        if (centrifugeId == localCentrifugeId) {
+            multiAdapter.file("adapters", centrifugeId, poolId, adapters);
+        } else {
+            gateway.send(
+                centrifugeId,
+                MessageLib.SetPoolAdapters({centrifugeId: centrifugeId, poolId: poolId.raw(), adapters: adapters})
```
lemunozm: Important note, this message should return `messagePoolId() => POOL_0`. But it can not be subsidized by POOL_0

lemunozm: Ref: https://kflabs.slack.com/archives/C07PG2EUR9C/p1755000251957629?thread_ts=1754998618.808749&cid=C07PG2EUR9C

src/common/MultiAdapter.sol
```diff
@@ -91,7 +102,8 @@ contract MultiAdapter is Auth, IMultiAdapter {
     }
 
     function _handle(uint16 centrifugeId, bytes calldata payload, IAdapter adapter_) internal {
-        Adapter memory adapter = _adapterDetails[centrifugeId][adapter_];
+        PoolId poolId = messageProperties.messagePoolId(payload);
```
wischli: Reminder that we also need handle pool independent messages which don't contain a pool id: `{Cancel, ScheduleUpgrade}`, `RecoverTokens` and`RegisterAsset`.

How do you envision configuring adapters for those? Using `safe` as recoverer?

lemunozm: Yeah, they will use the global adapters as they currently work. Using `Guardian.setAdapters()` and using safe as `adapter.wire()`. Also probably `safe` for recoverer

src/common/MultiAdapter.sol
```diff
@@ -148,28 +168,8 @@ contract MultiAdapter is Auth, IMultiAdapter {
...
-    /// @inheritdoc IMultiAdapter
-    function executeRecovery(uint16 centrifugeId, IAdapter adapter, bytes calldata payload) external {
-        bytes32 payloadHash = keccak256(payload);
-        uint256 recovery = recoveries[centrifugeId][adapter][payloadHash];
-
-        require(recovery != 0, RecoveryNotInitiated());
-        require(recovery <= block.timestamp, RecoveryChallengePeriodNotEnded());
-
-        delete recoveries[centrifugeId][adapter][payloadHash];
+    function executeRecovery(uint16 centrifugeId, PoolId poolId, IAdapter adapter, bytes calldata payload) external {
```
lemunozm: I propose to rename this as `executeHandle()`, recovery is now a usage of calling this method.

hieronx: Maybe just `execute`? But yeah renaming makes sense

src/common/interfaces/IGateway.sol
```diff
@@ -56,8 +57,8 @@ interface IGateway is IMessageHandler, IMessageSender, IRecoverable {
     /// @notice Dispatched when a batch that has not been underpaid is repaid.
     error NotUnderpaidBatch();
 
-    /// @notice Dispatched when a batch is repaid with insufficient funds.
-    error InsufficientFundsForRepayment();
+    /// @notice Dispatched when a batch is repaid with insufficient funds or the sending is blocked.
+    error CanNotBeRepaid();
```
hieronx: Nit:
```suggestion
    error CannotBeRepaid();
```

src/common/interfaces/IGatewaySenders.sol
```diff
@@ -9,6 +9,8 @@ import {ShareClassId} from "../types/ShareClassId.sol";
 import {VaultUpdateKind} from "../libraries/MessageLib.sol";
 
 interface ILocalCentrifugeId {
+    error CanNotBeSentLocally();
```
hieronx: Nit:
```suggestion
    error CannotBeSentLocally();
```

src/hub/Hub.sol
```diff
@@ -637,14 +644,43 @@ contract Hub is Multicall, Auth, Recoverable, IHub, IHubGatewayHandler, IHubGuar
...
         accounting.lock();
     }
 
+    /// @inheritdoc IHub
+    function setAdapters(
+        uint16 centrifugeId,
+        PoolId poolId,
+        IAdapter[] memory localAdapters,
+        bytes32[] memory remoteAdapters,
+        bytes32 remoteManager
```
hieronx: Why include manager here if we also have `setAdaptersManager`, and it could just be batched, right?

src/hub/Hub.sol
```diff
@@ -637,14 +644,43 @@ contract Hub is Multicall, Auth, Recoverable, IHub, IHubGatewayHandler, IHubGuar
     }
 
     /// @inheritdoc IHub
-    function updateJournal(PoolId poolId, JournalEntry[] memory debits, JournalEntry[] memory credits) external {
+    function updateJournal(PoolId poolId, JournalEntry[] memory debits, JournalEntry[] memory credits)
+        external
+        payable
```
hieronx, line 647: Oops very good catch!

src/hub/Hub.sol
```diff
@@ -111,6 +116,8 @@ contract Hub is Multicall, Auth, Recoverable, IHub, IHubGatewayHandler, IHubGuar
 
         IPoolEscrow escrow = poolEscrowFactory.newEscrow(poolId);
         gateway.setRefundAddress(poolId, escrow);
+
+        multiAdapter.setManager(poolId, admin);
```
hieronx: Why automatically set this here?

src/common/MultiAdapter.sol
```diff
@@ -211,23 +220,39 @@ contract MultiAdapter is Auth, IMultiAdapter {
...
 
         for (uint256 i; i < adapters_.length; i++) {
             total += adapters_[i].estimate(centrifugeId, i == PRIMARY_ADAPTER_ID - 1 ? payload : proof, gasLimit);
         }
     }
 
     /// @inheritdoc IMultiAdapter
-    function quorum(uint16 centrifugeId) external view returns (uint8) {
-        Adapter memory adapter = _adapterDetails[centrifugeId][adapters[centrifugeId][0]];
+    function enableSending(uint16 centrifugeId, PoolId poolId, bool canSend) external {
```
hieronx: Nit: would rename to `setOutgoing`

1. I think `enable*` is confusing since it might be disabling depending on the boolean param
2. Outgoing is a bit clearer than sending imo

hieronx: And would rename event and storage accordingly.

src/common/MultiAdapter.sol
```diff
@@ -211,23 +220,39 @@ contract MultiAdapter is Auth, IMultiAdapter {
...
+    }
+
+    //----------------------------------------------------------------------------------------------
+    // Getters
+    //----------------------------------------------------------------------------------------------
+
+    /// @inheritdoc IMultiAdapter
+    function quorum(uint16 centrifugeId, PoolId poolId) external view returns (uint8) {
+        Adapter memory adapter = _adapterDetails[centrifugeId][poolId][adapters[centrifugeId][poolId][0]];
         return adapter.quorum;
```
hieronx: Bit more readable:
```suggestion
        IAdapter adapter = adapters[centrifugeId][poolId][0];
        return _adapterDetails[centrifugeId][poolId][adapter].quorum;
```

hieronx: If you agree, would also change `activeSessionId`

src/common/MultiAdapter.sol
```diff
@@ -179,26 +180,34 @@ contract MultiAdapter is Auth, IMultiAdapter {
     //----------------------------------------------------------------------------------------------
 
     /// @inheritdoc IAdapter
-    function send(uint16 centrifugeId, bytes calldata payload, uint256 gasLimit, address refund)
+    function send(uint16 centrifugeId, bytes memory payload, uint256 gasLimit, address refund)
```
hieronx, line 183: Why this change?

src/common/MultiAdapter.sol
```diff
@@ -179,26 +180,34 @@ contract MultiAdapter is Auth, IMultiAdapter {
...
 
         bytes32 payloadHash = keccak256(payload);
         bytes32 payloadId = keccak256(abi.encodePacked(localCentrifugeId, centrifugeId, payloadHash));
-        bytes memory proof = payloadHash.serializeMessageProof();
 
         uint256 cost = adapters_[0].estimate(centrifugeId, payload, gasLimit);
         bytes32 adapterData = adapters_[0].send{value: cost}(centrifugeId, payload, gasLimit, refund);
         emit SendPayload(centrifugeId, payloadId, payload, adapters_[0], adapterData, refund);
 
+        // payload is now the proof (to avoid stack too deep issue)
```
hieronx: Took me a while to understand, maybe:
```suggestion
        // Override the payload variable to send the proof to the remaining adapters
```

src/common/MultiAdapter.sol
```diff
@@ -211,23 +220,39 @@ contract MultiAdapter is Auth, IMultiAdapter {
         view
         returns (uint256 total)
     {
-        IAdapter[] memory adapters_ = adapters[centrifugeId];
-        bytes memory proof = keccak256(payload).serializeMessageProof();
+        PoolId poolId = messageProperties.messagePoolId(payload);
+        IAdapter[] memory adapters_ = adapters[centrifugeId][poolId];
+
+        // If adapters not configured per pool, then use the global adapters
+        if (adapters_.length == 0) adapters_ = adapters[centrifugeId][GLOBAL_ID];
```
hieronx: Used twice, might be worth adding something like
```solidity
function adapters(uint16 centrifugeId, PoolId poolId) internal returns (IAdapter[] memory adapters_) {
        adapters_ = adapters[centrifugeId][poolId];

        // If adapters not configured per pool, then use the global adapters
        if (adapters_.length == 0) adapters_ = adapters[centrifugeId][GLOBAL_ID];
}
```

src/common/Gateway.sol
```diff
@@ -177,28 +188,29 @@ contract Gateway is Auth, Recoverable, IGateway {
...
-            transactionRefund != address(0) ? transactionRefund : address(subsidy[poolId].refund)
+            transactionRefund != address(0) ? transactionRefund : address(subsidy[paymentPoolId].refund)
         );
 
         return true;
     }
 
     /// @inheritdoc IGateway
-    function addUnpaidMessage(uint16 centrifugeId, bytes memory message) external auth {
+    function addUnpaidMessage(uint16 centrifugeId, bytes memory message, bool isSubsidized) external auth {
```
hieronx, line 207: When would `isSubsidized` ever be `true`?

src/common/Gateway.sol
```diff
@@ -152,23 +152,34 @@ contract Gateway is Auth, Recoverable, IGateway {
     }
 
     function _send(uint16 centrifugeId, bytes memory batch, uint128 batchGasLimit) internal returns (bool succeeded) {
-        PoolId poolId = processor.messagePoolIdPayment(batch);
+        PoolId paymentPoolId = processor.messagePoolIdPayment(batch);
+        PoolId adapterPoolId = processor.messagePoolId(batch);
```
hieronx: Ultra-aesthetics nit:
```suggestion
        PoolId adapterPoolId = processor.messagePoolId(batch);
        PoolId paymentPoolId = processor.messagePoolIdPayment(batch);
```

hieronx (commented):
Looks already really good!

src/hub/Hub.sol
lemunozm: Makes sense

src/hub/Hub.sol
lemunozm: We could add another Hub method for setting an adapter's manager. I feel like this manager should be the same as the pool manager in the same chain. The only reason we allow set a new manager is to set one in a remote chain where the the pool manager doesn't have access-

src/common/MultiAdapter.sol
lemunozm: A couple of notes:
1. `setX`, seems like you need to call the method in order to set something, but in this case, by default, it allows sending the messages
2. If I read `setOutgoing`, I would not know what it means and when it should be called.

Maybe, given 1. it makes sense to name it in the reverse way, which is more explanatory. i.e. `blockOutgoing`

src/common/MultiAdapter.sol
lemunozm: Ok 👍🏻

src/common/MultiAdapter.sol
lemunozm: I was getting stack too deep, and I think the less invasive is to use bytes `memory`, and reuse the payload for proofs. Other way is use `keccak256(payload)`, but this means iterate and compute the hash of all bytes of the message several times.

Happy if you find something cleaner!

src/common/MultiAdapter.sol
lemunozm: I really like it!

src/common/Gateway.sol
lemunozm: 2 cases:

- A subsidized message that can not be sent because there is no subsidized funds.
- Any message that can not be sent because sending is blocked.

src/common/MultiAdapter.sol
hieronx: Yeah I think this is the least worst option!

src/common/MultiAdapter.sol
hieronx: Yeah I think `blockOutgoing(.., bool isEnabled)` could work!

src/common/MultiAdapter.sol
```diff
@@ -211,24 +217,43 @@ contract MultiAdapter is Auth, IMultiAdapter {
...
+        require(msg.sender == manager[poolId], ManagerNotAllowed());
+        isSendingBlocked[centrifugeId][poolId] = !canSend;
+        emit EnableSending(centrifugeId, poolId, canSend);
+    }
+
+    //----------------------------------------------------------------------------------------------
+    // Getters
+    //----------------------------------------------------------------------------------------------
+
+    function _poolAdapters(uint16 centrifugeId, PoolId poolId) internal view returns (IAdapter[] memory adapters_) {
```
hieronx: My bad but would change this to external. And should add docs.
```suggestion
    /// @inheritdoc IMultiAdapter
    function poolAdapters(uint16 centrifugeId, PoolId poolId) external view returns (IAdapter[] memory adapters_) {
```

src/hub/Hub.sol
hieronx: I think this has two problems:
- It is now not updateable
- It is mapping between hub manager and adapter manager, but this is confusing since you can have multiple hub managers, but only 1 adapter manager

lemunozm: Ok, the second point makes a lot of sense! I will add another method to the hub

lemunozm:
Reminder for myself. I need to add the correct gas values in the `GasService` before merging this

src/common/MultiAdapter.sol
```diff
@@ -211,24 +217,43 @@ contract MultiAdapter is Auth, IMultiAdapter {
...
         }
     }
 
     /// @inheritdoc IMultiAdapter
-    function quorum(uint16 centrifugeId) external view returns (uint8) {
-        Adapter memory adapter = _adapterDetails[centrifugeId][adapters[centrifugeId][0]];
-        return adapter.quorum;
+    function blockOutgoing(uint16 centrifugeId, PoolId poolId, bool isBlocked) external {
+        require(msg.sender == manager[poolId], ManagerNotAllowed());
+        isSendingBlocked[centrifugeId][poolId] = isBlocked;
```
hieronx:
```suggestion
        isOutgoingBlocked[centrifugeId][poolId] = isBlocked;
```

src/common/MultiAdapter.sol
```diff
@@ -19,21 +21,28 @@ contract MultiAdapter is Auth, IMultiAdapter {
...
     IMessageHandler public gateway;
+    IMessageProperties public messageProperties;
 
-    mapping(uint16 centrifugeId => IAdapter[]) public adapters;
-    mapping(uint16 centrifugeId => mapping(IAdapter adapter => Adapter)) internal _adapterDetails;
+    mapping(PoolId => address) public manager;
+    mapping(uint16 centrifugeId => mapping(PoolId => bool)) public isSendingBlocked;
+    mapping(uint16 centrifugeId => mapping(PoolId => IAdapter[])) public adapters;
+    mapping(uint16 centrifugeId => mapping(PoolId => mapping(IAdapter adapter => Adapter))) internal _adapterDetails;
     mapping(uint16 centrifugeId => mapping(bytes32 payloadHash => Inbound)) public inbound;
```
hieronx: Ordering nit:
```suggestion
    mapping(uint16 centrifugeId => mapping(PoolId => IAdapter[])) public adapters;
mapping(uint16 centrifugeId => mapping(bytes32 payloadHash => Inbound)) public inbound;
    mapping(uint16 centrifugeId => mapping(PoolId => mapping(IAdapter adapter => Adapter))) internal _adapterDetails;
```

src/common/Gateway.sol
hieronx: In both cases though, the internal `_addUnpaidBatch` is called rather than the external `addUnpaidMessage`, right?

src/common/Gateway.sol
```diff
@@ -207,11 +219,17 @@ contract Gateway is Auth, Recoverable, IGateway {
...
 
-        PoolId poolId = processor.messagePoolIdPayment(batch);
-        if (msg.value > 0) subsidizePool(poolId);
-
         underpaid_.counter--;
-        require(_send(centrifugeId, batch, underpaid_.gasLimit), InsufficientFundsForRepayment());
+
+        if (!underpaid_.isSubsidized) {
+            _startTransactionPayment(msg.sender);
+        }
```
hieronx: I think this is more readable as one liners:
```suggestion
        if (!underpaid_.isSubsidized) _startTransactionPayment(msg.sender);
```

src/common/Gateway.sol
```diff
@@ -207,11 +219,17 @@ contract Gateway is Auth, Recoverable, IGateway {
...
+
+        require(_send(centrifugeId, batch, underpaid_.gasLimit), CannotBeRepaid());
+
+        if (!underpaid_.isSubsidized) {
+            _endTransactionPayment();
+        }
```
```suggestion
        if (!underpaid_.isSubsidized) _endTransactionPayment();
```

src/common/Gateway.sol
```diff
@@ -247,23 +265,31 @@ contract Gateway is Auth, Recoverable, IGateway {
...
 
     function _subsidizePool(PoolId poolId, address who, uint256 value) internal {
+        require(address(subsidy[poolId].refund) != address(0), RefundAddressNotSet());
         subsidy[poolId].value += value.toUint96();
         emit SubsidizePool(poolId, who, value);
     }
 
     /// @inheritdoc IGateway
-    function startTransactionPayment(address payer) external payable auth {
+    function startTransactionPayment(address payer) public payable auth {
```
hieronx: Why are these public rather than external, despite the internal functions all calling `_start/_end`?

src/common/Gateway.sol
lemunozm: For the case `isSubsidized == true` yes, only from the external one we call it with `false`. But I think it's valuable to be able to configure it from the external one, in case tomorrow we have some message that needs to be repaid with the subsidized funds instead.

src/common/Gateway.sol
lemunozm: Correct, they can be external 👍🏻

src/common/MultiAdapter.sol
lemunozm: Right!

src/common/Gateway.sol
hieronx: Fair point!

src/hub/Hub.sol
lemunozm: I think the less invasive way is to use `setAdaptersManager` with `localCentrifugeId` support: [6cc3d38](https://github.com/centrifuge/protocol-v3/pull/570/commits/6cc3d38de50a93374d7fc205d521fdd18a3c7963)

PR #573, opened 2025-08-13
chore: informational deployer improvements

wischli:
Fixes typos and improves debugging of deployments


hieronx (approved):

script/FullDeployer.s.sol
```diff
@@ -88,7 +90,8 @@ contract FullDeployer is ExtendedHubDeployer, ExtendedSpokeDeployer, AdaptersDep
             centrifugeId: centrifugeId,
             adminSafe: ISafe(vm.envAddress("ADMIN")),
             maxBatchGasLimit: uint128(maxBatchGasLimit),
-            version: keccak256(abi.encodePacked(vm.envOr("VERSION", string(""))))
+            // Default to bytes32(0) version to be consistent with CommonDeployer
+            version: bytes(versionString).length > 0 ? keccak256(abi.encodePacked(versionString)) : bytes32(0)
```
lemunozm, line 94: I think this is not need it? If `versionString` is length == 0, then it's the same as `bytes32(0)`, right?

test/adapters/Deployment.t.sol
```diff
@@ -32,9 +32,13 @@ contract AdaptersDeploymentInputTest is Test {
 }
 
 contract AdaptersDeploymentTest is AdaptersDeployer, CommonDeploymentInputTest, AdaptersDeploymentInputTest {
+    /// @dev Simple contract bytecode for mocking deployed contracts
+    bytes constant SIMPLE_CONTRACT = hex"6001600160005260206000f3";
```
lemunozm, line 36: Curious, from where does this number come?

wischli: Basically a random hex to fulfill `_validate*Input`.

script/FullDeployer.s.sol
wischli: AFAICT, Not quite. If versionString is empty, then
```solidity
keccak256(abi.encodePacked(versionString))
```

is not `bytes32(0)`. It’s the keccak256 of the empty byte string:
```solidity
0xc5d2460186f7233c927e7db2dcc703c0e500b653ca82273b7bfad8045d85a470
```

So without the conditional we’d store that non-zero constant for the “no version provided” case, which breaks our intent to use `bytes32(0)` as the sentinel (and to stay consistent with `CommonDeployer`).

lemunozm: Correct! Thanks for the explanation

PR #576, opened 2025-08-18
tests: refactor on-offrampmanager

wischli:
Created pure unit tests for `OnOfframpManager` that mock all dependencies instead of using the heavyweight BaseTest integration setup. Replaced real contract deployments with `vm.mockCall()`, added missing test cases to hit 100% coverage, and kept a simplified integration test with just the happy path. 

Ref: https://www.notion.so/Nice-to-have-On-offramp-Manager-test-refactor-20e2eac24e1780c4b66ace1b3faddef7?source=copy_link


hieronx (approved):
Thanks for the rewrite and cleanup!

lemunozm:
Love it! Thanks for simplifying this!

PR #577, opened 2025-08-18
Add adapter

hieronx:


wischli (approved):
Disregard this comment. Tried out something new and it failed :D

src/adapters/interfaces/IWormholeAdapter.sol
```diff
@@ -134,6 +134,8 @@ struct WormholeDestination {
 }
 
 interface IWormholeAdapter is IAdapter, IWormholeReceiver {
+    event Wire(uint16 indexed centrifugeId, uint16 indexed wormholeId, address adapter);
```
wischli, line 137: Great addition!

src/adapters/LayerZeroAdapter.sol
```diff
@@ -0,0 +1,149 @@
...
+    /// @inheritdoc ILayerZeroAdapter
+    function wire(uint16 centrifugeId, uint32 layerZeroEid, address adapter) external auth {
+        sources[layerZeroEid] = LayerZeroSource(centrifugeId, adapter);
+        destinations[centrifugeId] = LayerZeroDestination(layerZeroEid, adapter);
+        emit Wire(centrifugeId, layerZeroEid, adapter);
+    }
+
+    /// @dev Update the LayerZero delegate.
+    function setDelegate(address newDelegate) external auth {
+        endpoint.setDelegate(newDelegate);
```
wischli, line 60: Q: Should we add a sanity check for non-zero or self-address?

src/adapters/LayerZeroAdapter.sol
```diff
@@ -0,0 +1,149 @@
...
+
+    /// @inheritdoc ILayerZeroReceiver
+    function allowInitializePath(Origin calldata origin) external view override returns (bool) {
+        LayerZeroSource memory source = sources[origin.srcEid];
+        return source.addr != address(0) && source.addr == origin.sender.toAddressLeftPadded();
+    }
+
+    /// @inheritdoc ILayerZeroReceiver
+    function nextNonce(uint32, bytes32) external pure override returns (uint64 nonce) {
+        return 0;
```
wischli, line 88: Just to be sure: Based on the `ILayerZero` comment ([ref](https://github.com/centrifuge/protocol-v3/pull/577/files#diff-629942bad3c42e72a938c9c5f3b6038fdae9bf54ba343cf6fad751ffdd2eb9beR48)), we return 0 here to not enforce nonce ordering, right?

src/adapters/LayerZeroAdapter.sol
```diff
@@ -0,0 +1,149 @@
...
+        );
+    }
+
+    //----------------------------------------------------------------------------------------------
+    // Options builder
+    //----------------------------------------------------------------------------------------------
+
+    uint16 internal constant TYPE_3 = 3;
+    uint8 internal constant WORKER_ID = 1;
+    uint8 internal constant OPTION_TYPE_LZRECEIVE = 1;
```
wischli, line 138: Nit: Not a fan of storage values outside the header area

wischli, line 138: Since these values were opaque to me, I looked into the LayerZero-v2 ExecutorOptions to confirm these values for the most recent `TYPE_3 = 3`: https://github.com/LayerZero-Labs/LayerZero-v2/blob/200cda254120375f40ed0a7e89931afb897b8891/packages/layerzero-v2/evm/messagelib/contracts/libs/ExecutorOptions.sol#L10-L12

test/adapters/unit/LayerZeroAdapter.t.sol
```diff
@@ -0,0 +1,203 @@
...
+
+        return MessagingReceipt(bytes32(""), 0, MessagingFee(0, 0));
+    }
+
+    function setDelegate(address newDelegate) external {
+        values_address["delegate"] = newDelegate;
+    }
+
+    function quote(MessagingParams calldata, address) external pure returns (MessagingFee memory) {
+        // TODO: decode gas limit from params.options
```
wischli: IIUC, we could extract the gas limit from `options[6:]` right?

test/adapters/unit/LayerZeroAdapter.t.sol
```diff
@@ -0,0 +1,203 @@
...
+    address immutable newDelegate = makeAddr("newDelegate");
+
+    function testSetDelegateErrNotAuthorized() public {
+        vm.prank(makeAddr("NotAuthorized"));
+        vm.expectRevert(IAuth.NotAuthorized.selector);
+        adapter.setDelegate(newDelegate);
+    }
+
+    function testSetDelegate() public {
+        adapter.setDelegate(newDelegate);
```
wischli, line 105: Nit: Add event emission check

```suggestion
        vm.expectEmit();
        emit ILayerZeroAdapter.SetDelegate(newDelegate);
        adapter.setDelegate(newDelegate);
```

test/adapters/unit/LayerZeroAdapter.t.sol
```diff
@@ -0,0 +1,203 @@
...
+
+contract LayerZeroAdapterTestWire is LayerZeroAdapterTestBase {
+    function testWireErrNotAuthorized() public {
+        vm.prank(makeAddr("NotAuthorized"));
+        vm.expectRevert(IAuth.NotAuthorized.selector);
+        adapter.wire(CENTRIFUGE_ID, LAYERZERO_ID, REMOTE_LAYERZERO_ADDR);
+    }
+
+    function testWire() public {
+        adapter.wire(CENTRIFUGE_ID, LAYERZERO_ID, REMOTE_LAYERZERO_ADDR);
```
wischli, line 81: Nit: Add event emission check

```suggestion
        vm.expectEmit();
        emit ILayerZeroAdapter.Wire(CENTRIFUGE_ID, LAYERZERO_ID, REMOTE_LAYERZERO_ADDR);
        adapter.wire(CENTRIFUGE_ID, LAYERZERO_ID, REMOTE_LAYERZERO_ADDR);
```

test/adapters/unit/LayerZeroAdapter.t.sol
```diff
@@ -0,0 +1,203 @@
...
+        adapter.wire(CENTRIFUGE_ID, LAYERZERO_ID, makeAddr("DestinationAdapter"));
+
+        vm.deal(address(this), 0.1 ether);
+        vm.prank(address(GATEWAY));
+        adapter.send{value: 0.1 ether}(CENTRIFUGE_ID, payload, gasLimit, refund);
+
+        assertEq(endpoint.values_uint32("params.dstEid"), LAYERZERO_ID);
+        assertEq(endpoint.values_bytes32("params.receiver"), makeAddr("DestinationAdapter").toBytes32LeftPadded());
+        assertEq(endpoint.values_bytes("params.message"), payload);
+        // assertEq(endpoint.values_bytes("params.options"), ""); // TODO
```
wischli: Here it is 

```suggestion
        bytes memory expectedOptions = abi.encodePacked(
            uint16(3), // TYPE_3
            uint8(1), // WORKER_ID
            uint16(17), // uint128 gasLimit byte length + 1
            uint8(1), // OPTION_TYPE_LZ
            uint128(gasLimit)
        );
        assertEq(endpoint.values_bytes("params.options"), expectedOptions);
```

wischli (dismissed):
Just a bunch of post-approval questions, minor nits and test additions. Great job!

src/adapters/LayerZeroAdapter.sol
hieronx: We don't do it anywhere else (e.g. rely/deny), so wouldn't no.

src/adapters/LayerZeroAdapter.sol
hieronx: Indeed

src/adapters/LayerZeroAdapter.sol
hieronx: Personally think this is a case of making an exception since it works better in this specific case. These aren't really storage variables, just configuration vars that only apply to the options. For that reason to me it seems more logical to have it grouped together.

test/adapters/unit/LayerZeroAdapter.t.sol
hieronx: Yeah

src/adapters/interfaces/IWormholeAdapter.sol
lemunozm: Agree! my fault actually don't adding this!

src/adapters/LayerZeroAdapter.sol
```diff
@@ -0,0 +1,152 @@
...
+
+    mapping(uint32 layerZeroEid => LayerZeroSource) public sources;
+    mapping(uint16 centrifugeId => LayerZeroDestination) public destinations;
+
+    constructor(IMessageHandler entrypoint_, address endpoint_, address delegate, address deployer) Auth(deployer) {
+        entrypoint = entrypoint_;
+        endpoint = ILayerZeroEndpointV2(endpoint_);
+
+        endpoint.setDelegate(delegate);
+        emit SetDelegate(delegate);
```
lemunozm, line 44: NIT, maybe we can call to `setDelegate()` directly

src/adapters/LayerZeroAdapter.sol
```diff
@@ -0,0 +1,152 @@
...
+/// @notice Routing contract that integrates with LayerZero V2.
+/// @dev    A delegate is set on deployment, to configure the DVN and executor
+///         settings as well as the send/receive libraries.
+///
+///         Message ordering is not enforced.
+contract LayerZeroAdapter is Auth, ILayerZeroAdapter {
+    using CastLib for *;
+    using MathLib for *;
+
+    IMessageHandler public immutable entrypoint;
```
lemunozm, line 33: Maybe we want to set this as mutable in order to avoid migrating this in v3.1

src/adapters/LayerZeroAdapter.sol
```diff
@@ -0,0 +1,152 @@
...
+
+        MessagingReceipt memory receipt =
+            endpoint.send{value: msg.value}(_params(destination, payload, gasLimit), refund);
+        adapterData = receipt.guid;
+    }
+
+    /// @inheritdoc IAdapter
+    function estimate(uint16 centrifugeId, bytes calldata payload, uint256 gasLimit) external view returns (uint256) {
+        LayerZeroDestination memory destination = destinations[centrifugeId];
+        MessagingFee memory fee = endpoint.quote(_params(destination, payload, gasLimit), address(this));
```
lemunozm, line 113: Could the estimation fail because `address(this)` is not payable?

Maybe we should pass the `refund` address to this method too, to correctly estimate everything. Refunding will add some cost

src/adapters/LayerZeroAdapter.sol
lemunozm: I understand what you want is name those "random" values. It's true it looks a bit different. Maybe we can just add variables to the method below if only used there.

test/adapters/unit/LayerZeroAdapter.t.sol
```diff
@@ -0,0 +1,217 @@
...
+
+        return MessagingReceipt(bytes32(""), 0, MessagingFee(0, 0));
+    }
+
+    function setDelegate(address newDelegate) external {
+        values_address["delegate"] = newDelegate;
+    }
+
+    function quote(MessagingParams calldata, address) external pure returns (MessagingFee memory) {
+        // TODO: decode gas limit from params.options
```
lemunozm: A reminder of this TODO

lemunozm (commented):
Looks good! Pretty similar scheme! Just some few comment, mostly NITs

src/adapters/LayerZeroAdapter.sol
hieronx: This doesn't work, similar as https://github.com/centrifuge/protocol-v3/blob/main/src/common/Gateway.sol#L58, deployer is not auth

src/adapters/LayerZeroAdapter.sol
hieronx: All adapters have this now as immutable so would keep as is for now. Something to reconsider for v3.1 upgrade yeah

src/adapters/LayerZeroAdapter.sol
hieronx: Since this is already audited, would not change this now. But yes could have been variables in the function too.

test/adapters/unit/LayerZeroAdapter.t.sol
hieronx: Yeah I will remove this TODO. It is a bit complex, and not really much value. More valuable to add the integration tests that our auditors wrote to the codebase (https://gist.github.com/gvladika/ad74ef148840879b0b60fba96278a50e)

src/adapters/LayerZeroAdapter.sol
hieronx: The last param here is not the refund address, but the sender. See https://github.com/centrifuge/protocol-v3/blob/layerzero-adapter/src/adapters/interfaces/ILayerZeroAdapter.sol#L43

test/adapters/unit/LayerZeroAdapter.t.sol
hieronx: Tracking ticket: https://github.com/centrifuge/protocol-v3/issues/583

wischli (approved):

lemunozm:
As a note, the configuration is missing in `WireAdapters.s.sol`

PR #578, opened 2025-08-20
fix: hook `isRedeemClaim` classification instead of `isRedeemFulfillment `

wischli:
Fixes `BaseTransferHook` classification bug in redemption flow: Currently, `isRedeemClaim` is triggered during `fulfilledRedeemRequest` when it should be `isRedeemFulfillment`.  

Based on my analysis, the root cause is that the `redeemSource` constructor parameter was incorrectly set to `AsyncRequestManager` instead of `BalanceSheet`.  I could not identify a non-auth transfer of shares with `from == AsyncRequestManager`. However, during `revokeShares()`, shares are transferred from the `BalanceSheet` to `address(0)` indicating a burn. Since `AsyncRequestManager` never directly burns shares visible to the hook, `BalanceSheet` should be the `redeemSource` IMO. As a solution, I updated `HooksDeployer.s.sol` to use `address(balanceSheet)` instead of `address(asyncRequestManager)` as the `redeemSource` parameter.  

## Share Token Transfer Flows

For a quick recap, here is the share token flow for async investments. The critical part is the callback from `hub.revokeShares()` at the bottom.

### Deposit Flow

1. `hub.issueShares()` callback  
   - Triggers `AsyncRequestManager.issuedShares()`  
   - `address(0) → globalEscrow` (mint shares)
   - Classification: `isDepositFulfillment`  

2. `vault.mint(maxMint, investor)`  
   - Triggers `AsyncRequestManager.mint() → _processDeposit()`  
   - `globalEscrow → investor`
   - Classification: `isDepositClaim`  

### Redeem Flow

1. `vault.requestRedeem(shares, controller, investor)`  
   - Triggers `AsyncRequestManager.requestRedeem()` with `transfer = true`  
   - `investor → globalEscrow` (via `balanceSheet.transferSharesFrom()`)
   - Classification: `isRedeemRequest`

2. `hub.revokeShares()` callback  
   - Triggers `AsyncRequestManager.revokedShares()`  
   - Three transfers in sequence:  
     1. `globalEscrow → asyncRequestManager` (`globalEscrow.authTransferTo()`)  
        - Classification: `isRedeemClaim` (transfer, not burn)  
     2. `asyncRequestManager → balanceSheet` (`balanceSheet.revoke() → internal authTransferFrom()`)  
        - ❗**No classification due to bypassing hook check `onERC20Transfer`** due to calling `hook.onERC20AuthTransfer` ([ref](https://github.com/centrifuge/protocol-v3/blob/42d9223d7cdf7262ce7628df7d267e9ade81f351/src/spoke/ShareToken.sol#L127-L133))
     3. `balanceSheet → address(0)` (burn inside `BalanceSheet.revoke()`)  
        - Classification: `isRedeemFulfillment` (the **current bug case** if `redeemSource == asyncRequestManager` wrong)

### DIFF
--- a/test/hooks/integration/BaseTransferHook.t.sol
+++ b/test/hooks/integration/BaseTransferHook.t.sol
@@ -1,0 +1,173 @@
+// SPDX-License-Identifier: BUSL-1.1
+pragma solidity 0.8.28;
+
+import {ESCROW_HOOK_ID} from "../../../src/common/interfaces/ITransferHook.sol";
+
+import {FullRestrictions} from "../../../src/hooks/FullRestrictions.sol";
+
+import {FullDeployer, FullActionBatcher, CommonInput} from "../../../script/FullDeployer.s.sol";
+
+import "forge-std/Test.sol";
+
+import {IntegrationConstants} from "../../integration/utils/IntegrationConstants.sol";
+
+contract BaseTransferHookIntegrationTest is FullDeployer, Test {
+    uint16 constant LOCAL_CENTRIFUGE_ID = IntegrationConstants.LOCAL_CENTRIFUGE_ID;
+    address immutable ADMIN = address(adminSafe);
+    uint256 constant GAS = IntegrationConstants.INTEGRATION_DEFAULT_SUBSIDY;
+    address constant USER = address(0x1234);
+
+    FullRestrictions public correctHook;
+    FullRestrictions public wrongHook;
+
+    function setUp() public {
+        CommonInput memory input = CommonInput({
+            centrifugeId: LOCAL_CENTRIFUGE_ID,
+            adminSafe: adminSafe,
+            maxBatchGasLimit: uint128(GAS) * 100,
+            version: bytes32(0)
+        });
+
+        FullActionBatcher batcher = new FullActionBatcher();
+        super.labelAddresses("");
+        super.deployFull(input, noAdaptersInput(), batcher);
+        super.removeHubDeployerAccess(batcher);
+
+        vm.startPrank(ADMIN);
+        correctHook =
+            new FullRestrictions(address(root), address(balanceSheet), address(globalEscrow), address(spoke), ADMIN);
+        wrongHook = new FullRestrictions(
+            address(root), address(asyncRequestManager), address(globalEscrow), address(spoke), ADMIN
+        );
+        vm.stopPrank();
+    }
+
+    function testBalanceSheetBurns() public view {
+        assertTrue(
+            correctHook.isRedeemFulfillment(address(balanceSheet), address(0)), "balanceSheet burn is fulfillment"
+        );
+        assertFalse(correctHook.isRedeemClaim(address(balanceSheet), address(0)), "balanceSheet burn not claim");
+
+        assertFalse(
+            wrongHook.isRedeemFulfillment(address(balanceSheet), address(0)), "wrong hook: balanceSheet not fulfillment"
+        );
+        assertTrue(wrongHook.isRedeemClaim(address(balanceSheet), address(0)), "wrong hook: balanceSheet is claim");
+    }
+
+    function testAsyncRequestManagerBurns() public view {
+        assertFalse(
+            correctHook.isRedeemFulfillment(address(asyncRequestManager), address(0)),
+            "asyncRequestManager not fulfillment"
+        );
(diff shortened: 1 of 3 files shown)


test/hooks/integration/BaseTransferHook.t.sol
```diff
@@ -0,0 +1,173 @@
...
+    function testBalanceSheetBurns() public view {
+        assertTrue(
+            correctHook.isRedeemFulfillment(address(balanceSheet), address(0)), "balanceSheet burn is fulfillment"
+        );
+        assertFalse(correctHook.isRedeemClaim(address(balanceSheet), address(0)), "balanceSheet burn not claim");
+
+        assertFalse(
+            wrongHook.isRedeemFulfillment(address(balanceSheet), address(0)), "wrong hook: balanceSheet not fulfillment"
+        );
+        assertTrue(wrongHook.isRedeemClaim(address(balanceSheet), address(0)), "wrong hook: balanceSheet is claim");
```
hieronx, line 54: We should possibly rename `isRedeemClaim` to `isRevocation`, and `isDepositRequest` to `isIssuance`, since it is impossible to separate `balanceSheet.issue` from deposit requests, and `balanceSheet.revoke` from redeem claims. Wdyt?

hieronx (approved):

test/hooks/integration/BaseTransferHook.t.sol
wischli: Agree! Will do in follow-up. Wanted to keep this PR's non-test diff as low as possible

test/hooks/integration/BaseTransferHook.t.sol
```diff
@@ -0,0 +1,173 @@
...
+
+import {FullRestrictions} from "../../../src/hooks/FullRestrictions.sol";
+
+import {FullDeployer, FullActionBatcher, CommonInput} from "../../../script/FullDeployer.s.sol";
+
+import "forge-std/Test.sol";
+
+import {IntegrationConstants} from "../../integration/utils/IntegrationConstants.sol";
+
+contract BaseTransferHookIntegrationTest is FullDeployer, Test {
```
lemunozm, line 14: Curious, why are these tests integration tests?

test/hooks/integration/BaseTransferHook.t.sol
```diff
@@ -0,0 +1,173 @@
...
+        super.labelAddresses("");
+        super.deployFull(input, noAdaptersInput(), batcher);
+        super.removeHubDeployerAccess(batcher);
+
+        vm.startPrank(ADMIN);
+        correctHook =
+            new FullRestrictions(address(root), address(balanceSheet), address(globalEscrow), address(spoke), ADMIN);
+        wrongHook = new FullRestrictions(
+            address(root), address(asyncRequestManager), address(globalEscrow), address(spoke), ADMIN
+        );
```
lemunozm, line 41: Should not be enough to only perform the tests over the `correctHook`? I mean, once found the issue, we don't need to test/maintain the previous error

lemunozm (commented):
AFAIU, after your good explanation. I agree with the solution!

Just an informative question, why here:

> - Three transfers in sequence:  
>     1. `globalEscrow → asyncRequestManager` (`globalEscrow.authTransferTo()`)  
>        - Classification: `isRedeemClaim` (transfer, not burn) 

the classification is `isRedeemClaim` if we're just revoking?

test/hooks/integration/BaseTransferHook.t.sol
wischli: Yeah, naming should be updated: https://github.com/centrifuge/protocol-v3/pull/578#discussion_r2288783957

wischli: https://github.com/centrifuge/protocol-v3/issues/586

test/hooks/integration/BaseTransferHook.t.sol
wischli: In general, I agree. No strong opinion. I worked on them in tandem and thought its better to keep for more coverage and to ensure any other refactor pre v3.1 launch cannot reintroduce silently.

lemunozm: Yeah, I see the value as a proof that the fix is valid, but maybe if in the future we need to refactor this or a new programmer reads this, it adds more noise. If someone reintroduces the bug, then the `correctHook` will fail, so for me it seems it suffices

PR #580, opened 2025-08-21
Feat: QueueManager

onnovisser:
QueueManager to submit queued assets and shares. Permissionless by design.


src/managers/QueueManager.sol
```diff
@@ -0,0 +1,109 @@
...
+    function sync(PoolId poolId, ShareClassId scId, AssetId[] calldata assetIds) external {
+        ShareClassState storage sc = shareClassState[poolId][scId];
+        require(
+            sc.lastSync == 0 || sc.minDelay == 0 || block.timestamp >= sc.lastSync + sc.minDelay, MinDelayNotElapsed()
+        );
+
+        (uint128 delta,, uint32 queuedAssetCounter,) = balanceSheet.queuedShares(poolId, scId);
+        require(delta > 0 || queuedAssetCounter > 0, NoUpdates());
+
+        bool maybeSubmitShares = delta > 0 && assetIds.length >= queuedAssetCounter;
```
lemunozm: Should not be || instead? At least if `delta > 0` (independently of the queuedAssetCounter), we want to submitQueuedShares IIUC, but maybe I'm wrong in this understanding

src/managers/QueueManager.sol
```diff
@@ -0,0 +1,109 @@
...
+                TransientStorageLib.tloadBool(keccak256(abi.encode("assetSeen", assetIds[i]))) == false,
+                DuplicateAsset()
+            );
+            (uint128 deposits, uint128 withdrawals) = balanceSheet.queuedAssets(poolId, scId, assetIds[i]);
+            if (deposits == 0 && withdrawals == 0) {
+                // noop by just reading a property
+                // doing a noop call instead of reverting, to prevent DoS
+                cs[i] = abi.encodeWithSelector(balanceSheet.root.selector);
+            } else {
+                cs[i] = abi.encodeWithSelector(balanceSheet.submitQueuedAssets.selector, poolId, scId, assetIds[i], 0);
```
lemunozm: We should add here extraGasLimit

src/managers/QueueManager.sol
```diff
@@ -0,0 +1,109 @@
...
+            }
+            TransientStorageLib.tstore(keccak256(abi.encode("assetSeen", assetIds[i])), true);
+        }
+
+        if (maybeSubmitShares) {
+            if (actualSubmissions < queuedAssetCounter) {
+                // We didn't actually submit all queued assets, so don't submit shares
+                cs[submissions] = abi.encodeWithSelector(balanceSheet.root.selector);
+            } else {
+                cs[submissions] = abi.encodeWithSelector(balanceSheet.submitQueuedShares.selector, poolId, scId, 0);
```
lemunozm: and here

src/managers/QueueManager.sol
lemunozm: The user from the app, should at least satisfy `(submissions + 1) * extraGasLimit` as gas to ensure the worst case computation

src/managers/QueueManager.sol
```diff
@@ -0,0 +1,109 @@
...
+        for (uint256 i; i < submissions; i++) {
+            require(
+                TransientStorageLib.tloadBool(keccak256(abi.encode("assetSeen", assetIds[i]))) == false,
+                DuplicateAsset()
+            );
+            (uint128 deposits, uint128 withdrawals) = balanceSheet.queuedAssets(poolId, scId, assetIds[i]);
+            if (deposits == 0 && withdrawals == 0) {
+                // noop by just reading a property
+                // doing a noop call instead of reverting, to prevent DoS
+                cs[i] = abi.encodeWithSelector(balanceSheet.root.selector);
```
lemunozm: I think we can get rid of noops if you do a slice of the `cs` as `cs[0:activeSubmitShares+1]`. Although you need to recreate a smaller `cs` array in memory to do so (and a minor helper), it should be more efficient than calling several `root()` methods.

lemunozm (commented):
Good job! Agree, the solution is not easy, but both the noop thing and the subsidized stuff complicate this a lot, so not sure if it can be simplified much more. I left an idea for the noop thing

src/managers/QueueManager.sol
```diff
@@ -0,0 +1,113 @@
...
+
+/// @dev minDelay can be set to a non-zero value, for cases where assets or shares can be permissionlessly modified
+///      (e.g. if the on/off ramp manager is used, or if sync deposits are enabled). This prevents spam.
+contract QueueManager is IQueueManager, IUpdateContract {
+    using CastLib for *;
+
+    address public immutable contractUpdater;
+    IBalanceSheet public immutable balanceSheet;
+
+    struct ShareClassState {
```
wischli: Ultra nit: The name is quite generic. Not sure if we will use this struct outside the contract anywhere. If so, I'd prefer adding some context to it like `ShareClassQueueState` or `ScQueueState`

wischli: Another nit: Could we move this to `IQueueManager` or at least above the `immutable` storage declaration in this contract to keep all storage declarations on a block? Thanks!

src/managers/QueueManager.sol
```diff
@@ -0,0 +1,113 @@
...
+    }
+
+    //----------------------------------------------------------------------------------------------
+    // Sync
+    //----------------------------------------------------------------------------------------------
+
+    /// @dev It is the caller's responsibility to ensure all asset IDs have a non-zero delta,
+    ///      and `sync` is called n times up until the moment all asset IDs are included, and the shares
+    ///      get synced as well.
+    function sync(PoolId poolId, ShareClassId scId, AssetId[] calldata assetIds) external {
```
wischli: Which external contract(s) will be calling this? Are we sure we don't want need to add the `auth` modifier?

src/managers/QueueManager.sol
```diff
@@ -0,0 +1,113 @@
...
+        ShareClassState storage sc = shareClassState[poolId][scId];
+        require(
+            sc.lastSync == 0 || sc.minDelay == 0 || block.timestamp >= sc.lastSync + sc.minDelay, MinDelayNotElapsed()
```
wischli, line 81: Note for myself to add a property which ensures `lastSync` is strictly monotone increasing.

src/managers/QueueManager.sol
```diff
@@ -0,0 +1,113 @@
...
+        (uint128 delta,, uint32 queuedAssetCounter,) = balanceSheet.queuedShares(poolId, scId);
+        require(delta > 0 || queuedAssetCounter > 0, NoUpdates());
+
+        bool maybeSubmitShares = delta > 0 && assetIds.length >= queuedAssetCounter;
+        uint256 submissions = assetIds.length;
+        uint32 actualSubmissions = 0;
+        bytes[] memory cs = new bytes[](maybeSubmitShares ? submissions + 1 : submissions);
+
+        for (uint256 i; i < submissions; i++) {
+            require(
```
wischli: A gas saving alternative, that would be limited to 256 max assets per sync, would be using a bitmap instead of transient storage.

Draft
```solidity

  // First pass: Validation
  uint256 validBitmap = 0;
  uint32 validCount = 0;
  for (uint256 i; i < assetIds.length; i++) {
      (uint128 deposits, uint128 withdrawals) = balanceSheet.queuedAssets(...);
      // Gas: ~2,600 (external call + SLOAD)

      if (deposits > 0 || withdrawals > 0) {
          validBitmap |= (1 << i);
          // Gas: ~20 (bitwise ops in memory)
          validCount++;
          // Gas: ~3 (increment)
      }
  }

  // Second pass: Iterate only for valid assets
  bytes[] memory cs = new bytes[](validCount + 1);
  uint256 csIndex = 0;
  for (uint256 i; i < assetIds.length; i++) {
      if (validBitmap & (1 << i) != 0) {
          // Gas: ~10 (bitmap check)
          cs[csIndex++] = abi.encodeWithSelector(...);
          // Gas: ~800 (encoding)
      }
  }

  // Multicall with NO noop operations
  IMulticall(address(balanceSheet)).multicall(cs);
```

AFAICT, rough gas difference could be
* Current
  - Valid asset: 1,000 (TLOAD) + 2,600 (query) + 800 (encode) + 900 (TSTORE) = ~5,300 gas
  - Invalid asset: 1,000 (TLOAD) + 2,600 (query) + 500 (noop encode) + 900 (TSTORE) + 5,000 (noop execution) = ~10,000 gas
* Bitwise
  - Valid asset: 2,600 (query) + 23 (bitmap) + 10 (check) + 800 (encode) = ~3,433 gas
  - Invalid asset: 2,600 (query) + 23 (bitmap) + 10 (check) = ~2,633 gas (no multicall execution!)

Different examples
1. All Valid Assets (10 assets)
  - Current: 10 × 5,300 = 53,000 gas
  - Bitmap: 10 × 3,433 = 34,330 gas
  - Savings: 18,670 gas (35% reduction)
2. Mixed (7 valid, 3 invalid)
  - Current: (7 × 5,300) + (3 × 10,000) = 67,100 gas
  - Bitmap: (7 × 3,433) + (3 × 2,633) = 31,930 gas
  - Savings: 35,170 gas (52% reduction)
3. Mostly Invalid (2 valid, 8 invalid)
  - Current: (2 × 5,300) + (8 × 10,000) = 90,600 gas
  - Bitmap: (2 × 3,433) + (8 × 2,633) = 27,930 gas
  - Savings: 62,670 gas (69% reduction)

src/managers/interfaces/IQueueManager.sol
```diff
@@ -0,0 +1,20 @@
...
+        PoolId indexed poolId, ShareClassId indexed scId, uint64 newMinDelay, uint128 newExtraGasLimit
+    );
+
+    error NotContractUpdater();
+    error NoUpdates();
+    error MinDelayNotElapsed();
+    error DuplicateAsset();
+    error NoUpdateForAsset();
+
+    function sync(PoolId poolId, ShareClassId scId, AssetId[] calldata assetIds) external;
```
wischli: Nit: Would like to see natspec for this call.

wischli (commented):
Great job on this fuzzy unexplored topic of our protocol, especially with the 100% test coverage!

src/managers/QueueManager.sol
lemunozm: Love it! I think this can even be extended to any length if using bytes at a minor extra cost. 

Is 256 enough for us @hieronx?

hieronx: You can use https://github.com/centrifuge/protocol-v3/blob/main/src/misc/libraries/BitmapLib.sol for this

hieronx: > Is 256 enough for us @hieronx?

I think it is very unlikely that the gas cost of a batch of >256 submissions can even fit into the max gas limit, so yes. And if we ever exceed that limit, we can develop a `LargeBatchQueueManager` :)

src/managers/QueueManager.sol
onnovisser: I think `&&` is right. If `queuedAssetCounter` is 0 and `assetIds.length` is 0, we can only submit queued shares. if `queuedAssetCounter` is > 0 and `assetIds.length` is 0, we don't want to actually submit queued shares

src/managers/QueueManager.sol
onnovisser: @hieronx  The BitmapLib only supports uint128, so only 128 submissions. Would that be enough still, or should I expand the BitmapLib with uint256?

hieronx: I would expand bitmaplib since it's a simple copy paste

src/managers/QueueManager.sol
onnovisser: it's permissionless, though having it permissioned would make things simpler

src/managers/QueueManager.sol
```diff
@@ -0,0 +1,113 @@
...
+        // Deduplicate and validate using bitmap for valid indices
+        uint256 validBitmap = 0; // Each bit represents if that index is valid
+        uint256 validCount = 0;
+
+        for (uint256 i = 0; i < assetIds.length; i++) {
+            bytes32 key = bytes32(uint256(AssetId.unwrap(assetIds[i])));
+            if (TransientStorageLib.tloadBool(key)) {
+                continue; // Skip duplicate
+            }
+            TransientStorageLib.tstore(key, true);
```
lemunozm: I think this is no longer need it, can we not check the bitmap `validBitmap`?

src/managers/QueueManager.sol
```diff
@@ -0,0 +1,113 @@
...
+        // Build exactly-sized array
+        bool submitShares = delta > 0 && validCount >= queuedAssetCounter;
+        bytes[] memory cs = new bytes[](validCount + (submitShares ? 1 : 0));
+
+        uint256 csIndex = 0;
+        for (uint256 i = 0; i < assetIds.length; i++) {
+            if (validBitmap.getBit(i)) {
+                // Check if index i is valid
+                cs[csIndex++] = abi.encodeWithSelector(
+                    balanceSheet.submitQueuedAssets.selector, poolId, scId, assetIds[i], sc.extraGasLimit
```
lemunozm: Theoretically, `extraGasLimit` should be a parameter of `sync()`. 

The explanation: although the snapshot hook can be the same over time, how the snapshot hook behaves internally based on the message could differ a lot. Putting this value in the `sync()` method, allows the caller to play with that value in cases where some extra computation requires more gas

src/misc/libraries/BitmapLib.sol
```diff
@@ -3,16 +3,16 @@ pragma solidity 0.8.28;
 
 /// @title  BitmapLib
 library BitmapLib {
```
hieronx, line 5: I just noticed we have no tests for this library. Could you add some very basic unit test ones, that then also use both `uint128` and `uint256`?

src/managers/QueueManager.sol
hieronx: I think the pool manager should just set the worst case `extraGasLimit` then

Problem with making it a parameter is that this method is permissionless, so then callers can just set 0, and it will still fail with out of gas on the target chain.

lemunozm: Mmm... I understand. Ok then 👍🏻

src/managers/QueueManager.sol
onnovisser: I realized the Bitmap helps with avoiding the noop, but it still needs to check duplicates. I didn't see a way to check duplicates with the Bitmap alone

lemunozm: Ah right!

wischli: Yeah I was also missing that a bit. Discussed with Onno and to the best of my knowledge, transient is the most readable gas-efficient duplicate check.

lemunozm: Maybe there is another one. Just an idea (that not need to be implemented at all)

We can force the list of `assetsId` to be ordered. It means in terms of computation:
- Check that the list is not ordered is just a check with the previous element
- Check duplication is also a check with the previous element

But not sure, from integrations, how easy it is to always order the `assetIds`. I think it is from UX but not from other contracts

src/managers/QueueManager.sol
```diff
@@ -0,0 +1,113 @@
...
+        require(delta > 0 || queuedAssetCounter > 0, NoUpdates());
+
+        require(assetIds.length <= 256, TooManyAssets()); // Bitmap limit
+
+        // Deduplicate and validate using bitmap for valid indices
+        uint256 validBitmap = 0; // Each bit represents if that index is valid
+        uint256 validCount = 0;
+
+        for (uint256 i = 0; i < assetIds.length; i++) {
+            bytes32 key = bytes32(uint256(AssetId.unwrap(assetIds[i])));
```
lemunozm: Just to follow the same pattern we use in the rest of the codebase
```suggestion
            bytes32 key = bytes32(uint256(assetIds[i].raw()));
```

test/managers/integration/QueueManager.t.sol
```diff
@@ -0,0 +1,446 @@
...
+import {ShareClassId} from "../../../src/common/types/ShareClassId.sol";
+
+import "../../spoke/integration/BaseTest.sol";
+
+import {IUpdateContract} from "../../../src/spoke/interfaces/IUpdateContract.sol";
+import {UpdateContractMessageLib} from "../../../src/spoke/libraries/UpdateContractMessageLib.sol";
+
+import {IQueueManager} from "../../../src/managers/QueueManager.sol";
+
+abstract contract QueueManagerBaseTest is BaseTest {
```
lemunozm, line 13: Is it possible to isolate all/most of the test cases into a UT?

test/integration/EndToEnd.t.sol
```diff
@@ -991,34 +1001,28 @@ contract EndToEndFlows is EndToEndUtils {
 
     function _testUpdateAccountingAfterDeposit(bool sameChain, bool afterAsyncDeposit, bool nonZeroPrices) public {
         (afterAsyncDeposit) ? _testAsyncDeposit(sameChain, nonZeroPrices) : _testSyncDeposit(sameChain, nonZeroPrices);
-
-        vm.startPrank(BSM);
-        s.balanceSheet.submitQueuedAssets(POOL_A, SC_1, s.usdcId, EXTRA_GAS);
-        s.balanceSheet.submitQueuedShares(POOL_A, SC_1, EXTRA_GAS);
```
lemunozm: I would revert the changes on EndToEnd. The idea here is to have a good understanding of how the core works. IMO, if populated with integrations, it can be harder to find a future issue, because you need another layer of understanding/behavior (the integration).

Not fully sure if we need an EndToEnd for this integration when the integration is fully based on `BalanceSheet`. I think we don't need it. In case it is, I would create another test case, but I really think it doesn't need it.

lemunozm: My understanding of what I said is, if `BalanceSheet` methods and `updateContract` are already tested, we don't need to test them again with an extra wrapper. As long as this extra wrapped is tested (which it is)

test/managers/unit/QueueManager.t.sol
```diff
@@ -0,0 +1,536 @@
...
+            address(balanceSheet),
+            abi.encodeWithSelector(IBalanceSheet.submitQueuedAssets.selector, POOL_A, SC_1, ASSET_1, 0)
+        );
+
+        // No calls for non-queued assets
+        vm.expectCall(
+            address(balanceSheet),
+            abi.encodeWithSelector(IBalanceSheet.submitQueuedAssets.selector, POOL_A, SC_1, ASSET_2, 0),
+            0
+        );
```
lemunozm, line 478: NIT. Maybe this can be into a `_submitQueuedAssets()` aux method, or even inside `_mockQueuedAssets()` if the expectation always matches

The same for the Shares version

test/managers/unit/QueueManager.t.sol
```diff
@@ -0,0 +1,536 @@
...
+import {IUpdateContract} from "../../../src/spoke/interfaces/IUpdateContract.sol";
+import {UpdateContractMessageLib} from "../../../src/spoke/libraries/UpdateContractMessageLib.sol";
+
+import {Multicall} from "../../../src/misc/Multicall.sol";
+
+import "forge-std/Test.sol";
+
+contract IsContract {}
+
+contract MockBalanceSheet is Multicall {
```
lemunozm: I think this is no longer needed (?)

onnovisser: it's still used for mocking the multicall and provides 0 as default values for queued assets/shares

src/managers/QueueManager.sol
```diff
@@ -58,20 +70,21 @@ contract QueueManager is IQueueManager, IUpdateContract {
...
-    function sync(PoolId poolId, ShareClassId scId, AssetId[] calldata assetIds) external {
+    function sync(PoolId poolId, ShareClassId scId, AssetId[] calldata assetIds) external payable {
         ShareClassQueueState storage sc = scQueueState[poolId][scId];
         require(
             sc.lastSync == 0 || sc.minDelay == 0 || block.timestamp >= sc.lastSync + sc.minDelay, MinDelayNotElapsed()
         );
 
         (uint128 delta,, uint32 queuedAssetCounter,) = balanceSheet.queuedShares(poolId, scId);
-        require(delta > 0 || queuedAssetCounter > 0, NoUpdates());
+        // require(delta > 0 || queuedAssetCounter > 0, NoUpdates());
```
lemunozm: Is this mandatory?

src/managers/QueueManager.sol
```diff
@@ -85,12 +98,17 @@ contract QueueManager is IQueueManager, IUpdateContract {
...
                 validBitmap = validBitmap.withBit(i, true);
                 validCount++;
+                gasLimit += sendUpdateAssetsGasLimit + sc.extraGasLimit;
             }
         }
 
         // Build exactly-sized array
         bool submitShares = delta > 0 && validCount >= queuedAssetCounter;
-        bytes[] memory cs = new bytes[](validCount + (submitShares ? 1 : 0));
+        uint256 callCount = validCount + (submitShares ? 1 : 0);
```
lemunozm: Why `&&`, should not be `||` ?

src/managers/QueueManager.sol
```diff
@@ -106,8 +124,17 @@ contract QueueManager is IQueueManager, IUpdateContract {
             cs[validCount] =
                 abi.encodeWithSelector(balanceSheet.submitQueuedShares.selector, poolId, scId, sc.extraGasLimit);
             sc.lastSync = uint64(block.timestamp);
+            gasLimit += sendUpdateSharesGasLimit + sc.extraGasLimit;
         }
 
+        // Split because stack too deep
```
lemunozm: As another option, FYI, you can also use the following pattern to avoid stack too deep issue without splitting the method:
```
function ... {
    code
    { 
        // foo = ...
    }
    // foo out of scope, no longer exists from here and no counts in the stack
}
```

For example, you can surround `csIndex` wiht `{ }`

src/managers/QueueManager.sol
```diff
@@ -106,8 +124,17 @@ contract QueueManager is IQueueManager, IUpdateContract {
...
+        _sync(poolId, cs, gasLimit);
+    }
+
+    function _sync(PoolId poolId, bytes[] memory cs, uint128 gasLimit) internal {
+        uint256 estimate = adapter.estimate(poolId.centrifugeId(), cs[0], gasLimit);
```
lemunozm: CR. You should concatenate all elements of `cs` to perform a correct estimation.

I think we should also offer an `estimate()` method that computes the min gas used by `sync()` to be able to call it with an accurate value.

src/managers/QueueManager.sol
```diff
@@ -24,12 +27,21 @@ contract QueueManager is IQueueManager, IUpdateContract {
...
     mapping(PoolId => mapping(ShareClassId => ShareClassQueueState)) public scQueueState;
 
     constructor(address contractUpdater_, IBalanceSheet balanceSheet_) {
         contractUpdater = contractUpdater_;
         balanceSheet = balanceSheet_;
+        gateway = balanceSheet_.gateway();
+        adapter = gateway.adapter();
+        IGasService gasService = gateway.gasService();
+        sendUpdateAssetsGasLimit = gasService.updateHoldingAmount();
+        sendUpdateSharesGasLimit = gasService.updateShares();
```
lemunozm: We can migrate gasService with a new value returned by `updateHoldingAmount` or `updateShares`, so these values needs to be fetched each time we call `sync()`

src/managers/QueueManager.sol
```diff
@@ -24,12 +27,21 @@ contract QueueManager is IQueueManager, IUpdateContract {
...
+    uint128 public immutable sendUpdateSharesGasLimit;
 
     mapping(PoolId => mapping(ShareClassId => ShareClassQueueState)) public scQueueState;
 
     constructor(address contractUpdater_, IBalanceSheet balanceSheet_) {
         contractUpdater = contractUpdater_;
         balanceSheet = balanceSheet_;
+        gateway = balanceSheet_.gateway();
+        adapter = gateway.adapter();
+        IGasService gasService = gateway.gasService();
```
lemunozm: @hieronx I think we need a contract that wraps these 3 contracts to let integrators internally call in one go:

```
uint256 gas = multiAdapter.estimate(.., gasService.foo())
gateway.subsidizePool{value: gas}(..);
```

And also do it over the latest gateway/adapter/gasService contracts in case we need to upgrade one

lemunozm: I also think we should do it for `hub` and `balanceSheet`, otherwise, all integrators that use i.e. `hub`, will break if we need to migrate the hub for some bugfix or minor change.

Note that we can not support 2 hubs at the same time, given we need to process the messages only in one of them

src/managers/QueueManager.sol
onnovisser: yes, I wasn't sure about it. but then wouldn't I need to call `bs.gateway().adapter().gasService()` ? Because what if the gateway or adapter also changes. Then it may call the wrong GasService right?

src/managers/QueueManager.sol
onnovisser: will remove

src/managers/QueueManager.sol
lemunozm: Correct, I left a comment below regarding this. That depends on how probable is that something will be migrated. But for sure, `gasService` is highly probable. We could argue that `Gateway` or an `Adapter` is not likely to be migrated

src/managers/QueueManager.sol
onnovisser: I tried it but am still getting the error

src/managers/QueueManager.sol
onnovisser: it should only submit shares after submitting all assets, so if in this call we already have, or are about to submit all pending assets.

src/managers/QueueManager.sol
```diff
@@ -0,0 +1,139 @@
...
+contract QueueManager is IQueueManager, IUpdateContract {
+    using CastLib for *;
+    using BitmapLib for *;
+
+    address public immutable contractUpdater;
+    IBalanceSheet public immutable balanceSheet;
+    IGateway public immutable gateway;
+    IAdapter public immutable adapter;
+    uint128 public immutable sendUpdateAssetsGasLimit;
+    uint128 public immutable sendUpdateSharesGasLimit;
```
hieronx: Nit:
```suggestion
    IAdapter public immutable adapter;
    IGateway public immutable gateway;
    address public immutable contractUpdater;
    IBalanceSheet public immutable balanceSheet;
    uint128 public immutable sendUpdateAssetsGasLimit;
    uint128 public immutable sendUpdateSharesGasLimit;
```

hieronx (commented):
Looks good!

src/managers/QueueManager.sol
lemunozm: Yeah, I think you're right

src/managers/QueueManager.sol
wischli: I also had that idea but I don't think this is a reasonable assumption.

src/managers/QueueManager.sol
```diff
@@ -0,0 +1,139 @@
...
+                cs[csIndex++] = abi.encodeWithSelector(
+                    balanceSheet.submitQueuedAssets.selector, poolId, scId, assetIds[i], sc.extraGasLimit
+                );
+            }
+        }
+
+        if (submitShares) {
+            cs[validCount] =
+                abi.encodeWithSelector(balanceSheet.submitQueuedShares.selector, poolId, scId, sc.extraGasLimit);
+            sc.lastSync = uint64(block.timestamp);
```
wischli, line 104: Can you remind me why we only update when submitting shares but not just assets?

test/misc/unit/libraries/BitmapLib.t.sol
```diff
@@ -0,0 +1,65 @@
+// SPDX-License-Identifier: GPL-2.0-or-later
+pragma solidity 0.8.28;
+
+import {BitmapLib} from "../../../../src/misc/libraries/BitmapLib.sol";
+
+import "forge-std/Test.sol";
+
+contract BitmapLibTest is Test {
```
wischli, line 8: Thanks for adding this with 100% coverage, awesome!

wischli (commented):
Apart from the discussed gas estimation LGTM! One tiny question I need to be refreshed on. Great work!

src/managers/QueueManager.sol
onnovisser: you're not required to submit all assets. so someone could submit one asset and DoS

src/managers/QueueManager.sol
```diff
@@ -69,16 +78,18 @@ contract QueueManager is IQueueManager, IUpdateContract, Auth {
 
     /// @inheritdoc IQueueManager
     function sync(PoolId poolId, ShareClassId scId, AssetId[] calldata assetIds) external payable {
+        gateway.depositSubsidy{value: msg.value}(poolId);
+        crosschainBatcher.execute(abi.encodeWithSelector(QueueManager.syncCallback.selector, poolId, scId, assetIds));
+    }
```
lemunozm: Is it possible to add withdrawSubsidy at the end?

onnovisser: I can I suppose, but then it needs to be a gateway manager right?

src/managers/QueueManager.sol
```diff
@@ -0,0 +1,112 @@
...
+        }
+    }
+
+    //----------------------------------------------------------------------------------------------
+    // Sync
+    //----------------------------------------------------------------------------------------------
+
+    /// @inheritdoc IQueueManager
+    function sync(PoolId poolId, ShareClassId scId, AssetId[] calldata assetIds) external payable {
+        gateway.withBatch{value: msg.value}(
```
xmxanuel, line 72: You could ensure that the `msg.value` is sufficient here. Otherwise, someone could move them on purpose to underpaid and it would require additional `Gateway.repay` calls.

lemunozm: After the latest system refactor, that's no longer possible. If not fully paid, it will revert

src/managers/QueueManager.sol
```diff
@@ -0,0 +1,112 @@
...
+
+            // Check if valid
+            (uint128 deposits, uint128 withdrawals) = balanceSheet.queuedAssets(poolId, scId, assetIds[i]);
+            if (deposits > 0 || withdrawals > 0) {
+                balanceSheet.submitQueuedAssets(poolId, scId, assetIds[i], sc.extraGasLimit, address(0));
+                validCount++;
+            }
+        }
+
+        bool submitShares = delta > 0 && validCount >= queuedAssetCounter;
```
xmxanuel: How can validCount be greater than queuedAssetCounter?

Is this a valid case or could the expression be simplified to `validCount == queuedAssetCounter`?

xmxanuel: Why not `submitShares` always when `validCount == queuedAssetCounter`. This would update the snapshot flag for the shares even when `delta` equals zero.

src/managers/QueueManager.sol
```diff
@@ -0,0 +1,112 @@
...
+import {IGateway} from "../common/interfaces/IGateway.sol";
+import {ShareClassId} from "../common/types/ShareClassId.sol";
+
+import {IBalanceSheet} from "../spoke/interfaces/IBalanceSheet.sol";
+import {IUpdateContract} from "../spoke/interfaces/IUpdateContract.sol";
+import {UpdateContractMessageLib, UpdateContractType} from "../spoke/libraries/UpdateContractMessageLib.sol";
+
+/// @dev minDelay can be set to a non-zero value, for cases where assets or shares can be permissionlessly modified
+///      (e.g. if the on/off ramp manager is used, or if sync deposits are enabled). This prevents spam.
+contract QueueManager is IQueueManager, IUpdateContract, Auth {
```
hieronx: Nit: implementations before interfaces:
```suggestion
contract QueueManager is Auth, IQueueManager, IUpdateContract {
```

src/managers/QueueManager.sol
```diff
@@ -0,0 +1,112 @@
...
+        require(msg.sender == address(gateway), NotAuthorized());
+        ShareClassQueueState storage sc = scQueueState[poolId][scId];
+        require(
+            sc.lastSync == 0 || sc.minDelay == 0 || block.timestamp >= sc.lastSync + sc.minDelay, MinDelayNotElapsed()
+        );
+
+        (uint128 delta,, uint32 queuedAssetCounter,) = balanceSheet.queuedShares(poolId, scId);
+
+        uint256 validCount = 0;
+
```
hieronx: Ultra-nit, some unnecessary spacing here imo:
```suggestion
        uint256 validCount = 0;
```

src/managers/QueueManager.sol
```diff
@@ -0,0 +1,112 @@
...
+        for (uint256 i = 0; i < assetIds.length; i++) {
+            bytes32 key = keccak256(abi.encode(scId.raw(), assetIds[i].raw()));
+            if (TransientStorageLib.tloadBool(key)) {
+                continue; // Skip duplicate
+            }
```
hieronx:
```suggestion
            if (TransientStorageLib.tloadBool(key)) continue; // Skip duplicate
```

src/managers/QueueManager.sol
onnovisser: 1. You're right, `validCount` can never be greater, will make it a simpler `==`
2. The BalanceSheet will set the snapshot flag when submitting assets, if the shares delta is 0 https://github.com/centrifuge/protocol/blob/46207fe7a422a0534f673dd58dafa1caf9f269be/src/spoke/BalanceSheet.sol#L214

src/managers/QueueManager.sol
```diff
@@ -0,0 +1,112 @@
...
+            // Check if valid
+            (uint128 deposits, uint128 withdrawals) = balanceSheet.queuedAssets(poolId, scId, assetIds[i]);
+            if (deposits > 0 || withdrawals > 0) {
+                balanceSheet.submitQueuedAssets(poolId, scId, assetIds[i], sc.extraGasLimit, address(0));
+                validCount++;
+            }
+        }
+
+        bool submitShares = delta > 0 && validCount >= queuedAssetCounter;
+
```
hieronx:
```suggestion
```

src/managers/QueueManager.sol
xmxanuel: Ah, okay. I thought getting the `isSnapshot` true for the shareQueue in a separate transaction is different. Then it is fine, since `isSnapshot=true` for the last asset update.

hieronx (approved):

PR #584, opened 2025-08-27
Avoid manual payment for transfering shares if Hub is in the same chain as origin Spoke and minor fixes.

lemunozm:
Ref: https://kflabs.slack.com/archives/C07PG2EUR9C/p1756245526774999

### Changes
- No need to repay a message if the network topology of the transaction is: 
`Spoke chain X => Hub chain X => Spoke chain Y`
- Fix a bug in the Gateway where the extraGas wasn't taken into account for external unpaid messages (using `addUnpaidMessage()` method)
- Add missing `PrepareMessage` event to `addUnpaidMessage()`
- Fix gas values for the case of: Spoke on chain X => hub on chain Y => spoke on chain Y

### DIFF
--- a/src/common/GasService.sol
+++ b/src/common/GasService.sol
@@ -57,8 +57,8 @@ contract GasService is IGasService {
         notifyPricePoolPerAsset = BASE_COST + 35759;
         notifyShareMetadata = BASE_COST + 13343;
         updateShareHook = BASE_COST + 6415;
-        initiateTransferShares = BASE_COST + 52195;
-        executeTransferShares = BASE_COST + 70267;
+        initiateTransferShares = BASE_COST + 89689;
+        executeTransferShares = BASE_COST + 70428;
         updateRestriction = BASE_COST + 35992;
         updateContract = BASE_COST + 53345;
         requestCallback = BASE_COST + 186947; // approve deposit case
--- a/src/common/Gateway.sol
+++ b/src/common/Gateway.sol
@@ -187,7 +187,7 @@ contract Gateway is Auth, Recoverable, IGateway {
 
     /// @inheritdoc IGateway
     function addUnpaidMessage(uint16 centrifugeId, bytes memory message) external auth {
-        _addUnpaidBatch(centrifugeId, message, gasService.messageGasLimit(centrifugeId, message));
+        _addUnpaidBatch(centrifugeId, message, gasService.messageGasLimit(centrifugeId, message) + extraGasLimit);
     }
 
     function _addUnpaidBatch(uint16 centrifugeId, bytes memory message, uint128 gasLimit) internal {
--- a/src/common/Gateway.sol
+++ b/src/common/Gateway.sol
@@ -189,6 +189,7 @@ contract Gateway is Auth, Recoverable, IGateway {
     function addUnpaidMessage(uint16 centrifugeId, bytes memory message) external auth {
         uint128 gasLimit = gasService.messageGasLimit(centrifugeId, message) + extraGasLimit;
         extraGasLimit = 0;
+        emit PrepareMessage(centrifugeId, processor.messagePoolId(message), message);
         _addUnpaidBatch(centrifugeId, message, gasLimit);
     }
 
--- a/src/common/Gateway.sol
+++ b/src/common/Gateway.sol
@@ -187,7 +187,9 @@ contract Gateway is Auth, Recoverable, IGateway {
 
     /// @inheritdoc IGateway
     function addUnpaidMessage(uint16 centrifugeId, bytes memory message) external auth {
-        _addUnpaidBatch(centrifugeId, message, gasService.messageGasLimit(centrifugeId, message) + extraGasLimit);
+        uint128 gasLimit = gasService.messageGasLimit(centrifugeId, message) + extraGasLimit;
+        extraGasLimit = 0;
+        _addUnpaidBatch(centrifugeId, message, gasLimit);
     }
 
     function _addUnpaidBatch(uint16 centrifugeId, bytes memory message, uint128 gasLimit) internal {
--- a/test/integration/ThreeChainEndToEnd.t.sol
+++ b/test/integration/ThreeChainEndToEnd.t.sol
@@ -125,7 +125,7 @@ contract ThreeChainEndToEndDeployment is EndToEndUseCases {
         emit IHub.ForwardTransferShares(sC.centrifugeId, POOL_A, SC_1, INVESTOR_A.toBytes32(), amount);
 
         // If hub is not source, then message will be pending as unpaid on hub until repaid
-        if (direction != CrossChainDirection.FromHub) {
+        if (direction == CrossChainDirection.WithIntermediaryHub) {
             vm.expectEmit(true, false, false, false);
             emit IGateway.UnderpaidBatch(sC.centrifugeId, bytes(""));
         } else {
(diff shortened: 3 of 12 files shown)


src/common/Gateway.sol
```diff
@@ -187,7 +187,9 @@ contract Gateway is Auth, Recoverable, IGateway {
 
     /// @inheritdoc IGateway
     function addUnpaidMessage(uint16 centrifugeId, bytes memory message) external auth {
-        _addUnpaidBatch(centrifugeId, message, gasService.messageGasLimit(centrifugeId, message));
+        uint128 gasLimit = gasService.messageGasLimit(centrifugeId, message) + extraGasLimit;
+        extraGasLimit = 0;
+        _addUnpaidBatch(centrifugeId, message, gasLimit);
```
lemunozm: This was a bug. The unpaid message should take into account any extra gas added when it was created

lemunozm:
@wischli I've simplified a bit the ThreeChainEndToEnd configuration because we never use the spoke on A: [c0ace7e](https://github.com/centrifuge/protocol-v3/pull/584/commits/c0ace7e4063ffb2d7e2af31c9cb5496b4106f66c), tell me if you have any thoughts agains this.

I think, if you agree, given we only have 2 spokes for the test cases we do, we can call them origin/target or so, which helps reading in the direction of the cross transfer done.

src/common/GasService.sol
```diff
@@ -57,8 +57,8 @@ contract GasService is IGasService {
         notifyPricePoolPerAsset = BASE_COST + 35759;
         notifyShareMetadata = BASE_COST + 13343;
         updateShareHook = BASE_COST + 6415;
-        initiateTransferShares = BASE_COST + 52195;
-        executeTransferShares = BASE_COST + 70267;
+        initiateTransferShares = BASE_COST + 89689;
+        executeTransferShares = BASE_COST + 70428;
```
lemunozm, line 61: NOTE: these changes would not have occurred in any gas change (despite minimals).

These both paradigms should take the same gas:
- `Spoke on chain X => hub on chain Y => spoke on chain Z` (paid in Y for run Z)
- `Spoke on chain X => hub on chain X => spoke on chain Y` (paid in X for run Z)

But we had an issue where the first case:
- `Spoke on chain X => hub on chain Y => spoke on chain Y`, where we need to pay in X to run the hub and the spoke in Y (previously only the hub cost was considered)

TLDR: We had wrongly computed the `initiateTransferShares`, independently of this change, but now it's fixed.

lemunozm:
I'm excedding the Hub contract size by just 1B haha, any ideas?
- To avoid refactor everything, I propose to remove this line, that surely it will not be used: https://github.com/centrifuge/protocol-v3/blob/main/src/hub/Hub.sol#L81, meaning if we need to upload a new `HubHelpers`, it means we probably 99.9% need to upload again the `Hub`.

hieronx:
> I'm excedding the Hub contract size by just 1B haha, any ideas? To avoid refactor everything, I propose to remove this line, that surely it will not be used: https://github.com/centrifuge/protocol-v3/blob/main/src/hub/Hub.sol#L81, meaning if we need to upload a new HubHelpers, it means we probably 99.9% need to upload again the Hub.

Seems fine as temporary fix, we want to change Hub more significantly before v3.1 anyways

hieronx (dismissed):
LGTM

src/common/Gateway.sol
```diff
@@ -187,7 +187,10 @@ contract Gateway is Auth, Recoverable, IGateway {
 
     /// @inheritdoc IGateway
     function addUnpaidMessage(uint16 centrifugeId, bytes memory message) external auth {
-        _addUnpaidBatch(centrifugeId, message, gasService.messageGasLimit(centrifugeId, message));
+        uint128 gasLimit = gasService.messageGasLimit(centrifugeId, message) + extraGasLimit;
+        extraGasLimit = 0;
```
wischli, line 191: Q: Couldn't this reset to 0 cause serious issues for batch processing of multiple messages?

test/integration/ThreeChainEndToEnd.t.sol
```diff
@@ -25,8 +25,8 @@ import "forge-std/Test.sol";
 
 enum CrossChainDirection {
     WithIntermediaryHub, // C -> A -> B (Hub is on A)
-    FromHub, // C == A -> B (Hub is on C)
-    ToHub // C -> B == A (Hub is on B)
+    FromHub, // C => A -> B (Hub is on A)
+    ToHub // C -> A => B (Hub is on A)
```
wischli, line 29: Can you explain the difference of `=>` vs `->` here? From the context, I thought that the fat arrow `=>` indicates a transfer, for which I used `->`. The `C == A` here means that both chains are equal, e.g. they have the same centrifuge id, spoke and hub.

test/integration/ThreeChainEndToEnd.t.sol
```diff
@@ -69,39 +68,29 @@ contract ThreeChainEndToEndDeployment is EndToEndUseCases {
     }
 
     function _setSpokes(CrossChainDirection direction) internal {
+        // NOTE: Hub is always in deployA.
         if (direction == CrossChainDirection.WithIntermediaryHub) {
-            _setSpoke(deployA, CENTRIFUGE_ID_A, sA);
```
wischli, line 73: Why did you remove setting the spoke here? Doesn't this leave chain A without a spoke?

wischli (commented):
Thanks for looking into this and the improvements!

AFAICT, you haven't touched the incorrect transfer direction introduced by me initially right? For instance, the `FromHub` test emits `InitiateTransferShares(centrifugeId: 5, ...)` when it should be `InitiateTransferShares(centrifugeId: 6, ...)`, so `FromHub` and `ToHub` are swapped right now because the test always executes the transfer on chain `B` with `C` being the recipient.

src/common/Gateway.sol
lemunozm: `extraGasLimit` is set per message. So if you're in a batch, the process will be:

- setExtraGasLimit(1);
- send(m1); // extraGasLimit reset to 0
- setExtraGasLimit(2);
- addUnpaidMessage(m2); // extraGasLimit reset to 0
- setExtraGasLimit(3);
- send(m3); // extraGasLimit reset to 0
- ...

test/integration/ThreeChainEndToEnd.t.sol
lemunozm: - `=>` : local transfer
- `->` : remote transfer

I was getting a mess because from some PoV you have how the message is sent, and from other PoV you have the deployments (it is, if the hub is in the same chain as the spoke).

I understood that this enum wanted to express the directionality of the message sent. Open to suggestions anyway. Maybe a longer explanation like: `Spoke on chain A => hub on chain A => spoke on chain B` helps more.

test/integration/ThreeChainEndToEnd.t.sol
lemunozm: It was not used and was creating confusion (at least for me) to understand the topology of the network over the test case. 

I think on one side we want to have a full ThreeChain set up, but by the other side, we have different deployments depending of the usage/directionality of the cross transfer.

Given we only need to test the cross transfer (and probably nothing else with 3 chains), I wanted to reduce the configuration overhead to mimic the test case. Tell me if you're against this and I can revert.

NOTE: also removing this spoke, helps if later I rename sB and sC to origin/target to express better from where the things go

lemunozm:
> AFAICT, you haven't touched the incorrect transfer direction introduced by me initially right? For instance, the FromHub test emits InitiateTransferShares(centrifugeId: 5, ...) when it should be InitiateTransferShares(centrifugeId: 6, ...), so FromHub and ToHub are swapped right now because the test always executes the transfer on chain B with C being the recipient.

Yeah, I didn't touch this yet, probably in another PR to avoid more noise in this one

src/common/Gateway.sol
wischli: Thanks for reminding me of that!

test/integration/ThreeChainEndToEnd.t.sol
wischli: Thanks for the explanation! Agree it makes sense to cut the noise. Apologies for the previous state.

wischli (approved):

test/integration/ThreeChainEndToEnd.t.sol
lemunozm: Not at all! I think it has been mutated over the time

Issue #586, opened 2025-08-27
Rename `BaseTransferHook` transfer type helpers


wischli:
https://github.com/centrifuge/protocol-v3/pull/578#discussion_r2288783957

PR #588, opened 2025-08-28
ref: hook transfer type names

wischli:
https://github.com/centrifuge/protocol-v3/issues/586

Also removes the `BaseTransferHook` integration test checks for the pre https://github.com/centrifuge/protocol-v3/pull/578 `wrongHook` as rightfully pointed out by @lemunozm to reduce maintenance burden ([ref](https://github.com/centrifuge/protocol-v3/pull/578#discussion_r2304978133))


src/hooks/FreelyTransferable.sol
```diff
@@ -27,10 +27,10 @@ contract FreelyTransferable is BaseTransferHook {
         returns (bool)
     {
         if (isSourceOrTargetFrozen(from, to, hookData)) return false;
-        if (isDepositRequest(from, to)) return isTargetMember(to, hookData);
+        if (isIssuance(from, to)) return isTargetMember(to, hookData);
```
lemunozm: Question: We no longer have a hook when the user deposits the request?

wischli: We never did for `requestDeposit` because that just transfers assets from the user to the global escrow. The hooks act as a share token filter.

hieronx: This still triggers also on deposit request, because of https://github.com/centrifuge/protocol-v3/blob/57c998640024884daf70a9ff88c467e1c8199d34/src/vaults/AsyncRequestManager.sol#L82

wischli: Ah yes 🙈 WDYT about  `isDepositRequestOrIssuance` and `isRedeemRequestOrRevocation` to signal that?

hieronx: Yeah I think we need to do that. But `isRedeemRequestOrRevocation` => `isRedeemClaimOrRevocation`

wischli: [31bffc1](https://github.com/centrifuge/protocol-v3/pull/588/commits/31bffc100c3377d0046fed6debf09eb64f22dc29)

hieronx (approved):

PR #589, opened 2025-08-28
New messages for adapter per pool

lemunozm:
- Add new messages
- Add new message property: `messagePoolIdPayment` which only refers to who pays for the message

### DIFF
--- a/src/common/GasService.sol
+++ b/src/common/GasService.sol
@@ -20,6 +20,8 @@ contract GasService is IGasService {
     uint128 public immutable cancelUpgrade;
     uint128 public immutable recoverTokens;
     uint128 public immutable registerAsset;
+    uint128 public immutable initiateSetPoolAdapters;
+    uint128 public immutable executeSetPoolAdapters;
     uint128 public immutable request;
     uint128 public immutable notifyPool;
     uint128 public immutable notifyShareClass;
@@ -50,6 +52,8 @@ contract GasService is IGasService {
         cancelUpgrade = BASE_COST + 8861;
         recoverTokens = BASE_COST + 82906;
         registerAsset = BASE_COST + 34329;
+        initiateSetPoolAdapters = BASE_COST + 1000000; // TODO
+        executeSetPoolAdapters = BASE_COST + 1000000; // TODO
         request = BASE_COST + 86084; // request deposit case
         notifyPool = BASE_COST + 1154806; // create escrow case
         notifyShareClass = BASE_COST + 1775916;
@@ -85,6 +89,9 @@ contract GasService is IGasService {
         if (kind == MessageType.ScheduleUpgrade) return scheduleUpgrade;
         if (kind == MessageType.CancelUpgrade) return cancelUpgrade;
         if (kind == MessageType.RecoverTokens) return recoverTokens;
+        if (kind == MessageType.RegisterAsset) return registerAsset;
+        if (kind == MessageType.InitiateSetPoolAdapters) return initiateSetPoolAdapters;
+        if (kind == MessageType.ExecuteSetPoolAdapters) return executeSetPoolAdapters;
         if (kind == MessageType.RegisterAsset) return registerAsset;
         if (kind == MessageType.Request) return request;
         if (kind == MessageType.NotifyPool) return notifyPool;
--- a/src/common/GasService.sol
+++ b/src/common/GasService.sol
@@ -92,7 +92,6 @@ contract GasService is IGasService {
         if (kind == MessageType.RegisterAsset) return registerAsset;
         if (kind == MessageType.InitiateSetPoolAdapters) return initiateSetPoolAdapters;
         if (kind == MessageType.ExecuteSetPoolAdapters) return executeSetPoolAdapters;
-        if (kind == MessageType.RegisterAsset) return registerAsset;
         if (kind == MessageType.Request) return request;
         if (kind == MessageType.NotifyPool) return notifyPool;
         if (kind == MessageType.NotifyShareClass) return notifyShareClass;
--- a/src/common/libraries/MessageLib.sol
+++ b/src/common/libraries/MessageLib.sol
@@ -164,6 +164,10 @@ library MessageLib {
 
         if (kind <= uint8(MessageType.RecoverTokens)) {
             return 0; // Non centrifugeId associated
+        } else if (kind == uint8(MessageType.InitiateSetPoolAdapters)) {
+            return message.messagePoolIdPayment().centrifugeId();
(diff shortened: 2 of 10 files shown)


lemunozm:
@hieronx as a note. Not sure who should pay the ExecuteSetPoolAdapters, because in the first instance, there is no pool to subsidize in the Spoke side to pay for that message.

Of course that can be just fixed with our discussion here: https://kflabs.slack.com/archives/C07PG2EUR9C/p1756366924951689

src/common/GasService.sol
```diff
@@ -86,6 +90,9 @@ contract GasService is IGasService {
         if (kind == MessageType.CancelUpgrade) return cancelUpgrade;
         if (kind == MessageType.RecoverTokens) return recoverTokens;
         if (kind == MessageType.RegisterAsset) return registerAsset;
+        if (kind == MessageType.InitiateSetPoolAdapters) return initiateSetPoolAdapters;
+        if (kind == MessageType.ExecuteSetPoolAdapters) return executeSetPoolAdapters;
+        if (kind == MessageType.RegisterAsset) return registerAsset;
```
hieronx:
```suggestion
```

src/common/libraries/MessageLib.sol
```diff
@@ -140,9 +144,19 @@ library MessageLib {
...
         if (kind >= uint8(MessageType.NotifyPool)) {
             return PoolId.wrap(message.toUint64(1));
-        } else {
-            return PoolId.wrap(0);
         }
+
+        return PoolId.wrap(0);
+    }
+
+    function messagePoolIdPayment(bytes memory message) internal pure returns (PoolId poolId) {
```
hieronx, line 152: Should add tests for this

hieronx:
> as a note. Not sure who should pay the ExecuteSetPoolAdapters, because in the first instance, there is no pool to subsidize in the Spoke side to pay for that message.

If we don't manage to go with the other solution, I think having this unpaid, and require manual repayment, for the very first setup, is acceptable

lemunozm:
NOTE. We could _theoretically_ create batches for the same pool that contains different messages with different payers. The first message in the batch is who will pay for the entire batch.

For example, the following theoretical batch will be paid by POOL_0
- RegisterAsset
- InitiateSetPoolAdapter

But this theoretical batch will be paid by POOL_A
- InitiateSetPoolAdapter
- RegisterAsset

This is not a problem right not because we don't allow mixing them in a batch, but something to keep in mind if this precondition changes in the future.

src/common/GasService.sol
lemunozm: Oops, fixed!

src/common/libraries/MessageLib.sol
lemunozm: And for the new messages actually! Done

src/common/libraries/MessageLib.sol
```diff
@@ -140,16 +144,30 @@ library MessageLib {
...
+
+        return messagePoolId(message);
     }
 
     function messageSourceCentrifugeId(bytes memory message) internal pure returns (uint16) {
         uint8 kind = message.toUint8(0);
 
         if (kind <= uint8(MessageType.RecoverTokens)) {
             return 0; // Non centrifugeId associated
+        } else if (kind == uint8(MessageType.InitiateSetPoolAdapters)) {
```
hieronx, line 167: I think this case could be removed as `message.messagePoolId().centrifugeId();` would work for this?

lemunozm: I don't think so because `message.messagePoolId().centrifugeId()` for that message is 0, but we're able to ensure the source of that message will always be the payer

Note that these new messages belong to POOL_0

hieronx: > because message.messagePoolId().centrifugeId() for that message is 0
> Note that these new messages belong to POOL_0

Why? These are used to set adapters for a specific pool right? So can be pool ID 0 but also any other pool?

lemunozm: `messageSourceCentrifugeId` is used to determine from where `centrifugeId` the message was originated.

- If I choose `message.messagePoolId().centrifugeId()`, it will be 0, which means we can not determine from where the message comes.
- If I choose `message.messagePoolIdPayment().centrifugeId()`, it will be the poolId associated with the payer, which always belongs to the centrifuge chain where the pool was created. So we can use that value to double-check the origin of the message when it's received.

hieronx: > If I choose message.messagePoolId().centrifugeId(), it will be 0

This part I'm not following 🤔

lemunozm: > These are used to set adapters for a specific pool right? 

Yes, `messagePoolId` is used for that.

> So can be pool ID 0 but also any other pool?

The message contains the pool (always != 0), but we forcefully need to send this through the general adapters to be able to set the communication the first time

hieronx: I get it now. Wonder if this will create confusion on the indexing side, since from a user perspective, `InitiateSetPoolAdapters` is definitely related to a specific pool. Almost think we need a separate `messagePoolIdAdapters()`....

lemunozm: To achieve the above, we need `message.messagePoolId().centrifugeId() == 0`, to en route them through the general adapters.

hieronx: Yeah I meant that we'd also change the gateway logic to route based on `messagePoolIdAdapters`

lemunozm: Yeah, right now the routing is still though `messagePoolId` (which I can rename to `messagePoolIdAdapters`).

For me, this message is just a general message that carries poolId information, but it does not belong to it. But true that it's somehow in the middle...

If we finally remove the subsidized pool, we don't need anything regarding this, given that this method only exists for subsidizing this message. Maybe another reason in favor

hieronx: I didn't mean removing it, I meant having another separate method. The other usage of `messagePoolId` is in the `PrepareMessage` events, which the indexer uses. So the indexer should show the actual pool ID here in the case of `InitiateSetPoolAdapters`

hieronx: Will approve PR but let's discuss this minor detail later.

hieronx (approved):

src/common/libraries/MessageLib.sol
lemunozm: Will give it a thought!

PR #592, opened 2025-08-28
Sepolia tests to check gas values

lemunozm:
Check if Wormhole and LayerZero works fine sending gas


lemunozm:
I have verified that it worked. Both events: https://sepolia.etherscan.io/address/0xc9d550e6a3b0d68dcfe5bb9a159fae5805ea898b#events

PR #595, opened 2025-09-03
Add admin-controlled asset price oracle

hieronx:
### Product goals
* Hub managers can define 1 or more feeders, accounts which can set any price
* Prices should be set for assets, denominated in the pool currency (there is no use case within Centrifuge pools for setting prices denominated in other assets)
* 0.0 prices should be supported

### Design notes
* `BaseValuation` was removed as this provides little value, while abstracting away the integration with `HubRegistry` (hard to know this is `IERC6909Decimals`).
* `IValuation` interface was changed to allow using a single oracle valuation contract deployment for multiple pools, and to remove the need to query the pool currency


script/ValuationsDeployer.s.sol
```diff
@@ -37,13 +35,24 @@ contract ValuationsDeployer is HubDeployer {
...
+            create3(
+                generateSalt("oracleValuation"),
+                abi.encodePacked(
+                    type(OracleValuation).creationCode,
+                    abi.encode(hub, address(batcher), hubRegistry, input.centrifugeId, address(batcher))
+                )
             )
         );
 
         batcher.engageValuations(_valuationsReport());
```
lemunozm: NIT. We can remove this to avoid signed an extra TX (similar pattern as `ManagersDeployer.s.sol`)

src/common/interfaces/IValuation.sol
```diff
@@ -1,14 +1,17 @@
 // SPDX-License-Identifier: GPL-2.0-or-later
 pragma solidity >=0.5.0;
 
+import {PoolId} from "../types/PoolId.sol";
 import {AssetId} from "../types/AssetId.sol";
+import {ShareClassId} from "../types/ShareClassId.sol";
 
-/// Based on [ERC-7726](https://eips.ethereum.org/EIPS/eip-7726): Common Quote Oracle, but adapted for ERC6909.
```
lemunozm, line 6: Fully agree with removing our soft dependency with ERC-7726!

src/hub/HubHelpers.sol
```diff
@@ -259,7 +259,7 @@ contract HubHelpers is Auth, IHubHelpers {
         AssetId poolCurrency = hubRegistry.currency(poolId);
         uint128 assetUnitAmount = (10 ** hubRegistry.decimals(assetId.raw())).toUint128();
         uint128 poolUnitAmount = (10 ** hubRegistry.decimals(poolCurrency.raw())).toUint128();
-        uint128 poolAmountPerAsset = valuation.getQuote(assetUnitAmount, assetId, poolCurrency);
+        uint128 poolAmountPerAsset = valuation.getQuote(poolId, scId, assetId, assetUnitAmount);
 
         // Retrieve price by normalizing by pool denomination
         return d18(poolAmountPerAsset, poolUnitAmount);
```
lemunozm, line 265: Now we don't follow ERC-7726 maybe it makes sense to have `valuation.getPrice(poolId, scId, assetId)` and simplify this non-trivial computation into one readable line

src/valuations/OracleValuation.sol
```diff
@@ -0,0 +1,80 @@
...
+
+    //----------------------------------------------------------------------------------------------
+    // Administration
+    //----------------------------------------------------------------------------------------------
+
+    /// @inheritdoc IOracleValuation
+    function updateFeeder(PoolId poolId, address feeder_, bool canFeed) external {
+        require(hubRegistry.manager(poolId, msg.sender), NotHubManager());
+        feeder[poolId][feeder_] = canFeed;
+    }
```
lemunozm: I think we need to add an `UpdateFeeder` event here

test/common/mocks/MockValuation.sol
```diff
@@ -1,39 +1,47 @@
...
 
-    mapping(AssetId base => mapping(AssetId quote => Price)) public price;
+    constructor(IHubRegistry hubRegistry_) {
+        hubRegistry = hubRegistry_;
+    }
 
-    function setPrice(AssetId base, AssetId quote, D18 price_) external {
-        price[base][quote] = Price(price_, true);
-        D18 reciprocal = (price_.raw() != 0) ? price_.reciprocal() : d18(0);
-        price[quote][base] = Price(reciprocal, true);
```
lemunozm, line 28: Happy to know we were not using the reciprocal anywhere in the tests, I think at some point that was a requirement

src/valuations/OracleValuation.sol
```diff
@@ -0,0 +1,80 @@
...
+///         Prices should be denominated in the pool currency.
+///         Quorum is always 1, i.e. there is no aggregation of prices across multiple feeders.
+/// @dev    To set up, add a price feeder using hub.updateContract(), set this contract as the valuation
+///         for one or more assets, and set this contract as a hub manager, so it can call
+///         hub.updateHoldingValue().
+contract OracleValuation is IOracleValuation {
+    IHub public immutable hub;
+    address public immutable contractUpdater;
+    IHubRegistry public immutable hubRegistry;
+    uint16 public immutable localCentrifugeId;
```
lemunozm: `contractUpdater` and `localCentrifugeId` are unused. 

As a note, `contractUpdater` could only be used for spoke integrations

src/valuations/OracleValuation.sol
```diff
@@ -0,0 +1,80 @@
...
+
+import {IHub} from "../hub/interfaces/IHub.sol";
+import {IHubRegistry} from "../hub/interfaces/IHubRegistry.sol";
+
+/// @notice Provides an implementation for valuation of assets by trusted price feeders.
+///         Prices should be denominated in the pool currency.
+///         Quorum is always 1, i.e. there is no aggregation of prices across multiple feeders.
+/// @dev    To set up, add a price feeder using hub.updateContract(), set this contract as the valuation
+///         for one or more assets, and set this contract as a hub manager, so it can call
+///         hub.updateHoldingValue().
```
lemunozm:
```suggestion
/// @dev    To set up, add a price feeder using updateFeeder(), set this contract as the valuation
///         for one or more assets, and set this contract as a hub manager, so it can call
///         hub.updateHoldingValue().
```

lemunozm: Also double-checking if we really need to have `hub.updateHoldingValue()` only used by the pool manager. Could it not be permissionless? 🤔 cc @mustermeiszer

test/valuations/unit/OracleValuation.t.sol
```diff
@@ -0,0 +1,400 @@
...
+    address notManager = makeAddr("notManager");
+
+    OracleValuation valuation;
+
+    function setUp() public virtual {
+        _setupMocks();
+        _deployValuation();
+    }
+
+    function _setupMocks() internal {
```
lemunozm, line 44: Thanks to embrace the mocking style in the tests. I know at first seems more painful to set up, but IMO it's a win in the long term (fully decoupling tests and allows better to reach the 100% coverage in the contract)

test/valuations/Deployment.t.sol
```diff
@@ -14,14 +14,13 @@ contract ValuationsDeploymentTest is ValuationsDeployer, CommonDeploymentInputTe
...
-        assertEq(identityValuation.wards(nonWard), 0);
+    function testIdentityValuation() public view {
+        // dependencies set correctly
+        assertEq(address(identityValuation.hubRegistry()), address(hubRegistry));
+    }
 
+    function testOracleValuation() public view {
         // dependencies set correctly
-        assertEq(address(identityValuation.erc6909()), address(hubRegistry));
+        assertEq(address(oracleValuation.hubRegistry()), address(hubRegistry));
```
lemunozm, line 24:
```suggestion
        assertEq(address(oracleValuation.hubRegistry()), address(hubRegistry));
        assertEq(address(oracleValuation.hub()), address(hub));
```

lemunozm (commented):
I like the new contract but what's really a win is the Valuation refactor!!

test/valuations/unit/OracleValuation.t.sol
hieronx: Claude loved it!

src/valuations/OracleValuation.sol
hieronx: Good catch!

src/valuations/OracleValuation.sol
hieronx: Yeah I think it can potentially be permissionless, but not 100% sure

Regardless, I think we should automatically batch it. Just operationally much easier if this automatically happens.

src/hub/HubHelpers.sol
hieronx: `Holdings.update` requires `getQuote` though. So then we'd need to have both methods, which doesn't seem worth it. Would rather have this complexity here in the core, since valuations are modular contracts that should be easy to write for integrators.

script/ValuationsDeployer.s.sol
```diff
@@ -14,17 +15,14 @@ struct ValuationsReport {
 }
 
 contract ValuationsActionBatcher is HubActionBatcher {
-    function engageValuations(ValuationsReport memory report) public onlyDeployer {
-        report.identityValuation.rely(address(report.hub.common.root));
-    }
+    function engageValuations(ValuationsReport memory report) public onlyDeployer {}
```
wischli, line 18: Maybe-CR: Curious why we keep empty implementations here, maybe a missed TODO? In the same file, we are later using `batcher.engageValuations(_valuationsReport())` which does nothing.

src/common/interfaces/IValuation.sol
```diff
@@ -1,14 +1,17 @@
...
-/// Interface for asset conversions.
+/// Interface for valuation of assets, denominated in the pool currency.
 interface IValuation {
     /// @notice Returns the value of baseAmount of base in quote terms, e.g. 10 ETH (base) in USDC (quote).
-    /// @param base The asset in which the baseAmount is denominated in
-    /// @param quote The asset in which the user needs to value the baseAmount
+    /// @param assetId The asset in which the baseAmount is denominated in
     /// @param baseAmount The amount of base in base terms.
-    function getQuote(uint128 baseAmount, AssetId base, AssetId quote) external view returns (uint128 quoteAmount);
+    function getQuote(PoolId poolId, ShareClassId scId, AssetId assetId, uint128 baseAmount)
```
wischli, line 13: Nit: Natspec is missing the new params. Absolutely no blocker as we have many code paths in the repo with incomplete natspec. Can address this pre audit for the entire repo with the help of Claude Code.

src/valuations/OracleValuation.sol
```diff
@@ -0,0 +1,77 @@
...
+///         Quorum is always 1, i.e. there is no aggregation of prices across multiple feeders.
+/// @dev    To set up, add a price feeder using hub.updateFeeder(), set this contract as the valuation
+///         for one or more assets, and set this contract as a hub manager, so it can call
+///         hub.updateHoldingValue().
+contract OracleValuation is IOracleValuation {
+    IHub public immutable hub;
+    IHubRegistry public immutable hubRegistry;
+
+    mapping(PoolId => mapping(address => bool)) public feeder;
+    mapping(PoolId => mapping(ShareClassId => mapping(AssetId base => Price))) public price;
```
wischli: Nit question: Could we rename this to `pricePoolPerAsset` to be more explicit? Without the `OracleValuation` contract documentation, its not fully obvious IMO.

test/common/mocks/MockValuation.sol
wischli: Not sure if it was a real requirement but the reason was that using reciprocal has higher likelihood of reducing precision. So definitely a win!

test/valuations/unit/OracleValuation.t.sol
wischli: Such a beauty and 100% coverage 🚀

wischli (dismissed):
Amazing PR! Loving the concise description. Some questions and nits from my side, no blockers.

src/common/interfaces/IValuation.sol
hieronx: Yeah I like the idea of using Claude Code pre audit for this!

src/valuations/OracleValuation.sol
hieronx: Definitely!

script/ValuationsDeployer.s.sol
hieronx: I just left it for potential future additions here. Think that is okay but let me know what you prefer @wischli .

`batcher.engageValuations(_valuationsReport())` was already removed.

src/common/interfaces/IValuation.sol
wischli: Created a [Notion Ticket](https://www.notion.so/Update-natspec-docs-2642eac24e17803da479e455b96d2e05?source=copy_link) and marked myself as owner.

script/ValuationsDeployer.s.sol
wischli: Let's keep then!

wischli (approved):
Thanks for the renaming!

PR #596, opened 2025-09-04
Include CreateX-Forge inline to remove logs

hieronx:
I got tired of all the `Etching missing CreateX on chain: 313` logs in tests, and it's also not great for integrators.


lemunozm (approved):
Something that was already on my mind. Thanks for tackling this!

PR #468, opened 2025-06-24
Add invariant test suite

wischli:
## Summary

Adds a comprehensive invariant testing suite under `test/integration/recon-end-to-end/` using the [Recon/Chimera framework](https://book.getrecon.xyz). The suite deploys the full Hub + Spoke stack on-chain and uses Echidna (assertion mode) to explore multi-step call sequences that could violate protocol guarantees.

**29 new files, ~10,200 lines** covering 131 properties across vault operations, ERC-7540 compliance, deposit/redeem lifecycle, share accounting, escrow balance integrity, and authorization.

Also adds root-level config: `echidna.yaml`, `medusa.json`, `foundry.toml` remappings, and `lib/setup-helpers` submodule.

## What This Tests

The fuzzer calls target functions in random order with random inputs, then checks that protocol invariants still hold. Properties are organized into categories:

| Category | Count | Examples |
|----------|-------|---------|
| Vault max* correctness | 20 | deposit(maxDeposit) blocks further deposits; maxMint consistent with maxDeposit |
| ERC-7540 compliance | 13 | max* never reverts; preview* always reverts; claim > max always reverts |
| Deposit/redeem lifecycle | 24 | pending amounts track correctly; cancel amounts bounded; claim deltas match escrow |
| Share accounting | 12 | sum of balances = totalSupply; supply consistency; PPS never changes after user ops |
| Price & proportionality | 6 | deposit/redeem prices bounded by fulfillment price; asset-share proportionality |
| BalanceSheet queue state | 10 | queue counters consistent; nonace monotonic; reserve/unreserve integrity |
| Authorization | 3 | ward checks can't be bypassed; auth level correct; auth changes tracked |
| Doomsday edge cases | 12 | zero price handled gracefully; operations below max always succeed; accountValue never reverts |
| Inline target assertions | 31 | escrow balance deltas on issue/revoke, deposit/redeem claims, transfer restrictions |

Full table: [`test/integration/recon-end-to-end/properties-table.md`](test/integration/recon-end-to-end/properties-table.md)
Suite README: [`test/integration/recon-end-to-end/README.md`](test/integration/recon-end-to-end/README.md)

## Echidna Results

128 of 131 properties pass. The 3 failures are **acknowledged false positives**, not protocol bugs:

| Property | Status | Why |
|----------|--------|-----|
| `property_authorizationBypass` | ❌ Acknowledged | Admin can bypass BalanceSheet auth by skipping `updateBalanceSheetManager` — unrealistic operational mistake |
[...]

PR #599, opened 2025-09-04
Invariants: Investigate & fix outstanding critical & potential issues

wischli:
Fixed all outstanding critical and potential issues from #468, June 2025 (pre final v3 audits). Based against intermediary https://github.com/centrifuge/protocol-v3/pull/598 to reduce `main`-rebase noise. Note: There might be new issues emerging from running the suite due to the heavy amount of changes and rebases.

### Invariant Test Improvements

Zero Price Edge Case Handling
- Applied comprehensive zero price handling across all properties
- Prevents division by zero errors in price calculations

Async Vault Properties
- Corrected `maxDeposit` and `maxWithdraw` property calculations
- Fixed `maxMint` property for async vaults
- Added proper escrow edge case handling with meaningful catch cases

Infrastructure Improvements
- Removed mock `MessageProcessor` and `MessageDispatcher` in favor of real implementations
- Fixed asset ID resolution using proper `spoke.vaultToAssetId()` instead of hardcoded **pool currency**
- Updated setup permissions and managers configuration

Code Cleanup
- Removed outdated v2 invariant "fuzzing" suite
- Removed unnecessary `root_{cancel, schedule}Rely` function
- Refactored HubTargets after notify refactor
- Fix broken formatting
- Fix warnings 

### TODO

- [x] Receive review/guidance from Recon
- [x] Run Echidna locally for limited period
- [x] Inspect coverage
- [x] Run with suite & debug potentially emerging issues
- [ ] Format & apply import ordering


test/integration/recon-end-to-end/targets/SpokeTargets.sol
```diff
@@ -148,20 +148,20 @@ abstract contract SpokeTargets is BaseTargetFunctions, Properties {
 
     // Step 5 - set the request manager
     function spoke_setRequestManager(address vault) public asAdmin {
-        IBaseVault vaultInstance = IBaseVault(_getVault());
+        IBaseVault vaultInstance = IBaseVault(vault);
```
nican0r, line 151: @wischli could you explain the reason for this change here? 

The initial setup had been created so that the `ReconVaultManager` was the source of truth for the currently set vault. It seems like allowing a fuzzed address here for the vault would make it less likely to correctly set a valid request manager.

PR #602, opened 2025-09-09
Add hook managers

hieronx:
### Product requirements

* Allow users locally on the spoke chain (e.g. smart contracts) to manage members and freeze status. One use case is custom authentication contracts that should be reusable across all existing hooks.

### Design notes

* Hardcodes the spoke contract, but since this is 1) unlikely to be migrated after v3.1, 2) if absolutely necessary, hooks can be migrated quite easily, this seems okay.


onnovisser (approved):
LGTM

PR #603, opened 2025-09-10
Feat/gas failure protection

lemunozm:
This PR, takes into account:

1. Leave some gas to be able to enqueue a failed message. [ref](https://kflabs.slack.com/archives/C07PG2EUR9C/p1754943134412569)
2. Protect against DoS by not enough gas (based on EIP-150). [ref](https://kflabs.slack.com/archives/C07PG2EUR9C/p1757405470818009

**NOTE: 2. is already fixed by fixing 1.**

### DIFF
--- a/src/common/Gateway.sol
+++ b/src/common/Gateway.sol
@@ -112,7 +112,7 @@ contract Gateway is Auth, Recoverable, IGateway {
         } catch (bytes memory err) {
             bytes32 messageHash = keccak256(message);
             failedMessages[centrifugeId][messageHash]++;
-            emit FailMessage(centrifugeId, message, err);
+            emit FailMessage(centrifugeId, message, messageHash, err);
         }
     }
 
--- a/src/common/Gateway.sol
+++ b/src/common/Gateway.sol
@@ -107,10 +107,10 @@ contract Gateway is Auth, Recoverable, IGateway {
     }
 
     function _process(uint16 centrifugeId, bytes memory message) internal {
+        bytes32 messageHash = keccak256(message);
         try processor.handle{gas: gasleft() - GAS_FAIL_MESSAGE_STORAGE}(centrifugeId, message) {
-            emit ExecuteMessage(centrifugeId, message);
+            emit ExecuteMessage(centrifugeId, message, messageHash);
         } catch (bytes memory err) {
-            bytes32 messageHash = keccak256(message);
             failedMessages[centrifugeId][messageHash]++;
             emit FailMessage(centrifugeId, message, messageHash, err);
         }
@@ -124,7 +124,7 @@ contract Gateway is Auth, Recoverable, IGateway {
         failedMessages[centrifugeId][messageHash]--;
         processor.handle(centrifugeId, message);
 
-        emit ExecuteMessage(centrifugeId, message);
+        emit ExecuteMessage(centrifugeId, message, messageHash);
     }
 
     //----------------------------------------------------------------------------------------------
--- a/src/common/Gateway.sol
+++ b/src/common/Gateway.sol
@@ -102,12 +102,11 @@ contract Gateway is Auth, Recoverable, IGateway {
             uint256 executionGas = gasService.messageGasLimit(localCentrifugeId, message);
             require(gasleft() >= executionGas + GAS_FAIL_MESSAGE_STORAGE, NotEnoughGasToProcess());
 
-            _process(centrifugeId, message);
+            _process(centrifugeId, message, keccak256(message));
         }
     }
 
-    function _process(uint16 centrifugeId, bytes memory message) internal {
-        bytes32 messageHash = keccak256(message);
+    function _process(uint16 centrifugeId, bytes memory message, bytes32 messageHash) internal {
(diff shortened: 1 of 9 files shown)


src/common/Gateway.sol
```diff
@@ -31,6 +31,7 @@ contract Gateway is Auth, Recoverable, IGateway {
     using BytesLib for bytes;
     using TransientStorageLib for bytes32;
 
+    uint256 public constant GAS_FAIL_MESSAGE_STORAGE = 50_000;
```
lemunozm: I’ve seen that in different Ethereum version, different gas cost is used for some EVM instructions. Taking that into account, maybe we should make this configurable by Auth? Or moved to gasService?

src/common/Gateway.sol
```diff
@@ -31,9 +31,12 @@ contract Gateway is Auth, Recoverable, IGateway {
...
```
hieronx: Note: to be measured

src/common/Gateway.sol
```diff
@@ -96,7 +100,13 @@ contract Gateway is Auth, Recoverable, IGateway {
...
             remaining = remaining.slice(length, remaining.length - length);
 
-            try processor_.handle(centrifugeId, message) {
+            uint256 remainingGas = gasleft() - GAS_FAIL_MESSAGE_STORAGE;
+            uint256 executionGas = gasService.messageGasLimit(localCentrifugeId, message);
+
+            // We only want to fail inside `processor.handle()` by logic stuff, not by out of gas issues
+            require(remainingGas >= executionGas, NotEnoughGasToProcess());
+
+            try processor_.handle{gas: remainingGas}(centrifugeId, message) {
```
lemunozm: Important note, we only need to double check there is enough gas to process the message, but we must leave all the gas available to processor.handle() passing remainingGas.

Not used gas is useful again after `processor.handle()`

src/common/Gateway.sol
```diff
@@ -96,7 +100,13 @@ contract Gateway is Auth, Recoverable, IGateway {
             bytes memory message = remaining.slice(0, length);
             remaining = remaining.slice(length, remaining.length - length);
 
-            try processor_.handle(centrifugeId, message) {
+            uint256 remainingGas = gasleft() - GAS_FAIL_MESSAGE_STORAGE;
```
hieronx: This will fail with underflow error if `GAS_FAIL_MESSAGE_STORAGE < gasleft()`. Should we add an explicit error for this?

src/common/Gateway.sol
```diff
@@ -96,7 +100,13 @@ contract Gateway is Auth, Recoverable, IGateway {
...
+            uint256 executionGas = gasService.messageGasLimit(localCentrifugeId, message);
+
+            // We only want to fail inside `processor.handle()` by logic stuff, not by out of gas issues
+            require(remainingGas >= executionGas, NotEnoughGasToProcess());
```
hieronx: I think this is clear on its own:
```suggestion
            require(remainingGas >= executionGas, NotEnoughGasToProcess());
```

src/common/Gateway.sol
lemunozm: Probably the same error as the next line is enough. Good catch!

src/common/Gateway.sol
lemunozm: I left a value of 40K: 32-34K measured by benchmarking + 6-8K (some margin + the check after asking `gasleft()`). 

I think that extra is also needed because a super-long message can take more time to compute `keccak(message)`. For reference, the benchmark for a message of 5 bytes takes 32K gas, but for a message of 250 bytes it takes 34K gas

src/common/Gateway.sol
```diff
@@ -88,21 +92,27 @@ contract Gateway is Auth, Recoverable, IGateway {
...
 
-            try processor_.handle(centrifugeId, message) {
-                emit ExecuteMessage(centrifugeId, message);
-            } catch (bytes memory err) {
-                bytes32 messageHash = keccak256(message);
-                failedMessages[centrifugeId][messageHash]++;
-                emit FailMessage(centrifugeId, message, err);
-            }
+            uint256 executionGas = gasService.messageGasLimit(localCentrifugeId, message);
+            require(gasleft() >= executionGas + GAS_FAIL_MESSAGE_STORAGE, NotEnoughGasToProcess());
```
hieronx, line 103: Much more readable!

test/integration/ThreeChainEndToEnd.t.sol
```diff
@@ -153,6 +153,7 @@ contract ThreeChainEndToEndDeployment is EndToEndFlows {
 
         // C: Verify shares were minted
         assertEq(shareTokenB.balanceOf(INVESTOR_A), 0, "Shares should still be burned on chain B");
+        assertEq(shareTokenC.balanceOf(INVESTOR_A), amount, "Shares minted on chain C");
```
lemunozm, line 156: Unrelated, I'vet just realized that this check wasn't there

hieronx (approved):
🔥

PR #604, opened 2025-09-11
Ref: MultiAdapter approach without explicit proofs

lemunozm:
Thread: https://kflabs.slack.com/archives/C07PG2EUR9C/p1757574013006859?thread_ts=1756404586.718779&cid=C07PG2EUR9C


src/common/interfaces/IMultiAdapter.sol
```diff
@@ -28,7 +28,6 @@ interface IMultiAdapter is IAdapterBlockSendingExt, IMessageHandler {
         uint16[MAX_ADAPTER_COUNT] votes;
         /// @notice Each time adapters are updated, a new session starts which invalidates old votes
         uint64 sessionId;
-        bytes pending;
```
lemunozm, line 31: Note, we no longer need to store this, which can be big if it's a batch

test/common/unit/MultiAdapter.t.sol
```diff
@@ -245,278 +243,188 @@ contract MultiAdapterTestSetManager is MultiAdapterTest {
...
         multiAdapter.handle(REMOTE_CENT_ID, MESSAGE_1);
     }
 
-    function testErrEmptyMessage() public {
-        multiAdapter.setAdapters(REMOTE_CENT_ID, POOL_0, oneAdapter);
-
-        vm.prank(address(payloadAdapter));
-        vm.expectRevert(BytesLib.SliceOutOfBounds.selector);
-        multiAdapter.handle(REMOTE_CENT_ID, new bytes(0));
-    }
```
lemunozm, line 261: Right now, an empty message no longer dispatch `SliceOutOfBounds` because we don't need to read anything from the message in `MultiAdapter`. I think theoretically it should be ok to send an empty to the `Gateway` (which will discard it without fail), and I don't see any risk in this. 

Actually, it could be a way to test in production that messages through the adapters work without affecting the system.

src/common/interfaces/IMultiAdapter.sol
hieronx: Yeah this will be a significant gain, in terms of storage costs.

test/common/unit/MultiAdapter.t.sol
hieronx: Agree this should be fine!

src/common/MultiAdapter.sol
```diff
@@ -16,13 +15,11 @@ import {BytesLib} from "../misc/libraries/BytesLib.sol";
...
     using CastLib for *;
-    using MessageProofLib for *;
     using BytesLib for bytes;
     using ArrayLib for uint16[8];
     using MathLib for uint256;
 
     PoolId public constant GLOBAL_ID = PoolId.wrap(0);
-    uint8 public constant PRIMARY_ADAPTER_ID = 1;
 
     uint16 public immutable localCentrifugeId;
```
hieronx, line 24: Unrelated to this PR but:
```suggestion
    uint16 public immutable localCentrifugeId;
    
```

src/common/MultiAdapter.sol
```diff
@@ -190,22 +161,14 @@ contract MultiAdapter is Auth, IMultiAdapter {
...
-        emit SendPayload(centrifugeId, payloadId, payload, adapters_[0], adapterData, refund);
-
-        // Override the payload variable to send the proof to the remaining adapters
-        payload = MessageProofLib.createMessageProof(poolId, payloadHash);
-        for (uint256 i = 1; i < adapters_.length; i++) {
-            cost = adapters_[i].estimate(centrifugeId, payload, gasLimit);
-            adapterData = adapters_[i].send{value: cost}(centrifugeId, payload, gasLimit, refund);
-            emit SendProof(centrifugeId, payloadId, payloadHash, adapters_[i], adapterData);
+        bytes32 payloadId = keccak256(abi.encodePacked(localCentrifugeId, centrifugeId, keccak256(payload)));
+
```
hieronx: Nit:
```suggestion
```

src/common/MultiAdapter.sol
```diff
@@ -16,15 +15,14 @@ import {BytesLib} from "../misc/libraries/BytesLib.sol";
 
 contract MultiAdapter is Auth, IMultiAdapter {
     using CastLib for *;
-    using MessageProofLib for *;
     using BytesLib for bytes;
     using ArrayLib for uint16[8];
     using MathLib for uint256;
```
hieronx: Sorry, final nit while I'm going over it:
```suggestion
using MathLib for uint256;
using ArrayLib for uint16[8];
```

hieronx (approved):

src/common/MultiAdapter.sol
hieronx: Just committed it as I assume you are fine with it :)

PR #606, opened 2025-09-11
Refactor SCM to new Hub Request Manager

hieronx:
Should be merged after: https://github.com/centrifuge/protocol-v3/pull/617

### Product requirements

* Support setting a custom request manager per pool

### Design notes

* Request managers are set per `centrifugeId`. There is a hub request manager, stored in hub registry, and a spoke request manager, stored in spoke.
* Logic not related to requests and only related to share classes, has been moved to `ShareClassManager`. This is a shared contract across pools.
* Existing request management logic from `ShareClassManager` is moved into `HubRequestManager`. This is part of `src/vaults`, since it is closely tied to the `AsyncRequestManager`.
* Will require migrating `ShareClassManager` and `HubRegistry`
* Remaining misc fixes:
  * [x] Fix reentrancy protection

### TODOs

- [x] Fix fork tests

### Next PR
- [ ] Refactor deployment scripts to better support cross hub-spoke contracts (needed for `src/vaults` now but also `src/managers` soon)
- [x] Verify `HubRequestManager.t.sol` and `ShareClassManager.t.sol` unit tests still cover every path the old unit tests did
- [x] Consider allowing approve/issue/cancel etc methods to be called directly in the `BatchRequestManager`
- [x] Move out incoming gateway methods to another contract (`SpokeHandler`)
- [x] Add missing events and ID checks in `Hub`
- [x] Review whether `HubHelpers` can be merged back into `Hub`


script/TestData.s.sol
```diff
@@ -88,31 +91,44 @@ contract TestData is FullDeployer {
         _deploySyncDepositVault(centrifugeId, token, assetId);
     }
 
+    struct VaultState {
```
hieronx, line 95: Needed because of stack too deep issues

src/hub/interfaces/IHubRegistry.sol
```diff
@@ -11,8 +11,9 @@ interface IHubRegistry is IERC6909Decimals {
     event NewPool(PoolId poolId, address indexed manager, AssetId indexed currency);
     event UpdateManager(PoolId indexed poolId, address indexed manager, bool canManage);
     event SetMetadata(PoolId indexed poolId, bytes metadata);
-    event UpdateDependency(bytes32 indexed what, address dependency);
+    event UpdateDependency(PoolId indexed poolId, bytes32 indexed what, address dependency);
```
hieronx, line 16: I originally changed dependencies to be per pool, to use this to store the hub request manager. Later decided to introduce separate new storage. But I still feel having dependencies be indexed per pool makes sense, for more flexibilty.

script/FullDeployer.s.sol
```diff
@@ -15,15 +16,43 @@ import {
...
 
-contract FullActionBatcher is ExtendedHubActionBatcher, ExtendedSpokeActionBatcher, AdaptersActionBatcher {}
+struct FullReport {
+    HubReport hub;
+    HubRequestManager hubRequestManager;
+}
+
+contract FullActionBatcher is ExtendedHubActionBatcher, ExtendedSpokeActionBatcher, AdaptersActionBatcher {
+    function engageFull(FullReport memory report) public onlyDeployer {
+        // TODO: should be re-organized
```
hieronx, line 30: @lemunozm would appreciate your input on how to best structure this.

src/common/interfaces/IGatewayHandlers.sol
```diff
@@ -60,6 +60,20 @@ interface IHubGatewayHandler {
...
+    /// @param  assetId The asset id
+    /// @param  payload The payload to be processed by the request callback
+    /// @param  extraGasLimit Extra gas limit for the callback
+    function requestCallback(
+        PoolId poolId,
+        ShareClassId scId,
+        AssetId assetId,
+        bytes calldata payload,
+        uint128 extraGasLimit
+    ) external returns (uint256 cost);
```
lemunozm: I think this should not go in `IGatewayHandlers` interface, because this method is not called by the `Gateway` (or better said, `MessageProcessor`) as the others in the same interface. This is called from the `hubRequestManager` to the `Hub` to send a message.

I think it should be in its own interface that `Hub` implements, and `HubRequestManager` calls inside `hub/interfaces/` module

src/hub/Hub.sol
```diff
@@ -724,11 +625,8 @@ contract Hub is Multicall, Auth, Recoverable, IHub, IHubGatewayHandler, IHubGuar
     /// @dev Ensure the sender is authorized
     function _auth() internal auth {}
 
-    /// @dev Protect against reentrancy
-    function _protected() internal protected {}
-
-    /// @dev Ensure the method can be used without reentrancy issues, and the sender is a pool admin
-    function _isManager(PoolId poolId) internal protected {
+    /// @dev Ensure the sender is a pool admin
+    function _isManager(PoolId poolId) internal view {
```
lemunozm, line 605: Is it ok that `_isManager` is no longer `protected`?

src/hub/interfaces/IHubRegistry.sol
lemunozm: Agree!

src/hub/HubRegistry.sol
```diff
@@ -19,8 +19,9 @@ contract HubRegistry is Auth, IHubRegistry {
 
     mapping(PoolId => bytes) public metadata;
     mapping(PoolId => AssetId) public currency;
-    mapping(bytes32 => address) public dependency;
     mapping(PoolId => mapping(address => bool)) public manager;
+    mapping(PoolId => mapping(bytes32 => address)) public dependency;
+    mapping(PoolId => mapping(uint16 centrifugeId => address)) public hubRequestManager;
```
lemunozm: We know here all addresses will be `IHubRequestManager`:
```suggestion
    mapping(PoolId => mapping(uint16 centrifugeId => IHubRequestManager)) public hubRequestManager;
```

Later, it saves us some castings in `hub` and `hubHelpers`

src/vaults/HubRequestManager.sol
```diff
@@ -0,0 +1,853 @@
...
+contract HubRequestManager is Auth, IHubRequestManager {
+    using MathLib for *;
+    using CastLib for *;
+    using BytesLib for bytes;
+    using RequestMessageLib for *;
+    using RequestCallbackMessageLib for *;
+
+    IHubRegistry public immutable hubRegistry;
+
+    address public hub;
```
lemunozm: Regarding my first comment above, this could be:
```suggestion
    IHubRequestManagerCallback public hub;
```

src/hub/interfaces/IHubRequestManager.sol
```diff
@@ -0,0 +1,345 @@
+// SPDX-License-Identifier: BUSL-1.1
```
lemunozm: I think this file should only contain the strictly common stuff used in all hub request managers by the core. Which are only `request()`, `claimDeposit()` and `claimRedeem()`. Other methods are no longer mandatory by the core, and the rest of the interface will be in `src/vaults`, along with the implementation

test/integration/EndToEnd.t.sol
```diff
@@ -654,13 +663,28 @@ contract EndToEndFlows is EndToEndUtils {
...
+        uint32 depositEpochId = hub.hubRequestManager.nowDepositEpoch(shareClassId, assetId);
+        D18 pricePoolPerAsset = hub.hubHelpers.pricePoolPerAsset(poolId, shareClassId, assetId);
+        hub.hub.callRequestManager(
+            poolId,
+            assetId.centrifugeId(),
+            abi.encodeCall(
+                IHubRequestManager.approveDeposits,
+                (poolId, shareClassId, assetId, depositEpochId, amount, pricePoolPerAsset)
+            )
+        );
```
lemunozm, line 629: Because these calls seem complex and error prone with the encoding arguments, why not allow in `IHubRequestManager.approveDeposits` to be called by me FM directly?

approveDeposits (and similar methods) could check inside: 
```
require(`msg.sender == hub || hub.isManager(msg.sender))
```
or
```
require(`wards[msg.sender] == 1 || hub.isManager(msg.sender))
```

It means, you can add it to the multicall if used through `hub.callRequestManager()` or you can use it directly by just calling `hubRequestDeposit.approveDeposits()`

script/FullDeployer.s.sol
lemunozm: Ok... I was thinking about this. And I think we should not place HubRequestManager in `src/vaults`, it should be in another module (name to be decided).

The reason is that right now, `VaultDeployer` expects to inherit only from `SpokeDeployer`, and everything inside src/vaults should commit to this restriction. I think we should still keep things split in this regard.

This manager will live in the hub, so it should only inherit from `HubDeployer`. I would create `HubRequestManagersDeployer` (or other name) at the same level as `ValuationsDeployer`, and I will do later `ExtendedHubDeployer` extends from it:

Image

lemunozm: I understand some integrations will use later hub and spoke at the same time. I agree with this and this should be below `FullDeployer`, maybe `FullDeployerWithIntegrations` or so, and move `run()` method to `FullDeployerWithIntegrations`. 

But for me, it is important to distinguish that those integrations are different from the hubRequestManagers, They probably will use a hubRequestManager and other spoke contract to build themself. They seem to be in another layer on top of this one. 

So there is stuff under `src/vaults`, stuff under `src/<name-to-be-decided>`, and stuff under `src/integrations` that use both things and are built on top of them. Maybe `src/integrations` is already `src/managers`, which is a folder with different integrations, some of them using spoke, others hub, others both, but all these integrations are in another higher layer than hubRequestManagers

lemunozm (commented):
A huge refactor! But I agree it will leave the things super modular and more organized in smaller pieces

src/common/interfaces/IGatewayHandlers.sol
hieronx: Agree will update

src/hub/Hub.sol
hieronx: I think so, I don't believe there was ever any issues with this, just a precaution.

src/hub/HubRegistry.sol
hieronx: Makes sense will change

src/hub/interfaces/IHubRequestManager.sol
hieronx: Agree. I wonder if we could even remove `claimDeposit/Redeem` from `Hub`, and just expose these directly in `HubRequestManager`.

test/integration/EndToEnd.t.sol
hieronx: Hm, I see the advantage of having typed calls to the `hubRequestManager`, but I also think the downside is that it creates two paths to calling these methods, which is less than ideal. Will think about it more.

src/hub/interfaces/IHubRequestManager.sol
lemunozm: I also thought about that when reviewing this, I agree!

test/integration/EndToEnd.t.sol
lemunozm: I understand. To extend a bit more, the canonical way should always be to call them from the `hubRequestManager`, and only if you need to amalgamate in the same multicall, use `callRequestManager` as a tool offered to do that.

src/hub/Hub.sol
wischli: AFAICT, this change doesn't add any new attack vector indeed because `protected` only provides same-caller multicall protection.

src/hub/interfaces/IHubRequestManager.sol
wischli: > Agree. I wonder if we could even remove `claimDeposit/Redeem` from `Hub`, and just expose these directly in `HubRequestManager`.

IIUC, we would still have to relay the request callbacks then via the hub in order to not introduce the hub request manager as ward to the message dispatcher, right?  Also in favor of this change to reduce `Hub` and improve gas cost of notifications by reducing contract calls. 

Curious if you intend to call `notify` directly or still via `hub.callRequestManager`. In case of the former, requires adding `ReentrancyProtection` to `HubRequestManager` to add `protected` for the notify calls.

test/hub/integration/mocks/MockHubRequestManager.sol
```diff
@@ -0,0 +1,205 @@
...
+        return IHubGatewayHandler(hub).requestCallback(
+            poolId,
+            scId,
+            assetId,
+            RequestCallbackMessageLib.ApprovedDeposits(approvedAmount, pricePoolPerAsset.raw()).serialize(),
+            0
+        );
+    }
+
+    function approveRedeems(PoolId, ShareClassId, AssetId, uint32, uint128, D18) external override {}
```
wischli: Placeholder due to empty impl. Shouldn't this be utilized in some test?

wischli (commented):
Great and clean refactor! I need to put more time into the tests. AFAICT, they are insufficient compared to the pre-refactor state. Happy to pick up that task and add more tests!

src/hub/interfaces/IHubRequestManager.sol
hieronx: > IIUC, we would still have to relay the request callbacks then via the hub in order to not introduce the hub request manager as ward to the message dispatcher, right? Also in favor of this change to reduce Hub and improve gas cost of notifications by reducing contract calls.

Yeah like this is already the case for approveDeposits/Redeems etc.

> Curious if you intend to call notify directly or still via hub.callRequestManager. In case of the former, requires adding ReentrancyProtection to HubRequestManager to add protected for the notify calls.

I think adding protected modifier here would be good, yes.

test/integration/EndToEnd.t.sol
hieronx: Let's consider this in a followup PR.

script/FullDeployer.s.sol
```diff
@@ -16,33 +16,34 @@ import {
...
         // Rely others
-        report.hubRequestManager.rely(address(report.hub.hub));
-        report.hubRequestManager.rely(address(report.hub.hubHelpers));
+        report.batchRequestManager.rely(address(report.hub.hub));
+        report.batchRequestManager.rely(address(report.hub.hubHelpers));
 
         // File methods
-        report.hubRequestManager.file("hub", address(report.hub.hub));
+        report.batchRequestManager.file("hub", address(report.hub.hub));
+        report.batchRequestManager.file("gateway", address(report.hub.common.gateway));
```
lemunozm, line 40: To avoid migrating this, maybe we can do `hub.gateway()` in the few places we need the gateway (?)

script/FullDeployer.s.sol
```diff
@@ -16,33 +16,34 @@ import {
...
 
         // Rely Root
-        report.hubRequestManager.rely(address(report.hub.common.root));
+        report.batchRequestManager.rely(address(report.hub.common.root));
 
         // Rely others
-        report.hubRequestManager.rely(address(report.hub.hub));
-        report.hubRequestManager.rely(address(report.hub.hubHelpers));
+        report.batchRequestManager.rely(address(report.hub.hub));
+        report.batchRequestManager.rely(address(report.hub.hubHelpers));
```
lemunozm: I think we no longer need `hubHelpers`

src/hub/interfaces/IHubRequestManagerCallback.sol
```diff
@@ -0,0 +1,22 @@
+// SPDX-License-Identifier: GPL-2.0-or-later
+pragma solidity >=0.5.0;
+
+import {PoolId} from "../../common/types/PoolId.sol";
+import {AssetId} from "../../common/types/AssetId.sol";
+import {ShareClassId} from "../../common/types/ShareClassId.sol";
+
+interface IHubRequestManagerCallback {
+    /// @notice Handles a request callback originating from a spoke request.
```
lemunozm: NIT. Not sure if this doc line is accurate, this comes from hubRequestManager and goes to spoke

src/hub/Hub.sol
```diff
@@ -295,7 +246,7 @@ contract Hub is Multicall, Auth, Recoverable, IHub, IHubGatewayHandler, IHubGuar
     {
         _isManager(poolId);
 
-        hubRegistry.setHubRequestManager(poolId, centrifugeId, hubManager);
+        hubRegistry.setHubRequestManager(poolId, centrifugeId, IHubRequestManager(hubManager));
```
lemunozm: NIT. The API method could also use `IHubRequestManager`

test/hub/integration/mocks/MockHubRequestManager.sol
```diff
@@ -6,32 +6,32 @@ import {D18} from "../../../../src/misc/types/D18.sol";
...
     using RequestCallbackMessageLib for *;
 
     uint32 private constant MOCK_EPOCH = 1;
-    address public hub;
+    IHubRequestManagerCallback public hub;
 
     constructor(address hub_) {
-        hub = hub_;
+        hub = IHubRequestManagerCallback(hub_);
     }
```
lemunozm, line 22: Ultra NIT
```suggestion
constructor(IHubRequestManagerCallback hub_) {
        hub = hub_;
}
```

test/hub/integration/Cases.t.sol
```diff
@@ -103,152 +103,152 @@ contract TestCases is BaseTest {
...
-
-        RequestCallbackMessageLib.FulfilledRedeemRequest memory cb1 =
-            RequestCallbackMessageLib.deserializeFulfilledRedeemRequest(m1.payload);
-        assertEq(cb1.investor, INVESTOR);
-        assertEq(cb1.fulfilledAssetAmount, revokedAssetAmount);
-        assertEq(cb1.fulfilledShareAmount, APPROVED_SHARE_AMOUNT);
-        assertEq(cb1.cancelledShareAmount, SHARE_AMOUNT - APPROVED_SHARE_AMOUNT);
-    }
+    // /// forge-config: default.isolate = true
+    // function testDeposit() public returns (PoolId poolId, ShareClassId scId) {
```
lemunozm: I think we can remove those. EndToEnd should cover it. It's true that this ensures each parameter in the messages are correct, but the investment process can not work if it's not correct... Not sure what others think about this.

Ideally, I think this can be definitelly removed with a good BatchManagerRequest UTs

lemunozm (dismissed):
Minor comments and NITs that can be done in a following PR. Amazing job Jeroen!

script/FullDeployer.s.sol
hieronx: Let's consider in followup.

test/hub/integration/Cases.t.sol
hieronx: Removed!

lemunozm (approved):
💯

wischli (commented):
FWIW, post-mortem approval!

PR #607, opened 2025-09-11
Remove `scId` and `assetId` params from `setRequestManager`

hieronx:


src/spoke/interfaces/ISpoke.sol
```diff
@@ -27,10 +30,6 @@ struct ShareClassDetails {
 }
 
 struct ShareClassAsset {
-    /// @dev Manager that can send requests, and handles the request callbacks.
-    IRequestManager manager;
-    /// @dev Number of linked vaults.
-    uint32 numVaults;
     /// @dev The price per pool unit in asset denomination (POOL_UNIT/ASSET_UNIT)
     Price pricePoolPerAsset;
```
hieronx: Since only 1 field in this struct remaining, should we move it out of the struct? Wdyt @lemunozm ?

src/common/MessageDispatcher.sol
```diff
@@ -254,18 +254,13 @@ contract MessageDispatcher is Auth, IMessageDispatcher {
     }
 
     /// @inheritdoc IHubMessageSender
-    function sendSetRequestManager(PoolId poolId, ShareClassId scId, AssetId assetId, bytes32 manager) external auth {
-        if (assetId.centrifugeId() == localCentrifugeId) {
-            spoke.setRequestManager(poolId, scId, assetId, IRequestManager(manager.toAddress()));
+    function sendSetRequestManager(PoolId poolId, bytes32 manager) external auth {
+        if (poolId.centrifugeId() == localCentrifugeId) {
```
lemunozm: I think this is not correct; this will always be true if sending from the Hub. I think you need to pass the `centrifugeId` from outside, in the Hub you'll need to specify the chain to send this message

src/spoke/interfaces/ISpoke.sol
lemunozm: Would be great, yeah!

src/spoke/Spoke.sol
```diff
@@ -207,15 +207,10 @@ contract Spoke is Auth, Recoverable, ReentrancyProtection, ISpoke, ISpokeGateway
     }
 
     /// @inheritdoc ISpokeGatewayHandler
-    function setRequestManager(PoolId poolId, ShareClassId scId, AssetId assetId, IRequestManager manager)
-        public
-        auth
-    {
-        _shareClass(poolId, scId); // Check existence
-        ShareClassAsset storage assetInfo_ = assetInfo[poolId][scId][assetId];
-        require(assetInfo_.numVaults == 0, MoreThanZeroLinkedVaults());
```
lemunozm, line 216: Should this check still be there to avoid the pool manager to modify the vaults request manager in the middle of some investment process?

test/spoke/unit/Spoke.t.sol
```diff
@@ -632,19 +633,19 @@ contract SpokeTestSetRequestManager is SpokeTest {
         spoke.linkVault(POOL_A, SC_1, ASSET_ID_6909_1, vault);
 
         vm.prank(AUTH);
-        vm.expectRevert(ISpoke.MoreThanZeroLinkedVaults.selector);
-        spoke.setRequestManager(POOL_A, SC_1, ASSET_ID_6909_1, requestManager);
+        // This test is no longer applicable since request manager is now at pool level
```
lemunozm: I think it's still applicable by my above comment, if finally not needed, I would remove the whole test case, given the success test is already checked

lemunozm (commented):
Like the simplification!

Just one question, what's the reason this requirement has changed?

src/common/MessageDispatcher.sol
hieronx: Yeah I noticed too, just updated!

src/spoke/Spoke.sol
hieronx: Oops yes

src/spoke/interfaces/ISpoke.sol
```diff
@@ -16,7 +16,10 @@ import {IVaultFactory} from "../factories/interfaces/IVaultFactory.sol";
 
 /// @dev Centrifuge pools
 struct Pool {
-    uint256 createdAt;
+    /// @dev Timestamp of pool creation.
+    uint64 createdAt;
+    /// @dev Number of linked vaults.
+    uint32 numVaults;
```
onnovisser: if `numVaults` is no longer checked when setting the request manager, is it still used anywhere at all?

src/spoke/interfaces/ISpoke.sol
hieronx: Done!

src/spoke/interfaces/ISpoke.sol
hieronx: It is checked now again! Should have been

onnovisser:
Do we keep track of breaking changes somewhere, so it's easier to update the SDK after this is released?

hieronx:
> Do we keep track of breaking changes somewhere, so it's easier to update the SDK after this is released?

Good reminder, added labels so we can easily filter later.

onnovisser (approved):

mustermeiszer (commented):
@wischli @lemunozm @hieronx 

Do we have a ticket for writing the migratin from the existing storage we have in prod?

mustermeiszer:
Added [here](https://www.notion.so/Migrate-AsyncRequestManagers-2702eac24e17804fae96db18ae64ccac)

PR #611, opened 2025-09-15
Ref: Move adapter manager to Gateway

lemunozm:
### Changes

- Move adapter's manager to the `Gateway`
- Move `blockOutgoing()` method to the `Gateway`
- Remove `MultiAdapter.execute()`
- Add missing auth to `MessageDispatcher`


src/common/Gateway.sol
```diff
@@ -41,14 +41,18 @@ contract Gateway is Auth, Recoverable, IGateway {
     IRoot public immutable root;
     IGasService public gasService;
     IMessageProcessor public processor;
-    IAdapterBlockSendingExt public adapter;
+    IAdapter public adapter;
+
+    // management
```
```suggestion
    // Management
```

src/common/Gateway.sol
```diff
@@ -41,14 +41,18 @@ contract Gateway is Auth, Recoverable, IGateway {
...
+    mapping(PoolId => address) public manager;
 
     // Outbound & payments
     bool public transient isBatching;
     uint256 public transient fuel;
     address public transient transactionRefund;
     uint128 public transient extraGasLimit;
     mapping(PoolId => Funds) public subsidy;
```
hieronx, line 53: Unrelated nit:
```suggestion
uint256 public transient fuel;
bool public transient isBatching;
uint128 public transient extraGasLimit;
address public transient transactionRefund;
```

src/common/Gateway.sol
```diff
@@ -338,6 +348,13 @@ contract Gateway is Auth, Recoverable, IGateway {
...
     }
 
+    /// @inheritdoc IGateway
+    function blockOutgoing(PoolId poolId, bool isBlocked) external {
+        require(msg.sender == manager[poolId], ManagerNotAllowed());
+        isOutgoingBlocked[poolId] = isBlocked;
+        emit BlockOutgoing(poolId, isBlocked);
+    }
+
     //----------------------------------------------------------------------------------------------
```
hieronx, line 357: Entirely unrelated to this PR but now that the proof logic was removed, I think `_process` can be merged back into `handle`

src/common/MultiAdapter.sol
```diff
@@ -149,13 +141,6 @@ contract MultiAdapter is Auth, IMultiAdapter {
         }
     }
 
-    /// @inheritdoc IMultiAdapter
```
hieronx, line 152: With the removal of recovery executions, I would merge `_handle` back into `handle`

src/common/Gateway.sol
```diff
@@ -3,11 +3,11 @@ pragma solidity 0.8.28;
...
 import {IRoot} from "./interfaces/IRoot.sol";
+import {IAdapter} from "./interfaces/IAdapter.sol";
 import {IGateway} from "./interfaces/IGateway.sol";
 import {IGasService} from "./interfaces/IGasService.sol";
 import {IMessageSender} from "./interfaces/IMessageSender.sol";
 import {IMessageHandler} from "./interfaces/IMessageHandler.sol";
-import {IAdapterBlockSendingExt} from "./interfaces/IAdapter.sol";
 import {IMessageProcessor} from "./interfaces/IMessageProcessor.sol";
 
 import {Auth} from "../misc/Auth.sol";
```
hieronx, line 13: I noticed the natspec comment for this comment still mentions proofs, can you update this?

src/common/MultiAdapter.sol
```diff
@@ -26,9 +26,7 @@ contract MultiAdapter is Auth, IMultiAdapter {
     IMessageHandler public gateway;
```
hieronx, line 26: Another comment unrelated to this PR: it would be good to document in the interface file in `setAdapters`,  what a recovery adapter is. Maybe adding the example we discussed in Slack. I think most readers of the code will struggle to understand the difference between quorum, threshold, and recovery index.

src/common/Gateway.sol
```diff
@@ -41,14 +41,18 @@ contract Gateway is Auth, Recoverable, IGateway {
     IRoot public immutable root;
```
hieronx, line 40: Can you move the line with `GAS_FAIL_MESSAGE_STORAGE` below `GLOBAL_POT`?

hieronx (commented):
No substantial comments on the PR, looks great! Just some nits and unrelated improvement suggestions.

src/common/Gateway.sol
lemunozm: The why `_process` is extracted out is to be able to benchmark correctly the worst case: https://github.com/centrifuge/protocol-v3/blob/main/test/common/unit/Gateway.t.sol#L350 without the overhead that `handle` adds at the beginning

src/common/MultiAdapter.sol
lemunozm: Good point! Finally!

src/common/Gateway.sol
```diff
@@ -19,8 +19,7 @@ import {TransientStorageLib} from "../misc/libraries/TransientStorageLib.sol";
 import {Recoverable, IRecoverable, ETH_ADDRESS} from "../misc/Recoverable.sol";
 
 /// @title  Gateway
-/// @notice Routing contract that forwards outgoing messages to multiple adapters (1 full message, n-1 proofs)
-///         and validates that multiple adapters have confirmed a message.
+/// @notice Routing contract that forwards outgoing messages through an adapters
```
```suggestion
/// @notice Routing contract that forwards outgoing messages through one or more adapters.
```

src/common/Gateway.sol
```diff
@@ -19,8 +19,7 @@ import {TransientStorageLib} from "../misc/libraries/TransientStorageLib.sol";
...
 ///
 ///         Supports batching multiple messages, as well as paying for methods manually or through pool-level subsidies.
 ///
```
hieronx, line 25: Also need to update `and proofs` on line 27

hieronx (dismissed):

src/common/Gateway.sol
lemunozm: From gateway perspective, it's always one adapter

src/common/Gateway.sol
lemunozm: I think you're seeing an old commit, that should be fixed in the latest, but agree with "and proofs" which I missed

src/common/Gateway.sol
hieronx: "an adapter." then

hieronx (approved):

PR #612, opened 2025-09-16
Feat: add recovery adapter impl and end-to-end tests

lemunozm:
- Add simple recovery adapter implementation
- Add EndToEnd that verifies the recovery system works properly with the new recovery adapter

Blocked by: https://github.com/centrifuge/protocol-v3/pull/611


src/adapters/RecoveryAdapter.sol
```diff
@@ -0,0 +1,31 @@
...
+    IMessageHandler public immutable entrypoint;
+
+    constructor(IMessageHandler entrypoint_, address deployer) Auth(deployer) {
+        entrypoint = entrypoint_;
+    }
+
+    /// @inheritdoc IMessageHandler
+    function handle(uint16 centrifugeId, bytes calldata message) external auth {
+        entrypoint.handle(centrifugeId, message);
+    }
```
lemunozm, line 20: I would say we don't need any event here because it will be emitted in MultiAdapter, and we can know if it was a recovery or not by the adapter address, but I wanted to double-check

hieronx: Agree seems unnecessary

hieronx (dismissed):

lemunozm:
Will merge after #611

hieronx (approved):

PR #613, opened 2025-09-16
root messages only from mainnet

lemunozm:
Task: https://www.notion.so/Allow-ScheduleRely-message-only-from-Ethereum-mainnet-22d2eac24e1780059962d2d606af4df8?source=copy_link

Protect the `scheduleRely` from being used by chains that could be vulnerable.

### DIFF
--- a/src/common/MessageProcessor.sol
+++ b/src/common/MessageProcessor.sol
@@ -26,6 +26,8 @@ import {CastLib} from "../misc/libraries/CastLib.sol";
 import {BytesLib} from "../misc/libraries/BytesLib.sol";
 import {IRecoverable} from "../misc/interfaces/IRecoverable.sol";
 
+uint16 constant MAINNET_CENTRIFUGE_ID = 1;
+
 contract MessageProcessor is Auth, IMessageProcessor {
     using CastLib for *;
     using MessageLib for *;
@@ -68,11 +70,12 @@ contract MessageProcessor is Auth, IMessageProcessor {
     /// @inheritdoc IMessageHandler
     function handle(uint16 centrifugeId, bytes calldata message) external auth {
         MessageType kind = message.messageType();
-        uint16 sourceCentrifugeId = message.messageSourceCentrifugeId();
 
+        uint16 sourceCentrifugeId = message.messageSourceCentrifugeId();
         require(sourceCentrifugeId == 0 || sourceCentrifugeId == centrifugeId, InvalidSourceChain());
 
         if (kind == MessageType.ScheduleUpgrade) {
+            require(centrifugeId == MAINNET_CENTRIFUGE_ID, OnlyFromMainnet());
             MessageLib.ScheduleUpgrade memory m = message.deserializeScheduleUpgrade();
             root.scheduleRely(m.target.toAddress());
         } else if (kind == MessageType.CancelUpgrade) {
--- a/src/common/MessageProcessor.sol
+++ b/src/common/MessageProcessor.sol
@@ -26,12 +26,12 @@ import {CastLib} from "../misc/libraries/CastLib.sol";
 import {BytesLib} from "../misc/libraries/BytesLib.sol";
 import {IRecoverable} from "../misc/interfaces/IRecoverable.sol";
 
-uint16 constant MAINNET_CENTRIFUGE_ID = 1;
-
 contract MessageProcessor is Auth, IMessageProcessor {
     using CastLib for *;
     using MessageLib for *;
     using BytesLib for bytes;
+
+    uint16 public constant MAINNET_CENTRIFUGE_ID = 1;
 
     IRoot public immutable root;
     ITokenRecoverer public immutable tokenRecoverer;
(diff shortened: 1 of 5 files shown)


src/common/MessageProcessor.sol
```diff
@@ -68,11 +70,12 @@ contract MessageProcessor is Auth, IMessageProcessor {
     /// @inheritdoc IMessageHandler
     function handle(uint16 centrifugeId, bytes calldata message) external auth {
         MessageType kind = message.messageType();
-        uint16 sourceCentrifugeId = message.messageSourceCentrifugeId();
 
+        uint16 sourceCentrifugeId = message.messageSourceCentrifugeId();
         require(sourceCentrifugeId == 0 || sourceCentrifugeId == centrifugeId, InvalidSourceChain());
```
lemunozm, line 78: I think we no longer need it and is difficult to understand and maintain correctly for all messages. @hieronx do you think this is still needed or can be removed?

src/common/MessageProcessor.sol
```diff
@@ -26,6 +26,8 @@ import {CastLib} from "../misc/libraries/CastLib.sol";
 import {BytesLib} from "../misc/libraries/BytesLib.sol";
 import {IRecoverable} from "../misc/interfaces/IRecoverable.sol";
 
+uint16 constant MAINNET_CENTRIFUGE_ID = 1;
```
hieronx: Any specific reason to put this outside instead of inside the contract?

src/common/MessageProcessor.sol
```diff
@@ -68,11 +70,12 @@ contract MessageProcessor is Auth, IMessageProcessor {
...
-        uint16 sourceCentrifugeId = message.messageSourceCentrifugeId();
 
+        uint16 sourceCentrifugeId = message.messageSourceCentrifugeId();
         require(sourceCentrifugeId == 0 || sourceCentrifugeId == centrifugeId, InvalidSourceChain());
 
         if (kind == MessageType.ScheduleUpgrade) {
+            require(centrifugeId == MAINNET_CENTRIFUGE_ID, OnlyFromMainnet());
             MessageLib.ScheduleUpgrade memory m = message.deserializeScheduleUpgrade();
             root.scheduleRely(m.target.toAddress());
         } else if (kind == MessageType.CancelUpgrade) {
```
hieronx, line 84: We should also enforce only `MAINNET_CENTRIFUGE_ID` for `CancelUpgrade`, otherwise an exploited network can block all upgrades from mainnet.

src/common/MessageProcessor.sol
hieronx: This is still a way to prevent an exploited network from affecting pools which are not deployed right? Or maybe it's not needed anymore because of per pool adapters. Need to think about this more.

src/common/MessageProcessor.sol
hieronx: Actually, we should also only allow `MAINNET_CENTRIFUGE_ID` for `RecoverTokens`. Again, otherwise a rogue network can move all tokens.

hieronx: In the same direction: how do we prevent against a rogue network doing arbitrary `SetPoolAdaptersManager` calls for other pools? 🤔

lemunozm: We're protected actually, thanks to `messageSourceCentrifugeId`, because it checks that the pool used is received from the correct hub chain. So that's a reason not to remove that.

The same applies to `SetPoolAdapters` message.

src/common/MessageProcessor.sol
lemunozm: Actually there is a reason: https://github.com/centrifuge/protocol-v3/pull/613#discussion_r2352584928

src/common/MessageProcessor.sol
lemunozm: I feel constants outside the contract are better later when accessing them from tests, but I don't have a strong opinion. Do you prefer inside?

src/common/MessageProcessor.sol
hieronx: > The same applies to SetPoolAdapters message.

But this i snot true because of https://github.com/centrifuge/protocol-v3/blob/e776912409b25341deedff12cd74f9c215592621/src/common/libraries/MessageLib.sol#L169, right?

hieronx: (also, shouldn't we rename `SetPoolAdaptersManager` to `SetGatewayManager`?)

src/common/MessageProcessor.sol
hieronx: I think we use constants within the contract everywhere else now. Benefit is just that it is readable by other contracts (through a view method).

src/common/MessageProcessor.sol
lemunozm: Yeah, that was wrong, but it's fixed in https://github.com/centrifuge/protocol-v3/pull/611/files#diff-aabc3827b354a2c9bdb5ac997f51c78f3f32309e56b43335833d609ccb1c5a1cL168

src/common/MessageProcessor.sol
lemunozm: Good argument

hieronx (approved):

lemunozm:
For tests have failed due http errors, so should be good

PR #615, opened 2025-09-17
fix: Transfer shares

onnovisser:
Update issuance in ShareClassManager, call SnapshotHook and add extraGasLimit


test/hooks/mocks/MockSnapshotHook.sol
```diff
@@ -7,8 +7,21 @@ import {ISnapshotHook} from "../../../src/common/interfaces/ISnapshotHook.sol";
...
+
+    function onTransfer(
+        PoolId poolId,
+        ShareClassId scId,
+        uint16 fromCentrifugeId,
+        uint16 toCentrifugeId,
+        uint128 sharesTransferred
+    ) external {
+        transfers[poolId][scId][fromCentrifugeId] += sharesTransferred;
+        transfers[poolId][scId][toCentrifugeId] += sharesTransferred;
```
lemunozm: Should not be instead: 
```suggestion
        transfers[poolId][scId][fromCentrifugeId][toCentrifugeId] += sharesTransferred;
```
?

test/integration/ThreeChainEndToEnd.t.sol
```diff
@@ -124,7 +130,7 @@ contract ThreeChainEndToEndDeployment is EndToEndFlows {
 
         vm.prank(INVESTOR_A);
         sB.spoke.crosschainTransferShares{value: GAS}(
-            sC.centrifugeId, POOL_A, SC_1, INVESTOR_A.toBytes32(), amount, SHARE_HOOK_GAS
+            sC.centrifugeId, POOL_A, SC_1, INVESTOR_A.toBytes32(), amount, 0, SHARE_HOOK_GAS
```
lemunozm: NIT. I would rename SHARE_HOOK_GAS and I would use it instead of 0 for reference:

```suggestion
            sC.centrifugeId, POOL_A, SC_1, INVESTOR_A.toBytes32(), amount, HOOG_GAS, HOOK_GAS
```

test/integration/ThreeChainEndToEnd.t.sol
```diff
@@ -124,7 +130,7 @@ contract ThreeChainEndToEndDeployment is EndToEndFlows {
...
         );
         assertEq(shareTokenB.balanceOf(INVESTOR_A), 0, "Shares should be burned on chain B");
```
lemunozm: I think we need to add here a check to confirm `onTransfer()` is called using the mock. Something like:

```
assertEq(snapshotHook.transfers(..) == amount)
```

test/hooks/mocks/MockSnapshotHook.sol
onnovisser: makes sense 👍

test/integration/ThreeChainEndToEnd.t.sol
onnovisser: 👍

lemunozm (dismissed):
LGTM!

lemunozm:
The tests are failing by a previous non deterministic issue:
To fix it add to `hub/spoks/vaults /Deployment.t.sol` The following:

```diff
    function testGatewayExt(address nonWard) public view {
        // permissions set correctly
        vm.assume(nonWard != address(root)); // From common
        vm.assume(nonWard != address(messageDispatcher)); // From common
        vm.assume(nonWard != address(messageProcessor)); // From common
+       vm.assume(nonWard != address(guardian)); // From common
        vm.assume(nonWard != address(multiAdapter)); // From common
        vm.assume(nonWard != address(spoke)); // From spoke
        vm.assume(nonWard != address(balanceSheet)); // From spoke
        vm.assume(nonWard != address(vaultRouter));

        assertEq(gateway.wards(address(vaultRouter)), 1);
        assertEq(gateway.wards(nonWard), 0);
    }

**Review by lemunozm — DISMISSED** [review]

File: src/spoke/Spoke.sol [hunk 4]
```diff
@@ -96,6 +96,7 @@ contract Spoke is Auth, Recoverable, ReentrancyProtection, ISpoke, ISpokeGateway
         ShareClassId scId,
         bytes32 receiver,
         uint128 amount,
+        uint128 extraGasLimit,
```
hieronx, line 93: We need to keep supporting the current interface. Can you add an overloaded function with the old interface, where `extraGasLimit` is passed as 0?

hieronx (changes requested):

src/spoke/Spoke.sol
hieronx: It is required to support https://github.com/grove-labs/grove-alm-controller/blob/dev/src/libraries/CentrifugeLib.sol#L102

lemunozm: But we will break their integrations with the new spoke, isn't? Why is this integration special?

hieronx: They are not hardcoding the spoke so we won't break it. They cannot migrate these contracts.

src/spoke/Spoke.sol
```diff
@@ -96,23 +96,22 @@ contract Spoke is Auth, Recoverable, ReentrancyProtection, ISpoke, ISpokeGateway
...
-        IShareToken share = IShareToken(shareToken(poolId, scId));
-        require(centrifugeId != sender.localCentrifugeId(), LocalTransferNotAllowed());
-        require(
-            share.checkTransferRestriction(msg.sender, address(uint160(centrifugeId)), amount),
-            CrossChainTransferNotAllowed()
-        );
-
-        share.authTransferFrom(msg.sender, msg.sender, address(this), amount);
-        share.burn(address(this), amount);
+        _crosschainTransferShares(centrifugeId, poolId, scId, receiver, amount, 0, remoteExtraGasLimit);
```
lemunozm: NIT. you can call directly to `crosschainTransferShares` is tagged as public (so you don't need the private one)

lemunozm (dismissed):

src/spoke/Spoke.sol
onnovisser: will the duplicate `payTransaction` modifier be a problem?

lemunozm: It should be only in the lower one. But never mind, I will remove this in my PR and I will organize it there.

lemunozm: @onnovisser if you rebase, you probably can get this simplified, because `payTransaction` was removed

src/spoke/Spoke.sol
lemunozm: @hieronx the incomming implementation will refund the excess payment to the caller using `gateway.withdrawSubsidy`. IIUC, grove will use the exact cost upfront for sending that message. I ask this because if they add more gas, the refund will fail because of: https://github.com/centrifuge/protocol-v3/blob/feat/allow-multiple-gateway-mgrs/src/common/Gateway.sol#L278 and by the fact that grove impl does not have a `receive()` method in their contract.

I think it's ok, because the caller needs to estimate the cost, so the value will always match exactly. Just double-checking this

lemunozm (approved):
Thanks for this Onno!

hieronx (approved):

PR #616, opened 2025-09-17
Minor gateway changes

lemunozm:
- Makes block outgoing to dispatch an error instead of add the message to unpaid storage
- Add a way to withdraw funds from subsidized pools

### DIFF
--- a/src/common/interfaces/IGateway.sol
+++ b/src/common/interfaces/IGateway.sol
@@ -37,6 +37,7 @@ interface IGateway is IMessageHandler, IMessageSender, IRecoverable {
 
     event SetRefundAddress(PoolId poolId, IRecoverable refund);
     event SubsidizePool(PoolId indexed poolId, address indexed sender, uint256 amount);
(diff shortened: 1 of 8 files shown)


src/common/interfaces/IGateway.sol
```diff
@@ -111,6 +118,9 @@ interface IGateway is IMessageHandler, IMessageSender, IRecoverable {
     /// @notice Pay upfront to later be able to subsidize messages associated to a pool
     function subsidizePool(PoolId poolId) external payable;
```
hieronx: Maybe a bit cleaner: `depositSubsidy` and `withdrawSubsidy`?

hieronx (dismissed):
LGTM

src/common/interfaces/IGateway.sol
lemunozm: I like it! I wasn't confortable with the naming

hieronx (approved):

PR #617, opened 2025-09-18
Ref: only subsidized payments

lemunozm:
Blocked by: https://github.com/centrifuge/protocol-v3/pull/616

Major changes:

- [x] Remove transactional payments from the Gateway and related contracts
- [x] Actions that send a message will return the transaction cost to later ease integrators to determine the final cost.
  - [x] Send returning cost
  - [x] MessageDispatching returning message cost
  - [x] ~endBatching() returning cost~ (next PR: https://github.com/centrifuge/protocol-v3/pull/618)
- [x] `registerAsset()` and `crosschainTransferShares()` requires payment that will subsidize under the hood.

NOTE: In a next PR I will do the batching permissionless which means more clean ups removing gateway dependencies. Please when review, take into account that the batching process will be simplified in the next PR: https://github.com/centrifuge/protocol-v3/pull/618


src/hub/Hub.sol
```diff
@@ -125,8 +117,6 @@ contract Hub is Multicall, Auth, Recoverable, IHub, IHubGatewayHandler, IHubGuar
     /// @inheritdoc IHub
     function notifyDeposit(PoolId poolId, ShareClassId scId, AssetId assetId, bytes32 investor, uint32 maxClaims)
         external
-        payable
```
lemunozm, line 128: I understand this method does not require protection againts someone calling to drain the subsidized funds, because they can only be called if there are shares/assets, and once called, it will not send the message again.

I think those fees must be taken up front in the investment process by the pool manager to the investor.

lemunozm:
Fork tests failing because: https://kflabs.slack.com/archives/C07PG2EUR9C/p1758210962120689

src/common/interfaces/IGuardianActions.sol
```diff
@@ -5,7 +5,5 @@ import {PoolId} from "../types/PoolId.sol";
 import {AssetId} from "../types/AssetId.sol";
 
 interface IHubGuardianActions {
-    /// @notice Creates a new pool.
```
hieronx, line 8: Why remove natspec here?

src/common/Gateway.sol
```diff
@@ -169,61 +167,49 @@ contract Gateway is Auth, Recoverable, IGateway {
...
-            require(cost <= fuel, NotEnoughTransactionGas());
-            fuel -= cost;
-        } else {
-            // Subsidized pool payment
-            if (cost > subsidy[paymentPoolId].value) {
-                _requestPoolFunding(paymentPoolId);
-            }
+        if (cost > subsidy[paymentPoolId].value) {
+            _requestPoolFunding(paymentPoolId);
+        }
```
```suggestion
        if (cost > subsidy[paymentPoolId].value) _requestPoolFunding(paymentPoolId);
```

src/common/Gateway.sol
```diff
@@ -295,40 +280,6 @@ contract Gateway is Auth, Recoverable, IGateway {
         emit WithdrawSubsidy(poolId, to, amount);
     }
 
-    /// @inheritdoc IGateway
```
hieronx, line 298: Love the cleanup here

src/spoke/Spoke.sol
```diff
@@ -108,6 +102,8 @@ contract Spoke is Auth, Recoverable, ReentrancyProtection, ISpoke, ISpokeGateway
         share.authTransferFrom(msg.sender, msg.sender, address(this), amount);
         share.burn(address(this), amount);
 
+        gateway.depositSubsidy{value: msg.value}(poolId);
```
hieronx: Could we actually return the unused cost here, at this layer?

hieronx (dismissed):

src/common/interfaces/IGuardianActions.sol
lemunozm: A typo!

src/spoke/Spoke.sol
lemunozm: Yeah, I was back and forth with that here. Still not sure what's better:

1. We can call sendInitiateTransferShares first and get the cost, and later subsidize with the exact value.
   - PROS:
     - Exact amount
   - CONS:
     - If there is no subsidized amount in the pool, the call will fail even though the user has paid correctly. This could be not very pleasan, although it should be sporadic.
2. Or we can subsidize with msg.value and later call sendIntitateTransferShares (current impl).
   - PROS:
     - If the gas provided is enough, we're sure the message will be sent independently of the subsidized funds.
   - CONS:
     - We don't know the exact amount to pay. So the caller must estimate this. But anyway they will need to estimate it in 1. right?
    
BTW, the current implementation is wrong. We need to check `msg.value > cost`

lemunozm: > If there is no subsidized amount in the pool, the call will fail even though the user has paid correctly. This could be not very pleasan, although it should be sporadic.

The big question, is, are we ok with this cons? It could be pretty annoying in terms of UX, to see the transaction was successful but unpaid, but actually it was paid (gas lost for the user).

lemunozm: I have a fresh idea with best of both approaches, will upload a commit with this

hieronx: Thanks for taking me along in your live thought process 😄

lemunozm: Before submitting, I think the most correct solution is:

```solidity
uint256 cost = sender.sendRegisterAsset(centrifugeId, assetId, decimals);
        if (cost == 0) {
            payable(msg.sender).call{value: msg.value}("");
        } else {
            gateway.depositSubsidy{value: cost}(PoolId.wrap(0));
        }
```

- If the user doesn't pay `cost` and the message was sent, it will fail.
- If the user pays `cost` but not subsidized funds, the message is added to the unpaid, and the user will be refunded, to later allow them to call `repay()` without losing funds.
- The pool is subsidized with `cost` if the message was sent.


The other option is to fail if no subsidized funds. Tell me if you agree.

lemunozm: Although, in most cases, the user will call `crosschainTransferShares{value: exactCost}()` which makes me lean towards solution 2. again because it will never fail due a unsubsidized pool:

```solidity
gateway.depositSubsidy{value: cost}(PoolId.wrap(0));
uint256 cost = sender.sendRegisterAsset(centrifugeId, assetId, decimals);
require(msg.value > cost);
```

lemunozm: Sorry for doing so many back and forth with this. After testing almost every possible solution. I lean towards for option 2 like:

```solidity
gateway.depositedFunds{value: msg.value}()
uint256 cost = sender.send*();
require(msg.value >= cost)
```

- The thing is that we never want the state where some integrator call fails because lack of subsidized funds if they're already providing those in the call.
- Also, the caller needs a way to estimate the call to at least `cost` value.

The point is that using `cost` is to save us from estimate inside the method, but it will not save the caller to anyways estimate if the need to pass some gas.

lemunozm: It would be cool to provide a mechanism to refund any value > cost from the gateway to the caller. Something like:

```solidity
gateway.withdrawFunds(msg.value - cost, msg.sender)
```

As long as the integrator is also a gateway manager, they can do that.

hieronx: Agree with going to option 2!

hieronx (approved):

PR #619, opened 2025-09-21
Feat: allow multiple gateway managers

lemunozm:
Change the gateway manager storage to allow multiple managers in the form of: `PoolId => address who => bool canManager`

With this change, we can allow the integrator to also be a gateway manager, allowing gas payments to be refunded in integrations using `gateway.withdrawSubsidy(msg.sender)`, which I think is a good feature.

Right now, any excess in the payment for setRegisterAsset and crosschainTransferShares is refunded to the caller

- Extra: Added SavantChat issue fixes


src/common/Gateway.sol
```diff
@@ -70,7 +70,12 @@ contract Gateway is Auth, Recoverable, IGateway {
     }
 
     modifier onlyManager(PoolId poolId) {
-        require(msg.sender == manager[poolId], ManagerNotAllowed());
+        require(manager[poolId][msg.sender], NotAuthorized());
+        _;
+    }
+
+    modifier onlyAuthOrManager(PoolId poolId) {
```
hieronx, line 72: This is actually a good point: we should make `blockOutgoing` into `onlyAuthOrManager` as well, for more flexibility.

src/spoke/Spoke.sol
```diff
@@ -108,6 +108,7 @@ contract Spoke is Auth, Recoverable, ReentrancyProtection, ISpoke, ISpokeGateway
         uint256 cost =
             sender.sendInitiateTransferShares(centrifugeId, poolId, scId, receiver, amount, remoteExtraGasLimit);
         require(msg.value >= cost, NotEnoughGas());
+        if (msg.value > cost) gateway.withdrawSubsidy(poolId, msg.sender, msg.value - cost);
```
hieronx, line 114: If a FM forgets to set the Spoke as gateway manager, this will always revert. Not sure we should do this 🤔

lemunozm: It works out of the box because spoke is auth for gateway

hieronx: Right!

hieronx (dismissed):
LGTM!

hieronx (approved):
LGTM!

PR #620, opened 2025-09-22
tests: remove fork tests temporarily

wischli:
Removes fork tests from CI and locally to reduce the noise from failed fork tests pre v3.1 audit kickoff. Goal is to re-enable them afterwards in order to reuse them for future spells. TBD: Decouple from EndToEnd infra.

After discussion with @lemunozm , successor of #622 which decoupled e2e and fork tests. Fork tests of #622 are still failing due to the breaking HUB ABI but this will be addressed later, when reverting this PR and adding back fork tests.


lemunozm (dismissed):
Thanks!

wischli:
@lemunozm I also removed artifacts in [2b07c99](https://github.com/centrifuge/protocol-v3/pull/620/commits/2b07c9988c159cdc2081906aaabb5d9b861939e5) in order to streamline e2e development until we merge back fork tests. Do you think this is beneficial or we could keep the artifacts?

test/integration/EndToEnd.t.sol
```diff
@@ -547,16 +546,13 @@ contract EndToEndFlows is EndToEndUtils {
...
+        bool nonZeroPrices
     ) internal {
         // Configure prices
         _configurePricesForFlow(hub, spoke, poolId, shareClassId, assetId, poolManager, nonZeroPrices);
 
-        // Deploy or get existing vault (with fallback for fork tests)
-        IAsyncVault vault =
-            _ensureAsyncVaultExists(hub, spoke, poolId, shareClassId, assetId, poolManager, existingVault);
+        // Deploy or get existing vault
+        IAsyncVault vault = _ensureAsyncVaultExists(hub, spoke, poolId, shareClassId, assetId, poolManager);
```
lemunozm: We no longer need this, we can call directly to `_getAsyncVault()`, and also remove the request manager v3_0 from there. I think the same applies to `_ensureDepositEpochsAligned` and `_ensureRedeemEpochsAligned`

lemunozm (commented):
Agree with the simplification! And proposed a couple of them more

test/integration/EndToEnd.t.sol
wischli: With `_getAsyncVault` some tests throw because they only create the vault at this code path, not before. Wondering if we should keep `_ensureAsyncVaultExists` here and rename it to something like `_getOrDeployAsyncVault`

lemunozm: Makes sense! I think then it can be simplified, removing the fork part 👍🏻

lemunozm (dismissed):
Thanks for the clean up!

lemunozm (approved):

test/integration/fork/ChainConfigs.sol
```diff
@@ -1,113 +0,0 @@
...
-                wormholeId: IntegrationConstants.ETH_WORMHOLE_ID,
-                axelarId: IntegrationConstants.ETH_AXELAR_ID,
-                usdc: IntegrationConstants.ETH_USDC,
-                poolAdmin: IntegrationConstants.ETH_DEFAULT_POOL_ADMIN,
-                adminSafe: IntegrationConstants.ETH_ADMIN_SAFE,
-                hasAxelar: true
-            }),
-            ChainConfig({
-                name: "Base",
-                envVarName: "BASE_RPC_URL",
```
lemunozm, line 39: .

PR #621, opened 2025-09-22
Ref remove extraGasLimit transient storage

lemunozm:
Minor refactor to get rid of `extraGasLimit` transient storage in the Gateway. 

This was annoying because it created an invisible side effect between the `Gateway` and `MessageDispatcher` . It's very easy to forget the call to `gateway.setExtraGasLimit` or from the gateway forgot to assume the caller has called `setExtraGasLimit` in some scenarios, and clean the state after reading.

Now this relation is clearly established using a parameter in `send()`, and the compiler forces  it

### DIFF
--- a/src/common/Gateway.sol
+++ b/src/common/Gateway.sol
@@ -6,7 +6,6 @@ import {IRoot} from "./interfaces/IRoot.sol";
 import {IAdapter} from "./interfaces/IAdapter.sol";
 import {IGateway} from "./interfaces/IGateway.sol";
 import {IGasService} from "./interfaces/IGasService.sol";
-import {IMessageSender} from "./interfaces/IMessageSender.sol";
 import {IMessageHandler} from "./interfaces/IMessageHandler.sol";
 import {IMessageProcessor} from "./interfaces/IMessageProcessor.sol";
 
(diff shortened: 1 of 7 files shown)


src/common/interfaces/IMessageSender.sol
```diff
@@ -1,9 +0,0 @@
-// SPDX-License-Identifier: GPL-2.0-or-later
-pragma solidity >=0.5.0;
-
-/// @notice Generic interface for entities that handles outgoing messages
-interface IMessageSender {
```
lemunozm, line 5: Note, only implemented by the Gateway, and no mocks/tests use this, so I've moved it to the Gateway because it fully depends on how Gateway works

src/common/Gateway.sol
```diff
@@ -142,16 +140,20 @@ contract Gateway is Auth, Recoverable, IGateway {
...
         require(message.length > 0, EmptyMessage());
 
         PoolId poolId = processor.messagePoolId(message);
 
         emit PrepareMessage(centrifugeId, poolId, message);
 
         uint128 gasLimit = gasService.messageGasLimit(centrifugeId, message) + extraGasLimit;
-        extraGasLimit = 0;
 
         if (isBatching) {
```
hieronx, line 156: Nit, some unnecessary spacing here imo:
```suggestion
PoolId poolId = processor.messagePoolId(message);
emit PrepareMessage(centrifugeId, poolId, message);

uint128 gasLimit = gasService.messageGasLimit(centrifugeId, message) + extraGasLimit;
if (isBatching) {
```

hieronx (dismissed):
Another nice cleanup!

src/common/Gateway.sol
lemunozm: Fully agree

hieronx (approved):
Another nice cleanup!

PR #622, opened 2025-09-22
Tests: decouple fork tests

lemunozm:
Except several simplifications that can be done in EndToEnd, this PR decouples fork tests so that they do not depend on EndToEnd, which may be useful to recover fork tests after 3.1 if we go in that direction.


wischli (approved):
Thanks for decoupling these and leaving a cleaner state for the post-audit-reenable of the fork tests!

PR #623, opened 2025-09-23
Fix: incorrect payment method call

lemunozm:
Just realising we're calling the incorrect method in `MessageProcessor`


wischli (approved):
Good catch. Thanks for the quick fix!

PR #624, opened 2025-09-23
Automatic gas limit benchmarking & CI tools

lemunozm:
## Part 1: MessageBenchmarker

Generate automatic snapshot values for all gas limits required for the messages. The values are getting from `*EndToEnd*` tests and stored in `napshots/MessageGasLimits.json`

It works by adding a `MessageBenchmarker` contract between `Gateway` and `MessageProcessor` that will measure the cost of each message, choosing the highest value found.

NOTE: the built-in foundry snapshot system can not be used, because here, we're mixing snapshots with the same name from different tests (which makes the values to be reset between test suites), and also, we cannot choose the highest one.

To generate these values, we need two special env variables:
- `RAYON_NUM_THREADS = 1` to force Foundry to run the tests sequentially. If not, there are concurrence issues with writing/reading the JSON snapshot file.
- `BENCHMARKING_RUN_ID = <unique value>` to be able to reset the values only the first time in the execution of all tests. We want to take the higher value of each message, but we don't want to take into account values from previous runs/versions. Each new run should take a new value

Example of execution:
```sh
RAYON_NUM_THREADS=1 BENCHMARKING_RUN_ID="$(date +%s)" forge test EndToEnd
```

## Part 2: Scripting

- Added `update_gas_service_values.py` => SHOUTOUT to @gpmayorga, which updates GasService values with the snapshot json 
- Added `benchmarks.sh` with two flavours, `apply` (for updating GasService.sol contract), and `check` (that just verify for CI runs).
- Added `check-benchmarks` as CI job.

- [ ] QUESTION: Should this CI job be optional and only run for releases? Could be interesting to see all diffs in the code to double-check performance issues

## TL;DR

Use:
```sh
./script/utils/benchmarks.sh apply
```

If you PR modifies source code to update `GasService` and pass the CI job.


src/common/GasService.sol
```diff
@@ -47,34 +47,34 @@ contract GasService is IGasService {
...
-        maxAssetPriceAge = BASE_COST + 27260;
-        maxSharePriceAge = BASE_COST + 26032;
-        updateGatewayManager = BASE_COST + 26287;
+        // NOTE: Below values should be updated using script/utils/benchmark.sh
+        scheduleUpgrade = BASE_COST + 34062;
+        cancelUpgrade = BASE_COST + 14469;
+        recoverTokens = BASE_COST + 87322;
+        registerAsset = BASE_COST + 44795;
+        setPoolAdapters = BASE_COST + 416884; // using MAX_ADAPTER_COUNT
+        request = BASE_COST + 336276;
```
hieronx: Wonder why this got so much higher (336k vs 86k) 🤔

src/common/GasService.sol
```diff
@@ -47,34 +47,34 @@ contract GasService is IGasService {
...
+        notifyPool = BASE_COST + 1166002; // create escrow case
+        notifyShareClass = BASE_COST + 1784490;
+        notifyPricePoolPerShare = BASE_COST + 42424;
+        notifyPricePoolPerAsset = BASE_COST + 45788;
+        notifyShareMetadata = BASE_COST + 54228;
+        updateShareHook = BASE_COST + 31780;
+        initiateTransferShares = BASE_COST + 188316;
```
hieronx: Also a lot higher (188k vs 89k)

hieronx (dismissed):
Really nice automation!

test/integration/EndToEnd.t.sol
```diff
@@ -1009,17 +1026,18 @@ contract EndToEndUseCases is EndToEndFlows, VMLabeling {
...
         s.balanceSheet.submitQueuedShares(POOL_A, SC_1, EXTRA_GAS);
-        assertEq(address(s.balanceSheet.escrow(POOL_A)).balance, h.gasService.updateShares());
+        assertEq(address(s.balanceSheet.escrow(POOL_A)).balance, refunded * 2);
         assertEq(address(s.gateway).balance, DEFAULT_SUBSIDY);
 
         // This message is fully paid with refunded amount
         s.balanceSheet.submitQueuedShares(POOL_A, SC_1, EXTRA_GAS);
-        assertEq(address(s.balanceSheet.escrow(POOL_A)).balance, h.gasService.updateShares() / 2);
-        assertEq(address(s.gateway).balance, DEFAULT_SUBSIDY);
+        console.log("aa");
```
wischli:
```suggestion
```

test/integration/EndToEnd.t.sol
```diff
@@ -998,8 +1013,10 @@ contract EndToEndUseCases is EndToEndFlows, VMLabeling {
         h.hub.notifySharePrice(POOL_A, SC_1, s.centrifugeId);
         h.hub.setSnapshotHook(POOL_A, h.snapshotHook);
 
-        // Each message will return half of the gas wasted
-        adapterBToA.setRefundedValue(h.gasService.updateShares() / 2);
+        uint128 refunded = h.gasService.updateShares() - 100;
+        adapterBToA.setRefundedValue(refunded);
+
+        console.log(h.gasService.updateShares());
```
wischli: Can be removed, right?

snapshots/MessageGasLimits.json
```diff
@@ -0,0 +1,30 @@
+{
+  "BENCHMARKING_RUN_ID": 1758731414,
```
wischli: I like the idea of running benchmarks on every PR but I would remove this field because it doesn't provide value but will always cause trouble.

test/integration/utils/MessageBenchmarker.sol
```diff
@@ -0,0 +1,124 @@
...
+
+        uint256 prev = _getPreviousRegisteredValue(json, string.concat("$.", name));
+        uint256 before = gasleft();
+
+        messageProcessor.handle(centrifugeId, message);
+
+        uint256 new_ = before - gasleft();
+        uint256 higher = prev > new_ ? prev : new_;
+
+        //NOTE: If add a new entry, add first the name in the file, i.e: "newEntry" : 0
```
wischli: Ultra-nit:
```suggestion
        // NOTE: If add a new entry, add first the name in the file, i.e: "newEntry" : 0
```

wischli (commented):
Love it! Awesome job @lemunozm and @gpmayorga . I suppose we can deprecate the current snapshots then?

test/integration/EndToEnd.t.sol
lemunozm: Correct, thanks for checking my dirty

snapshots/MessageGasLimits.json
lemunozm: Sadly, this is mandatory to ensure all tests belong to the same run. If not tests from different contracts will override themself taken wrong values

src/common/GasService.sol
lemunozm: Thanks for highlight this, I thought at first that I was chosen the wrong request previously, and now the good one, but after another thought I think it's because one processing is taken into account the next message, because everything is sync.

Good catch, I need to filter this in `MessageBenchmarker`

src/common/GasService.sol
lemunozm: Same as above

lemunozm:
> I suppose we can deprecate the current snapshots then?

I think we could, although they could be different paths. These are only the ones computed after receiving the messages. The computing before sending the message is not measured here.

Not sure how important is to measure those

src/common/GasService.sol
lemunozm: This should be fixed now after https://github.com/centrifuge/protocol-v3/pull/631

src/common/GasService.sol
```diff
@@ -47,34 +47,34 @@ contract GasService is IGasService {
...
+        updateRestriction = BASE_COST + 81844;
+        updateContract = BASE_COST + 99339;
+        requestCallback = BASE_COST + 225405; // approve deposit case
+        updateVaultDeployAndLink = BASE_COST + 2820311;
+        updateVaultLink = BASE_COST + 152664;
+        updateVaultUnlink = BASE_COST + 101376;
+        setRequestManager = BASE_COST + 68197;
+        updateBalanceSheetManager = BASE_COST + 71761;
+        updateHoldingAmount = BASE_COST + 271358;
+        updateShares = BASE_COST + 151411;
```
lemunozm: - High increment because of calling the snapshot (added by Onno's PR)

src/common/GasService.sol
```diff
@@ -47,34 +47,34 @@ contract GasService is IGasService {
...
+        notifyPricePoolPerShare = BASE_COST + 74705;
+        notifyPricePoolPerAsset = BASE_COST + 78425;
+        notifyShareMetadata = BASE_COST + 87934;
+        updateShareHook = BASE_COST + 64017;
+        initiateTransferShares = BASE_COST + 250624;
+        executeTransferShares = BASE_COST + 144865;
+        updateRestriction = BASE_COST + 81844;
+        updateContract = BASE_COST + 99339;
+        requestCallback = BASE_COST + 225405; // approve deposit case
+        updateVaultDeployAndLink = BASE_COST + 2820311;
```
lemunozm: Still have not found the cause of this issue in mainnet, if found again, we can invest and update `GasService`

src/common/GasService.sol
```diff
@@ -47,34 +47,34 @@ contract GasService is IGasService {
...
-        maxAssetPriceAge = BASE_COST + 27260;
-        maxSharePriceAge = BASE_COST + 26032;
-        updateGatewayManager = BASE_COST + 26287;
+        // NOTE: Below values should be updated using script/utils/benchmark.sh
+        scheduleUpgrade = BASE_COST + 66469;
+        cancelUpgrade = BASE_COST + 46876;
+        recoverTokens = BASE_COST + 123959;
+        registerAsset = BASE_COST + 76846;
+        setPoolAdapters = BASE_COST + 452173; // using MAX_ADAPTER_COUNT
+        request = BASE_COST + 186928;
```
lemunozm: The high increment is due to the latest refactor in BRM + pass the message back to Spoke and enqueue as unpaid.

src/common/GasService.sol
```diff
@@ -12,7 +12,7 @@ contract GasService is IGasService {
     using MessageLib for *;
 
     /// @dev Takes into account Adapter + Gateway processing + some mismatch happened regarding the input values
-    uint128 public constant BASE_COST = 200_000;
+    uint128 public constant BASE_COST = 150_000;
```
lemunozm: All values have increased by about `~50K` because now the measurement is taken into account `Gateway` + `MessageProcessor`. That's why I've removed that from the offset too.

lemunozm (commented):
All params show expected gas values after rebasing https://github.com/centrifuge/protocol/pull/636

src/common/GasService.sol
hieronx: Can this be reduced even more now that there is the `RECEIVE_COST` in adapters?

src/common/interfaces/IAdapter.sol
```diff
@@ -1,9 +1,7 @@
 // SPDX-License-Identifier: GPL-2.0-or-later
 pragma solidity >=0.5.0;
 
-import {IAuth} from "../../misc/interfaces/IAuth.sol";
-
-interface IAdapter is IAuth {
+interface IAdapter {
```
hieronx, line 4: Good catch

hieronx (dismissed):

src/common/GasService.sol
lemunozm: Yes, probably we can cut more (not sure how more)

The problem is that we have no way to estimate the computation done in the `Gateway` and `Multiadapter` when a batch is received. We can receive long messages that require more computation in the Gateway as a whole. Events will store more bytes, etc...

But maybe we can reduce a bit more yeah, `100000` is ok? Take into account `MultiAdapter` is not measured

hieronx: I'd say, let's do 100.000 and then do a bunch of testing on sepolia in the next few weeks

hieronx (dismissed):

hieronx (approved):

PR #618, opened 2025-09-18
Ref: pool permissioned batching

lemunozm:
Blocked by: https://github.com/centrifuge/protocol-v3/pull/619

Enable batching perssioned by pool, which simplify quite a lot the usage and gateway interaction from other contracts

PR #625, opened 2025-09-23
Idea: Permissionless batching by aux contract

lemunozm:
As a different implementation than https://github.com/centrifuge/protocol-v3/pull/618 (which has the UX issue discussed here: https://kflabs.slack.com/archives/C07PG2EUR9C/p1758624177416539?thread_ts=1758619310.258729&cid=C07PG2EUR9C)

Batching methods `startBatching` and `endBatching` are always called by `GatewayBatcher`, a simple contract that just forces both calls to happen. Any contract can call `GatewayBatcher.withBatch(bytes memory data)` to call any method from itself with batching support

### DIFF
--- a/src/common/CrosschainBatcher.sol
+++ b/src/common/CrosschainBatcher.sol
@@ -7,7 +7,7 @@ import {ICrosschainBatcher} from "./interfaces/ICrosschainBatcher.sol";
 import {Auth} from "../misc/Auth.sol";
 
 /// @title  CrosschainBatcher
-/// @dev    Helper contract that enables integrations to automatically batch multiple cross-chain tansactions.
+/// @dev    Helper contract that enables integrations to automatically batch multiple cross-chain transactions.
 ///         Should be used like:
 ///         ```
 ///         contract Integration {
--- a/src/common/CrosschainBatcher.sol
+++ b/src/common/CrosschainBatcher.sol
@@ -2,10 +2,9 @@
 pragma solidity 0.8.28;
 
 import {IGateway} from "./interfaces/IGateway.sol";
+import {ICrosschainBatcher} from "./interfaces/ICrosschainBatcher.sol";
 
 import {Auth} from "../misc/Auth.sol";
-
-import {ICrosschainBatcher} from "./interfaces/ICrosschainBatcher.sol";
 
 /// @title  CrosschainBatcher
 /// @dev    Helper contract that enables integrations to automatically batch multiple messages. Should be used like:
--- a/src/common/CrosschainBatcher.sol
+++ b/src/common/CrosschainBatcher.sol
@@ -1,17 +1,29 @@
 // SPDX-License-Identifier: BUSL-1.1
 pragma solidity 0.8.28;
 
-import {PoolId} from "./types/PoolId.sol";
-import {AssetId} from "./types/AssetId.sol";
 import {IGateway} from "./interfaces/IGateway.sol";
-import {ShareClassId} from "./types/ShareClassId.sol";
 
 import {Auth} from "../misc/Auth.sol";
 
-import {IGatewayBatchCallback} from "./interfaces/IGatewayBatchCallback.sol";
+import {ICrosschainBatcher} from "./interfaces/ICrosschainBatcher.sol";
 
-// Gateway trust on GatewayBatcher
-contract GatewayBatchCallback is Auth, IGatewayBatchCallback {
+/// @title  CrosschainBatcher
+/// @dev    Helper contract that enables integrations to automatically batch multiple messages. Should be used like:
+///         ```
+///         contract Integration {
+///             ICrosschainBatcher batcher;
+///
+///             function doSomething(PoolId poolId) external {
+///                 uint256 cost = batcher.batch(abi.encodeWithSelector(Integration._execute.selector, poolId));
+///             }
+///
+///             function _execute(PoolId poolId) external {
+///                 require(batcher.sender() == address(this));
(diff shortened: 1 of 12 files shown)


src/common/Gateway.sol
```diff
@@ -292,7 +292,7 @@ contract Gateway is Auth, Recoverable, IGateway {
             bytes32 outboundBatchSlot = _outboundBatchSlot(centrifugeId, poolId);
             uint128 gasLimit = _gasLimitSlot(centrifugeId, poolId).tloadUint128();
 
-            _send(centrifugeId, TransientBytesLib.get(outboundBatchSlot), gasLimit);
+            cost += _send(centrifugeId, TransientBytesLib.get(outboundBatchSlot), gasLimit);
```
hieronx, line 295: This is anyways a nice addition

src/common/GatewayBatchCallback.sol
```diff
@@ -0,0 +1,49 @@
...
+import {PoolId} from "./types/PoolId.sol";
+import {AssetId} from "./types/AssetId.sol";
+import {IGateway} from "./interfaces/IGateway.sol";
+import {ShareClassId} from "./types/ShareClassId.sol";
+
+import {Auth} from "../misc/Auth.sol";
+
+import {IGatewayBatchCallback} from "./interfaces/IGatewayBatchCallback.sol";
+
+// Gateway trust on GatewayBatcher
```
hieronx: From the user's/integrator's perspective, they might not care about that this is batching related to the gateway. What matters is that it will batch multiple cross-chain methods. So I would rename this to `CrosschainBatcher`. Wdyt @lemunozm ?

hieronx: Given this is entirely focused on integrators, I would even add a code sample here, to make this super clear:
```suggestion
/// @title  CrosschainBatcher
/// @dev    Helper contract that enables integrations to automatically batch multiple messages. Should be used like:
///         ```
///         contract Integration {
///                function doSomething(PoolId poolId) external {
///                    crosschainBatcher.withCallback(abi.encodeWithSelector(Integration.execute.selector, poolId));
///                }
///
///                function execute(PoolId poolId) external {
///                    require(gatewayBatcher.sender() == address(this));
///                    // Call several hub, balance sheet, or spoke methods that trigger cross-chain transactions
///                }
///         }
///         ```
```

src/common/CrosschainBatcher.sol
```diff
@@ -0,0 +1,60 @@
...
+///                 uint256 cost = batcher.batch(abi.encodeWithSelector(Integration._execute.selector, poolId));
+///             }
+///
+///             function _execute(PoolId poolId) external {
+///                 require(batcher.sender() == address(this));
+///                 // Call several hub, balance sheet, or spoke methods that trigger cross-chain transactions
+///             }
+///         }
+///         ```
+contract CrosschainBatcher is Auth, ICrosschainBatcher {
```
hieronx, line 25: @lemunozm want to add this to the deploy scripts already in this PR? Likely needed for Onno to add integration tests for the queue manager.

src/common/CrosschainBatcher.sol
```diff
@@ -0,0 +1,60 @@
...
+/// @dev    Helper contract that enables integrations to automatically batch multiple messages. Should be used like:
+///         ```
+///         contract Integration {
+///             ICrosschainBatcher batcher;
+///
+///             function doSomething(PoolId poolId) external {
+///                 uint256 cost = batcher.batch(abi.encodeWithSelector(Integration._execute.selector, poolId));
+///             }
+///
+///             function _execute(PoolId poolId) external {
```
hieronx: I don't think there should be an underscore prefix here, if it's not an internal method

src/common/CrosschainBatcher.sol
```diff
@@ -0,0 +1,60 @@
...
+import {Auth} from "../misc/Auth.sol";
+
+/// @title  CrosschainBatcher
+/// @dev    Helper contract that enables integrations to automatically batch multiple messages. Should be used like:
+///         ```
+///         contract Integration {
+///             ICrosschainBatcher batcher;
+///
+///             function doSomething(PoolId poolId) external {
+///                 uint256 cost = batcher.batch(abi.encodeWithSelector(Integration._execute.selector, poolId));
```
hieronx: Nit: personally think `batcher.batch` is a bit odd, would prefer `batcher.execute` or `batcher.doCallback` or something else

hieronx: `uint256 cost = ` part can probably be removed here, not really relevant.

src/common/CrosschainBatcher.sol
```diff
@@ -0,0 +1,60 @@
+// SPDX-License-Identifier: BUSL-1.1
+pragma solidity 0.8.28;
+
+import {IGateway} from "./interfaces/IGateway.sol";
+import {ICrosschainBatcher} from "./interfaces/ICrosschainBatcher.sol";
+
+import {Auth} from "../misc/Auth.sol";
+
+/// @title  CrosschainBatcher
+/// @dev    Helper contract that enables integrations to automatically batch multiple messages. Should be used like:
```
hieronx: My bad but:
```suggestion
/// @dev    Helper contract that enables integrations to automatically batch multiple cross-chain transactions. Should be used like:
```

src/common/CrosschainBatcher.sol
hieronx: (if you disagree, feel free to ignore)

hieronx (dismissed):

hieronx (approved):

PR #627, opened 2025-09-24
tests: full BatchRequestManager coverage post SCM-Refactor

wischli:
Follow-up of https://github.com/centrifuge/protocol-v3/pull/606 to add missing coverage from previous `ShareClassManager` tests (including precision edge cases) and have 100% `BatchRequestManager` coverage. Also extends some previous tests by asserting amounts from event log parsing which was done previously with return amounts of external calls that now only return the cost.

TODO

- [x] Get back to 100% full coverage after rebase

### DIFF
--- a/foundry.toml
+++ b/foundry.toml
@@ -27,7 +27,7 @@ ignored_warnings_from = [
   "src/misc/ReentrancyProtection.sol", # Issue: https://github.com/ethereum/solidity/issues/14359
 ]
 
-[lint]
+[profile.lint]
 severity = ["high"]
 exclude_lints = ["erc20-unchecked-transfer", "incorrect-shift"]
 
--- a/foundry.toml
+++ b/foundry.toml
@@ -27,7 +27,7 @@ ignored_warnings_from = [
   "src/misc/ReentrancyProtection.sol", # Issue: https://github.com/ethereum/solidity/issues/14359
 ]
 
-[lint]
+[profile.default.lint]
 severity = ["high"]
 exclude_lints = ["erc20-unchecked-transfer", "incorrect-shift"]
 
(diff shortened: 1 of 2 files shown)


wischli:
Unfortunately, we cannot reach true 100% coverage due the fact that when Solidity compiles `require(condition, error)`, it generates optimized bytecode that the coverage instrumentation can struggle to hook into, which is the case here. The `if/revert` pattern creates more explicit control flow that Forge's tooling can properly instrument but I did not want to change anything in the BRM just because of that. This line is not recognized as fully covered by Foundry but sufficient tests exist: https://github.com/centrifuge/protocol-v3/blob/4f69794392c44cdbf6d27121fd37a2c6ec9eeced/src/vaults/BatchRequestManager.sol#L591

lemunozm:
> Unfortunately, we cannot reach true 100% coverage due the fact that when Solidity compiles `require(condition, error)`, it generates optimized bytecode that the coverage instrumentation can struggle to hook into, which is the case here. The `if/revert` pattern creates more explicit control flow that Forge's tooling can properly instrument but I did not want to change anything in the BRM just because of that. This line is not recognized as fully covered by Foundry but sufficient tests exist: https://github.com/centrifuge/protocol-v3/blob/4f69794392c44cdbf6d27121fd37a2c6ec9eeced/src/vaults/BatchRequestManager.sol#L591

Could it be because the coverage tool expects the code to pass through 3 different cases:
- Generate the error
- Have msg.value > cost
- Have msg.value = cost

just theorizing

wischli:
> Could it be because the coverage tool expects the code to pass through 3 different cases:
> 
> * Generate the error
> * Have msg.value > cost
> * Have msg.value = cost
> 
> just theorizing

AFAICT, that's not the case. Coverage tools don't distinguish between `>` and `==` when both satisfy the same condition `>=`, they are both the `true` branch. It's really about how `require()` compiles into optimized bytecode whereas the `if/revert` pattern generates an explicit `JUMPI` instruction which the coverage tool can recognize.

hieronx (approved):

PR #631, opened 2025-09-25
Sending messages while processing messages to unpaid

lemunozm:
Thread: https://kflabs.slack.com/archives/C07PG2EUR9C/p1758793226119319

TLDR: The gateway enters a mode where each call to `send` becomes unpaid while processing a message. That way we get out of the box the `addUnpaidMessage` behavior for sending any message.

### DIFF
--- a/test/integration/EndToEnd.t.sol
+++ b/test/integration/EndToEnd.t.sol
@@ -12,7 +12,7 @@ import {MathLib} from "../../src/misc/libraries/MathLib.sol";
 import {ETH_ADDRESS} from "../../src/misc/interfaces/IRecoverable.sol";
 
 import {Root} from "../../src/common/Root.sol";
-import {Gateway} from "../../src/common/Gateway.sol";
+import {IGateway, Gateway} from "../../src/common/Gateway.sol";
 import {Guardian} from "../../src/common/Guardian.sol";
 import {PoolId} from "../../src/common/types/PoolId.sol";
 import {GasService} from "../../src/common/GasService.sol";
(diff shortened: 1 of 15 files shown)


test/integration/EndToEnd.t.sol
```diff
@@ -874,6 +888,9 @@ contract EndToEndFlows is EndToEndUtils {
         uint128 shares = uint128(s.spoke.shareToken(POOL_A, SC_1).balanceOf(INVESTOR_A));
         vault.requestRedeem(shares, INVESTOR_A, INVESTOR_A);
         vault.cancelRedeemRequest(PLACEHOLDER_REQUEST_ID, INVESTOR_A);
+
+        if (!sameChain) h.gateway.repay{value: GAS}(s.centrifugeId, _getLastUnpaidMessage());
```
lemunozm, line 903: Maybe the name `repay()` no longer makes sense in this context, because it was never paid actually. Should we rename this to just `pay()`/`payAndSend()` or something else?

wischli: I think we should rename that yes to reduce confusion for future auditors. I prefer `payAndSend` due to its expressiveness.

wischli: Also wondering whether we should add an event for `repay` calls. Do you see any value in that @hieronx ?

wischli (approved):
Such an elegant, concise solution - incredible job @lemunozm 🎉 Couldn't even find a nit.

test/integration/EndToEnd.t.sol
lemunozm: There is already a `RepayBatch`, or do you mean a different thing?

hieronx: Yeah there's already an event right?

hieronx: I wouldn't rename, since this would just create unnecessary complexity when migrating indexer etc.

lemunozm: Ok, we would need to explain in the docs that this means the first time paid.

PR #632, opened 2025-09-25
ci: fix forge fmt

wischli:
Fixes CI lint and CI test after foundry upgrade that happened around merging #606. 

### Changes

1. Apply `forge fmt` with the latest foundry version (two lines had to be disabled due to being non-deterministic, alternated after each fmt between two states)
2. Fix compiler warnings in `IdentityValuation.getPrice`

### CI Error description

CI uses latest version `forge 1.3.6-nightly (d6daa0ac83 2025-09-25T08:01:37.285831000Z)` to which you should upgrade 

```bash
foundryup nightly 
foundryup -u nightly
```

### DIFF
--- a/foundry.toml
+++ b/foundry.toml
@@ -27,7 +27,7 @@ ignored_warnings_from = [
   "src/misc/ReentrancyProtection.sol", # Issue: https://github.com/ethereum/solidity/issues/14359
 ]
 
-[lint]
+[profile.default.lint]
 severity = ["high"]
 exclude_lints = ["erc20-unchecked-transfer", "incorrect-shift"]
 
@@ -100,7 +100,7 @@ int_types = "long"
 multiline_func_header = "attributes_first"
 quote_style = "double"
 number_underscore = "preserve"
-wrap_comments = true
+wrap_comments = false
 ignore = [
   "test/*.sol"
 ]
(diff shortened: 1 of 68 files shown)


src/adapters/LayerZeroAdapter.sol
```diff
@@ -20,9 +20,9 @@ import {MathLib} from "../misc/libraries/MathLib.sol";
 
 import {IMessageHandler} from "../common/interfaces/IMessageHandler.sol";
 
-/// @title  LayerZero Adapter
+/// @title LayerZero Adapter
 /// @notice Routing contract that integrates with LayerZero V2.
-/// @dev    A delegate is set on deployment, to configure the DVN and executor
```
hieronx, line 25: Can we disable this? I think the old formatting is better

wischli: Done in [fd108b8](https://github.com/centrifuge/protocol-v3/pull/632/commits/fd108b85774136d39d986838db939c0c37bf9cca)

lemunozm: Can we actually force the indentation?

lemunozm (approved):
Thanks for taking care of this

src/adapters/LayerZeroAdapter.sol
wischli: > Can we actually force the indentation?

Most likely not. There is no configuration related to natspec and for general comments, there is only wrapping. We would have to write a regex based script similar to the import one.

PR #628, opened 2025-09-25
Allow direct BRM calls

hieronx:

PR #633, opened 2025-09-25
tests: add manager permission checks for `BatchRequestManager`

wischli:
Adds missing tests of new public calls of BRM added in https://github.com/centrifuge/protocol-v3/pull/628


lemunozm (approved):
LGTM!

PR #636, opened 2025-09-29
Ref: Subsidized funds at vaults level & new payment logic

lemunozm:
### Problem

The previous solution was guided by two requirements coming from how vaults need to work:
- A vault can not be paid explicitly.
- A vault can never fail if there is no funds (automatically as unpaid if not enough for paying the message)

This, along with ping-pong messages makes us built the subsidized funds design at the core level. This works fine if all messages has the same priviledges as the pool, if not, some problems arise:
- We need to ensure the permissionless user pays what they need to pay to not drain funds from the subsidized pool
- Everything is subsidized, but in the end, a permissionless user needs to pay for the calls => mixing two payment methods at the core level
- If a multicall caller mixes permissionless and pool permissioned methods, it's not clear (and never fully solved) how to pay and mix those messages.
  - Actually we have current issues with this like this [burrasec issue](https://github.com/burrasec/2025-09-Centrifuge/issues/2)
  - Other issues that have not yet highlighted where we were not able to multicall with payment methods

### Solution

With the change of avoiding ping-pong messages, we can now move the vaults' requirements where they need to be, at the `vaults` level. It means, now `AsyncRequestManager` offers a subsidized funds mechanism only for the vaults, but outside of the core level. All payments can now be done explicitly, allowing any mix of messages in integrators and multicalls.

#### Main design points
- Composability is the major point of this design, because everything is paid in the methods, and the user has full control of the refund, they can compose integrators on top of integrators without worrying about different payment levels.
- Any extra gas given in the payment is refunded to the user to allow easy composability for integrators using the `refund` parameter omnipresent in the API, it means:
  - If the action is overpaid, extra eth is refunded
  - If the gas limit used is not fully consumed, it is refunded asynchronously by the adapter
  - If the message can not be paid and `unpaidMode` is enabled, any remaining eth is refunded to the user
  - If the message is sent to a local chain, the payment is refunded to the user
    - [ ] Still undecided if we should revert here or not, not sure what's the best UX here.
  - If the payment is done in batching mode, it reverts as an indication of bad usage of the API.  
- If the gateway is configured in `unpaidMode`, it means: It will try to send the message if they have funds to do it; if not, instead of reverting, it will add the message as unpaid.
- The payment in multicall is only done in `endBatching`
- If batching, the `refund` parameter of the calls is ignored, only the one passed in endBatching is the used one


src/spoke/Spoke.sol
```diff
@@ -164,22 +162,27 @@ contract Spoke is Auth, Recoverable, ReentrancyProtection, ISpoke, ISpokeGateway
 
         emit RegisterAsset(centrifugeId, assetId, asset, tokenId, name, symbol, decimals, isInitialization);
 
-        gateway.depositSubsidy{value: msg.value}(GLOBAL_POOL);
-        uint256 cost = sender.sendRegisterAsset(centrifugeId, assetId, decimals);
-        require(msg.value >= cost, NotEnoughGas());
-        if (msg.value > cost) gateway.withdrawSubsidy(GLOBAL_POOL, refund, msg.value - cost);
+        gateway.setPayment{value: msg.value}(refund);
+        sender.sendRegisterAsset(centrifugeId, assetId, decimals);
```
lemunozm: All calls that send messages just need to be implemented with these two lines.

The gateway will refund with the reminder, no gateway subsidized checks, no issues if the subsidized is out of funds, no explicit refunding, and fully composable

lemunozm: It became simpler, passing the refund as param

src/common/PoolEscrow.sol
```diff
@@ -21,10 +21,6 @@ contract PoolEscrow is Escrow, Recoverable, IPoolEscrow {
         poolId = poolId_;
     }
 
-    receive() external payable {
```
hieronx, line 24: Why not leave this?

src/hub/Hub.sol
```diff
@@ -619,4 +641,8 @@ contract Hub is Multicall, Auth, Recoverable, IHub, IHubRequestManagerCallback,
     function _isManager(PoolId poolId) internal view {
         require(hubRegistry.manager(poolId, msg.sender), IHub.NotManager());
     }
+
+    function _payment() internal view returns (uint256 value) {
+        return gateway.isBatching() ? 0 : msg.value;
+    }
```
lemunozm, line 647: We can only pay for messages if we're not in a multicall, if not the first message in the multicall will take all `msg.value` making rest methods to revert:

Related to this: https://kflabs.slack.com/archives/C07PG2EUR9C/p1759183383159389

src/spoke/Spoke.sol
```diff
@@ -106,12 +106,9 @@ contract Spoke is Auth, Recoverable, ReentrancyProtection, ISpoke, ISpokeGateway
 
         emit InitiateTransferShares(centrifugeId, poolId, scId, msg.sender, receiver, amount);
 
-        gateway.depositSubsidy{value: msg.value}(poolId);
-        uint256 cost = sender.sendInitiateTransferShares(
-            centrifugeId, poolId, scId, receiver, amount, extraGasLimit, remoteExtraGasLimit
-        );
-        require(msg.value >= cost, NotEnoughGas());
-        if (msg.value > cost) gateway.withdrawSubsidy(poolId, msg.sender, msg.value - cost);
```
lemunozm, line 114: Now integrators are much more simpler and fully composable

src/vaults/VaultRouter.sol
```diff
@@ -63,19 +63,19 @@ contract VaultRouter is Auth, Multicall, Recoverable, IVaultRouter {
         super.multicall(data);
 
         if (!wasBatching) {
-            gateway.endBatching();
+            gateway.endBatching{value: msg.value}(msg.sender);
```
lemunozm, line 66: Note for VaultRouter actions:
- If no batching, the messages are subsidized at the vault level.
- If batching, the messages are paid in the multicall.

src/vaults/AsyncRequestManager.sol
```diff
@@ -151,7 +168,15 @@ contract AsyncRequestManager is Auth, Recoverable, IAsyncRequestManager {
...
-        spoke.request(vault_.poolId(), vault_.scId(), spoke.vaultDetails(vault_).assetId, payload);
+        IRefundEscrow refund = refundEscrowFactory.get(vault_.poolId());
+        require(address(refund).code.length > 0, RefundEscrowNotDeployed());
+
+        refund.withdrawFunds(address(this), address(refund).balance); // All funds goes to this contract
+
+        // It use all funds for the message, and the rest is refunded again to the RefundEscrow
+        spoke.request{
+            value: balanceSheet.gateway().isBatching() ? 0 : address(this).balance
+        }(vault_.poolId(), vault_.scId(), spoke.vaultDetails(vault_).assetId, payload, address(refund), true);
```
hieronx, line 186: I'm not fully following this. By setting `unpaid=true`, aren't we not actually paying for the tx with subsidized funds?

src/common/PoolEscrow.sol
lemunozm: Is no longer used, but I can leave it

src/vaults/AsyncRequestManager.sol
lemunozm: By setting `unpaid=true`, you're configuring the gateway to not revert if no funds. If funds are available, the message is still sent; if no funds, the message is added to the unpaid storage.

script/TestData.s.sol
```diff
@@ -104,6 +105,8 @@ contract TestData is FullDeployer {
     function _deployAsyncVault(uint16 centrifugeId, ERC20 token, AssetId assetId) internal {
         VaultState memory state;
         state.poolId = hubRegistry.poolId(centrifugeId, 1);
+        asyncRequestManager.depositSubsidy{value: 0.5 ether}(state.poolId);
```
hieronx, line 108: This shouldn't be needed as it's all local, right?

src/common/Gateway.sol
```diff
@@ -93,9 +88,7 @@ contract Gateway is Auth, Recoverable, IGateway {
         emit UpdateManager(poolId, who, canManage);
     }
 
-    receive() external payable {
-        _depositSubsidy(GLOBAL_POT, msg.sender, msg.value);
-    }
+    receive() external payable {}
```
hieronx: I think this is not recoverable, so might as well be removed, right?

src/common/Gateway.sol
```diff
@@ -168,34 +162,35 @@ contract Gateway is Auth, Recoverable, IGateway {
...
         }
+    }
 
-        adapter.send{value: cost}(centrifugeId, batch, batchGasLimit, address(subsidy[paymentPoolId].refund));
-
-        return cost;
+    function _refund(address refund, uint256 fuel) internal {
+        if (fuel > 0) {
+            (bool success,) = payable(refund).call{value: fuel}("");
+            require(success, CannotRefund());
```
hieronx, line 190: I wonder what would be the downside if we just removed this require. Then if a smart contract calls any method without having a `receive` method, they would not receive the refund, but the tx should still succeed. Wdyt @lemunozm ?

src/common/Gateway.sol
```diff
@@ -142,11 +135,11 @@ contract Gateway is Auth, Recoverable, IGateway {
     //----------------------------------------------------------------------------------------------
 
     /// @inheritdoc IGateway
-    function send(uint16 centrifugeId, bytes calldata message, uint128 extraGasLimit)
+    function send(uint16 centrifugeId, bytes calldata message, uint128 extraGasLimit, address refund)
```
wischli, line 136: Curious if we should add a safety-check for `refund != address(0)` when `msg.value > 0` to mitigate loss of funds, particularly because there are code paths with `refund == address(0)` that assume no payment due to unpaid mode.

src/vaults/AsyncRequestManager.sol
```diff
@@ -151,7 +168,21 @@ contract AsyncRequestManager is Auth, Recoverable, IAsyncRequestManager {
...
-        spoke.request(vault_.poolId(), vault_.scId(), spoke.vaultDetails(vault_).assetId, payload);
+        IRefundEscrow refund;
+        uint256 payment;
+
+        if (!balanceSheet.gateway().isBatching()) {
+            refund = refundEscrowFactory.get(vault_.poolId());
+            require(address(refund).code.length > 0, RefundEscrowNotDeployed());
+
+            refund.withdrawFunds(address(this), address(refund).balance); // All funds goes to this contract
+            payment = address(this).balance;
```
wischli: Curious if it can ever happen that `AsyncRequestManager` has residual ETH from other source. Then, the pool subsidy would use those funds as well. In case of that, we should rather do the following, though I suppose my question is purely theoretical

```suggestion
            uint256 availableSubsidy = address(refund).balance;
            refund.withdrawFunds(address(this), availableSubsidy);
            payment = availableSubsidy;
```

src/vaults/RefundEscrow.sol
```diff
@@ -0,0 +1,19 @@
...
+import {Auth} from "../misc/Auth.sol";
+
+contract RefundEscrow is Auth, IRefundEscrow {
+    constructor(address owner) Auth(owner) {}
+
+    receive() external payable {}
+
+    function depositFunds() external payable auth {}
+
+    function withdrawFunds(address to, uint256 value) external auth {
```
wischli, line 17: Should we dispatch an event when withdrawing funds?

src/vaults/RefundEscrow.sol
```diff
@@ -0,0 +1,19 @@
...
+        (bool success,) = to.call{value: value}("");
```
wischli, line 18: AFAICT, this would fail with an unclear error if `value > balance`. 
```suggestion
        require(address(this).balance >= value, InsufficientBalance());
        (bool success,) = to.call{value: value}("");
```

src/common/Gateway.sol
lemunozm: The Gateway is Recoverable, although if everything works well, right now everything is correctly refunded. So we could remove that and also no longer inherit from Recoverable if you agree

script/TestData.s.sol
lemunozm: Good question! I need this as a way to deploy the escrow, but I could pass no value.

src/common/Gateway.sol
lemunozm: Correct. The thing is that we open a way for the user to lose their funds if something is not correctly set.

Right now, the caller is not required to have `receive()` so there is no problem there. Therefore, I think it should be a requirement for `refund` to have it. Note that propably `refund` will come from the highest API level, so integrators in between cannot ensure `refund` is payable (or not easily)

src/common/PoolEscrow.sol
lemunozm: @hieronx should I keep this or is it ok?

src/common/Gateway.sol
hieronx: Actually yeah let's keep it as Recoverable then!

src/common/Gateway.sol
hieronx: I can see that, makes sense.

src/common/Gateway.sol
lemunozm: Agree receive() can be removed anyways!

src/hub/Hub.sol
```diff
@@ -74,7 +72,6 @@ contract Hub is Multicall, Auth, Recoverable, IHub, IHubRequestManagerCallback,
         else if (what == "holdings") holdings = IHoldings(data);
         else if (what == "shareClassManager") shareClassManager = IShareClassManager(data);
         else if (what == "gateway") gateway = IGateway(data);
-        else if (what == "poolEscrowFactory") poolEscrowFactory = IPoolEscrowFactory(data);
```
wischli, line 77: Great to see this gone!

src/vaults/factories/RefundEscrowFactory.sol
```diff
@@ -0,0 +1,41 @@
...
+    address public controller;
+
+    constructor(address deployer) Auth(deployer) {}
+
+    function file(bytes32 what, address data) external auth {
+        if (what == "controller") controller = data;
+        else revert FileUnrecognizedParam();
+        emit File(what, data);
+    }
+
```
wischli, line 23: Ultra-Nit: Missing natspec inheritance everywhere
```suggestion
    /// @inheritdoc IRefundEscrowFactory
```

wischli (commented):
Incredible job implementing this so cleanly in such a short time frame! Really happy to see the all the obscure payment artifacts gone and having a unified, easy-to-follow mechanism on both sides of the protocol! A couple of questions from my side.

src/common/Gateway.sol
lemunozm: Good question. I don't think so because if the user choose address(0) as refunds it means: burns any reminder.

Right now, the codepaths that use `address(0)` don't have any payment. But if it would have it, they didn't know where to refund it, so `address(0)` should be the default as way to say "I don't care"

src/vaults/AsyncRequestManager.sol
lemunozm: Good theoretical question. Should not have any eth, but I agree this is more correct

src/vaults/RefundEscrow.sol
lemunozm: I thought about this, and I think not at this level, because this is used internally back and forth for each payment. It should be in the incoming withdrawDeposit in `AsyncRequestManager`

src/vaults/RefundEscrow.sol
lemunozm: The caller ensures that this never happen, already passing `address(refund).balance`, so the error will never be dispatched

src/hub/Hub.sol
lemunozm: Yeah, this was tormenting me at nights

wischli (approved):
Thanks for the quick responses and update! LSGTM!

PR #637, opened 2025-09-29
Add ERC-165 support to BRM

hieronx:


wischli (dismissed):

wischli (approved):

src/hub/interfaces/IHubRequestManager.sol
```diff
@@ -1,11 +1,30 @@
...
+        bytes32 investor,
+        uint32 maxClaims
+    ) external payable returns (uint256 cost);
+
+    /// @notice Notify a redemption for an investor address located in the chain where the asset belongs
+    function notifyRedeem(PoolId poolId, ShareClassId scId, AssetId payoutAssetId, bytes32 investor, uint32 maxClaims)
+        external
+        payable
+        returns (uint256 cost);
+}
```
lemunozm, line 30: Should not go under `vaults` because these methods are not used by Hub? I mean, Hub doesn't need to know them

hieronx: I would say multiple hub request mgr implementations can/should use this interface, hence it is a shared interface.

lemunozm: But it's shared across `vaults` level, isn't? I mean, right now it belongs to the core when theoretically it does not belong to

hieronx: I disagree. If there are multiple `vaults` implementations, they would be in different folders. The current `vaults` is one implementation, so it should not be in there.

lemunozm: It reminds me to `BaseTransferHook`. Seems like it's in some place in between then. But ok for now to leave it there, and see how it evolves.

PR #641, opened 2025-09-29
Gateway with permissionless batch

lemunozm:
### Product requirements

* Enable batching permissionless for integrators

### Design notes

* I've renamed execute to `withBatch` because `execute` can mean executing a message from the gateway


src/hub/Hub.sol
```diff
@@ -82,16 +82,7 @@ contract Hub is Multicall, Auth, Recoverable, IHub, IHubRequestManagerCallback,
...
-        if (!wasBatching) {
-            gateway.startBatching();
-        }
-
-        super.multicall(data);
-
-        if (!wasBatching) {
-            gateway.endBatching();
-        }
+        gateway.withBatch(abi.encodeWithSelector(super.multicall.selector, data));
```
lemunozm: Sadly, `super.multicall.selector` doesn't work because the multicall method is overridden and always takes the child implementation, calling itself.

Possible solutions are:
- Just remove multicall implementation and let the users do it at UX level (super.multicall will be still there)
- ~Rename `multicall` to `multicallWithBatch`~

src/hub/Hub.sol
```diff
@@ -639,10 +624,8 @@ contract Hub is Multicall, Auth, Recoverable, IHub, IHubRequestManagerCallback,
...
     /// @dev Ensure the sender is a pool admin
     function _isManager(PoolId poolId) internal view {
-        require(hubRegistry.manager(poolId, msg.sender), IHub.NotManager());
-    }
-
-    function _payment() internal view returns (uint256 value) {
-        return gateway.isBatching() ? 0 : msg.value;
+        address batcher = gateway.batcher();
+        address caller = address(batcher) != address(0) ? batcher : msg.sender;
+        require(hubRegistry.manager(poolId, caller), IHub.NotManager());
```
lemunozm, line 629: The downside of this implementation, is that the caller to the multicall is no longer `msg.sender`, it's the gateway, so we need to check which sender is who called the `gateway` (it is, call gateway.batcher()`)

The same applies to BalanceSheet, and also to VaultRouter

PR #643, opened 2025-09-30
Hook improvements

hieronx:
### Product requirements

* Fix `maxMint/maxDeposit` behaviour to match deposit claiming checks in `BaseTransferHook`
* Allow separating cross-chain transfer executions from deposit requests and issuances

### TODOs

- [x] Fix Spoke test


src/hooks/FullRestrictions.sol
```diff
@@ -41,6 +41,7 @@ contract FullRestrictions is BaseTransferHook {
         if (isRedeemFulfillment(from, to)) return true;
         if (isRedeemClaimOrRevocation(from, to)) return true;
         if (isCrosschainTransfer(from, to)) return true;
+        if (isCrosschainTransferExecution(from, to)) return isTargetMember(to, hookData);
```
wischli, line 39: Shouldn't this just return true based on your comment ([ref](https://kflabs.slack.com/archives/C07PG2EUR9C/p1759216879358349?thread_ts=1759173152.483119&cid=C07PG2EUR9C))?

>  [F]or deRWA tokens, [...] require[s] being a member for deposit requests and balance sheet issuances, but don't require being a member for transfers, and cross-chain transfers should be unrestricted as well

```suggestion
        if (isCrosschainTransferExecution(from, to)) return true;
```

hieronx: deRWA tokens use the `FreelyTransferable` hook. This is the `FullRestrictions` hook, which should only allow cross-chain transfers to other members (just like it restricts normal transfers).

wischli (approved):

PR #645, opened 2025-09-30
Gateway with permissionless batch and multicall abstraction

lemunozm:
### Product requirements

* Enable batching permissionless for integrators

### Design notes

* I've renamed execute to withBatch because execute can mean executing a message from the gateway
- Abstract multicall with batching support


src/common/BatchedMulticall.sol
```diff
@@ -0,0 +1,34 @@
...
+
+        super.multicall(data);
+
+        gateway.endBatching{value: msg.value}(msg.sender);
+        _isBatching = false;
+    }
+
+    /// @dev gives the current msg.value depending on the batching state
+    function _payment() internal view returns (uint256 value) {
+        return _isBatching ? 0 : msg.value;
```
lemunozm, line 33: We use a transient and not directly `gateway.isBatching()` for performance reasons and to avoid mixing this with `gateway.withBatch`.

If we use `gateway.isBatching()`, and someone uses a hub method inside `withBatch`, any payment done to that hub method can be lost without notification.

src/common/Gateway.sol
```diff
@@ -248,9 +256,32 @@ contract Gateway is Auth, Recoverable, IGateway {
             _gasLimitSlot(centrifugeId, poolId).tstore(uint256(0));
         }
 
+        isBatching = false;
```
lemunozm, line 260: Not an issue but it's more safe if moved after sending

src/common/interfaces/IBatchedMulticall.sol
```diff
@@ -0,0 +1,9 @@
+// SPDX-License-Identifier: GPL-2.0-or-later
+pragma solidity >=0.5.0;
+
+import {IGateway} from "./IGateway.sol";
+
+/// @notice Simple escrow that can be used to deposit and withdraw native tokens
```
hieronx: Nat spec to be updated

src/common/interfaces/IGateway.sol
```diff
@@ -113,6 +120,28 @@ interface IGateway is IMessageHandler, IRecoverable {
...
+    ///         so any method call inside that requires messaging will be batched.
+    /// @param  data encoding data for the callback method
+    /// @dev    Helper contract that enables integrations to automatically batch multiple cross-chain transactions.
+    ///         Should be used like:
+    ///         ```
+    ///         contract Integration {
+    ///             ICrosschainBatcher batcher;
+    ///
+    ///             function doSomething(PoolId poolId) external {
+    ///                 batcher.withBatch(abi.encodeWithSelector(Integration.callback.selector, poolId));
```
hieronx: Example needs to be updated

src/common/BatchedMulticall.sol
```diff
@@ -0,0 +1,35 @@
...
+    function multicall(bytes[] calldata data) public payable override protected {
+        require(!gateway.isBatching(), IGateway.AlreadyBatching());
+        _isBatching = true;
+
+        gateway.startBatching();
+
+        super.multicall(data);
+
+        gateway.endBatching{value: msg.value}(msg.sender);
+        _isBatching = false;
```
hieronx, line 28: Nit:
```suggestion
        require(!gateway.isBatching(), IGateway.AlreadyBatching());
        
        _isBatching = true;
        gateway.startBatching();

        super.multicall(data);

        gateway.endBatching{value: msg.value}(msg.sender);
        _isBatching = false;
```

src/common/Gateway.sol
```diff
@@ -223,16 +224,23 @@ contract Gateway is Auth, Recoverable, IGateway {
     }
 
     /// @inheritdoc IGateway
-    function startBatching() external auth {
+    function startBatching() public auth {
+        _startBatching();
+    }
+
+    function _startBatching() internal {
```
hieronx, line 231: Is the internal method still needed?

src/common/Gateway.sol
```diff
@@ -223,16 +224,23 @@ contract Gateway is Auth, Recoverable, IGateway {
...
         isBatching = true;
     }
 
     /// @inheritdoc IGateway
-    function endBatching(address refund) external payable auth {
+    function endBatching(address refund) public payable auth {
+        _endBatching(refund);
+    }
+
+    function _endBatching(address refund) internal {
```
hieronx, line 241: Same

src/common/Gateway.sol
hieronx: Or even: are these separate methods needed, can they just be integrated in `withBatch`?

src/common/Gateway.sol
```diff
@@ -248,9 +256,32 @@ contract Gateway is Auth, Recoverable, IGateway {
...
         }
 
+        isBatching = false;
+
         _refund(refund, msg.value - cost);
     }
 
+    /// @inheritdoc IGateway
+    function withBatch(bytes memory data, address refund) external payable {
+        require(batcher == address(0), AlreadyBatching());
```
wischli, line 267: Wondering if we should add `require(!isBatching, AlreadyBatching());` below as well to be fully re-entrancy safe as we set `batcher  = address(0)` before `endBatching(refund)`. I know this is purely theoretical.

src/common/Gateway.sol
lemunozm: Both are need it, sadly, one is called from the Hub/BalanceSheet and VaultRouter, the other one is called by the own gateway inside `withBatch`

hieronx: But `withBatch` calls the public methods, so then the internal methods can be merged into these, right?

test/common/unit/BatchedMulticall.t.sol
```diff
@@ -0,0 +1,65 @@
...
+        vm.mockCall(address(gateway), abi.encodeWithSelector(gateway.endBatching.selector), abi.encode());
+
+        bytes[] memory calls = new bytes[](2);
+        calls[0] = abi.encodeWithSelector(multicall.add.selector, 2);
+        calls[1] = abi.encodeWithSelector(multicall.add.selector, 3);
+
+        vm.prank(ANY);
+        multicall.multicall(calls);
+
+        assertEq(multicall.isBatching(), false);
```
wischli, line 67: Nit: Shouldn't we also assert that `total == 5`?

wischli (dismissed):
Really elegant clean-up, incredible work!

test/common/unit/BatchedMulticall.t.sol
lemunozm: Yes! I forgot to check that

src/common/Gateway.sol
lemunozm: Well, maybe the check you propose should be inside startBatching actually, will think on it!

lemunozm: I think I could start a multicall and inside the multicall I can call withBatch right now, when it shouldn't. Good catch!

src/common/BatchedMulticall.sol
lemunozm: I like your nits!

src/common/Gateway.sol
lemunozm: Ouch, that's wrong! It should call the inner ones. Some rebase had break that. Good catch! Actually, I should provide a better test to catch those

lemunozm: Gateway tests assume `auth` by default if not `vm.prank`. I will change that next week, because that is usually a source of issues.

lemunozm:
Added better tests for `withBatch` which was not correctly covered, thanks for your highlights @hieronx and @wischli

src/common/interfaces/IGateway.sol
```diff
@@ -113,6 +120,31 @@ interface IGateway is IMessageHandler, IRecoverable {
     /// @notice Finalize batching messages and send the resulting batch message
     function endBatching(address refund) external payable;
 
+    /// @notice Calls a method that should be in the same contract as the caller, as a callback.
+    ///         The method called will be wrapped inside startBatching and endBatching,
+    ///         so any method call inside that requires messaging will be batched.
+    /// @param  data encoding data for the callback method
```
hieronx: `@param` should be below `@dev`

hieronx (approved):

PR #646, opened 2025-09-30
Fix burrasec issue regarding gas process.

lemunozm:
### Issue

* The payment expected in handle side can be less than the payment used when sending. This fix skips that check to ensure it's always higher


src/common/Gateway.sol
```diff
@@ -101,9 +101,7 @@ contract Gateway is Auth, Recoverable, IGateway {
             bytes memory message = remaining.slice(0, length);
             remaining = remaining.slice(length, remaining.length - length);
 
-            uint256 executionGas = gasService.messageGasLimit(localCentrifugeId, message);
-            require(gasleft() >= executionGas + GAS_FAIL_MESSAGE_STORAGE, NotEnoughGasToProcess());
-
+            require(gasleft() > GAS_FAIL_MESSAGE_STORAGE, NotEnoughGasToProcess());
```
hieronx, line 106: Unrelated note: do we still need the internal `_process` method? Think I asked this in a previous PR, but it's now even simpler.

hieronx (dismissed):

src/common/Gateway.sol
lemunozm: It's split to correctly perform a measurement from the tests the value needed for populate `GAS_FAIL_MESSAGE_STORAGE`. We could remove it, but we can no longer test the gas in that section...

hieronx: Why can't we use a harness contract to test, like  https://github.com/centrifuge/protocol/blob/main/test/vaults/unit/BatchRequestManager.t.sol#L78

hieronx: See https://getfoundry.sh/guides/best-practices/writing-tests/#test-harnesses

lemunozm: It's what is done now, called `GatewatExt` (probably should rename it): https://github.com/centrifuge/protocol/blob/main/test/common/unit/Gateway.t.sol#L119-L121

hieronx: Ah got it. Could we reimplement `_process` in that method in `GatewayExt`? To be able to measure it.

lemunozm: We can copy the code yeah

lemunozm: Done [1cd1834](https://github.com/centrifuge/protocol/pull/646/commits/1cd1834303a4233b167a9208705d5e94bc5a961c)

src/common/Gateway.sol
```diff
@@ -100,18 +100,17 @@ contract Gateway is Auth, Recoverable, IGateway {
...
-        }
-    }
 
-    function _process(uint16 centrifugeId, bytes memory message, bytes32 messageHash) internal {
-        try processor.handle{gas: gasleft() - GAS_FAIL_MESSAGE_STORAGE}(centrifugeId, message) {
-            emit ExecuteMessage(centrifugeId, message, messageHash);
-        } catch (bytes memory err) {
-            failedMessages[centrifugeId][messageHash]++;
-            emit FailMessage(centrifugeId, message, messageHash, err);
+            // This "catch" branch below code must be computed in GAS_FAIL_MESSAGE_STORAGE gas units
```
hieronx: I personally think this is pretty obvious, but okay to leave if you feel it's dont.

```suggestion
```

test/common/unit/Gateway.t.sol
```diff
@@ -117,7 +117,12 @@ contract GatewayExt is Gateway {
     }
 
     function process(uint16 centrifugeId, bytes memory message, bytes32 messageHash) public {
-        _process(centrifugeId, message, messageHash);
+        try processor.handle{gas: gasleft() - GAS_FAIL_MESSAGE_STORAGE}(centrifugeId, message) {
```
hieronx, line 121: Here I would actually add a comment :)
```suggestion
        // Copied only the try catch from `handle`, to be able to measure the gas cost
        try processor.handle{gas: gasleft() - GAS_FAIL_MESSAGE_STORAGE}(centrifugeId, message) {
```

hieronx (dismissed):

hieronx (approved):

PR #647, opened 2025-09-30
LayerZero gas change

lemunozm:
As an idea of the fix


src/adapters/LayerZeroAdapter.sol
```diff
@@ -70,11 +72,14 @@ contract LayerZeroAdapter is Auth, ILayerZeroAdapter {
         external
         payable
     {
+        require(payload.toUint64(0) >= gasleft());
```
lemunozm, line 75: Maybe we should add some offset for the current waste gas:
```
require(payload.toUint64(0) >= gasleft() - OFFSET);
```

hieronx (commented):
Yeah this could work!

PR #648, opened 2025-09-30
Fix: adapter session id collistion

lemunozm:
### Product requirements

* We need to have different session ids for global pool and pool adapters. Related issue: https://github.com/burrasec/2025-09-Centrifuge/issues/1

### Design notes

* Start counting from global adapters sessionId because pool adapters can never be removed


snapshots/AsyncVault.json
```diff
@@ -1,4 +1,4 @@
 {
-  "fulfillDepositRequest": "806029",
```
lemunozm: I think we're measuring something wrong because this change should not affect here 🤔

hieronx (approved):
LGTM!

PR #649, opened 2025-10-01
Withdraw vaults subsidity

lemunozm:
### Product requirements

* The pool manager should be able to withdraw subsidized funds used in AsyncRequestManager

### Design notes

* Added a new message to UpdateContract to achive this


src/vaults/AsyncRequestManager.sol
```diff
@@ -171,8 +187,12 @@ contract AsyncRequestManager is Auth, Recoverable, IAsyncRequestManager {
         IRefundEscrow refund;
         uint256 payment;
 
-        if (!balanceSheet.gateway().isBatching()) {
-            refund = refundEscrowFactory.get(vault_.poolId());
+        PoolId poolId = vault_.poolId();
+        AssetId assetId = spoke.vaultDetails(vault_).assetId;
+        bool isLocal = poolId.centrifugeId() == assetId.centrifugeId();
+
+        if (!balanceSheet.gateway().isBatching() || isLocal) {
```
lemunozm: Just an addition to improve the performance when the message is sent locally

src/common/MessageDispatcher.sol
```diff
@@ -677,7 +677,7 @@ contract MessageDispatcher is Auth, IMessageDispatcher {
 
     function _refund(address refund) internal {
         if (msg.value > 0) {
-            (bool success,) = payable(refund).call{value: msg.value}(new bytes(0));
+            (bool success,) = payable(refund).call{value: msg.value}("");
```
hieronx, line 683: For my context: why?

src/vaults/AsyncRequestManager.sol
hieronx: Should be `!isLocal`, right?

lemunozm: Thanks for our eagle view! and `&&`

src/common/MessageDispatcher.sol
lemunozm: Is the same to the compiler, more legible to me. But no strong opinion

test/integration/EndToEnd.t.sol
```diff
@@ -1248,4 +1249,23 @@ contract EndToEndUseCases is EndToEndFlows, VMLabeling {
 
         assertEq(s.spoke.pool(POOL_A), block.timestamp); // 2 of 2 received and processed
     }
+
+    /// forge-config: default.isolate = true
```
wischli, line 1255: Nit: I feel like we use this restriction too excessively as default. Is there any particular reason for it to be required here? Just an observation from the past reviews and since I don't have anything else to say here, I finally spoke up 😅

wischli (commented):
Apart from Jeroen's spotted required change LGTM!

test/integration/EndToEnd.t.sol
lemunozm: Because we're replicating a real scenario where each call means a different transaction. 

The initial reason was some transients that were expected to be reset in some conditions, I don't remember whose. Not sure if regarding the accounting when receiving messages or so, or regarding epochs in ShareClassManager

Maybe we can try removing all of them and see if something fails. Maybe it's no longed need it

lemunozm:
`AsyncRequestManager` exceeds the max size, maybe we can remove `Recoverable` from it, given that the only one who should call it is the refundEscrow, no one else.

Of course, a random person can send funds to it, but that should never happen, and not being this in the core, we could not provide a tool to recover this

hieronx:
> AsyncRequestManager exceeds the max size, maybe we can remove Recoverable from it, given that the only one who should call it is the refundEscrow, no one else.
> Of course, a random person can send funds to it, but that should never happen, and not being this in the core, we could not provide a tool to recover this

Yeah I think that's reasonable.

wischli (approved):

hieronx (approved):

PR #650, opened 2025-10-01
Vault Registry

hieronx:
### Product requirements

* Separate vault logic from core spoke, to allow future migrations.


src/spoke/Spoke.sol
```diff
@@ -266,6 +240,12 @@ contract Spoke is Auth, Recoverable, ReentrancyProtection, ISpoke, ISpokeGateway
         emit ExecuteTransferShares(poolId, scId, receiver.toAddress(), amount);
     }
 
+    /// @inheritdoc ISpoke
+    function setShareTokenVault(PoolId poolId, ShareClassId scId, address asset, address vault) external auth {
+        IShareToken token = shareToken(poolId, scId);
+        token.updateVault(asset, vault);
```
wischli, line 246: Wondering if we should add a check if the given vault address is actually linked.

src/vaults/VaultRouter.sol
```diff
@@ -36,18 +37,20 @@ contract VaultRouter is Auth, BatchedMulticall, Recoverable, IVaultRouter {
     uint256 private constant REQUEST_ID = 0;
 
     ISpoke public immutable spoke;
+    IVaultRegistry public immutable vaultRegistry;
```
wischli, line 40: I suppose we need to deploy a new VaultRouter anyways when we upgrade the VaultRegistry?

src/spoke/VaultRegistry.sol
```diff
@@ -0,0 +1,172 @@
...
+/// @title  VaultRegistry
+/// @notice This contract manages vault deployment, linking, and unlinking operations for pools and share classes
+contract VaultRegistry is Auth, Recoverable, IVaultRegistry, IVaultRegistryGatewayHandler {
+    mapping(
+        PoolId poolId => mapping(ShareClassId scId => mapping(AssetId assetId => mapping(IRequestManager => IVault)))
+    ) public vault;
+
+    mapping(IVault => VaultDetails) internal _vaultDetails;
+
+    ISpoke public spoke;
```
wischli: Nit: Shouldn't this be at top of the storage ordering by our convention? And maybe also `_vaultDetails` in the middle.

script/VaultsDeployer.s.sol
```diff
@@ -36,6 +36,10 @@ contract VaultsActionBatcher is SpokeActionBatcher {
         report.syncDepositVaultFactory.rely(address(report.spoke.spoke));
         report.asyncRequestManager.rely(address(report.spoke.spoke));
 
+        // Rely VaultRegistry
+        report.asyncVaultFactory.rely(address(report.spoke.vaultRegistry));
```
wischli, line 38: Above these lines we have a ward from the vault factories to the spoke which can be removed AFAICT

```diff
-        report.asyncVaultFactory.rely(address(report.spoke.spoke));
-        report.syncDepositVaultFactory.rely(address(report.spoke.spoke));
```

src/common/interfaces/IVaultRegistryGatewayHandler.sol
```diff
@@ -0,0 +1,24 @@
+// SPDX-License-Identifier: GPL-2.0-or-later
+pragma solidity >=0.5.0;
+
+import {PoolId} from "../types/PoolId.sol";
+import {AssetId} from "../types/AssetId.sol";
+import {ShareClassId} from "../types/ShareClassId.sol";
+import {VaultUpdateKind} from "../libraries/MessageLib.sol";
+
+/// @notice Interface for VaultRegistry methods called by messages
+interface IVaultRegistryGatewayHandler {
```
wischli: Nit: Curious why we didn't add this to the exisiting `src/common/interfaces/IGatewayHandlers`?

wischli (commented):
Overall LGTM!

src/common/interfaces/IVaultRegistryGatewayHandler.sol
lemunozm: Agree

test/spoke/unit/VaultRegistry.t.sol
```diff
@@ -0,0 +1,510 @@
...
+    uint8 constant DECIMALS = 18;
+    string constant NAME = "name";
+    string constant SYMBOL = "symbol";
+    bytes32 constant SALT = "salt";
+    bytes constant PAYLOAD = "payload";
+
+    D18 immutable PRICE = d18(42e18);
+    uint128 constant AMOUNT = 200;
+    uint64 immutable MAX_AGE = 10_000;
+    uint64 immutable PAST_OLD = 0;
```
lemunozm: NIT. A lot of constants and mocks can be cleaned from here. Similar thing but opposite in `Spoke.t.sol`

lemunozm (commented):
Apart from William comments, LGTM!

src/vaults/VaultRouter.sol
lemunozm: Yes, we already have changed multicall stuff there, and added a couple of methods too

src/spoke/Spoke.sol
lemunozm: I don't think so, because this method is also used to unlink, so `vault == addres(0)`

src/spoke/VaultRegistry.sol
hieronx: Normally would group only within a section, so public and internal separate. But given there are only two here, agree it makes sense to group both.

wischli (dismissed):

lemunozm (approved):
LGTM!

PR #651, opened 2025-10-01
feat: add update hub contract flow

wischli:
### Product requirements

- Enable external protocols (such as oracles) on spoke chains to trigger arbitrary contract updates on hub-side contracts similar to current `UpdateContract` flow from hub to spoke
- Support both same-chain and cross-chain execution paths
- Maintain same payment flow as other message types (payable + refund pattern from PR #636)
- Preserve sender identity through `bytes32` conversion for pure hub-side authorization (permissionless dispatch on spoke side)

### Design notes

- New message type: `UpdateHubContract` with same fields of `UpdateContract` and an additional `bytes32 sender`
- Entry point: `Spoke.updateHubContract()`
- Executor: new `HubContractUpdater`

### Why introduce as separate message instead of extending UpdateContract?

**Option A (Chosen): Separate `UpdateHubContract `message**
```solidity
// Spoke -> Hub direction
UpdateHubContract(poolId, scId, sender, target, payload)
// Hub → Spoke direction  
UpdateContract(centrifugeId, poolId, scId, target, payload)
```

**Option B (Alternative): Add `sender` field to `UpdateContract`**
```solidity
UpdateContract(centrifugeId, poolId, scId, sender, target, payload)
```

1. Adding unused `sender` field to `UpdateContract` seemed more hacky to me (but no strong opinion)
2. Authorization model: `UpdateContract` is authenticated at Hub level permissioned to manager whereas `UpdateHubContract` is authenticated at target contract level
3. Separate messages make direction explicit: `UpdateContract` is Hub -> Spoke with manager pushing updates whereas `UpdateHubContract` is Spoke -> Hub for external protocol pulling updates

### TODO

- [ ] Maybe find better name or rename `UpdateContract` to `UpdateSpokeContract`?


src/messaging/libraries/MessageLib.sol
```diff
@@ -130,6 +132,8 @@ library MessageLib {
             length += 2 + message.toUint16(length); //payloadLength
         } else if (kind == uint8(MessageType.UpdateContract)) {
             length += 2 + message.toUint16(length); //payloadLength
+        } else if (kind == uint8(MessageType.UpdateHubContract)) {
```
lemunozm, line 135: We need special treatment of this inside `messageSourceCentrifugeId` because the poolId in this message does not correlate with the original chain. I think we need to return 0 in `messageSourceCentrifugeId` for this message.

src/messaging/MessageDispatcher.sol
```diff
@@ -600,6 +603,37 @@ contract MessageDispatcher is Auth, IMessageDispatcher {
...
+        bytes32 target,
+        bytes32 sender,
+        bytes calldata payload,
+        uint128 extraGasLimit,
+        address refund
+    ) external payable auth {
+        uint16 hubCentrifugeId = poolId.centrifugeId();
+
+        if (hubCentrifugeId == localCentrifugeId) {
+            hubContractUpdater.execute(poolId, scId, sender.toAddress(), target.toAddress(), payload);
```
lemunozm, line 623: FYI, the order or sender and target does not correlate the message. I think we need to normalize those to avoid issues.

lemunozm (commented):
Also agree renaming `UpdateContract` to `UpdateSpokeContract` and similar references, including `ContractUpdater` to `SpokeContractUpdater`

src/core/hub/HubContractUpdater.sol
```diff
@@ -0,0 +1,30 @@
...
+
+import {Auth} from "../misc/Auth.sol";
+
+import {PoolId} from "../common/types/PoolId.sol";
+import {ShareClassId} from "../common/types/ShareClassId.sol";
+import {IUpdateHubContractGatewayHandler} from "../common/interfaces/IGatewayHandlers.sol";
+
+/// @title HubContractUpdater
+/// @notice Central executor for spoke-to-hub contract updates
+contract HubContractUpdater is Auth, IUpdateHubContractGatewayHandler {
```
lemunozm, line 14: Not sure if we need this because we already have `HubHandler` as source of truth in the `Hub`. I think this can be removed

hieronx:
Replaced by https://github.com/centrifuge/protocol/pull/663

src/messaging/libraries/MessageLib.sol
lemunozm: Fixed here: https://github.com/centrifuge/protocol/pull/665

PR #652, opened 2025-10-01
protect batcher against reentrancy

lemunozm:
Thread: https://kflabs.slack.com/archives/C04HF8SHRRC/p1759326168913709


src/common/Gateway.sol
```diff
@@ -275,8 +273,14 @@ contract Gateway is Auth, Recoverable, IGateway {
...
-        batcher = address(0);
         _endBatching(refund);
+        _batcher = address(0);
+    }
+
+    function lockCallback() external returns (address caller) {
+        require(address(_batcher) != address(0), CallbackIsLocked());
+        caller = _batcher;
+        _batcher = address(0);
     }
```
lemunozm, line 287: Calling this method on top of the callback means 3 things:
- The callback that uses this can only be called once.
- The callback is called from the gateway under `withBatch`, otherwise `_batcher` will be 0.
- The callback is called from the same contract, because withBatch force `msg.sender`

src/common/Gateway.sol
```diff
@@ -275,8 +273,14 @@ contract Gateway is Auth, Recoverable, IGateway {
             }
         }
 
-        batcher = address(0);
         _endBatching(refund);
+        _batcher = address(0);
+    }
+
+    function lockCallback() external returns (address caller) {
```
wischli, line 283: Nit: 
```suggestion
    /// @inheritdoc IGateway
    function lockCallback() external returns (address caller) {
```

wischli (commented):
AFAICT, provides sufficient protection!

src/common/Gateway.sol
lemunozm: Really like your attention to doc stuff! Thanks

src/common/Gateway.sol
```diff
@@ -275,8 +273,15 @@ contract Gateway is Auth, Recoverable, IGateway {
             }
         }
 
-        batcher = address(0);
         _endBatching(refund);
+        _batcher = address(0);
```
hieronx: Shouldn't we check here `require(_batcher == address(0))`, to force `lockCallback` to have been called?

src/common/Gateway.sol
```diff
@@ -44,7 +44,7 @@ contract Gateway is Auth, Recoverable, IGateway {
     mapping(PoolId => mapping(address => bool)) public manager;
 
     // Outbound & payments
-    address public transient batcher;
+    address transient _batcher;
```
hieronx: Why is this _batcher instead of batcher now?

src/common/Gateway.sol
```diff
@@ -275,8 +273,15 @@ contract Gateway is Auth, Recoverable, IGateway {
             }
         }
 
-        batcher = address(0);
         _endBatching(refund);
+        _batcher = address(0);
+    }
+
+    /// @inheritdoc IGateway
+    function lockCallback() external returns (address caller) {
```
hieronx, line 283: Is it even helpful to return the `caller` here? Should always be the callback destination right?

src/common/Gateway.sol
lemunozm: Because batcher is private, to ensure it's correctly read, you need to go through `lockCallback()`

src/common/Gateway.sol
lemunozm: The caller here is essential, because it is the original `msg.sender` used to call `withBatch`. We don't have that info inside the callback. Inside the callback you need to use this `caller` instead of `msg.sender`

src/common/Gateway.sol
lemunozm: Great idea! Really nice idea!

src/common/Gateway.sol
hieronx: Can you add the `private` keyword, to make this explicit?

src/common/interfaces/IGateway.sol
```diff
@@ -139,17 +140,25 @@ interface IGateway is IMessageHandler, IRecoverable {
     ///             }
     ///
     ///             function callback(PoolId poolId) external {
-    ///                 require(gateway.batcher() == address(this));
+    ///                 // Avoid reentrancy and ensure it's called from withBatch in the same contract:
+    ///                 address doSomethingSender = gateway.lockCallback();
```
hieronx: More conventional (often used with OZ: https://github.com/OpenZeppelin/openzeppelin-contracts/blob/5eb047a0c10386b6634db56327a1f74e8a0432c2/contracts/utils/Context.sol#L17)
```suggestion
    ///                 address msgSender = gateway.lockCallback();
```

src/common/interfaces/IGateway.sol
```diff
@@ -139,17 +140,25 @@ interface IGateway is IMessageHandler, IRecoverable {
...
+    ///
     ///                 // Call several hub, balance sheet, or spoke methods that trigger cross-chain transactions
     ///             }
     ///         }
     ///         ```
-    /// @param  data encoding data for the callback method
-    function withBatch(bytes memory data, address refund) external payable;
+    ///
+    ///         NOTE: inside callback, `doSomethingSender` should be used instead of msg.sender
```
hieronx:
```suggestion
    ///         NOTE: inside callback, `msgSender` should be used instead of msg.sender
```

src/common/Gateway.sol
```diff
@@ -44,7 +44,7 @@ contract Gateway is Auth, Recoverable, IGateway {
     mapping(PoolId => mapping(address => bool)) public manager;
 
     // Outbound & payments
-    address public transient batcher;
+    address internal transient _batcher;
     bool public transient isBatching;
     bool public transient unpaidMode;
```
```suggestion
bool public transient isBatching;
bool public transient unpaidMode;
address internal transient _batcher;
```

src/common/interfaces/IGateway.sol
lemunozm: makes sense

src/common/Gateway.sol
lemunozm: I made internal, in case in the future we want to improve test coverage

hieronx (approved):

xmxanuel (commented):
looks good

PR #653, opened 2025-10-01
normalize hub API params

lemunozm:
### Product requirements

* Smooth and predictable API

### Design notes

* Normalize all `Hub` methods to start with the `PoolId` param which is the most relevant parameter


src/hub/Hub.sol
```diff
@@ -452,8 +452,8 @@ contract Hub is BatchedMulticall, Auth, Recoverable, IHub, IHubRequestManagerCal
```
hieronx, line 452: I think you missed `setMaxSharePriceAge

lemunozm: Correct!

hieronx (approved):

PR #654, opened 2025-10-01
feat: split Guardian into ProtocolGuardian & OpsGuardian

wischli:
### Product requirements

* Split the single `Guardian` contract into two separate contracts with distinct responsibilities: `OpsGuardian` for operational tasks and `ProtocolGuardian` for protocol-critical governance
* Move `createPool` to `OpsGuardian` as it is an operational (non-critical) function
* Move `setAdapters` and `blockOutgoing` to `ProtocolGuardian` as they are protocol-critical emergency controls
* Add `initAdapters` to `OpsGuardian` for first-time network setup only, which reverts if adapters already exist for the target `centrifugeId`
* Grant `ProtocolGuardian` direct ward access to `Gateway` to enable `blockOutgoing` across all pools without needing manager rights
* Extract adapter wiring functionality from `IAdapter` into a separate `IAdapterWiring` interface to clarify which adapters support cross-chain wiring configuration
* Add explicit `centrifugeId` parameter to `wire()` functions to improve clarity and eliminate redundant ABI decoding

### Design notes

* `OpsGuardian` handles operational tasks: `createPool` and `initAdapters` (first-time adapter setup only)
* `ProtocolGuardian` handles protocol-critical tasks: `setAdapters` (local-only), `blockOutgoing`, `pause/unpause`, scheduled relies, upgrades, and token recovery
  * Did not implement cross-chain message sending from `ProtocolGuardian.setAdapters` due to open question about this flow
* Removed `updateGatewayManager` entirely; `ProtocolGuardian` has direct Gateway ward access instead of using manager rights
* Individual adapters (Wormhole, Axelar, LayerZero) no longer grant ward access to any guardian contract, only to `Root` and `AdminSafe`
* Both guardians maintain the same `Safe`-based access control pattern from the original Guardian contract
* OpsGuardian implements self-revoking security pattern: both `wire()` and `initAdapters()` call `deny(address(this))` after completing one-time setup operations, ensuring OpsGuardian cannot modify adapters or MultiAdapter configuration after initial deployment
* Created `IAdapterWiring` interface in `src/admin/interfaces/` containing `wire()` and `isWired()` methods
  * Only bridge adapters (AxelarAdapter, LayerZeroAdapter, WormholeAdapter) implement `IAdapterWiring`
* Updated bridge adapter implementations to accept `centrifugeId` as explicit parameter instead of extracting it from ABI-encoded data

### Breaking changes

* `IAdapter.wire()` and `IAdapter.isWired()` removed from base adapter interface
* Bridge adapter `wire()` methods now take `centrifugeId` as separate parameter
* Guardian contracts split into two separate contracts with different responsibilities and permission models


script/CommonDeployer.s.sol
```diff
@@ -219,16 +233,25 @@ abstract contract CommonDeployer is Script, JsonRegistry, CreateXScript {
...
                     abi.encode(ISafe(address(batcher)), root, gateway, multiAdapter, messageDispatcher)
                 )
             )
         );
 
+        opsGuardian = OpsGuardian(
+            create3(
+                generateSalt("opsGuardian"),
+                abi.encodePacked(
+                    type(OpsGuardian).creationCode, abi.encode(ISafe(address(batcher)), address(0), multiAdapter)
```
wischli, line 245: Unfortunately, we cannot reference the `hub` here which is why we later file it in the HubDeployer

lemunozm: Yeah, the same issue we had with the previous `Guardian`, but I think, after the Deployer refactor, we could now pass the hub argument, which is great!

src/common/OpsGuardian.sol
```diff
@@ -0,0 +1,62 @@
...
+    //----------------------------------------------------------------------------------------------
+    // Adapter Management
+    //----------------------------------------------------------------------------------------------
+
+    /// @inheritdoc IOpsGuardian
+    function initAdapters(uint16 centrifugeId, IAdapter[] calldata adapters, uint8 threshold, uint8 recoveryIndex)
+        external
+        onlySafe
+    {
+        require(multiAdapter.quorum(centrifugeId, PoolId.wrap(0)) == 0, AdaptersAlreadyInitialized());
```
hieronx: Would add a constant for `GLOBAL_POOL = PoolId.wrap(0)`

src/common/OpsGuardian.sol
```diff
@@ -0,0 +1,62 @@
...
+    //----------------------------------------------------------------------------------------------
+    // Pool Management
+    //----------------------------------------------------------------------------------------------
+
+    /// @inheritdoc IOpsGuardian
+    function createPool(PoolId poolId, address admin, AssetId currency) external onlySafe {
+        hub.createPool(poolId, admin, currency);
+    }
+
+    //----------------------------------------------------------------------------------------------
```
hieronx: Would put this section below the modifier, like in all other contracts

src/admin/OpsGuardian.sol
```diff
@@ -0,0 +1,62 @@
...
+
+    /// @inheritdoc IOpsGuardian
+    function initAdapters(uint16 centrifugeId, IAdapter[] calldata adapters, uint8 threshold, uint8 recoveryIndex)
+        external
+        onlySafe
+    {
+        require(multiAdapter.quorum(centrifugeId, PoolId.wrap(0)) == 0, AdaptersAlreadyInitialized());
+        multiAdapter.setAdapters(centrifugeId, PoolId.wrap(0), adapters, threshold, recoveryIndex);
+    }
+
```
hieronx, line 58: Just realized, don't we need functions for the `file` methods in adapters, like https://github.com/centrifuge/protocol/blob/release-v3.0.1/src/common/adapters/WormholeAdapter.sol#L43, here as well?

wischli: Not necessary after https://github.com/centrifuge/protocol/pull/566, see [comment here](https://github.com/centrifuge/protocol/pull/566#discussion_r2257672049).

hieronx: Sorry not `file`, but we do need a function to call `wire`, right?

wischli: That's also not necessary as per #566 which added `safeAdmin` as ward of each adapter such that the previous `Guardian.wireAdapters` could be deprecated back then. The flow is 
1. `ProtocolGuardian.setAdapters` via safe
2. `{Axelar, LayerZero, Wormhole}Adapter.wire` via safe

https://github.com/centrifuge/protocol/blob/9ab0d076e017b33a60d81aa4770a128a371ff081/script/AdaptersDeployer.s.sol#L42-L56

hieronx: Thanks for reminding me! Aligned, makes sense.

Only challenge now is: I don't think the ops guardian safe can be a ward on the adapters, that is way too much access.

But I do think we need the ops guardian to be able to wire new networks, otherwise we cannot set up new networks at all without the involvement of the protocol guardian..

wischli: Then I suppose we should drop `ward` access of `OpsGuardian` to `MultiAdapter` after calling `multiAdapter.setAdapters`. WDYT?

hieronx: I don't follow 🤔 How would that solve the problem? We need to be able to call `*adapter.wire` for new networks from the OpsGuardian.

script/AdaptersDeployer.s.sol
```diff
@@ -43,17 +43,14 @@ contract AdaptersActionBatcher is CommonActionBatcher {
     function engageAdapters(AdaptersReport memory report) public onlyDeployer {
         if (address(report.wormholeAdapter) != address(0)) {
             report.wormholeAdapter.rely(address(report.common.root));
-            report.wormholeAdapter.rely(address(report.common.guardian));
```
lemunozm, line 46: Actually that was unused before, right?

script/CommonDeployer.s.sol
```diff
@@ -90,7 +102,8 @@ contract CommonActionBatcher {
 
     function postEngageCommon(CommonReport memory report) public onlyDeployer {
         // We override the deployer with the correct admin once everything is deployed
-        report.guardian.file("safe", address(report.adminSafe));
+        report.protocolGuardian.file("safe", address(report.adminSafe));
+        report.opsGuardian.file("safe", address(report.adminSafe));
```
lemunozm: Should we configure two different `adminSafe`?

script/CommonDeployer.s.sol
lemunozm: I think you don't need to pass any argument directly in the constructor if later it's filed later, but we can keep it as it is because after the deployer refactor we can pass the hub 👍🏻

src/admin/interfaces/IProtocolGuardian.sol
```diff
@@ -68,12 +53,24 @@ interface IGuardian {
...
     ) external payable;
 
-    /// @notice Set adapters into MultiAdapter. Check IMultiAdapter docs
+    /// @notice Set adapters locally for global pool
+    /// @dev Local-only operation, does not send cross-chain message
+    /// @param centrifugeId Target chain ID to configure adapters on
+    /// @param adapters Array of adapter contract addresses
+    /// @param threshold Minimum number of adapters that must agree
+    /// @param recoveryIndex Index of the recovery adapter in the array
     function setAdapters(uint16 centrifugeId, IAdapter[] calldata adapters, uint8 threshold, uint8 recoveryIndex)
```
lemunozm, line 56: Because set adapters only modify something already set IIUC, I would rename it to `updateAdapters` to follow the convention

src/common/interfaces/IProtocolGuardian.sol
```diff
@@ -68,12 +53,24 @@ interface IGuardian {
...
+    /// @dev Local-only operation for fast emergency response
+    /// @param centrifugeId Target chain ID to block/unblock
+    /// @param isBlocked True to block outgoing messages, false to unblock
+    function blockOutgoing(uint16 centrifugeId, bool isBlocked) external;
+
+    /// @notice Updates a contract parameter.
+    /// @param what Name of the parameter to update.
+    /// Accepts a `bytes32` representation of 'safe', 'hub', 'sender', 'gateway', or 'multiAdapter' string value.
+    /// @param data New value given to the `what` parameter
+    function file(bytes32 what, address data) external;
```
lemunozm: Ultra NIT. Could you move `file()` to the first method in the interface?

src/common/MessageDispatcher.sol
```diff
@@ -637,7 +637,7 @@ contract MessageDispatcher is Auth, IMessageDispatcher {
         address refund
     ) external payable auth {
         if (centrifugeId == localCentrifugeId) {
-            revert CannotBeSentLocally();
+            _refund(refund);
```
lemunozm: CR. This case is special, we want to perform an error here instead of success silently. setPoolAdapters must fail if used on the same network because it's logically an error from the caller. This is a special message that only makes sense across chains.

This differs from other calls. When they succeed, they must refund the user in case they pay

src/admin/ProtocolGuardian.sol
```diff
@@ -118,17 +98,35 @@ contract Guardian is IGuardian {
...
+    // Adapter Management
+    //----------------------------------------------------------------------------------------------
+
+    /// @inheritdoc IProtocolGuardian
     function setAdapters(uint16 centrifugeId, IAdapter[] calldata adapters, uint8 threshold, uint8 recoveryIndex)
         external
         onlySafe
     {
-        multiAdapter.setAdapters(centrifugeId, PoolId.wrap(0), adapters, threshold, recoveryIndex);
+        multiAdapter.setAdapters(centrifugeId, GLOBAL_POOL, adapters, threshold, recoveryIndex);
```
lemunozm, line 127: Q: Should this fail if it was never configured, as opposite to initAdapters

test/integration/EndToEnd.t.sol
```diff
@@ -1203,9 +1211,18 @@ contract EndToEndUseCases is EndToEndFlows, VMLabeling {
...
+        uint256 refundBalanceBefore = REFUND.balance;
+
         vm.startPrank(FM);
         h.hub.setAdapters{value: GAS}(h.centrifugeId, POOL_A, localAdapters, remoteAdapters, 1, 1, REFUND);
+
+        // Verify the gas was refunded
+        uint256 refundBalanceAfter = REFUND.balance;
+        assertEq(
+            refundBalanceAfter, refundBalanceBefore + GAS, "Gas should be refunded for local adapter configuration"
+        );
```
lemunozm: This is a programming error by the pool manager; it should not succeed

lemunozm (commented):
Awesome PR that for sure was tedious with additional operational complexity.

Everything looks great, just a CR

src/admin/OpsGuardian.sol
lemunozm: > But I do think we need the ops guardian to be able to wire new networks, otherwise we cannot set up new networks at all without the involvement of the protocol guardian.

Partially agree. I think we need the `safe` associated with `OpsGuardian` to be able to wire adapters. IIUC, there should be two `safe` accounts, right? one for each Guardian.

wischli: > I don't follow 🤔 How would that solve the problem? We need to be able to call `*adapter.wire` for new networks from the OpsGuardian.

Currently, we don't. We need to call it from the `safe` such that we can remove the ward after initializing. So the question is really the one @lemunozm raised: Should there be two safes?

src/admin/interfaces/IProtocolGuardian.sol
wischli: IMO, as a safety fallback this method should always work and not require any precondition. WDYT @hieronx ?

src/admin/OpsGuardian.sol
hieronx: Yes there should be separate safes.

src/common/OpsGuardian.sol
```diff
@@ -53,6 +54,18 @@ contract OpsGuardian is IOpsGuardian {
         multiAdapter.setAdapters(centrifugeId, GLOBAL_POOL, adapters, threshold, recoveryIndex);
     }
 
+    /// @inheritdoc IBaseGuardian
+    function wire(address adapter, bytes memory data) external onlySafe {
+        uint16 centrifugeId;
+        assembly {
```
hieronx: Good idea to check the actual centrifugeId that is passed this way!

I think you can also do `data.toUint16(0)` for this, to not use assembly.

src/adapters/AxelarAdapter.sol
```diff
@@ -40,13 +40,20 @@ contract AxelarAdapter is Auth, IAxelarAdapter {
     // Administration
     //----------------------------------------------------------------------------------------------
 
-    /// @inheritdoc IAxelarAdapter
-    function wire(uint16 centrifugeId, string calldata axelarId, string calldata adapter) external auth {
+    /// @inheritdoc IAdapter
+    function wire(bytes memory data) external auth {
```
lemunozm: Because `centrifugeId` always exists in all adapters, we could extract it from the encoding. Maybe it helps in readability:

````
function wire(uint16 centrifugeId, bytes memory data) external
```

src/admin/interfaces/IProtocolGuardian.sol
lemunozm: Discussed in the sync to let the same name and purpose

src/common/MultiAdapter.sol
```diff
@@ -56,6 +56,18 @@ contract MultiAdapter is Auth, IMultiAdapter {
...
+    function wire(bytes memory) external pure {
+        revert("MultiAdapter: wire not supported");
+    }
+
+    /// @inheritdoc IAdapter
+    /// @dev MultiAdapter is not a cross-chain adapter, so wire/isWired are not applicable
+    function isWired(uint16) external pure returns (bool) {
+        revert("MultiAdapter: isWired not supported");
+    }
+
```
lemunozm: Thinking now that this method should go under some new interface that only adapters that require them implement. These `wire` methods are never called from the `core`, only from `admin`, which is on top.

Maybe `IAdapterWiring` or so, an interface that exists only in `src/admin` and only adapters from `src/adapters` choose to implement.

Otherwise we're populating the core with something from outside...

src/common/OpsGuardian.sol
lemunozm: We can also pass explicitelly the `centrifugeId`, extracting this from the encoding:

`wire(address adapter, uint16 centrifugeId, bytes memory data)`. Being data, only the rest of the content, which is adapter-dependent

wischli: I went ahead with @lemunozm's proposal

script/CommonDeployer.s.sol
```diff
@@ -60,27 +65,30 @@ contract CommonActionBatcher {
...
-        report.messageDispatcher.rely(address(report.guardian));
         report.messageProcessor.rely(address(report.root));
+        report.messageDispatcher.rely(address(report.protocolGuardian));
         report.messageProcessor.rely(address(report.gateway));
         report.tokenRecoverer.rely(address(report.root));
+        report.tokenRecoverer.rely(address(report.protocolGuardian));
         report.tokenRecoverer.rely(address(report.messageDispatcher));
         report.tokenRecoverer.rely(address(report.messageProcessor));
         report.poolEscrowFactory.rely(address(report.root));
+        report.poolEscrowFactory.rely(address(report.protocolGuardian));
```
lemunozm: Where is this need it?

src/admin/OpsGuardian.sol
```diff
@@ -0,0 +1,77 @@
...
+        require(multiAdapter.quorum(centrifugeId, GLOBAL_POOL) == 0, AdaptersAlreadyInitialized());
+        multiAdapter.setAdapters(centrifugeId, GLOBAL_POOL, adapters, threshold, recoveryIndex);
+        IAuth(address(multiAdapter)).deny(address(this));
+    }
+
+    /// @inheritdoc IBaseGuardian
+    function wire(address adapter, uint16 centrifugeId, bytes memory data) external onlySafe {
+        require(!IAdapterWiring(adapter).isWired(centrifugeId), AdapterAlreadyWired());
+        IAdapterWiring(adapter).wire(centrifugeId, data);
+        IAuth(adapter).deny(address(this));
```
hieronx: This should be removed, since the wiring is per network
```suggestion
```

src/admin/ProtocolGuardian.sol
```diff
@@ -119,17 +115,26 @@ contract Guardian is IGuardian {
...
+    /// @inheritdoc IProtocolGuardian
+    function blockOutgoing(uint16 centrifugeId, bool isBlocked) external onlySafe {
+        gateway.blockOutgoing(centrifugeId, GLOBAL_POOL, isBlocked);
     }
 
-    /// @inheritdoc IGuardian
-    function updateGatewayManager(address who, bool canManage) external onlySafe {
-        gateway.updateManager(PoolId.wrap(0), who, canManage);
+    /// @inheritdoc IBaseGuardian
+    function wire(address adapter, uint16 centrifugeId, bytes memory data) external onlySafe {
```
lemunozm, line 136: NIT, maybe we can accept directly an `IAdapterWiring` here in the interface (?)

script/AdaptersDeployer.s.sol
```diff
@@ -43,18 +43,24 @@ contract AdaptersActionBatcher is CommonActionBatcher {
...
+            // Permanent ward for ongoing adapter maintenance
+            report.axelarAdapter.rely(address(report.common.protocolGuardian));
+            // Temporary ward for initial wire, self-denies after
+            report.axelarAdapter.rely(address(report.common.opsGuardian));
         }
         if (address(report.layerZeroAdapter) != address(0)) {
             report.layerZeroAdapter.rely(address(report.common.root));
-            report.layerZeroAdapter.rely(address(report.common.guardian));
-            report.layerZeroAdapter.rely(address(report.common.adminSafe));
+            // Permanent ward for ongoing adapter maintenance
```
hieronx, line 60: The `adminSafe` should still be relied directly `layerZeroAdapter`, to be able to call `setDelegate`

src/adapters/AxelarAdapter.sol
```diff
@@ -44,13 +46,19 @@ contract AxelarAdapter is Auth, IAxelarAdapter {
     // Administration
```
hieronx: Given it also includes the view method:
```suggestion
    // Network wiring
```

lemunozm (commented):
I really really really like the final result!!

Thanks for all the work on this (with the tedious rebasing part) and leave everything well tests and well documented, as always!

src/adapters/AxelarAdapter.sol
hieronx: (and also in the other adapters)

src/admin/OpsGuardian.sol
lemunozm: ~I think like William here, the `OpsGuardian` should never call again to `wire` here, isn't?~

wischli: Ah yes we cannot remove the ward for the entire adapter then, good catch! Same applies to the `deny` after `initAdapters`. Sorry!

src/admin/ProtocolGuardian.sol
wischli: I prefer the separate signature due to the explicit `address adapter`.

script/CommonDeployer.s.sol
wischli: Good catch, not needed and removed now

src/adapters/AxelarAdapter.sol
wischli: Done in [edaa246](https://github.com/centrifuge/protocol/pull/654/commits/edaa2469aabb5d723c442f95435b604da91b86a7)

script/AdaptersDeployer.s.sol
wischli: Done in [edaa246](https://github.com/centrifuge/protocol/pull/654/commits/edaa2469aabb5d723c442f95435b604da91b86a7)

src/admin/OpsGuardian.sol
wischli: Done in [edaa246](https://github.com/centrifuge/protocol/pull/654/commits/edaa2469aabb5d723c442f95435b604da91b86a7)

lemunozm (dismissed):

script/AdaptersDeployer.s.sol
```diff
@@ -61,6 +61,8 @@ contract AdaptersActionBatcher is CommonActionBatcher {
             report.layerZeroAdapter.rely(address(report.common.protocolGuardian));
             // Temporary ward for initial wire, self-denies after
```
hieronx: `self-denies after` should be removed now

hieronx: Actually comment in general should be rewritten, or maybe just removed.

wischli: [5e738d8](https://github.com/centrifuge/protocol/pull/654/commits/5e738d860b4d9fd3b443c735a156f8918db676bd)

lemunozm (approved):

script/FullDeployer.s.sol
```diff
@@ -128,6 +128,7 @@ contract FullDeployer is ExtendedHubDeployer, ExtendedSpokeDeployer, AdaptersDep
         CommonInput memory commonInput = CommonInput({
             centrifugeId: centrifugeId,
             adminSafe: ISafe(vm.envAddress("ADMIN")),
+            opsSafe: ISafe(vm.envAddress("ADMIN")),
```
hieronx: Need a new env var here:
```suggestion
            opsSafe: ISafe(vm.envAddress("OPS_ADMIN")),
```

script/FullDeployer.s.sol
```diff
@@ -128,6 +128,7 @@ contract FullDeployer is ExtendedHubDeployer, ExtendedSpokeDeployer, AdaptersDep
         CommonInput memory commonInput = CommonInput({
             centrifugeId: centrifugeId,
             adminSafe: ISafe(vm.envAddress("ADMIN")),
```
hieronx:
```suggestion
            adminSafe: ISafe(vm.envAddress("PROTOCOL_ADMIN")),
```

lemunozm (dismissed):
🚀

script/FullDeployer.s.sol
wischli: [e903e7c](https://github.com/centrifuge/protocol/pull/654/commits/e903e7c14488ee0b294f4ae3df247db8998dd6bf)

script/deploy/lib/load_config.py
```diff
@@ -137,20 +145,41 @@ def chain_id(self) -> str:
...
+        if protocol_admin_address := self._check_env_file("PROTOCOL_ADMIN"):
+            return protocol_admin_address
+
+        if self.is_testnet:
+            protocol_admin_address = "0x423420Ae467df6e90291fd0252c0A8a637C1e03f"
+        else:
+            if "protocolAdmin" in self.config["network"]:
+                protocol_admin_address = self.config["network"]["protocolAdmin"]
+            else:
+                protocol_admin_address = self.config["network"]["safeAdmin"]
```
lemunozm, line 161: What is now "safeAdmin"? Will this return null?

lemunozm (approved):
LGTM! Thanks for diving into the python part!

hieronx (approved):

script/deploy/lib/load_config.py
wischli: The one from the env files: https://github.com/centrifuge/protocol/blob/7d314fdd0077471762911e9752ff34a246e633e2/env/ethereum.json#L12

But it's a fair point that after the next deploy, `safeAdmin` will vanish from the env files. Need to think about this, but luckily this codepath is just a fallback when the corresponding admin has not been set.

PR #666, opened 2025-10-02
Add fee hook interface

hieronx:
### Product requirements

* Automated accrual of fees, so this does not need to be executed separately and is rather forced to be done by the manager


onnovisser:
should the fee hook have a view method for the outstanding fee amount? For if something like the NAVManager wants to use the outstanding amount the reduce the NAV

src/core/hub/Hub.sol
```diff
@@ -323,6 +328,8 @@ contract Hub is BatchedMulticall, Auth, Recoverable, IHub, IHubRequestManagerCal
         _isManager(poolId);
 
         shareClassManager.updateSharePrice(poolId, scId, pricePoolPerShare);
+
+        if (address(feeHook) != address(0)) feeHook.accrue(poolId, scId);
```
lemunozm, line 332: Can we take advance that EVM success when a method doesn't exist for some contract and avoid the `if`?

hieronx: It is not about whether the method exist, but whether the fee hook has been set.

hieronx:
> should the fee hook have a view method for the outstanding fee amount? For if something like the NAVManager wants to use the outstanding amount the reduce the NAV

Good idea, will add!

src/core/hub/Hub.sol
lemunozm: Yeah, but the method would not exists for `address(0)`.

Q: I understand this condition exists because we still does't know the implementation, right? But the idea is to always have this implemented along with the Hub

onnovisser (dismissed):

src/core/hub/Hub.sol
hieronx: > But the idea is to always have this implemented along with the Hub

Yes but we won't have it for the first few months.

lemunozm: Just an idea, what about uploading an empty impl, so that later we can change it for the real one?

hieronx: I think having the if statement is cleaner.

src/core/hub/Hub.sol
```diff
@@ -155,6 +158,8 @@ contract Hub is BatchedMulticall, Auth, Recoverable, IHub, IHubRequestManagerCal
         D18 pricePoolPerAsset_ = pricePoolPerAsset(poolId, scId, assetId);
         emit NotifyAssetPrice(assetId.centrifugeId(), poolId, scId, assetId, pricePoolPerAsset_);
         sender.sendNotifyPricePoolPerAsset{value: _payment()}(poolId, scId, assetId, pricePoolPerAsset_, refund);
+
+        if (address(feeHook) != address(0)) feeHook.accrue(poolId, scId);
```
lemunozm, line 162: A bit concerned that this method could not be enough later (i.e. it requires more params), or that we need to call it later from other places in the hub.

Q: What's the reason the integrator can not call later feeHook directly? on top of the Hub?

lemunozm (approved):
LGTM, except a concern, that probably comes because I don't have the whole picture of this feature

PR #667, opened 2025-10-02
ref: Call snapshot hook only from Holdings

onnovisser:
### Product requirements

* 

### Design notes

* 

### TODOs

- [ ]


src/core/hub/Holdings.sol
```diff
@@ -129,6 +129,18 @@ contract Holdings is Auth, IHoldings {
         _callOnSync(poolId, scId, centrifugeId, snapshot_);
     }
 
+    /// @inheritdoc IHoldings
+    function snapshotTransferShares(
```
lemunozm: Not blocker

Should we unify `snapshotTransferShares` and `checkSnapshot` names to indicate that both do more or less the same, but with different callbacks? Right now they are not symmetrical.

I propose: 
- `callOnSyncSnapshot`
- `callOnTransferSnapshot`

What do you think?

lemunozm (dismissed):
Thanks!

src/core/hub/Holdings.sol
hieronx: Agree with this! I think the `call` is implicit, would just go for `onSyncSnapshot` and `onTransferSnapshot`

lemunozm: But with onSyncSnapshot or onTransferSnapshot it seems the methods are called as a callback, when they are actually calling the callback (the other way arround)

hieronx: That's fair!

lemunozm (dismissed):
Thanks Onno for enduring all my tiny renamings and refactorizations regarding this!

hieronx (dismissed):

hieronx (approved):

PR #668, opened 2025-10-02
Ref: decoupling GasService & minor refs

lemunozm:
### Design notes

* Add `IMessageLimits` that IGasService implements.
* Move `IMessageProcessor`, `IMessageDispatcher` and `IGasService` to `src/messaging/interfaces`
* Move `TokenRecoverer` to `src/core`
* Remove maxBatchSizeLimit
  - This is no longer needed because the remote chain is protected by the gas check when the message fails.


snapshots/MessageGasLimits.json
```diff
@@ -1,30 +1,30 @@
...
-  "setRequestManager": 100432,
-  "updateBalanceSheetManager": 103996,
-  "updateHoldingAmount": 304247,
-  "updateShares": 183926,
-  "maxAssetPriceAge": 110037,
-  "maxSharePriceAge": 106923,
-  "updateGatewayManager": 87952
-}
+  "BENCHMARKING_RUN_ID": 1759443455,
+  "scheduleUpgrade": 98862,
```
lemunozm: All values have increased 5K because we now have two different accesses in `handle()`, will fix this... 😓

lemunozm: I don't find a better organization than this (that not imply creating new interfaces with no sense). The thing is that previously, when reading, only `processor` was read to call both `handle()` and `messagePoolId()`, whereas now `poolId` is read from the `messageProperties` contract (previous `gasService`), which adds another read. But conceptually makes much more sense now 🤔, because all message properties are gathered from the same contract/interface from the core

lemunozm: Not super happy, but the latest commits fix this

lemunozm: Need to say that with the most correct solution I talk about, we increase processing by 5K, but we also decrease sending by 2K

hieronx (dismissed):

hieronx (approved):

PR #669, opened 2025-10-02
Idea: Allow nested withBatch calls for composibility

lemunozm:
Just a simple and minor change that allows integrations to call other integrations that already use `withBatch`, which I think is super great for composability and UX. For example, it allows:

- IntegrationA.withBatch()
  - callback
    - IntegrationB.withBatch()
      - callback
    - IntegrationC.withBatch()
      - callback
  
The entire above schema can be batched in just one batch instead of reverting when `integrationB` is called (due to nested batches).

### How does it works?

If `withBatch` is called, and the `Gateway` was already batching, it will just skip `_startBatching` and `_endBatching` to continue batching by the previous `withBatch` or `multicall`.

### DIFF
--- a/src/core/Gateway.sol
+++ b/src/core/Gateway.sol
@@ -260,9 +260,10 @@ contract Gateway is Auth, Recoverable, IGateway {
 
     /// @inheritdoc IGateway
     function withBatch(bytes memory data, address refund) external payable {
-        _startBatching();
-        _batcher = msg.sender;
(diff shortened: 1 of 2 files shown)


lemunozm:
Actually I think we need this, because some hook can do `withBatch` (as we currently do). That means a Pool manager could not call `multicall` if the hook calls later `withBatch`

hieronx (approved):

PR #672, opened 2025-10-03
Feat: add gateway estimate method

lemunozm:
### Product requirements

* UX needs a way to easily compute the gas estimation

### Design notes

* The UI can now do:
  - Simulate a call() and check in the event the created batch.
  - Sum up all extraGasLimit added to the actions
  - call `gateway.estimate(remoteChain, batchFromTheEvent, SumUpExtraGasLimit)` to obtain the exact required value.


wischli (approved):
Really clean solution IMO!

lemunozm:
@hieronx Independenly of how Apps estimate, I think we should provide this, do you agree?

hieronx:
> @hieronx Independenly of how Apps estimate, I think we should provide this, do you agree?

This could be implemented in a separate contract, right? Call it `GatewayLens` (lens is common name for contracts that are purely view helpers for offchain).

I'm wary of adding more to the `Gateway` given it is now assumed to not be migratable.

lemunozm:
🤔 ok, not strong opinion. I did in the same contract because the way it parses the batch should be the same as `handle()` does. But not strong opinion

hieronx:
@lemunozm I would keep core minimal if we don't need to then.

PR #674, opened 2025-10-03
ref: move `Root` and `TokenRecoverer` to `admin` module

hieronx:


src/core/Gateway.sol
```diff
@@ -55,16 +55,16 @@ contract Gateway is Auth, Recoverable, IGateway {
...
     {
         localCentrifugeId = localCentrifugeId_;
-        root = root_;
+        pauser = pauser_;
         messageLimits = messageLimits_;
     }
 
     modifier pauseable() {
-        require(!root.paused(), Paused());
+        require(!pauser.paused(), Paused());
```
lemunozm, line 67: Love it!

src/core/spoke/interfaces/IEndorsements.sol
```diff
@@ -0,0 +1,10 @@
+// SPDX-License-Identifier: GPL-2.0-or-later
+pragma solidity >=0.5.0;
+
+interface IEndorsements {
+    event Endorse(address indexed user);
+    event Veto(address indexed user);
```
lemunozm, line 6: Should not the event be in `IRoot`, given these events are emitted by the methods from Root?

src/core/interfaces/IProtocolPauser.sol
```diff
@@ -0,0 +1,10 @@
+// SPDX-License-Identifier: GPL-2.0-or-later
+pragma solidity >=0.5.0;
+
+interface IProtocolPauser {
+    event Pause();
+    event Unpause();
```
lemunozm, line 6: Same here

lemunozm (commented):
Really like!

src/core/spoke/interfaces/IEndorsements.sol
hieronx: Conceptually I think events should generally be in the most generic interface, since any implementor of the endorsements interface should follow these, such that an indexer can track this.

Of course in practice it does not matter at all..

lemunozm: Ok, that's fair!

lemunozm (approved):
LGTM!

wischli (approved):

PR #675, opened 2025-10-03
Fix max functions in sync manager

hieronx:
### Product requirements

* `max` functions in sync vaults should account for transfer restrictions


src/vaults/SyncManager.sol
```diff
@@ -258,4 +251,11 @@ contract SyncManager is Auth, Recoverable, ISyncManager {
                 rounding
             );
     }
+
+    /// @dev    Checks transfer restrictions for the vault shares. Sender (from) and receiver (to) have to both pass
+    ///         the restrictions for a successful share transfer.
+    function _canTransfer(IBaseVault vault_, address from, address to, uint256 value) internal view returns (bool) {
+        IShareToken share = IShareToken(vault_.share());
+        return share.checkTransferRestriction(from, to, value);
```
lemunozm, line 259: How do you know here that you're checking shares or assets?

lemunozm: Should TO, FROM be different in the maxDeposit case?

hieronx: I don't really understand your question 🤔

lemunozm: In: 
`if (!_canTransfer(vault_, address(0), owner, assets)) return 0;` 

it looks like minting assets, because FROM == address(0) and TO == owner. If both should match to `BaseTransferHook.isDepositRequestOrIssuance`, then nothing to complain, but wanted to double-check

hieronx: Yeah this is a bit imprecise. It is matching AsyncRequestManager behaviour (see maxMint/maxDeposit there). Reason is that you otherwise easily get into overflow errors.

PR #676, opened 2025-10-03
docs: Diagram + README updates

onnovisser:
### Product requirements

* Update the UML diagrams with the recent changes

Core/Hub
image

Core/Spoke
image

Core/Gateway
image

Managers/Hub
image

Managers/Spoke
image

Admin
image

Hooks
image

Vault Request Managers (mostly unchanged)
image

### Design notes

* 

### TODOs

- [ ]


hieronx:
I think these are a great start, but I think they could use a bit of manual finetuning. AI generates all possible paths, and is not always making the best tradeoffs in readability.

E.g. compare
image

to
image

I think with the second one, the connection between the Holdings, `ISnapshotHook` interface, NAVManager, `INAVHook` interface and SimplePriceManager is much clearer.

onnovisser:
Yes, the fine-tuning I had done so far was reducing some of the noise, as AI was including Auth and interface inheritance everywhere.

onnovisser:
Also I was including all the hard dependencies, like SimplePriceManager needs Hub, Gateway, SCM and HubRegistry. But maybe that's also mostly just noise

docs/architecture/overview.puml
```diff
@@ -0,0 +1,133 @@
+@startuml overview
+title Centrifuge Protocol - Architecture Overview
```
wischli: I like this overview but the amount of arrows (50+) is too much. I think the contract dependencies have to be abstracted away here for the majority. Don't have an obvious solution at hand.

docs/architecture/core/spoke.puml
```diff
@@ -0,0 +1,93 @@
+@startuml spoke
+title Spoke
```
wischli, line 2: I'd remove at least ContractUpdater, IUpdateContract and VaultRouter because these don't provide any value. But that will leave the majority of the diagram still as noise.

docs/architecture/core/gateway.puml
```diff
@@ -0,0 +1,64 @@
+@startuml gateway
+title Gateway
```
wischli: The amount of arrows from/to the `Message*` contracts is overwhelming for me. Maybe for `Root` we can add a note that it has superuser power to remove some arrows. For adapters, I would prefer a single arrow from the `adapters` block to the Gatway over each single adapter.

docs/architecture/core/spoke.puml
onnovisser: What about something like this
image

docs/architecture/core/gateway.puml
onnovisser: image
something like this? removed the admin part entirely. the individual adapters as well

hieronx: Looks great! Think `IAdapter` should only be connected to `MultiAdapter` though

docs/architecture/core/spoke.puml
wischli: I really like this but the `IVault` to `ShareToken` relationship doesn't seem correct. IMO, `ShareToken` has a dependency to `IVault` but there is no impl relationship

docs/architecture/core/gateway.puml
wischli: I like it but Im not sure if we should remove the entire admin part as the Guardians are necessary for setting up new and updating the wiring of new networks.

docs/architecture/core/spoke.puml
onnovisser: Yes, I wasn't quite sure about that. I added it because, even though ShareToken doesn't directly reference IVault, it does have a `asset address => vault address` storage.

onnovisser:
Updated the Hub as well
image

docs/architecture/core/spoke.puml
wischli: Maybe change the relation to `ShareToken ..> IVault`?

hieronx: > Yes, I wasn't quite sure about that. I added it because, even though ShareToken doesn't directly reference IVault, it does have a asset address => vault address storage.

This is a relationship from ShareToken to IERC7575Share though, not to IVault from our system.

I would leave this out.

hieronx:
> Updated the Hub as well

Looks really great!

onnovisser:
Added Admin and Hooks

image

image

docs/architecture/core/gateway.puml
```diff
@@ -0,0 +1,31 @@
...
+
+Spoke <-> IMessageDispatcher
+HubHandler <-> IMessageDispatcher
+IMessageProcessor --> HubHandler
+IMessageProcessor --> Spoke
+
+IMessageDispatcher --> Gateway
+
+Gateway --> IMessageProcessor
+Gateway <-> MultiAdapter
```
lemunozm: Nit, can this line be vertical? I think it could improve the readability of this part a lot

```suggestion
Gateway <-down> MultiAdapter
```

lemunozm (commented):
Awesome!! Really like it Onno. You've became a plantuml master after this!

docs/architecture/overview.puml
lemunozm: What if we only show the arrows between modules?

onnovisser: Tried to make it much more minimal, using only interfaces, and only have arrows between modules. But not sure how useful the graph is like this.
image

lemunozm: Yeah, this is a thing I don't like about PlantUML, lines package to package are drawn very bad, so it forces you to link to an inner element inside the package, which means you lose legibility

onnovisser: Will delete this one, now that you have a better one

hieronx (dismissed):

hieronx (approved):

PR #677, opened 2025-10-03
Move messaging into core

hieronx:


src/messaging/libraries/UpdateContractMessageLib.sol
```diff
@@ -1,8 +1,8 @@
 // SPDX-License-Identifier: BUSL-1.1
 pragma solidity 0.8.28;
 
-import {CastLib} from "../../misc/libraries/CastLib.sol";
```
lemunozm, line 4: Could we move `UpdateContractMessageLib` out of the core? For example `src/utils`?

lemunozm (commented):
Thanks for doing this so fast! Just one request

src/messaging/libraries/UpdateContractMessageLib.sol
hieronx: Done @lemunozm

lemunozm (approved):
Thank you so much to be my AI proxy! 🙏🏻

PR #678, opened 2025-10-03
Testnet 3.1 deployments w/ LayerZero

gpmayorga:
This PR:

Deploys layerzero adapter and the most up to date code from main (v3.1)


script/AdaptersDeployer.s.sol
```diff
@@ -119,15 +118,14 @@ contract AdaptersDeployer is CommonDeployer {
...
-            require(adaptersInput.layerZero.delegate != address(0), "LayerZero delegate address cannot be zero");
 
             layerZeroAdapter = LayerZeroAdapter(
                 create3(
                     generateSalt("layerZeroAdapter"),
                     abi.encodePacked(
                         type(LayerZeroAdapter).creationCode,
                         abi.encode(
-                            multiAdapter, adaptersInput.layerZero.endpoint, adaptersInput.layerZero.delegate, batcher
+                            multiAdapter, adaptersInput.layerZero.endpoint, address(adminSafe), batcher
```
lemunozm: - Is our delegate the admin? 

- In case it's, should it be the protocol one or the ops one? (@wischli @hieronx)

hieronx: Should definitely be the protocol one (it has significant powers).

script/WireAdapters.s.sol
```diff
@@ -78,41 +73,76 @@ contract WireAdapters is Script {
...
+
+            try vm.parseJsonBool(remoteConfig, "$.adapters.wormhole.deploy") returns (bool dep) {
+                remoteWormhole = dep;
+            } catch {}
+            try vm.parseJsonBool(remoteConfig, "$.adapters.layerZero.deploy") returns (bool dep) {
+                remoteLayerZero = dep;
+            } catch {}
+            try vm.parseJsonBool(remoteConfig, "$.adapters.axelar.deploy") returns (bool dep) {
+                remoteAxelar = dep;
+            } catch {}
```
lemunozm: Maybe you can extract this into an utility (similar to the already exist on LaunchDeployer.sol)

```
    function _parseJsonBoolOrDefault(string memory config, string memory path) private pure returns (bool) {
        try vm.parseJsonBool(config, path) returns (bool value) {
            return value;
        } catch {
            return false;
        }
    }
``

File: script/WireAdapters.s.sol [hunk 3]
```diff
@@ -17,8 +17,6 @@ import {ILayerZeroAdapter} from "../src/adapters/interfaces/ILayerZeroAdapter.so
 ///
 ///      Intended for testnet use only.
 contract WireAdapters is Script {
-    IAdapter[] adapters; // Storage array like in CommonDeployer
```
lemunozm, line 16: Q: Why this was removed? I think this worked well given it can be increased by `push()`

script/WireAdapters.s.sol
```diff
@@ -78,41 +73,76 @@ contract WireAdapters is Script {
...
+                guardian.setAdapters(remoteCentrifugeId, adaptersForDst, count, threshold);
+                console.log("Registered MultiAdapter(", localNetwork, ") for", remoteNetwork, "with", count, "adapters");
+            } else {
+                console.log("Skipping registration for", remoteNetwork, "(no common adapters)");
+            }
 
-            // Wire WormholeAdapter
-            if (localWormholeAddr != address(0)) {
+            // Wire WormholeAdapter when present on both ends
+            if (localWormholeAddr != address(0) && remoteWormhole) {
```
lemunozm: I think we can assume `localWormholeAddr == address(0)` means that `remoteWormhole = false`, and simplify this

gpmayorga:
Somehow I F*** up when trying to merge main and apply my changes resulting in all shorts of conflicts and ultimately the wrong version of everything that went into main. Starting from scratch again

PR #681, opened 2025-10-04
Core refactor followups

hieronx:
Followup from https://github.com/centrifuge/protocol/pull/679

- Make gateway immutable on higher level contracts
- Add core interface for `IScheduleAuth` and move `ITokenRecoverer` to `src/core/messaging/interfaces`, remove dependencies on non-core from core
- Adapt networks array in `SimplePriceManager` to only restrict which networks are notified
- Add missing `deny` calls for `batcher` on `SimplePriceManager` and `NAVManager`
- Remove `CommonDeployer` artifact
- Replace `vm.assume`s with `bound`s to allow larger fuzz runs
- Remove `Hub.callRequestManager`


wischli (approved):
Thanks for the cleanup, LGTM!

PR #682, opened 2025-10-05
Extend natspec

hieronx:


wischli (dismissed):
Matches what Claude proposed for missing natspec as well. Can make a few minor additions tomorrow if desired.

PR #684, opened 2025-10-05
Extend interface natspec

hieronx:


src/core/spoke/interfaces/IBalanceSheet.sol
```diff
@@ -71,43 +74,122 @@ interface IBalanceSheet is IBatchedMulticall {
...
-    // --- Errors ---
+    //----------------------------------------------------------------------------------------------
+    // Errors
+    //----------------------------------------------------------------------------------------------
+
     error FileUnrecognizedParam();
     error CannotTransferFromEndorsedContract();
 
+    //----------------------------------------------------------------------------------------------
+    // View methods
```
wischli: Nit question: I suppose you decided on View > Admin order just because it lead to a smaller diff here?

wischli (commented):
I have a few diff changes on my branch that I will push here for completion.

src/core/spoke/interfaces/IBalanceSheet.sol
hieronx: Wasn't very intentional, agree ideally admin should be before view.

PR #685, opened 2025-10-05
feat: price updates with `asOfDate` for yield and accounting

mustermeiszer:
### Product requirements

* Yield calculations MUST know as of which date the price update was - e.g. Friday price might be submitted only monday

### Design notes

* No functional changes, just adapts the interface of the method and adds the passed argument to an evnt


wischli:
Closing due to superseding by #686

PR #686, opened 2025-10-05
Pass `computedAt` when updating price

hieronx:
Alternative proposal to https://github.com/centrifuge/protocol/pull/685

### Product requirements

* Yield calculations MUST know as of which date the price update was - e.g. Friday price might be submitted only monday

### Design notes

* Allow passing a custom `computedAt` which is `<= block.timestamp` when setting price per share
   * Thought about restricting it to `lastComputedAt <= computedAt`, but that seems needlessly restrictive for the core
* Use this price as the `computedAt` property to pass to the spoke
* Remove the generic `Metrics` property and separating total issuance and price, as there is no reason these should be combined
* Unrelated refactor: standardize `PoolId` without name and `ShareClassId` without name in storage mappings


mustermeiszer (approved):
Like that a lot!

PR #687, opened 2025-10-06
tests: #686 post-mortem

wischli:
Minor post-mortem of #686 which improves coverage and adds missing natspec.

### DIFF
--- a/src/core/hub/interfaces/IShareClassManager.sol
+++ b/src/core/hub/interfaces/IShareClassManager.sol
@@ -107,6 +107,8 @@ interface IShareClassManager {
     function pricePoolPerShare(PoolId poolId, ShareClassId scId) external view returns (D18 price, uint64 computedAt);
 
     /// @notice Returns the total issuance across all networks for a share class
+    /// @dev     This is only updated when queued shares on the spoke are updated to the hub, so can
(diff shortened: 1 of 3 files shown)


src/core/hub/interfaces/IShareClassManager.sol
```diff
@@ -97,10 +98,19 @@ interface IShareClassManager {
...
+    /// @notice Returns the current price per share and when it was computed
+    ///
+    /// @param poolId Identifier of the pool
+    /// @param scId Identifier of the share class
+    /// @return price The latest price per share (in pool currency denomination)
+    /// @return computedAt Timestamp when the price was computed (may be earlier than submission time)
     function pricePoolPerShare(PoolId poolId, ShareClassId scId) external view returns (D18 price, uint64 computedAt);
 
-    /// @notice TODO
+    /// @notice Returns the total issuance across all networks for a share class
```
hieronx, line 109:
```suggestion
    /// @notice Returns the total issuance across all networks for a share class
    /// @dev     This is only updated when queued shares on the spoke are updated to the hub, so can
    ///                maybe out of sync and not reflect the exact latest issuance across networks.
```

hieronx (approved):

PR #688, opened 2025-10-06
docs: more interface natspec

wischli:
Extends #684 by adding additional natspec with applied `forge fmt` to fix CI


src/managers/hub/interfaces/INAVManager.sol
```diff
@@ -31,6 +33,9 @@ interface INAVHook {
     ) external;
 }
 
+/// @title  INAVManager
+/// @notice Manager for multi-network net asset value (NAV) accounting and price calculations
+/// @dev    Tracks NAV across multiple networks using double-entry accounting with equity, liability, gain, and loss accounts
```
hieronx: Else we should also add asset and expense
```suggestion
/// @dev    Tracks NAV across multiple networks using double-entry accounting accounts
```

PR #690, opened 2025-10-06
fix: decouple refund escrow creation from controller

wischli:
### Product requirements

* Updating the controller of `RefundEscrowFactory` must not create to a new `RefundEscrow`

### Design notes

* Removed controller address parameter from RefundEscrow constructor, making CREATE2 address calculation dependent only on immutable bytecode
* Factory now grants initial ward permissions to itself during deployment, then transfers control to configured controller via rely/deny pattern
* CREATE2 hash now excludes constructor arguments, ensuring escrow addresses remain stable across controller migrations

### DIFF
--- a/src/vaults/factories/RefundEscrowFactory.sol
+++ b/src/vaults/factories/RefundEscrowFactory.sol
@@ -4,6 +4,7 @@ pragma solidity 0.8.28;
 import {IRefundEscrowFactory} from "./interfaces/IRefundEscrowFactory.sol";
 
 import {Auth} from "../../misc/Auth.sol";
+import {IAuth} from "../../misc/interfaces/IAuth.sol";
 
 import {PoolId} from "../../core/types/PoolId.sol";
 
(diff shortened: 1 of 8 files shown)


src/vaults/factories/RefundEscrowFactory.sol
```diff
@@ -23,21 +24,17 @@ contract RefundEscrowFactory is Auth, IRefundEscrowFactory {
 
     /// @inheritdoc IRefundEscrowFactory
     function newEscrow(PoolId poolId) external auth returns (IRefundEscrow escrow) {
-        escrow = new RefundEscrow{salt: bytes32(uint256(poolId.raw()))}(address(controller));
+        escrow = new RefundEscrow{salt: bytes32(uint256(poolId.raw()))}();
+        IAuth(address(escrow)).rely(controller);
```
hieronx, line 31: We need to rely root as well, so we can actually migrate the wards on the escrows

wischli: [87a4305](https://github.com/centrifuge/protocol/pull/690/commits/87a43059613bc327ae49ca65972120e79ccaa624)

lemunozm: good one!

lemunozm (approved):
Perfect fix!

hieronx (approved):

PR #691, opened 2025-10-06
Fix callRequestManager issue

lemunozm:
Fix of: https://kflabs.slack.com/archives/C07PG2EUR9C/p1759615256850889

Proposed solution: https://kflabs.slack.com/archives/C07PG2EUR9C/p1759742743208179?thread_ts=1759694985.378669&cid=C07PG2EUR9C


wischli (dismissed):
Didn't see any obvious issues but leaving this one of for @hieronx

src/core/hub/Hub.sol
```diff
@@ -502,6 +501,16 @@ contract Hub is BatchedMulticall, Auth, Recoverable, IHub, IHubRequestManagerCal
...
     }
 
+    /// @inheritdoc IHub
+    function callRequestManager(PoolId poolId, uint16 centrifugeId, bytes calldata data) external payable {
+        _isManager(poolId);
+
+        // We assume first parameter from data is always a PoolId
+        require(poolId == PoolId.wrap(data.toUint256(4).toUint64()), NotManager());
+
+        hubRegistry.hubRequestManager(poolId, centrifugeId).callFromHub(data);
```
hieronx: Why not just change the ABI of `callFromHub` to
```solidity
    function callFromHub(PoolId poolId, bytes calldata data) external auth {
```

lemunozm: Because we anyway need to pass the poolId inside `data` to perform the correct delegate call, so we're passing that twice.

Also could be a risk of someone encoding `callFromHub` inside data when call to `callRequestManager` but affecting to another PoolId

lemunozm: i.e: `callRequestManager(poolA, callFromHub(PoolA, encodedCallToPoolB))`

lemunozm: Ouch! with the current solution I think it can happen the same. We need to add the check inside the callFromHub too

hieronx: That's fair!

PR #696, opened 2025-10-07
ci: add ward check script

wischli:
### Product requirements

* Implement automated ward coverage validation to prevent authorization failures at runtime by verifying that all auth-protected contract dependencies have proper ward grants and test coverage
* Ensure every `Foo.file("bar")` parameter is initialized either in deployment scripts `foo.file("bar", ..)` or constructor or contract of `Foo` to prevent incomplete deployments
* Provide developers with actionable feedback on missing ward relationships without blocking PR merges

### Design notes

* Added `check_ward_coverage.py` script that performs static analysis of Solidity source files to extract file() parameters, constructor initializations, auth method calls, and expected ward relationships
* Implemented name normalization to handle both camelCase and UPPER_SNAKE_CASE conventions (e.g., `adminSafe` vs `ADMIN_SAFE`) when comparing deployment scripts against test assertions
* GitHub Actions workflow configured as optional check that posts informational PR comments but always passes to avoid blocking merges
* Added three new test functions (`testBatchRequestManager`, updated `testSimplePriceManager`) in `Deployer.t.sol` to cover previously missing ward grant assertions
* Constructor detection automatically identifies parameters set in constructors to eliminate false positives for uninitialized parameters


hieronx (approved):
🚢

lemunozm (approved):
Awesome improvement in our code confidence

PR #697, opened 2025-10-07
Fix gas service values taking into account EIP-150

lemunozm:
### Product requirements

* Gas values benchmarked in `GasService`, are taken into account from a hypothetical adapter calling to `MultiAdapter`. Nevertheless, the gas paid by some executor to the adapter is not taken into account EIP-150, which only forwards 63/64 of the total gas to the adapter. This PR fixes this taken into account the extra required gas

### Design notes

* Updated python script to the new format

### DIFF
--- a/src/core/messaging/GasService.sol
+++ b/src/core/messaging/GasService.sol
@@ -2,6 +2,7 @@
 pragma solidity 0.8.28;
 
 import {IGasService} from "./interfaces/IGasService.sol";
+import {GAS_FAIL_MESSAGE_STORAGE} from "../interfaces/IGateway.sol";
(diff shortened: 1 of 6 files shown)


src/core/messaging/GasService.sol
```diff
@@ -115,4 +117,8 @@ contract GasService is IGasService {
         if (kind == MessageType.UntrustedContractUpdate) return untrustedContractUpdate;
         revert InvalidMessageType(); // Unreachable
     }
+
+    function _gasValue(uint128 value) internal pure returns (uint128) {
```
hieronx, line 125: Would be helpful to add a `///@dev` description here

hieronx (dismissed):
Approving to merge to unblock Guille

hieronx (approved):

PR #699, opened 2025-10-07
feat: Chainlink CCIP Adapter

mustermeiszer:


mustermeiszer:
All kudos to https://github.com/cll-sa for draft that.

src/adapters/CCIPAdapter.sol
```diff
@@ -0,0 +1,122 @@
...
+    }
+
+    //----------------------------------------------------------------------------------------------
+    // Administration
+    //----------------------------------------------------------------------------------------------
+
+    /// @inheritdoc IAdapterWiring
+    function wire(uint16 centrifugeId, bytes memory data) external auth {
+        (uint64 chainSelector, address adapter) = abi.decode(data, (uint64, address));
+        require(ccipRouter.isChainSupported(chainSelector), UnsupportedChain());
```
lemunozm: Should we also add these checks to other adapters?

src/adapters/CCIPAdapter.sol
```diff
@@ -0,0 +1,122 @@
...
+
+        sources[chainSelector] = CCIPSource(centrifugeId, keccak256(abi.encodePacked(adapter)));
```
lemunozm: Why store the adapter as `bytes32` if the data is set as `address`?

src/adapters/CCIPAdapter.sol
```diff
@@ -0,0 +1,122 @@
...
+        CCIPDestination memory destination = destinations[centrifugeId];
+        require(destination.chainSelector != 0, UnknownChainId());
+
+        IClient.EVM2AnyMessage memory ccipMessage = IClient.EVM2AnyMessage({
+            receiver: abi.encode(destination.addr),
+            data: payload,
+            tokenAmounts: new IClient.EVMTokenAmount[](0),
+            feeToken: address(0),
+            extraArgs: _argsToBytes(IClient.EVMExtraArgsV1({gasLimit: gasLimit + RECEIVE_COST}))
+        });
```
lemunozm: NIT. Could we move this to some private function to ensure we use the same in `send` & `estimate`?

mustermeiszer:
closing in favor of https://github.com/centrifuge/protocol/pull/700

hieronx:
@mustermeiszer I merged https://github.com/centrifuge/protocol/pull/700 into this one already

src/adapters/CCIPAdapter.sol
hieronx: Good idea!

src/adapters/CCIPAdapter.sol
hieronx: I just added it because I noticed CCIP has an easy method for it. I think it can't hurt but is not very critical, ultimately an admin error. Would not change the adapters post-audit start now.

src/adapters/CCIPAdapter.sol
hieronx: Adapted

src/adapters/ChainlinkAdapter.sol
```diff
@@ -0,0 +1,130 @@
...
+
+import {IAdapterWiring} from "../admin/interfaces/IAdapterWiring.sol";
+
+/// @title  Chainlink Adapter
+/// @notice Routing contract that integrates with Chainlink Chainlink
+contract ChainlinkAdapter is Auth, IChainlinkAdapter {
+    using CastLib for *;
+
+    /// @dev Cost of executing `ccipReceive()` except entrypoint.handle()
+    uint256 public constant RECEIVE_COST = 4000;
```
wischli, line 29: I suppose this particular amount results from testnet testing?

src/adapters/ChainlinkAdapter.sol
```diff
@@ -0,0 +1,130 @@
...
+
+    function _createMessage(ChainlinkDestination memory destination, bytes calldata payload, uint256 gasLimit)
+        internal
+        pure
+        returns (IClient.EVM2AnyMessage memory)
+    {
+        return IClient.EVM2AnyMessage({
+            receiver: abi.encode(destination.addr),
+            data: payload,
+            tokenAmounts: new IClient.EVMTokenAmount[](0),
```
wischli, line 111: Nit: AFAICT, we could improve gas efficiency by moving this to a contract constant such that we don't need to create a new empty array on every call.

src/adapters/ChainlinkAdapter.sol
```diff
@@ -0,0 +1,130 @@
...
+        ccipRouter = IRouterClient(ccipRouter_);
+    }
+
+    //----------------------------------------------------------------------------------------------
+    // Administration
+    //----------------------------------------------------------------------------------------------
+
+    /// @inheritdoc IAdapterWiring
+    function wire(uint16 centrifugeId, bytes memory data) external auth {
+        (uint64 chainSelector, address adapter) = abi.decode(data, (uint64, address));
```
wischli, line 48: Curious if we should check `ccipRouter.isSourceChain(chainSelector)`.

src/adapters/ChainlinkAdapter.sol
lemunozm: For this contract, just testing, yes

src/adapters/ChainlinkAdapter.sol
hieronx: Had added it but removed it to be consistent with other adapters where we are not checking this either

src/adapters/ChainlinkAdapter.sol
hieronx: Seems like a very minor improvement not worth the reduced readability

test/integration/Deployer.t.sol
```diff
@@ -662,8 +669,10 @@ contract FullDeploymentTestPeripherals is FullDeploymentConfigTest {
 
     function testSimplePriceManager(address nonWard) public view {
         // permissions set correctly
+        vm.assume(nonWard != address(root));
         vm.assume(nonWard != address(navManager));
 
+        assertEq(simplePriceManager.wards(address(root)), 1);
```
lemunozm: Good catch!

test/integration/Deployer.t.sol
```diff
@@ -649,9 +654,11 @@ contract FullDeploymentTestPeripherals is FullDeploymentConfigTest {
 
     function testNavManager(address nonWard) public view {
         // permissions set correctly
+        vm.assume(nonWard != address(root));
         vm.assume(nonWard != address(holdings));
         vm.assume(nonWard != address(hubHandler));
 
+        assertEq(navManager.wards(address(root)), 1);
```
lemunozm: I think we wouldn't need rely on root it NAVManager or SimplePriceManager if instead of using `auth` in `onUpdate` and `OnTransfer`, we use the caller (holding/hubHandler), i.e:  `require(msg.sender == holdings)`. 

That I think would be easier for other integration builders because they don't need to use the Auth infrastructure or rely in its own root for something already constant

lemunozm (commented):
PR looks goods!

Just a non CCIP related comment

test/integration/Deployer.t.sol
wischli: Wondering if we should extend the ward script check for root wards.

lemunozm: I will be super cool to ensure this, due it's the highest critical ward. But not all contracts will have it, for example, simplePriceManager will not after the Onno's change

lemunozm:
We can close this, right?

PR #700, opened 2025-10-07
Update CCIP adapter

hieronx:


src/adapters/interfaces/ICCIPAdapter.sol
```diff
@@ -0,0 +1,128 @@
+// SPDX-License-Identifier: GPL-2.0-or-later
+pragma solidity >=0.5.0;
+
+import {IAdapter} from "../../core/interfaces/IAdapter.sol";
+
+import {IAdapterWiring} from "../../admin/interfaces/IAdapterWiring.sol";
+
+// Tag to indicate only a gas limit. Only usable for EVM as destination chain.
```
hieronx, line 8: We prefer inlining all external code over adding dependencies.

src/adapters/CcipAdapter.sol
```diff
@@ -65,58 +76,44 @@ contract CCIPAdapter is Auth, ICCIPAdapter {
...
+            receiver: abi.encode(destination.addr),
             data: payload,
-            tokenAmounts: new Client.EVMTokenAmount[](0),
+            tokenAmounts: new IClient.EVMTokenAmount[](0),
             feeToken: address(0), // Pay in native token
-            extraArgs: Client._argsToBytes(Client.EVMExtraArgsV1({gasLimit: gasLimit}))
+            extraArgs: _argsToBytes(IClient.EVMExtraArgsV1({gasLimit: gasLimit}))
         });
 
-        // Get exact fee amount needed
```
hieronx, line 86: The system always calls `adapter.estimate` first and then passes exactly this as `msg.value` to `adapter.send`. So there will never be any excess gas directly here.

There will often be excess gas on the destination chain as the gas limit is overestimated (significantly).

PR #702, opened 2025-10-07
Revert on estimate if destination is not setup

hieronx:
### Product requirements

* If no destination is set up, then the `estimate` method of adapters should revert

### DIFF
--- a/test/adapters/unit/WormholeAdapter.t.sol
+++ b/test/adapters/unit/WormholeAdapter.t.sol
@@ -105,7 +105,9 @@ contract WormholeAdapterTest is WormholeAdapterTestBase {
         assertEq(adapter.wards(address(this)), 1);
     }
 
-    function testEstimate(uint64 gasLimit) public view {
+    function testEstimate(uint64 gasLimit) public {
+        adapter.wire(CENTRIFUGE_CHAIN_ID, abi.encode(WORMHOLE_CHAIN_ID, REMOTE_WORMHOLE_ADDR));
+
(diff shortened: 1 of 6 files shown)


wischli (approved):

test/adapters/unit/WormholeAdapter.t.sol
```diff
@@ -105,7 +105,9 @@ contract WormholeAdapterTest is WormholeAdapterTestBase {
         assertEq(adapter.wards(address(this)), 1);
     }
 
-    function testEstimate(uint64 gasLimit) public view {
+    function testEstimate(uint64 gasLimit) public {
```
lemunozm, line 108: NIT, could we add `testErrUnknownChainId` cases?

hieronx: Let's do in followup PR, I think in general the adapter tests can be improved a bit.

lemunozm: I was thinking that maybe we can build some generic adapter test that acts over `IAdapter` and `IAdapterWiring`. So you only need to mock your specific adapter calls to some generic expected behavior and run this generic adapter tests for your new adapter implementation

hieronx: Interesting idea, makes sense to me yes!

hieronx: https://github.com/centrifuge/protocol/issues/703

PR #704, opened 2025-10-07
Tests: Simplify EndToEnd assuming cross-test constants

lemunozm:
Due to the fork-test split, we no longer need methods that are fully parameterized. We can assume for EndToEnd to have some constants across all test and reduce some boilerplate and increase traceability when testing. Assumed constants:
- Assume just 1 hub and 1 spoke, the spoke can be in the same `centrifugeId` if `sameChain == true`
- Spoke always the configured spoke once the mandatory call `setSpoke(bool)` is done (handle by `s`)
- Hub unique (handled by `h`)
- Pool always `POOL_A`
- ShareClassId always `SC_1`
- assetId always `s.usdcId`
- Pool manager is always `FM`
- investor always `INVESTOR_A`

It's true that some weird cases can use two investors or two pools. I think for those, we should create a new test case file and configure it. I think 99% of the cases can be implemented with the above assumptions.

Following PRs:
- Rename spokes in EndToEndThreeChain to adjust the direction of the test (and maybe figure out how we can set this up more clearly, along with EndToEnd, but that's a difficult problem to solve)
- Split tests in EndToEnd for the sameChain param, it is:
  - testCaseA_sameChain()
  - testCaseA_differentChain()
  instead of using the boolean parameter (ease debugging)


test/integration/EndToEnd.t.sol
```diff
@@ -558,322 +505,188 @@ contract EndToEndFlows is EndToEndUtils {
...
+        vm.startPrank(FM);
+        h.hub.updateVault{value: GAS}(
+            POOL_A, SC_1, s.usdcId, s.asyncVaultFactory, VaultUpdateKind.DeployAndLink, EXTRA_GAS, REFUND
+        );
+        vm.stopPrank();
+        IAsyncVault vault = IAsyncVault(address(s.vaultRegistry.vault(POOL_A, SC_1, s.usdcId, s.asyncRequestManager)));
+
+        _executeAsyncDepositRequest(vault, USDC_AMOUNT_1);
+        _processAsyncDepositApproval(USDC_AMOUNT_1);
+        _processAsyncDepositClaim(vault, USDC_AMOUNT_1);
```
lemunozm: These calls (and the homologous in `_testAsyncRedeem` / `_testSyncDeposit`) are only called once from here, wondering if we should unroll them here, to easily follow the steps along with the debugger.

For me it makes sense but want you opinion

wischli: I'd prefer if we could keep them because we use them in fork tests with intermediary steps that are necessary for the fork tests to ensure epochs are aligned:

https://github.com/centrifuge/protocol/pull/620/files#diff-393f2461cb9076cd0e9766da55076364d4e0c27a6fa0d3880f7b5bbd12fdb99cL85-L94

lemunozm: But after the fork-test split, it should no longer be a requirement, right?

wischli: Ah yes you are right. Then no objection from my side.

wischli (dismissed):
Overall I like the simplification. I think we should definitely add e2e for multiple investors in the future, though maybe invariant tests suffice to cover that area.

As a nit, I'd like the comments of L70-L140 to be updated as they reference the old structure and even CV instead of hub 😆

test/integration/EndToEnd.t.sol
lemunozm: Ok, if no one else has strong opinions agains this, this is the diff: [08499eb](https://github.com/centrifuge/protocol/pull/704/commits/08499ebb77eb7dc6504a77516d3f70b1715b0ab4)

wischli (approved):
Thanks for the further cleanup!

PR #707, opened 2025-10-08
fix: `MultiAdapter` uninitialized network view panic

wischli:
### Product requirements

Ref: https://kflabs.slack.com/archives/C07PG2EUR9C/p1759913667204739?thread_ts=1759913523.296029&cid=C07PG2EUR9C

  * Fix bug where MultiAdapter getter functions (`quorum`, `threshold`, `recoveryIndex`, `activeSessionId`) would panic with array index out-of-bounds error when called on uninitialized networks
  * Ensure proper error handling for OpsGuardian.initAdapters() when setting up adapters for the first time on a new network

### Design notes

  * Introduced `_getFirstAdapterDetails()` internal helper that returns zero-initialized Adapter struct when no adapters are configured, preventing array access panics
  * Refactored all four view functions to use the new helper
  * Added `MultiAdapterTestGetters` test suite with coverage for empty state getter behavior


src/core/messaging/MultiAdapter.sol
```diff
@@ -200,28 +200,38 @@ contract MultiAdapter is Auth, IMultiAdapter {
...
+    function _getFirstAdapterDetails(uint16 centrifugeId, PoolId poolId)
+        internal
+        view
+        returns (Adapter memory)
+    {
+        IAdapter[] memory adapters_ = poolAdapters(centrifugeId, poolId);
+        // Fallback for an uninitialized network
+        if (adapters_.length == 0) {
+            return Adapter(0, 0, 0, 0, 0);
+        }
```
hieronx: Nit: I think this comment is already clear from the dev comment above
```suggestion
        if (adapters_.length == 0) return Adapter(0, 0, 0, 0, 0);
```

src/core/messaging/MultiAdapter.sol
```diff
@@ -200,28 +200,38 @@ contract MultiAdapter is Auth, IMultiAdapter {
         }
     }
 
+    /// @dev Internal helper to get the first adapter's details for a pool, handling empty cases
+    function _getFirstAdapterDetails(uint16 centrifugeId, PoolId poolId)
```
hieronx: Nit: would put internal method below the last external view method

wischli: That was also my intuition but we have an internal view above this one already.

lemunozm: ~~Probably we should move both~~

hieronx: Yeah but the internal view method (`_poolAdapterDetails`) is below the external method (`poolAdapters`) that uses it. Here similarly, `_getFirstAdapterDetails` can be below the external methods that use it.

Or we can introduce a new `Internal methods`/`Helpers` section below this

lemunozm (dismissed):
Waiting for fixing Jeroen comments but perfectly clean!

src/core/messaging/MultiAdapter.sol
lemunozm: I prefer leaving internal methods just below the method that calls them. Otherwise it's more difficult to read the contract

wischli: Moved `_getFirstAdapterDetails` to bottom, kept `_poolAdapterDetails` where it was for lower diff: [135841e](https://github.com/centrifuge/protocol/pull/707/commits/135841ea82bf024e8477f16e7c341b0aa38f2ec0)

src/core/messaging/MultiAdapter.sol
wischli: [a4511ec](https://github.com/centrifuge/protocol/pull/707/commits/a4511ec5e84fcfb765af6751541b40030332107a)

hieronx (approved):

lemunozm (approved):

PR #708, opened 2025-10-08
fix: Return NAV as 0 when negative

onnovisser:
### Product requirements

* Return NAV as 0, instead of reverting by underflow. Still assumes accounts aren't misconfigured.

### Design notes

* 

### TODOs

- [ ]


lemunozm (dismissed):

test/managers/hub/unit/NAVManager.t.sol
```diff
@@ -393,16 +393,29 @@ contract NAVManagerNetAssetValueTest is NAVManagerTest {
...
+        assertEq(nav, 0);
+    }
+
+    function testNetAssetValueInvalid(
+        bool equityIsPositive,
+        bool gainIsPositive,
+        bool lossIsPositive,
+        bool liabilityIsPositive
+    ) public {
+        vm.assume(!equityIsPositive || !gainIsPositive || !lossIsPositive || !liabilityIsPositive);
```
wischli: Nit: No need to parameterize here when you restrict each to a single bit. Overall, bool fuzz parameters are questionable because the fuzzer still runs for the configured `fuzz-run` amount independent of whether it already covered the entire parameter space since each run is stateless.

lemunozm: Interesting... 🤔, how would you ensure all combinations except all true will need to be tested?

wischli: Create a helper and trigger it manually. Though I revoke my previous comment because it would just reduce readability. Sorry for the fuzz.

wischli (dismissed):

src/managers/hub/NAVManager.sol
```diff
@@ -208,9 +208,24 @@ contract NAVManager is INAVManager, Auth {
         (bool lossIsPositive, uint128 loss) = accounting.accountValue(poolId, lossAccount(centrifugeId));
         (bool liabilityIsPositive, uint128 liability) = accounting.accountValue(poolId, liabilityAccount(centrifugeId));
 
-        require(equityIsPositive && gainIsPositive && lossIsPositive && liabilityIsPositive, InvalidNAV());
+        uint128 totalPositive = 0;
+        uint128 totalNegative = 0;
 
-        return equity + gain - loss - liability;
+        if (equityIsPositive) totalPositive += equity;
```
hieronx, line 217:
```suggestion
        // Compute NAV = equity + gain - loss - liability
        if (equityIsPositive) totalPositive += equity;
```

hieronx (dismissed):

src/managers/hub/NAVManager.sol
```diff
@@ -208,9 +208,25 @@ contract NAVManager is INAVManager, Auth {
...
+        else totalNegative += equity;
+
+        if (gainIsPositive) totalPositive += gain;
+        else totalNegative += gain;
+
+        if (lossIsPositive) totalNegative += loss;
+        else totalPositive += loss;
+
+        if (liabilityIsPositive) totalNegative += liability;
+        else totalPositive += liability;
```
lemunozm, line 230: NIT. Maybe it's easier to read/audit if we do something like:

```solidity
struct Total {
    uint128 positive;
    uint128 negative;
}

function add(bool isPositive, uint128 value) {
    if (isPositive) positive += value;
    else negative += value;
}

using {add} for Total;

//Later in the method:

total.add(equityIsPositive, equity)
total.add(gainIsPositive, gain)
total.add(!lossIsPositive, loss)
total.add(!liabilityIsPositive, equity)
```

onnovisser: hmm, not sure if this would make it easier to read. then someone has to jump around the code to find what `Total` is and what `add` does. seems like it would lead to more mental overhead

lemunozm: Yeah, no strong opinion. Just the current impl seems a bit difficult to understand, or to make an error in the computations, but as long as we fully test this, we would be good.

wischli: Preferring @lemunozm 's proposal but no strong/blocking opinion.

wischli: I think there should be at least comments explaining why for loss and liability, we increment the negative part when `*isPositive == true` which is the contrary of equity and gain from naming. To me, it logically doesn't make sense to increment a variable suffixed with `negative` if `isPositve == true`.

wischli (dismissed):

hieronx (approved):

wischli (approved):

PR #709, opened 2025-10-08
Add cancelation flow to test data

hieronx:


script/TestData.s.sol
```diff
@@ -270,6 +270,21 @@ contract TestData is LaunchDeployer {
...
+        // Submit cancellation of the deposit request
+        state.vault.cancelDepositRequest(0, msg.sender);
+
+        // Force cancel the deposit request (manager action)
+        batchRequestManager.forceCancelDepositRequest(
+            state.poolId, state.scId, msg.sender.toBytes32(), assetId, msg.sender
+        );
+
+        // Claim the cancelled deposit request
+        state.vault.claimCancelDepositRequest(0, msg.sender, msg.sender);
```
wischli, line 287: Curious if we really want to clean it up or keep it in a pending state. But I suppose you have a clearer picture on the indexer needs.

wischli (approved):

script/TestData.s.sol
hieronx: Claiming is a separate event that the indexer also needs to track.

PR #713, opened 2025-10-10
Fix: Hub diagram

onnovisser:


hieronx (approved):

docs/architecture/core/hub.puml
```diff
@@ -23,19 +19,19 @@ class Accounting
...
 Hub --> Accounting
 Hub --> IMessageDispatcher
-Hub --> Gateway
 Hub .[dotted].> IFeeHook
 
 HubHandler --> Hub
 IMesssageProcessor --> HubHandler
 
-IBatchRequestManager --> Hub
+IHubRequestManager --> Hub
```
lemunozm, line 34: @hieronx I think this should be `IHubRequestManager <-- Hub` instead. Or bidirectional, assuming all implementations of IHubRequestManager also point to the hub.

PR #714, opened 2025-10-10
Add gas limit and gas paid to `HandlePayload` events

hieronx:
### Product requirements

* Enable offchain indexing of gas paid per adapter over time

### Design notes

* Adds an internal method to solve stack too deep issue


src/core/messaging/MultiAdapter.sol
```diff
@@ -154,14 +154,25 @@ contract MultiAdapter is Auth, IMultiAdapter {
...
+        uint16 centrifugeId,
+        bytes32 payloadId,
+        bytes calldata payload,
+        IAdapter adapter,
+        uint256 gasLimit,
+        address refund
+    ) internal {
+        uint256 cost = adapter.estimate(centrifugeId, payload, gasLimit);
+        bytes32 adapterData = adapter.send{value: cost}(centrifugeId, payload, gasLimit, refund);
+        emit SendPayload(centrifugeId, payloadId, payload, adapter, adapterData, refund, gasLimit, cost);
```
lemunozm: NIT. `refund` is always the latest param in all our API, I would keep that "soft requirement" here in the event too.

lemunozm (dismissed):
Q. Could not we get both `gasLimit` and `cost` from the adapter event?

wischli (dismissed):

src/core/messaging/MultiAdapter.sol
hieronx: Makes sense.

Re
> Q. Could not we get both gasLimit and cost from the adapter event?

There is no event in our adapters. We'd need to track each external adapter individually, that is much harder.

foundry.toml
```diff
@@ -17,6 +17,17 @@ fs_permissions = [
   { access = "read-write", path = "snapshots" },
 ]
 
+# {@see {@link https://github.com/foundry-rs/foundry/issues/4060} }
```
hieronx, line 20: I added a few flags that should make verification easier

lemunozm (approved):
Thanks for accepting my NIT!

PR #716, opened 2025-10-13
Fix: Gateway batch processing for same poolId

lemunozm:
### Product requirements

* A malicious adapter can send malformed batches with messages from different pools, forcing executions of other pool-related messages

### Design notes

* Check all messages of a batch belong to the same pool


hieronx:
@lemunozm can you check this Slack message? https://kflabs.slack.com/archives/C07PG2EUR9C/p1760099888713709

src/core/messaging/Gateway.sol
```diff
@@ -96,6 +96,7 @@ contract Gateway is Auth, Recoverable, IGateway {
 
     /// @inheritdoc IMessageHandler
     function handle(uint16 centrifugeId, bytes memory batch) public pauseable auth {
+        PoolId batchPoolId = processor.messagePoolId(batch);
```
lemunozm: This poolId is already validates by MultiAdapter which ensures that it came by the correct adapters

src/core/messaging/Gateway.sol
```diff
@@ -104,6 +105,9 @@ contract Gateway is Auth, Recoverable, IGateway {
             remaining = remaining.slice(length, remaining.length - length);
             bytes32 messageHash = keccak256(message);
 
+            PoolId messagePoolId = processor.messagePoolId(message);
+            require(batchPoolId == messagePoolId, MalformedBatch());
+
```
hieronx: Nit:
```suggestion
            require(batchPoolId == processor.messagePoolId(message), MalformedBatch());
```

lemunozm: Agree

src/core/messaging/interfaces/IMessageHandler.sol
```diff
@@ -1,21 +1,28 @@
 // SPDX-License-Identifier: GPL-2.0-or-later
 pragma solidity >=0.5.0;
 
+import {PoolId} from "../../types/PoolId.sol";
+
 /// @notice Generic interface for entities that handle incoming messages
```
wischli, line 6: Nit: To distinguish from `IPoolMessageHandler`, could extend natspec like

```suggestion
/// @notice Generic interface for entities that handle incoming messages
/// @dev Use this interface for components that process individual messages
///      where poolId can be extracted from the message itself (e.g., MessageProcessor, retry)
```

src/core/messaging/interfaces/IMessageHandler.sol
```diff
@@ -1,21 +1,28 @@
...
 
-    /// @notice Dispatched when an invalid message is trying to handle
-    error InvalidMessage(uint8 code);
+    /// @notice Handling incoming messages
+    /// @param centrifugeId Source chain
+    /// @param message Incoming message
+    function handle(uint16 centrifugeId, bytes calldata message) external;
+}
 
+/// @notice Generic interface for entities that handle incoming messages
```
wischli, line 18: Nit addition to top for easier difference to `IMessageHandler`:

```suggestion
/// @notice Generic interface for entities that handle incoming messages
/// @dev Use this interface for components that process batches and require
///      validated poolId context. The poolId parameter represents the validated
///      pool that all messages in the batch must belong to.
```

lemunozm:
Close in favor of: https://github.com/centrifuge/protocol/pull/718

PR #718, opened 2025-10-13
Fix: Gateway batch processing for same poolId (solution 1)

lemunozm:
### Product requirements

* A malicious adapter can send malformed batches with messages from different pools, forcing executions of other pool-related messages

### Design notes

* Check that all messages of a batch belong to the same pool
* Ensure we have 1 read of poolId per message to ensure gas values when batching remains constant


src/core/messaging/Gateway.sol
```diff
@@ -96,11 +96,18 @@ contract Gateway is Auth, Recoverable, IGateway {
...
     /// @inheritdoc IMessageHandler
     function handle(uint16 centrifugeId, bytes memory batch) public pauseable auth {
+        PoolId batchPoolId = processor.messagePoolId(batch);
         bytes memory remaining = batch;
 
         while (remaining.length > 0) {
             uint256 length = processor.messageLength(remaining);
             bytes memory message = remaining.slice(0, length);
+
+            if (remaining.length != batch.length) {
```
hieronx, line 106: Is this really necessary? The if statement itself adds gas, and the check is only a single equality check.

lemunozm: It saves 2000 gas for fetching the poolId through `processor`

I think we should kee it because it adds 2000 to each benchmarked message, so a batch of 20 messages will imply an extra cost of `2000 * 20`, when this is not really true for batches.

lemunozm: But we can assume that to be refunded in some cases, just an extra overhead upfront

hieronx: How is it 2000 gas for each message? I would think it's only 2000 for the first one since it's the first external call?

hieronx: That's a strong argument for just doing `message.toUint64(0)`.

lemunozm: > I would think it's only 2000 for the first one since it's the first external call?

Yeah, I also thought it was like you say

> That's a strong argument for just doing message.toUint64(0).

We can not do that, because depending on each message, the poolId could be there or not.

lemunozm: We could link again `MessageLib` to the `Gateway` (and MultiAdapter) to avoid those calls, but it seems risky in case tomorrow we need to perform some modification there for some reason

hieronx (approved):

PR #721, opened 2025-10-13
ref: rm payable approveRedeems

wischli:
### Product requirements

* Fix https://github.com/electisec/centrifuge-v31-report/issues/1
* Remove unnecessary payable modifier from approveRedeems function to prevent accidental ETH transfers during redemption approval operations 

### Design notes

* Removed `payable` modifier from `approveRedeems` function in BatchRequestManager.sol and its interface
* Function does not handle native ETH and should not accept value transfers
* No breaking changes to function signature or parameters, only removes incorrect modifier


hieronx (approved):

lemunozm (commented):
As a note, if we ending up adding multicall here, we would need to re-add this

PR #726, opened 2025-10-14
refactor: QueueManager gas savings

onnovisser:
Save a little bit of gas by removing a variable and a few expressions


src/managers/spoke/QueueManager.sol
```diff
@@ -71,27 +71,22 @@ contract QueueManager is Auth, IQueueManager, ITrustedContractUpdate {
         gateway.lockCallback();
 
         ShareClassQueueState storage sc = scQueueState[poolId][scId];
-        require(
-            sc.lastSync == 0 || sc.minDelay == 0 || block.timestamp >= sc.lastSync + sc.minDelay, MinDelayNotElapsed()
-        );
+        require(sc.lastSync == 0 || block.timestamp >= sc.lastSync + sc.minDelay, MinDelayNotElapsed());
```
lemunozm, line 66: IIUC, we don't need to check `minDelay` because it will automatically pass by `block.timestamp >= sc.lastSync`, right?

Should not also remove `sc.lastSync == 0` because `block.timestamp >= 0` will also passes?

src/managers/spoke/QueueManager.sol
```diff
@@ -71,27 +71,22 @@ contract QueueManager is Auth, IQueueManager, ITrustedContractUpdate {
...
 
-        (uint128 delta,, uint32 queuedAssetCounter,) = balanceSheet.queuedShares(poolId, scId);
-        uint256 validCount = 0;
         for (uint256 i = 0; i < assetIds.length; i++) {
-            bytes32 key = keccak256(abi.encode(scId.raw(), assetIds[i].raw()));
+            bytes32 key = keccak256(abi.encode(assetIds[i].raw()));
```
lemunozm: I remember vaguely another security issue related to this, that we say that thanks to using scId (which contains the poolId), the current impl was not an issue (maybe for some reentrancy stuff?).

Could you point me to that issue? Are we sure we're not opening that issue here?

lemunozm: I think a hook dispatched under submitQueue in a local chain could call again sync() with a different pool and mix the keys.

src/managers/spoke/QueueManager.sol
wischli: Assuming the delay won't be incredibly large, I suppose we can go with
```suggestion
        require(block.timestamp >= sc.lastSync + sc.minDelay, MinDelayNotElapsed());
```

Because for `sc.lastSync == 0` this just checks if the current block timestamp is larger than `minDelay` which we can assume to always be positive.

src/managers/spoke/QueueManager.sol
wischli: I'd also prefer keeping the `scId` key prefix. Otherwise, if within the same transaction multiple syncs happen for the same `assetId` but different `scId`, there will be collision and some assets are skipped when they should not.

hieronx (dismissed):

hieronx (approved):

PR #728, opened 2025-10-14
Fix: revert in closeGainLoss when accounts are negative

onnovisser:


hieronx (approved):

wischli (approved):

src/managers/hub/NAVManager.sol
```diff
@@ -173,8 +173,11 @@ contract NAVManager is INAVManager, Auth {
         AccountId gainAccount_ = gainAccount(centrifugeId);
         AccountId lossAccount_ = lossAccount(centrifugeId);
 
-        (, uint128 gainValue) = accounting.accountValue(poolId, gainAccount_);
-        (, uint128 lossValue) = accounting.accountValue(poolId, lossAccount_);
+        (bool gainIsPositive, uint128 gainValue) = accounting.accountValue(poolId, gainAccount_);
+        (bool lossIsPositive, uint128 lossValue) = accounting.accountValue(poolId, lossAccount_);
+
+        // Gain and loss accounts should never be negative
+        require(gainIsPositive && lossIsPositive, InvalidStateOfAccounts());
```
lemunozm, line 180: @onnovisser I'm late here, but can we have a negative 0? Should not be considered as successul here?

onnovisser: No, 0 is also considered positive

PR #729, opened 2025-10-14
Use `withBatch` callback in `BatchedMulticall`, add this to `BRM` and `VaultRouter`

hieronx:


src/vaults/VaultRouter.sol
```diff
@@ -29,39 +31,70 @@ import {IVaultRegistry} from "../core/spoke/interfaces/IVaultRegistry.sol";
...
+            (bool success, bytes memory returnData) = address(this).delegatecall(data[i]);
+            if (!success) {
+                uint256 length = returnData.length;
+                require(length != 0, CallFailedWithEmptyRevert());
+
+                assembly ("memory-safe") {
+                    revert(add(32, returnData), length)
+                }
+            }
+        }
```
lemunozm: I think you can inherit from `Multicall` instead of ReentrancyProtection and call `super.multicall(data)` here. Also easier for integrators

lemunozm (dismissed):
And you said I'm fast! 🏇🏻 

Despite the lines added, I really happy with the final result. It's not as hard as I thought. Do you think we can also implement this in the BRM? would be the best approach then 🤔

src/vaults/VaultRouter.sol
hieronx: Problem is that then we have to use a new interface for the main `VaultRouter.multicall` function. Which is not ideal.

src/vaults/VaultRouter.sol
```diff
@@ -55,15 +55,16 @@ contract VaultRouter is Auth, Multicall, Recoverable, IVaultRouter {
     }
 
     function multicall(bytes[] calldata data) public payable override protected {
```
lemunozm: I think you don’t need to use protected in VaultRouter.multicall, because super.multicall will use protected as sender . I think this is the correct fix where sender still comes from `gateway.lockCallback()`

src/misc/ReentrancyProtection.sol
```diff
@@ -8,20 +8,20 @@ pragma solidity ^0.8.28;
...
 
-    address internal _initiator;
+    address private transient _initiator;
 
     /// @dev The method is protected for reentrancy issues.
     modifier protected() {
         if (_initiator == address(0)) {
             // Single call re-entrancy lock
-            _initiator = msgSender();
+            _initiator = msg.sender;
```
lemunozm, line 19: I don't understand why this was causing the issue, could you expand?

src/vaults/VaultRouter.sol
```diff
@@ -29,39 +29,50 @@ import {IVaultRegistry} from "../core/spoke/interfaces/IVaultRegistry.sol";
...
+        gateway.withBatch{value: msg.value}(
+            abi.encodeWithSelector(VaultRouter.executeMulticall.selector, data), msg.sender
+        );
+    }
+
+    function executeMulticall(bytes[] calldata data) external payable {
+        gateway.lockCallback();
+        super.multicall(data);
+    }
+
```
lemunozm: I think we can implement `BatchedMulticall` as these two methods now, removing the boilerplate

src/vaults/VaultRouter.sol
hieronx: Actually we need `protected` here to set `_initiator` which will be used as `msgSender()`

src/misc/ReentrancyProtection.sol
```diff
@@ -21,8 +21,12 @@ abstract contract ReentrancyProtection {
             _initiator = address(0);
         } else {
             // Multicall re-entrancy lock
-            require(msg.sender == _initiator, UnauthorizedSender());
+            require(msgSender() == _initiator, UnauthorizedSender());
```
onnovisser: how is this checking anything? the `else` happens when `_initiator != address(0)` and then `msgSender()` returns `_initiator`, so this always evaluates to `true`, no?

src/vaults/VaultRouter.sol
hieronx: In `super.multicall`, the sender is gateway.

src/misc/ReentrancyProtection.sol
hieronx: True this doesn't work 🤔

src/vaults/VaultRouter.sol
lemunozm: ~But that's what we want, right? If not other methods will fail because they expect the gateway~

Well, if not, they will fail because of the wrong owner 🤔

src/misc/ReentrancyProtection.sol
lemunozm: Good catch Onno!

src/misc/ReentrancyProtection.sol
```diff
@@ -8,9 +8,10 @@ pragma solidity ^0.8.28;
 /// @notice Abstract contract that implements reentrancy protection using transient storage.
 abstract contract ReentrancyProtection {
     /// @notice Dispatched when there is a re-entrancy issue
-    error UnauthorizedSender();
+    error UnauthorizedSender(address expected, address actual);
 
     address private transient _initiator;
+    address internal transient _multicallSource;
```
lemunozm: I'm not fan of touching something from `misc` to support a custom behavior 🤔 

IIUC, the problem is that the sender for the methods must be `lockCallback()` (or the firs msg.sender), but the one for reentrancy must be the gateway, right?

src/core/hub/Hub.sol
```diff
@@ -498,7 +498,7 @@ contract Hub is BatchedMulticall, Auth, Recoverable, IHub, IHubRequestManagerCal
     ) external payable {
         IHubRequestManager manager = hubRegistry.hubRequestManager(poolId, assetId.centrifugeId());
         require(address(manager) != address(0), InvalidRequestManager());
-        require(msg.sender == address(manager), NotAuthorized());
+        require(msgSender() == address(manager), NotAuthorized());
```
lemunozm: I think this must not be changed, because it must be the manager

src/core/utils/BatchedMulticall.sol
```diff
@@ -13,28 +13,35 @@ import {IGateway} from "../messaging/interfaces/IGateway.sol";
...
+    }
 
-        gateway.endBatching{value: msg.value}(msg.sender);
-        _isBatching = false;
+    function msgSender() internal view virtual returns (address) {
+        return _sender != address(0) ? _sender : msg.sender;
     }
 
     /// @dev gives the current msg.value depending on the batching state
     function _payment() internal view returns (uint256 value) {
```
lemunozm: NIT. I think we could call this now `msgValue()` (or with `_` preffix). That way we have a pattern to use for the implementors: `msgSender()` and `msgValue()`, pretty symmetric and easy to remember

hieronx: Good idea!

src/core/utils/BatchedMulticall.sol
```diff
@@ -13,28 +13,34 @@ import {IGateway} from "../messaging/interfaces/IGateway.sol";
...
 
     constructor(IGateway gateway_) {
         gateway = gateway_;
     }
 
     /// @inheritdoc IMulticall
     /// @notice With extra support for batching
-    function multicall(bytes[] calldata data) public payable override protected {
+    function multicall(bytes[] calldata data) public payable override {
         require(!gateway.isBatching(), IGateway.AlreadyBatching());
```
lemunozm: I think this check is no longer needed.

hieronx: True!

src/core/utils/BatchedMulticall.sol
```diff
@@ -11,30 +11,39 @@ import {IGateway} from "../messaging/interfaces/IGateway.sol";
...
+    }
 
+    function executeMulticall(bytes[] calldata data) external payable protected {
+        gateway.lockCallback();
         super.multicall(data);
+    }
 
-        gateway.endBatching{value: msg.value}(msg.sender);
-        _isBatching = false;
+    /// @dev Integrators MUST use msgSender() instead of msg.sender, since this is replaced
```
wischli, line 38: Thanks for the explicit comment!

src/core/utils/BatchedMulticall.sol
```diff
@@ -11,30 +11,39 @@ import {IGateway} from "../messaging/interfaces/IGateway.sol";
...
+    ///      by the gateway address inside the multicall.
+    function msgSender() internal view virtual returns (address) {
+        return _sender != address(0) ? _sender : msg.sender;
     }
 
-    /// @dev gives the current msg.value depending on the batching state
-    function _payment() internal view returns (uint256 value) {
-        return _isBatching ? 0 : msg.value;
+    /// @dev Only the call to multicall should pass the msg.value, which is then passed
+    ///      in `gateway.withBatch`. All inner calls should not pass any msg.value.
```
wischli: Ultra ultra nit
```suggestion
    ///      in `gateway.withBatch`. No inner calls should pass any msg.value.
```

src/vaults/BatchRequestManager.sol
```diff
@@ -245,7 +249,7 @@ contract BatchRequestManager is Auth, ReentrancyProtection, IBatchRequestManager
         uint32 nowRedeemEpochId,
         uint128 approvedShareAmount,
         D18 pricePoolPerAsset
-    ) external authOrManager(poolId) {
+    ) external payable authOrManager(poolId) {
```
wischli, line 252: Wondering why this change doesn't require adding a gas payment to `approveRedeems` calls in EndToEnd.t.sol.

wischli (approved):
As a nit, I would have preferred the README update in a separate PR. Code LGTM but I can imagine e2e tests to fail due to the missing gas payment addition for `approveRedeems`.

src/vaults/BatchRequestManager.sol
hieronx: You can call a payable method without passing msg.value. Only reason for payable here is to allow multicall with other payable methods.

lemunozm: To extend, multicall uses a `delegatecall`, which means the same `msg.value` is passed through all method in the multicall. It means, if you want to use `approveRedeems` along with other methods, you need to support passing some `msg.value` (this value is not lost, just viewed from the method)

lemunozm (approved):
Everything in around 100 changed lines, sick!

onnovisser (approved):

lemunozm (approved):

PR #731, opened 2025-10-14
Use withBatch callback in VaultRouter (no reentrancy changes)

lemunozm:
based on https://github.com/centrifuge/protocol/pull/729

### DIFF
--- a/src/vaults/VaultRouter.sol
+++ b/src/vaults/VaultRouter.sol
@@ -40,6 +40,8 @@ contract VaultRouter is Auth, Multicall, Recoverable, IVaultRouter {
     IGateway public immutable gateway;
     IVaultRegistry public immutable vaultRegistry;
 
+    address internal transient _sender;
+
     /// @inheritdoc IVaultRouter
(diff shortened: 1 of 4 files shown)


src/vaults/VaultRouter.sol
```diff
@@ -52,13 +54,15 @@ contract VaultRouter is Auth, Multicall, Recoverable, IVaultRouter {
         vaultRegistry = vaultRegistry_;
     }
 
-    function multicall(bytes[] calldata data) public payable override protected {
+    function multicall(bytes[] calldata data) public payable override {
+        _sender = msg.sender;
```
wischli, line 58: Wouldn't it be possible for a malicious actor to re-enter `VaultRouter.multicall` in `VaultRouter.executeMulticall` which would then overwrite the `_sender`?

hieronx: `gateway.lockCallback` should block this

lemunozm:
Added to the original PR

PR #732, opened 2025-10-14
Feat: add gas multiplier to the adapters

lemunozm:
### Product requirements

* We need some extra gas proportional to the gasLimit to avoid price fluctiation issues between source and target chains

### Design notes

* Added a multiplier by centrifuge Id when wiring

### DIFF
--- a/script/WireAdapters.s.sol
+++ b/script/WireAdapters.s.sol
@@ -83,10 +83,11 @@ contract WireAdapters is Script {
             // Wire WormholeAdapter
             if (localWormholeAddr != address(0)) {
                 bytes memory wormholeData = abi.encode(
+                    GAS_MULTIPLIER,
                     uint16(vm.parseJsonUint(remoteConfig, "$.adapters.wormhole.wormholeId")),
                     vm.parseJsonAddress(remoteConfig, "$.contracts.wormholeAdapter")
                 );
-                opsGuardian.wire(localWormholeAddr, remoteCentrifugeId, GAS_MULTIPLIER, wormholeData);
+                opsGuardian.wire(localWormholeAddr, remoteCentrifugeId, wormholeData);
 
                 console.log("Wired WormholeAdapter from", localNetwork, "to", remoteNetwork);
             }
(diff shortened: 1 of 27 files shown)


src/admin/OpsGuardian.sol
```diff
@@ -60,9 +60,12 @@ contract OpsGuardian is IOpsGuardian {
...
 
     /// @inheritdoc IBaseGuardian
-    function wire(address adapter, uint16 centrifugeId, bytes memory data) external onlySafe {
+    function wire(address adapter, uint16 centrifugeId, uint8 gasBufferPercentage, bytes memory data)
+        external
+        onlySafe
+    {
         require(!IAdapterWiring(adapter).isWired(centrifugeId), AdapterAlreadyWired());
-        IAdapterWiring(adapter).wire(centrifugeId, data);
+        IAdapterWiring(adapter).wire(centrifugeId, gasBufferPercentage, data);
```
hieronx: Wormhole doesn't need it. Why not just include it only in the `data`?

lemunozm: Correct! 😅

script/WireAdapters.s.sol
```diff
@@ -82,6 +83,7 @@ contract WireAdapters is Script {
             // Wire WormholeAdapter
             if (localWormholeAddr != address(0)) {
                 bytes memory wormholeData = abi.encode(
+                    GAS_MULTIPLIER,
```
lemunozm, line 86: I think for testnets, a constant is enough. we don't need to configure this from the json

src/adapters/AxelarAdapter.sol
```diff
@@ -114,7 +117,11 @@ contract AxelarAdapter is Auth, IAxelarAdapter {
         require(bytes(destination.axelarId).length != 0, UnknownChainId());
 
         return axelarGasService.estimateGasFee(
-            destination.axelarId, destination.addr, payload, gasLimit + RECEIVE_COST, bytes("")
+            destination.axelarId,
+            destination.addr,
+            payload,
+            (gasLimit + RECEIVE_COST) * destination.gasBufferPercentage,
```
wischli: In order to make this percentage based, would have to divide by 100. Then, you should use `MathLib.mulDiv though. I think this is still better than using a `D18` gasBufferPercentage which comes with larger encoding.


```suggestion
            Math.mulDiv(gasLimit + RECEIVE_COST, 100 + destination.gasBufferPercentage, 100),
```

src/adapters/LayerZeroAdapter.sol
```diff
@@ -113,8 +115,9 @@ contract LayerZeroAdapter is Auth, ILayerZeroAdapter {
         LayerZeroDestination memory destination = destinations[centrifugeId];
         require(destination.layerZeroEid != 0, UnknownChainId());
 
-        MessagingReceipt memory receipt =
-            endpoint.send{value: msg.value}(_params(destination, payload, gasLimit + RECEIVE_COST), refund);
+        MessagingReceipt memory receipt = endpoint.send{value: msg.value}(
+            _params(destination, payload, (gasLimit + RECEIVE_COST) * destination.gasBufferPercentage), refund
```
wischli: Same as above

wischli (commented):
Apart from the factor != percentage fix, LGTM! Would love to have a short comment in code to explain why Wormhole does not need to utilize the factor.

src/adapters/AxelarAdapter.sol
lemunozm: Because we don't need precision at all here, could we just do: 
```
(gasLimit + RECEIVE_COST) * (100 + destination.gasBufferPercentage) / 100
```

`destination.gasBufferPercentag` is assumed to be a small number about 3 digits as much, so no overflow issues neither

hieronx (approved):

PR #733, opened 2025-10-14
Polish lockCallback behavior

lemunozm:
A very minor PR that polishes the lockCallback behavior.

Previously, it was returning the `batcher`, when it's not needed, because the batcher **is always** `address(this)` from the caller perspective. 

It created confusion, giving the false sense that it was the `msg.sender` to the method that later calls `withBatch`

### DIFF
--- a/src/core/messaging/Gateway.sol
+++ b/src/core/messaging/Gateway.sol
@@ -288,10 +288,9 @@ contract Gateway is Auth, Recoverable, IGateway {
     }
 
     /// @inheritdoc IGateway
-    function lockCallback() external returns (address caller) {
+    function lockCallback() external {
         require(_batcher != address(0), CallbackIsLocked());
         require(msg.sender == _batcher, CallbackWasNotFromSender());
-        caller = _batcher;
         _batcher = address(0);
     }
 
--- a/src/core/messaging/interfaces/IGateway.sol
+++ b/src/core/messaging/interfaces/IGateway.sol
@@ -165,8 +165,8 @@ interface IGateway is IMessageHandler, IRecoverable {
     ///             }
     ///
     ///             function callback(PoolId poolId) external {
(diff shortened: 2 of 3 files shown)


src/core/messaging/interfaces/IGateway.sol
```diff
@@ -165,8 +165,8 @@ interface IGateway is IMessageHandler, IRecoverable {
     ///             }
     ///
     ///             function callback(PoolId poolId) external {
-    ///                 // Avoid direct reentrancy to the callback and ensure it's called from withBatch in the same contract:
-    ///                 address msgSender = gateway.lockCallback();
+    ///                 // Avoid reentrancy to the callback and ensure it's called from withBatch in the same contract:
+    ///                 gateway.lockCallback();
```
lemunozm, line 169: This line was wrong, it's not `msgSender`, it's `address(this)`, which is not needed from this perspective, and adds confusion

src/core/messaging/Gateway.sol
```diff
@@ -288,10 +288,9 @@ contract Gateway is Auth, Recoverable, IGateway {
     }
 
     /// @inheritdoc IGateway
-    function lockCallback() external returns (address caller) {
+    function lockCallback() external {
         require(_batcher != address(0), CallbackIsLocked());
         require(msg.sender == _batcher, CallbackWasNotFromSender());
```
lemunozm, line 293: This check is which forces the previously returned address to always be `address(this)` from the caller's perspective. Then we do not need to return it

src/core/messaging/interfaces/IGateway.sol
hieronx: Completely agree this was confusing!

hieronx (approved):

wischli (approved):
Thanks for the simplification!

PR #734, opened 2025-10-14
Remove redundant zero address checks

hieronx:
* Fix https://github.com/electisec/centrifuge-v31-report/issues/11

### DIFF
--- a/src/core/spoke/Spoke.sol
+++ b/src/core/spoke/Spoke.sol
@@ -328,7 +328,6 @@ contract Spoke is Auth, Recoverable, ReentrancyProtection, ISpoke, ISpokeGateway
         bool unpaid
     ) external payable {
         IRequestManager manager = requestManager[poolId];
-        require(address(manager) != address(0), InvalidRequestManager());
         require(msg.sender == address(manager), NotAuthorized());
 
         gateway.setUnpaidMode(unpaid);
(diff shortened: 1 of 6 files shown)


test/core/spoke/unit/Spoke.t.sol
```diff
@@ -441,7 +441,7 @@ contract SpokeTestRegisterAsset is SpokeTest {
 contract SpokeTestRequest is SpokeTest {
     function testErrInvalidRequestManager() public {
         vm.prank(AUTH);
-        vm.expectRevert(ISpoke.InvalidRequestManager.selector);
+        vm.expectRevert(IAuth.NotAuthorized.selector);
```
lemunozm, line 444: I think this case is already tested by `testErrNotAuthorized`, and it can be removed

src/core/spoke/Spoke.sol
```diff
@@ -328,7 +328,6 @@ contract Spoke is Auth, Recoverable, ReentrancyProtection, ISpoke, ISpokeGateway
         bool unpaid
     ) external payable {
         IRequestManager manager = requestManager[poolId];
-        require(address(manager) != address(0), InvalidRequestManager());
```
lemunozm, line 331: Not sure about this finding and change. This check helps the user to know they have not configure the manager in a previous stage. If they obtain a `NotAuthorized` error, it looks like they did something wrong in the current call, when the error came previously by a lack of configuration. I think, to get a good UX, we need to keep it.

hieronx: That's a fair point!

wischli: That's also what I had in mind.

PR #737, opened 2025-10-15
BRM: use msgSender and msgValue & no auth

lemunozm:
Some additions over the multicall refactor alrrady merged:

- [x] Adapt to work with `BatchedMulticall`
- [x] Remove `authOrManager` and let only `manager` in `BRM`
- [x] Remove `authOrManager` and let only `manager` in `BalanceSheet`
- [x] `_sender` private


test/vaults/unit/BatchRequestManager.t.sol
```diff
@@ -363,9 +368,11 @@ abstract contract BatchRequestManagerBaseTest is Test {
         approvedAmountUsdc = uint128(bound(approvedAmountUsdc_, MIN_REQUEST_AMOUNT_USDC - 1, depositAmountUsdc));
         approvedPool = _intoPoolAmount(USDC, approvedAmountUsdc);
         batchRequestManager.requestDeposit(poolId, scId, depositAmountUsdc, investor, USDC);
+        vm.startPrank(MANAGER);
         batchRequestManager.approveDeposits{
             value: COST
         }(poolId, scId, USDC, _nowDeposit(USDC), approvedAmountUsdc, _pricePoolPerAsset(USDC), REFUND);
+        vm.stopPrank();
```
lemunozm, line 375: I hate the current solution, but I didn't find a better & faster way to do this without doing a big refactor.

The thing is that `_nowDeposit()` performs an external call, so it consumes anything you add in the prank.

I've also tried to do something like:
```
let now_ = _nowDeposit(USDC);
vm.prank(MANAGER);
approveDeposits(now);
```
But it adds a lot of variables and I reach stack too deep in several places.

test/vaults/unit/BatchRequestManager.t.sol
```diff
@@ -2687,115 +2831,6 @@ contract BatchRequestManagerRoundingEdgeCasesDeposit is BatchRequestManagerBaseT
...
-        hubRegistryMock.updateManager(poolId, poolManager, true);
-    }
-
-    function testApproveDepositsAsManager() public {
-        batchRequestManager.requestDeposit(poolId, scId, MIN_REQUEST_AMOUNT_USDC, investor, USDC);
-
-        vm.prank(poolManager);
-        batchRequestManager.approveDeposits{
-            value: COST
-        }(poolId, scId, USDC, _nowDeposit(USDC), MIN_REQUEST_AMOUNT_USDC, _pricePoolPerAsset(USDC), REFUND);
```
lemunozm, line 2708: As a note. These tests were not working before by the reason above. Now removed because they overlap with the rest of the tests

lemunozm (commented):
I hope @wischli these changes don't affect to much to the invariant work and can be adapted with a simple change

test/vaults/unit/BatchRequestManager.t.sol
wischli: Yeah I think this is the least invasive solution. Suppose for invariant tests it will be fine.

hieronx (approved):

wischli (approved):

hieronx (approved):

PR #740, opened 2025-10-15
Add explicit error for reserve underflow in pool escrow

hieronx:


src/core/spoke/PoolEscrow.sol
```diff
@@ -32,6 +32,8 @@ contract PoolEscrow is Escrow, Recoverable, IPoolEscrow {
     /// @inheritdoc IPoolEscrow
     function withdraw(ShareClassId scId, address asset, uint256 tokenId, uint128 value) external auth {
         Holding storage holding_ = holding[scId][asset][tokenId];
+        require(holding_.total >= holding_.reserved, InsufficientBalance(asset, tokenId, value, 0));
```
lemunozm, line 35: Super NIT. I think technically the error is `InsufficientReserve`, but something we can live on

lemunozm (approved):

src/core/spoke/PoolEscrow.sol
hieronx: Don't fully agree, the issue is that the balance is decreased by any reserve amount. So you have an insufficient balance to withdraw, from an integrator perspective. 

I would reuse the same error because if you want to withdraw 200, total is 300 and reserved is 200, you wouldn't have `total < reserved`, but effectively you still have an insufficient balance because of the reserved amount.

lemunozm (approved):

wischli (commented):
Invariant tests will love this

PR #741, opened 2025-10-15
Gateway: extend withBatch and some cleans

lemunozm:
### Product requirements

* Integrators using `withBatch` could want to use eth transfers inside the callback. Right now, that's unsupported, but we should let scale layers on top.

### Design notes

* Two flavors of withBatch, with and without `msg.value`
* Remove unused start/endBatching
* Remove unused check in startBatching


src/core/messaging/Gateway.sol
```diff
@@ -226,22 +226,11 @@ contract Gateway is Auth, Recoverable, IGateway {
         emit RepayBatch(centrifugeId, batch);
     }
 
-    /// @inheritdoc IGateway
-    function startBatching() public auth {
-        _startBatching();
-    }
-
     function _startBatching() internal {
-        require(!isBatching, AlreadyBatching());
```
lemunozm, line 235: The check is no longer needed, maybe we can also remove the internal method

hieronx: Was about to say, seems like we can merge all this into `withBatch`

hieronx (approved):
LGTM

src/core/messaging/Gateway.sol
```diff
@@ -227,21 +227,30 @@ contract Gateway is Auth, Recoverable, IGateway {
     }
 
     /// @inheritdoc IGateway
-    function startBatching() public auth {
-        _startBatching();
-    }
+    function withBatch(bytes memory data, uint256 value, address refund) public payable {
+        require(value <= msg.value, NotEnoughGas());
```
onnovisser: shouldn't this revert with `NotEnoughValueForCallback` instead?

lemunozm: Correct! Good catch

lemunozm: And also be checked in tests

onnovisser (approved):

PR #744, opened 2025-10-15
Remove gas buffer percentage in adapters

hieronx:


lemunozm (approved):
Super fast 🚀 , some Jeroen's clone should have been done this while I was discussing with you in slack 😆

onnovisser (approved):

src/core/messaging/GasService.sol
```diff
@@ -13,9 +13,9 @@ import {MessageLib, MessageType, VaultUpdateKind} from "./libraries/MessageLib.s
 contract GasService is IGasService {
     using MessageLib for *;
 
-    /// @dev Takes into account diverge computation from the base menchmarked value.
-    /// Also takens into account
-    uint128 public constant BASE_COST = 20_000;
+    /// @dev Takes into account diverge computation from the base benchmarked value.
+    uint128 public constant BASE_COST = 50_000;
+    uint16 public constant GAS_MULTIPLIER = 10; // 10%
```
lemunozm: Should this be used in `_gasValue()` method?

hieronx: I think we decided for now to remove it, pending feedback from sepolia testing if it needs to be added

PR #746, opened 2025-10-16
fix: ShareClassManager and SimplePriceManager - transfer shares underflow

onnovisser:
### Product requirements

* SCM and SimplePriceManager can underflow if shares are transferred before submitQueuedShares is called

### Design notes

* 

### TODOs

- [ ]


src/managers/hub/SimplePriceManager.sol
```diff
@@ -88,21 +88,34 @@ contract SimplePriceManager is ISimplePriceManager {
...
         Metrics storage metrics_ = metrics[poolId];
-        uint128 issuance = shareClassManager.issuance(poolId, scId, centrifugeId);
-
-        metrics_.issuance = metrics_.issuance + issuance - networkMetrics_.issuance;
-        metrics_.netAssetValue = metrics_.netAssetValue + netAssetValue - networkMetrics_.netAssetValue;
+        uint128 newIssuance = shareClassManager.issuance(poolId, scId, centrifugeId);
+
+        // When shares are transferred, the issuance in the SCM updates immediately,
+        // but in this contract they are tracked separately as transferredIn/Out.
+        // Here we get the diff between the old SPM issuance and the new SCM issuance,
```
wischli: Nit
```suggestion
        // Here we get the diff between the old SCM issuance and the new SCM issuance,
```

test/managers/hub/unit/SimplePriceManager.t.sol
```diff
@@ -379,4 +388,96 @@ contract SimplePriceManagerOnTransferTest is SimplePriceManagerTest {
...
+        (, globalIssuance) = priceManager.metrics(POOL_A);
+        (, fromIssuance, fromTransferredIn, fromTransferredOut,,) = priceManager.networkMetrics(POOL_A, CENTRIFUGE_ID_1);
+        (, toIssuance, toTransferredIn, toTransferredOut,,) = priceManager.networkMetrics(POOL_A, CENTRIFUGE_ID_2);
+
+        assertEq(globalIssuance, 300);
+        assertEq(fromIssuance, 50);
+        assertEq(toIssuance, 200);
+        assertEq(fromTransferredOut, 0);
+        assertEq(fromTransferredIn, 0);
+        assertEq(toTransferredIn, 50);
```
wischli, line 514: Q: IIUC, this is not 0 because we haven't called onUpdate for network 2 yet right?

wischli (commented):
Makes sense to me and seems feasible! Thanks for the fix. Overall, I'd love to see some assertions/tests in e2e in the future, definitely not this PR.

test/managers/hub/unit/SimplePriceManager.t.sol
onnovisser: yes exactly

src/core/hub/ShareClassManager.sol
```diff
@@ -21,7 +21,8 @@ contract ShareClassManager is Auth, IShareClassManager {
     mapping(PoolId => mapping(ShareClassId => uint128)) public totalIssuance;
     mapping(PoolId => mapping(ShareClassId => Price)) public pricePoolPerShare;
     mapping(PoolId => mapping(ShareClassId => ShareClassMetadata)) public metadata;
-    mapping(PoolId => mapping(ShareClassId => mapping(uint16 centrifugeId => uint128))) public issuance;
+    mapping(PoolId => mapping(ShareClassId => mapping(uint16 centrifugeId => IssuancePerNetwork))) internal
```
hieronx: Maybe we should make this public. Otherwise there is no way to know the individual sum of issuances and revocations, if it is reverting.

src/core/hub/ShareClassManager.sol
```diff
@@ -79,13 +80,16 @@ contract ShareClassManager is Auth, IShareClassManager {
...
-        uint128 newTotalIssuance =
-            isIssuance ? totalIssuance[poolId][scId_] + amount : totalIssuance[poolId][scId_] - amount;
-        totalIssuance[poolId][scId_] = newTotalIssuance;
-        uint128 newIssuancePerNetwork = isIssuance ? issuance[poolId][scId_][centrifugeId] + amount : issuance[poolId][scId_][centrifugeId] - amount; // forgefmt: disable-line
-        issuance[poolId][scId_][centrifugeId] = newIssuancePerNetwork;
+        IssuancePerNetwork storage ipn = _issuancePerNetwork[poolId][scId_][centrifugeId];
+
+        if (isIssuance) {
+            ipn.issuances += amount;
+            totalIssuance[poolId][scId_] += amount;
```
hieronx, line 88: I think we need to swap the order of 
```
shareClassManager.updateShares(originCentrifugeId, poolId, scId, amount, false);
shareClassManager.updateShares(targetCentrifugeId, poolId, scId, amount, true);
```

in `HubHandler`. Since right now, it is decreasing from the total issuance first, and then increasing. Which means you can have an underflow if the transfer is larger than the total submitted shares across networks.

onnovisser: good point

mustermeiszer (commented):
Good fix

src/managers/hub/SimplePriceManager.sol
hieronx: This is intentional (Single Price Manager = SPM)

src/managers/hub/SimplePriceManager.sol
```diff
@@ -88,21 +88,34 @@ contract SimplePriceManager is ISimplePriceManager {
...
+        (uint128 issuanceDelta, bool isIncrease) = _calculateIssuanceDelta(
+            networkMetrics_.issuance, newIssuance, networkMetrics_.transferredIn, networkMetrics_.transferredOut
+        );
 
-        D18 pricePoolPerShare_ = pricePoolPerShare(poolId);
+        if (isIncrease) {
+            metrics_.issuance = metrics_.issuance + issuanceDelta;
+        } else {
+            metrics_.issuance = metrics_.issuance - issuanceDelta;
+        }
```
hieronx:
```suggestion
        metrics_.issuance = isIncrease) ? metrics_.issuance + issuanceDelta : metrics_.issuance - issuanceDelta;
```

src/managers/hub/SimplePriceManager.sol
onnovisser: actually clarified that it's the `current stale SPM issuance`

src/managers/hub/SimplePriceManager.sol
```diff
@@ -140,4 +153,25 @@ contract SimplePriceManager is ISimplePriceManager {
...
+        // which is the same as (newIssuance + transferredOut) - (oldIssuance + transferredIn)
+        // and avoids potential underflow
+        uint128 adjustedNew = newIssuance + transferredOut;
+        uint128 adjustedOld = oldIssuance + transferredIn;
+
+        if (adjustedNew >= adjustedOld) {
+            return (adjustedNew - adjustedOld, true);
+        } else {
+            return (adjustedOld - adjustedNew, false);
+        }
```
hieronx:
```suggestion
        if (adjustedNew >= adjustedOld) return (adjustedNew - adjustedOld, true);
        else return (adjustedOld - adjustedNew, false);
```

hieronx (dismissed):

hieronx (approved):

PR #757, opened 2025-10-20
ci: Anvil deployments with multi-chain (wiring adapters)

gpmayorga:
Anvil will now detect errors in WireAdapters and deploy to a fork of both sepolia and arbitrum-sepolia


lemunozm:
Is this still valid @gpmayorga? It seems already incorporated in one of your PRs from `protocol-internal`

gpmayorga:
Yes I think this is already in protocol-intermal

PR #758, opened 2025-10-20
Refund always if WithBatch

lemunozm:
https://kflabs.slack.com/archives/C04HF8SHRRC/p1760963899305569?thread_ts=1760962717.727929&cid=C04HF8SHRRC

### DIFF
--- a/src/core/messaging/Gateway.sol
+++ b/src/core/messaging/Gateway.sol
@@ -227,14 +227,14 @@ contract Gateway is Auth, Recoverable, IGateway {
     }
 
     /// @inheritdoc IGateway
-    function withBatch(bytes memory data, uint256 value, address refund) public payable {
-        require(value <= msg.value, NotEnoughValueForCallback());
+    function withBatch(bytes memory data, uint256 callbackValue, address refund) public payable {
+        require(callbackValue <= msg.value, NotEnoughValueForCallback());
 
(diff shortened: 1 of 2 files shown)


src/core/messaging/Gateway.sol
```diff
@@ -247,16 +247,18 @@ contract Gateway is Auth, Recoverable, IGateway {
         // Force the user to call lockCallback()
         require(address(_batcher) == address(0), CallbackWasNotLocked());
 
-        if (!wasBatching) _endBatching(msg.value - value, refund);
+        uint256 cost;
+        if (!wasBatching) cost = _endBatching(msg.value - value, refund);
+
+        _refund(refund, msg.value - value - cost);
```
hieronx: This is really hard to follow. I would rename `wasBatching` to `isNested` and, then do:
```suggestion
        if (isNested) {
            _refund(refund, msg.value - value);
        } else {
            uint256 cost = _endBatching(msg.value - value, refund);
            _refund(refund, msg.value - value - cost);
        }
```

wischli (commented):
Logic looks correct to me but agreeing with @hieronx because I first thought there was something off and had to draw a diagram

hieronx (approved):

wischli (approved):

PR #765, opened 2026-02-04
Migration v3.1 cleaning

lemunozm:
### Product requirements

* Remove main fails/content related to v3.1 migration
* Remove the possibility of deploying using an existing `Root`. This is important to remove, because it was core specific for the migration v3.1. If we do some error deploying using `ROOT` envvar, we will deploy unaccessible contracts, forcing us to redeploy with a different create3 for that address. 

From this, I have some ideas to generalize the validation system to support any kind of spell/migration in a generic way (WIP). By now, we can remove these no longer used files.


onnovisser (approved):

script/spell/MigrationQueries.sol
```diff
@@ -4,14 +4,16 @@ pragma solidity 0.8.28;
 import {PoolId} from "../../src/core/types/PoolId.sol";
 import {AssetId} from "../../src/core/types/AssetId.sol";
 
-import {Root} from "../../src/admin/Root.sol";
```
wischli, line 7: What's the reason for keeping this file around despite removing the other ones?

wischli (approved):

script/spell/MigrationQueries.sol
lemunozm: Good question! I thought it could be interesting to still have some mechanism to fetch from the indexer, but it's true the current one is tied to the migration. Would be really cool to tell claude to have a whole graphQL parser of the current state 🤔 

By now probably we can probably remove, yes. I'll do in a follow-up PR!

lemunozm: Now remember the why, MigrationQueries is somehow injected into the validators base logic. I'm refactoring that in another PR where I could finally remove `MigrationQueries`

PR #768, opened 2026-02-05
Add EnvConfig parser

lemunozm:
### Product requirements

* Add EnvConfig parser as a source of truth for validating env/*.json files. 
* Benefits:
  * Allows to always parse the env files with same expectations and do proper checks.
  * Allows to check that all env files are consitent with our requirements (see `EnvFiles.t.sol`).
  * Allows a much easier integration of env files with our codebase.
    * Right now integrated to `ValidateContractsFromFactories.s.sol` script but the long term of this is to integrate also in `LaunchDeployer`, `OnlyAdapters`, spells, validators, and fork tests (follow up PRs)


script/utils/EnvConfig.s.sol
```diff
@@ -0,0 +1,362 @@
...
+    function isMainnet(NetworkConfig memory config) internal pure returns (bool) {
+        return keccak256(bytes(config.environment)) == keccak256("mainnet");
+    }
+
+    function graphQLApi(NetworkConfig memory config) internal pure returns (string memory) {
+        return config.isMainnet() ? GraphQLConstants.PRODUCTION_API : GraphQLConstants.TESTNET_API;
+    }
+
+    function _contains(string memory str, string memory substr) private pure returns (bool) {
+        return bytes(str).length >= bytes(substr).length && keccak256(bytes(str)) != keccak256(bytes(substr))
```
wischli: Can you explain to me why `keccak256(bytes(str)) != keccak256(bytes(substr)` is necessary here? If `str == substr` (e.g., `_contains("plume", "plume")`), this returns false.

script/utils/EnvConfig.s.sol
```diff
@@ -0,0 +1,362 @@
...
+            uint16 centrifugeId = remoteConfig.network.centrifugeId;
+            require(centrifugeId <= 31, "centrifugeId value higher than 31");
+
+            batchLimits[centrifugeId] = remoteConfig.network.batchLimit;
+        }
+    }
+
+    function rpcUrl(NetworkConfig memory config) internal view returns (string memory) {
+        string memory apiKey = _envString("ALCHEMY_API_KEY");
+        if (_contains(config.baseRpcUrl, "plume")) apiKey = _envString("PLUME_API_KEY");
```
wischli: AFAICT, `_envString` reverts if the env var is missing. So if you only have `PLUME_API_KEY` set (not `ALCHEMY_API_KEY`), calling `rpcUrl()` for Plume will revert on the first line before reaching the if. The old RPCComposer used an if/else pattern that didn't have this problem

script/utils/EnvConfig.s.sol
```diff
@@ -0,0 +1,362 @@
...
+        try vm.parseJsonAddress(json, string.concat(".contracts.", key, ".address")) returns (address addr) {
+            return addr;
+        } catch {
+            return address(0);
+        }
+    }
+}
+
+library EnvConfigLib {
+    function etherscanApiKey(EnvConfig memory) internal view returns (string memory) {
```
wischli, line 298: The parameter is redundant here right?
```suggestion
    function etherscanApiKey() internal view returns (string memory) {
```

script/utils/EnvConfig.s.sol
```diff
@@ -0,0 +1,362 @@
...
+        config.contracts = _parseContractsConfig(json);
+    }
+
+    function _parseNetworkConfig(string memory json) private pure returns (NetworkConfig memory config) {
+        config.chainId = vm.parseJsonUint(json, ".network.chainId");
+        config.environment = vm.parseJsonString(json, ".network.environment");
+        config.centrifugeId = uint16(vm.parseJsonUint(json, ".network.centrifugeId"));
+        config.protocolAdmin = vm.parseJsonAddress(json, ".network.protocolAdmin");
+        config.opsAdmin = vm.parseJsonAddress(json, ".network.opsAdmin");
+        config.baseRpcUrl = vm.parseJsonString(json, ".network.baseRpcUrl");
```
wischli, line 141: Nit: Maybe we want to check for unsafe truncations here as this script is supposed to function as a validation layer?

wischli (dismissed):
Great improvement! Have a few non-blocking questions though.

script/utils/EnvConfig.s.sol
lemunozm: Good catch! I think it was some remnant, I think actually we can call `indexOf`. Fixed: [358db7e](https://github.com/centrifuge/protocol/pull/768/commits/358db7e578612ab64987be9cc55a649aac3d2ad5)

script/utils/EnvConfig.s.sol
lemunozm: You're correct again! Thanks for your deep review. I was assuming all envvars are always accesible but could not be the same in some scenarios, so better have if/else pattern here. Fixed: [358db7e](https://github.com/centrifuge/protocol/pull/768/commits/358db7e578612ab64987be9cc55a649aac3d2ad5)

script/utils/EnvConfig.s.sol
lemunozm: Now yes, but it couldn't for future chains that don't use our `ETHERSCAN_API_KEY`, that's why I left the API waiting for that scenario. Actually, I think `hyper-evm-testnet` doesn't support that, but I didn't have a script to test this yet correctly. Also

script/utils/EnvConfig.s.sol
lemunozm: Yeah, I was thinking about that first, but later I realized that some unsafe truncation is as wrong as configuring the wrong value for the expected network. We don't have a way to validate correct values here because all of them are foreign, and only one value is correct

wischli (approved):

onnovisser (approved):
💯

PR #769, opened 2026-02-05
Deploy pool-specific hooks

onnovisser:
### Product requirements

* Deploy pool-specific hooks for mainnet pools that have their PoolEscrow set, to prevent falling into the try/catch

Mostly for review. Don't think we want to merge this

### Design notes

* 

### TODOs

- [ ]

### DIFF
--- a/script/PoolHooks.s.sol
+++ b/script/PoolHooks.s.sol
@@ -1,0 +1,236 @@
+// SPDX-License-Identifier: BUSL-1.1
+pragma solidity 0.8.28;
+
+import {PoolId} from "../src/core/types/PoolId.sol";
+import {ShareClassId} from "../src/core/types/ShareClassId.sol";
+
+import {Root} from "../src/admin/Root.sol";
+import {Spoke} from "../src/core/spoke/Spoke.sol";
+import {BalanceSheet} from "../src/core/spoke/BalanceSheet.sol";
+import {IShareToken} from "../src/core/spoke/interfaces/IShareToken.sol";
+import {IPoolEscrow} from "../src/core/spoke/interfaces/IPoolEscrow.sol";
+import {IPoolEscrowProvider} from "../src/core/spoke/factories/interfaces/IPoolEscrowFactory.sol";
+
+import {FreelyTransferable} from "../src/hooks/FreelyTransferable.sol";
+import {FullRestrictions} from "../src/hooks/FullRestrictions.sol";
+
+import {stdJson} from "forge-std/StdJson.sol";
+import {console} from "forge-std/console.sol";
+
+import {JsonRegistry} from "./utils/JsonRegistry.s.sol";
+import {GraphQLQuery} from "./utils/GraphQLQuery.s.sol";
+import {GraphQLConstants} from "./utils/GraphQLConstants.sol";
+import {CreateXScript} from "./utils/CreateXScript.sol";
+import {MessageDispatcher} from "../src/core/messaging/MessageDispatcher.sol";
+import {makeSalt} from "./CoreDeployer.s.sol";
+
+struct TokenInstanceData {
+    PoolId poolId;
+    ShareClassId scId;
+    address tokenAddress;
+}
+
+/// @title PoolHooks
+/// @notice Script to deploy pool-specific hooks with pool escrow addresses
+contract PoolHooks is JsonRegistry, GraphQLQuery, CreateXScript {
+    using stdJson for string;
+
+    bytes32 constant VERSION = "1";
+
+    uint16 public centrifugeId;
+
+    address public freelyTransferableHook;
+    address public fullRestrictionsHook;
+    Root public root;
+    Spoke public spoke;
+    BalanceSheet public balanceSheet;
+    IPoolEscrowProvider public poolEscrowFactory;
+
+    address public deployer;
+
+    function _graphQLApi() internal pure override returns (string memory) {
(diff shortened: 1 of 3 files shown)


script/PoolHooks.s.sol
```diff
@@ -0,0 +1,233 @@
...
+    ShareClassId scId;
+    address tokenAddress;
+}
+
+/// @title PoolHooks
+/// @notice Script to deploy pool-specific hooks with pool escrow addresses
+contract PoolHooks is JsonRegistry, GraphQLQuery, CreateXScript {
+    using stdJson for string;
+
+    bytes32 constant VERSION = "1";
```
wischli: I think this should be "v3.1" as configured in `env/*.json`

script/PoolHooks.s.sol
```diff
@@ -0,0 +1,233 @@
...
+        console.log("===");
+        console.log("Processing pool %d with hook %s", PoolId.unwrap(token.poolId), currentHook);
+        console.log("Pool escrow: %s", poolEscrowAddr);
+
+        if (currentHook == freelyTransferableHook) {
+            _deployFreelyTransferable(token.poolId, token.scId, poolEscrowAddr);
+        } else if (currentHook == fullRestrictionsHook) {
+            _deployFullRestrictions(token.poolId, token.scId, poolEscrowAddr);
+        } else {
+            console.log("Hook is not freelyTransferable or fullRestrictions, skipping");
```
wischli, line 148: I suppose none of the pools use `FreezeOnly` or `RedemptionRestrictions` hooks?

script/PoolHooks.s.sol
```diff
@@ -0,0 +1,233 @@
...
+    }
+
+    function _deployFullRestrictions(PoolId poolId, ShareClassId scId, address poolEscrow)
+        internal
+        returns (FullRestrictions hook)
+    {
+        string memory saltName = string.concat("fullRestrictions-", vm.toString(PoolId.unwrap(poolId)));
+        bytes32 salt = makeSalt(saltName, VERSION, deployer);
+
+        hook = FullRestrictions(
```
wischli: Wondering if we should make the script idempotent by adding a check whether it has been deployed already.

wischli (commented):
LGTM except for the potential incorrect `VERSION`. Script also worked for me locally, great job!

script/PoolHooks.s.sol
onnovisser: Exactly

script/PoolHooks.s.sol
onnovisser: Yes maybe a good idea. We may need to run the script again before we have the factory in place.

hieronx: That would be very useful yes!

script/test-pool-hooks-fork.sh
```diff
@@ -0,0 +1,63 @@
...
+
+anvil --fork-url "$REMOTE_RPC_URL" &
+ANVIL_PID=$!
+trap "kill $ANVIL_PID" EXIT
+
+LOCAL_RPC_URL="http://127.0.0.1:8545" #anvil
+
+sleep 3.0 # Wait ensuring Anvil is up
+
+SENDER="0xc1A929CBc122Ddb8794287D05Bf890E41f23c8cb"
```
lemunozm, line 44: As a note. This sender is the admin from testnet.

script/PoolHooks.s.sol
```diff
@@ -0,0 +1,233 @@
...
+
+    function _deployFreelyTransferable(PoolId poolId, ShareClassId scId, address poolEscrow)
+        internal
+        returns (FreelyTransferable hook)
+    {
+        string memory saltName = string.concat("freelyTransferable-", vm.toString(PoolId.unwrap(poolId)));
+        bytes32 salt = makeSalt(saltName, VERSION, deployer);
+
+        hook = FreelyTransferable(
+            create3(
```
lemunozm, line 168: I really think we don't need `create3` if we deploy with a custom deployer. We're not gaining anything, and can be confused when in other chains they have different addresses for the same pools.

I would go only with `create3` if we can ensure that we always have the same addresses in all chains.

The only way to have all hooks with the same addresses for the same pools is using a factory IIUC.

lemunozm (commented):
Except my two comments, the whole process looks really nice!

script/test-pool-hooks-fork.sh
onnovisser: ah yes I just picked a random address for testing

script/PoolHooks.s.sol
onnovisser: Right now, if you use the same account to deploy on all networks, the hooks for the same pool will all have the same address. If we need to deploy more for the same pool again in the future, we could use the same account again to get the same addresses AFAIK.

lemunozm: The problem is that future hooks will be deployed by pool managers, not us. This is only a first time initialization by us

onnovisser: True, though I'm not sure it's that important that between now and then, the addresses are the same. Just that now they are the same for the same pool seems convenient. 
Maybe @hieronx has thoughts on this?

lemunozm:
@onnovisser I think we need to merge this, because we can need this script again until we reach a solid solution, right?

lemunozm (approved):

PR #771, opened 2026-02-06
feat: TokenBridge

onnovisser:
### Product requirements

* TokenBridge compatible with Glacis Airlift 
https://www.notion.so/LiFi-bridging-contract-2b62eac24e178072a5a3ed17ebe770e3?source=copy_link

### Design notes

* Takes custody of user's share tokens before calling Spoke.crosschainTransferShares
* Remaining msg.value optionally gets sent to a relayer, which can be set by the ProtocolGuardian
* ProtocolGuardian has to file chain ID -> cent ID mapping, because Airlift only submits a destination chain ID
* extraGasLimit and remoteExtraGasLimit configurable by FM

### TODOs

- [ ] Maybe change cent ID file method to allow setting multiple IDs at once

### DIFF
--- a/script/FullDeployer.s.sol
+++ b/script/FullDeployer.s.sol
@@ -160,10 +160,7 @@ contract FullActionBatcher is CoreActionBatcher {
 
         report.batchRequestManager.rely(address(report.root));
 
-        if (address(report.layerZeroAdapter) != address(0)) report.layerZeroAdapter.rely(address(report.root));
-        if (address(report.wormholeAdapter) != address(0)) report.wormholeAdapter.rely(address(report.root));
-        if (address(report.axelarAdapter) != address(0)) report.axelarAdapter.rely(address(report.root));
-        if (address(report.chainlinkAdapter) != address(0)) report.chainlinkAdapter.rely(address(report.root));
+        _relyAdapters(report, address(report.root));
 
         // Rely spoke
         report.asyncRequestManager.rely(address(report.core.spoke));
@@ -188,25 +185,13 @@ contract FullActionBatcher is CoreActionBatcher {
         report.tokenRecoverer.rely(address(report.protocolGuardian));
         report.tokenBridge.rely(address(report.protocolGuardian));
         // Permanent ward for ongoing adapter maintenance
-        if (address(report.layerZeroAdapter) != address(0)) {
-            report.layerZeroAdapter.rely(address(report.protocolGuardian));
-        }
-        if (address(report.wormholeAdapter) != address(0)) {
-            report.wormholeAdapter.rely(address(report.protocolGuardian));
-        }
-        if (address(report.axelarAdapter) != address(0)) report.axelarAdapter.rely(address(report.protocolGuardian));
-        if (address(report.chainlinkAdapter) != address(0)) {
-            report.chainlinkAdapter.rely(address(report.protocolGuardian));
-        }
+        _relyAdapters(report, address(report.protocolGuardian));
 
         // Rely opsGuardian
         report.core.multiAdapter.rely(address(report.opsGuardian));
         report.core.hub.rely(address(report.opsGuardian));
         // Temporal ward for initial adapter wiring
-        if (address(report.layerZeroAdapter) != address(0)) report.layerZeroAdapter.rely(address(report.opsGuardian));
-        if (address(report.wormholeAdapter) != address(0)) report.wormholeAdapter.rely(address(report.opsGuardian));
-        if (address(report.axelarAdapter) != address(0)) report.axelarAdapter.rely(address(report.opsGuardian));
-        if (address(report.chainlinkAdapter) != address(0)) report.chainlinkAdapter.rely(address(report.opsGuardian));
+        _relyAdapters(report, address(report.opsGuardian));
 
         // Rely tokenRecoverer
         report.root.rely(address(report.tokenRecoverer));
@@ -351,6 +336,14 @@ contract FullActionBatcher is CoreActionBatcher {
         if (address(report.axelarAdapter) != address(0)) report.axelarAdapter.deny(address(this));
         if (address(report.layerZeroAdapter) != address(0)) report.layerZeroAdapter.deny(address(this));
         if (address(report.chainlinkAdapter) != address(0)) report.chainlinkAdapter.deny(address(this));
+    }
+
+    /// @dev helper function to save some bytes
+    function _relyAdapters(FullReport memory report, address ward) internal {
+        if (address(report.layerZeroAdapter) != address(0)) report.layerZeroAdapter.rely(ward);
+        if (address(report.wormholeAdapter) != address(0)) report.wormholeAdapter.rely(ward);
+        if (address(report.axelarAdapter) != address(0)) report.axelarAdapter.rely(ward);
+        if (address(report.chainlinkAdapter) != address(0)) report.chainlinkAdapter.rely(ward);
     }
 
     function _setLayerZeroUlnConfig(LayerZeroAdapter adapter, uint32 eid, SetConfigParam[] memory params) internal {
--- a/script/FullDeployer.s.sol
+++ b/script/FullDeployer.s.sol
@@ -39,6 +39,7 @@ import {SyncDepositVaultFactory} from "../src/vaults/factories/SyncDepositVaultF
 
 import "forge-std/Script.sol";
 
+import {TokenBridge} from "../src/bridge/TokenBridge.sol";
 import {SubsidyManager} from "../src/utils/SubsidyManager.sol";
 import {AxelarAdapter} from "../src/adapters/AxelarAdapter.sol";
 import {WormholeAdapter} from "../src/adapters/WormholeAdapter.sol";
@@ -101,6 +102,7 @@ struct FullReport {
     TokenRecoverer tokenRecoverer;
     ProtocolGuardian protocolGuardian;
     OpsGuardian opsGuardian;
+    TokenBridge tokenBridge;
     SubsidyManager subsidyManager;
     RefundEscrowFactory refundEscrowFactory;
     AsyncVaultFactory asyncVaultFactory;
@@ -141,6 +143,7 @@ contract FullActionBatcher is CoreActionBatcher {
     ) public onlyDeployer {
         // Rely Root
         report.tokenRecoverer.rely(address(report.root));
+        report.tokenBridge.rely(address(report.root));
 
         report.subsidyManager.rely(address(report.root));
         report.refundEscrowFactory.rely(address(report.root));
@@ -183,6 +186,7 @@ contract FullActionBatcher is CoreActionBatcher {
         report.core.messageDispatcher.rely(address(report.protocolGuardian));
         report.root.rely(address(report.protocolGuardian));
         report.tokenRecoverer.rely(address(report.protocolGuardian));
+        report.tokenBridge.rely(address(report.protocolGuardian));
         // Permanent ward for ongoing adapter maintenance
         if (address(report.layerZeroAdapter) != address(0)) {
             report.layerZeroAdapter.rely(address(report.protocolGuardian));
@@ -261,10 +265,10 @@ contract FullActionBatcher is CoreActionBatcher {
         report.batchRequestManager.file("hub", address(report.core.hub));
 
         // Endorse methods
-
         report.root.endorse(address(report.core.balanceSheet));
         report.root.endorse(address(report.asyncRequestManager));
         report.root.endorse(address(report.vaultRouter));
+        report.root.endorse(address(report.tokenBridge));
 
         // Connect adapters
         for (uint256 i; i < connectionList.length; i++) {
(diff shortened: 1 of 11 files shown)


lemunozm:
Original PR for reference: https://github.com/centrifuge/protocol-internal/pull/74

src/bridge/TokenBridge.sol
```diff
@@ -0,0 +1,96 @@
...
+import {SafeTransferLib} from "../misc/libraries/SafeTransferLib.sol";
+
+import {PoolId} from "../core/types/PoolId.sol";
+import {ISpoke} from "../core/spoke/interfaces/ISpoke.sol";
+import {ShareClassId} from "../core/types/ShareClassId.sol";
+import {ITrustedContractUpdate} from "../core/utils/interfaces/IContractUpdate.sol";
+
+/// @title  TokenBridge
+/// @notice Wrapper contract for cross-chain token transfers compatible with Glacis Airlift
+contract TokenBridge is Auth, ITokenBridge {
```
wischli, line 17: Q for all of us: Should this contract implement `Recoverable` to protect against accidently sent tokens?

src/bridge/TokenBridge.sol
```diff
@@ -0,0 +1,96 @@
...
+
+    /// @inheritdoc ITokenBridge
+    function file(bytes32 what, uint256 evmChainId, uint16 centrifugeId) external auth {
+        if (what == "centrifugeId") chainIdToCentrifugeId[evmChainId] = centrifugeId;
+        else revert FileUnrecognizedParam();
+        emit File(what, evmChainId, centrifugeId);
+    }
+
+    /// @inheritdoc ITrustedContractUpdate
+    function trustedCall(PoolId poolId, ShareClassId scId, bytes memory payload) external auth {
```
wischli, line 47: Nit: We can use `calldata` payload. I discovered this as part of optimizing ARM size code. We still have several code paths in the repo using `memory` despite the interface declaration of `calldata`.
```suggestion
    function trustedCall(PoolId poolId, ShareClassId scId, bytes calldata payload) external auth {
```

src/bridge/TokenBridge.sol
```diff
@@ -0,0 +1,96 @@
...
+        }
+    }
+
+    //----------------------------------------------------------------------------------------------
+    // Bridging
+    //----------------------------------------------------------------------------------------------
+
+    /// @inheritdoc ITokenBridge
+    function send(address token, uint256 amount, bytes32 receiver, uint256 destinationChainId, address refundAddress)
+        public
```
wischli, line 68: Should this really be `public` or instead `external`?

src/bridge/TokenBridge.sol
```diff
@@ -0,0 +1,96 @@
...
+
+        (PoolId poolId, ShareClassId scId) = spoke.shareTokenDetails(token);
+        GasLimits memory limits = gasLimits[poolId][scId];
+
+        spoke.crosschainTransferShares{value: msg.value}(
+            centrifugeId,
+            poolId,
+            scId,
+            receiver,
+            uint128(amount),
```
wischli, line 88: CR: Should use safe method from Mathlib to improve debugging in case this edge case occurs

```suggestion
            amount.toUint128(),
```

src/bridge/TokenBridge.sol
```diff
@@ -0,0 +1,96 @@
...
+    {
+        uint16 centrifugeId = chainIdToCentrifugeId[destinationChainId];
+        require(centrifugeId != 0, InvalidChainId());
+
+        SafeTransferLib.safeTransferFrom(token, msg.sender, address(this), amount);
+        if (IERC20(token).allowance(address(this), address(spoke)) == 0) {
+            SafeTransferLib.safeApprove(token, address(spoke), type(uint256).max);
+        }
+
+        (PoolId poolId, ShareClassId scId) = spoke.shareTokenDetails(token);
```
wischli, line 80: Nit: We should move this up before the actual transfer because it is theoretically fallible such that we can save costs in case this check fails.

src/bridge/TokenBridge.sol
```diff
@@ -0,0 +1,96 @@
...
+        public
+        payable
+        returns (bytes memory)
+    {
+        uint16 centrifugeId = chainIdToCentrifugeId[destinationChainId];
+        require(centrifugeId != 0, InvalidChainId());
+
+        SafeTransferLib.safeTransferFrom(token, msg.sender, address(this), amount);
+        if (IERC20(token).allowance(address(this), address(spoke)) == 0) {
+            SafeTransferLib.safeApprove(token, address(spoke), type(uint256).max);
```
wischli, line 77: I am curious why this is necessary. I might be missing something but AFAICT the Spoke moves TokenBridge's tokens only using its ward privilege not via a standard ERC20 allowance and because it passes `msg.sender` as both the `sender` and `from` arguments, the allowance is never checked?

src/bridge/interfaces/ITokenBridge.sol
```diff
@@ -0,0 +1,72 @@
...
+interface ITokenBridge is ITrustedContractUpdate {
+    event File(bytes32 indexed what, address data);
+    event File(bytes32 indexed what, uint256 evmChainId, uint16 centrifugeId);
+    event UpdateGasLimits(
+        PoolId indexed poolId, ShareClassId indexed scId, uint128 extraGasLimit, uint128 remoteExtraGasLimit
+    );
+
+    error FileUnrecognizedParam();
+    error InvalidChainId();
+    error InvalidRelayer();
```
wischli, line 17: Wanted to flag these are not yet used in the codebase

src/bridge/interfaces/ITokenBridge.sol
```diff
@@ -0,0 +1,72 @@
...
+    error InvalidToken();
+    error UnknownTrustedCall();
+    error ShareTokenDoesNotExist();
+    error FailedToTransferToRelayer();
```
wischli, line 21: Same here

test/bridge/unit/TokenBridge.t.sol
```diff
@@ -0,0 +1,292 @@
...
+    }
+
+    function testSendInvalidChainId() public {
+        uint256 invalidChainId = 999;
+
+        vm.expectRevert(ITokenBridge.InvalidChainId.selector);
+        bridge.send(shareToken1, DEFAULT_AMOUNT, receiver.toBytes32(), invalidChainId, user);
+    }
+
+    function testSendInvalidToken() public {
```
wischli, line 276: Great tests! Could you add one for revoking a chain mapping by setting centrifugeId to 0 and then verifying `send()` reverts? And maybe also one for catching `uint128(amount)` overflow.

test/core/unit/ProtocolGuardian.t.sol
```diff
@@ -310,3 +314,50 @@ contract ProtocolGuardianTestFile is ProtocolGuardianTest {
...
+        protocolGuardian.fileTokenBridgeRelayer(relayer);
+    }
+
+    function testFileRelayerRevertWhenNotSafe() public {
+        vm.prank(UNAUTHORIZED);
+        vm.expectRevert(IProtocolGuardian.NotTheAuthorizedSafe.selector);
+        protocolGuardian.fileTokenBridgeRelayer(makeAddr("relayer"));
+    }
+
+    function testfileCentrifugeIdSuccess() public {
```
wischli, line 341: Nit: Incorrect camelcase

```suggestion
    function testFileCentrifugeIdSuccess() public {
```

test/core/unit/ProtocolGuardian.t.sol
```diff
@@ -310,3 +314,50 @@ contract ProtocolGuardianTestFile is ProtocolGuardianTest {
...
+        vm.expectCall(
+            address(tokenBridge),
+            abi.encodeWithSignature("file(bytes32,uint256,uint16)", bytes32("centrifugeId"), evmChainId, CENTRIFUGE_ID)
+        );
+
+        vm.prank(address(SAFE));
+        protocolGuardian.fileTokenBridgeCentrifugeId(evmChainId, CENTRIFUGE_ID);
+    }
+
+    function testfileCentrifugeIdRevertWhenNotSafe() public {
```
wischli, line 358:
```suggestion
    function testFileCentrifugeIdRevertWhenNotSafe() public {
```

wischli (commented):
Thanks for re-opening and good job on putting this all together. I have a few questions before merging though.

script/FullDeployer.s.sol
```diff
@@ -348,6 +338,14 @@ contract FullActionBatcher is CoreActionBatcher {
         if (address(report.chainlinkAdapter) != address(0)) report.chainlinkAdapter.deny(address(this));
     }
 
+    /// @dev helper function to save some bytes
+    function _relyAdapters(FullReport memory report, address ward) internal {
+        if (address(report.layerZeroAdapter) != address(0)) report.layerZeroAdapter.rely(ward);
+        if (address(report.wormholeAdapter) != address(0)) report.wormholeAdapter.rely(ward);
+        if (address(report.axelarAdapter) != address(0)) report.axelarAdapter.rely(ward);
+        if (address(report.chainlinkAdapter) != address(0)) report.chainlinkAdapter.rely(ward);
+    }
```
lemunozm, line 347: Nice idea to reduce bytecode!

src/bridge/TokenBridge.sol
```diff
@@ -0,0 +1,96 @@
...
+        else revert FileUnrecognizedParam();
+        emit File(what, data);
+    }
+
+    /// @inheritdoc ITokenBridge
+    function file(bytes32 what, uint256 evmChainId, uint16 centrifugeId) external auth {
+        if (what == "centrifugeId") chainIdToCentrifugeId[evmChainId] = centrifugeId;
+        else revert FileUnrecognizedParam();
+        emit File(what, evmChainId, centrifugeId);
+    }
```
lemunozm, line 44: From my experience, `file()` methods for things other than dependencies can be somewhat messy in the end and difficult to handle. Could we use here a "normal" name like `setDestination(evmChainId, centrifugeId)` or something similar?

src/bridge/interfaces/ITokenBridge.sol
```diff
@@ -0,0 +1,72 @@
...
+    /// @param centrifugeId The corresponding Centrifuge chain ID
+    function file(bytes32 what, uint256 evmChainId, uint16 centrifugeId) external;
+
+    //----------------------------------------------------------------------------------------------
+    // Bridging
+    //----------------------------------------------------------------------------------------------
+
+    /// @notice Send a token from chain A to chain B after approving this contract with the token
+    /// @dev This function transfers tokens from the caller and initiates a cross-chain transfer
+    /// @dev These methods match the expected interface from Glacis Airlift for cross-chain token transfers
```
lemunozm, line 53: Maybe then we should move `send(..)` to its own interface like `IGlacisAirliftTransfer` or so? to reflect that and ensure this method signature remains immutable.

Do you have a reference for where this method is defined?

src/bridge/TokenBridge.sol
lemunozm: Yes, I also think the allowance is never checked. I think approval is something the user of `TokenBridge` should do before call to `TokenBridge` for leave `TokenBridge` to handle the transfer funds.

src/bridge/TokenBridge.sol
```diff
@@ -0,0 +1,96 @@
...
+
+        spoke.crosschainTransferShares{value: msg.value}(
+            centrifugeId,
+            poolId,
+            scId,
+            receiver,
+            uint128(amount),
+            limits.extraGasLimit,
+            limits.remoteExtraGasLimit,
+            relayer != address(0) ? relayer : refundAddress // Transfer remaining ETH to relayer if set
```
lemunozm, line 91: Why not always to the `refundAddress`? No fully understand the need of the `relayer`, could you extend?

src/bridge/TokenBridge.sol
lemunozm: Looks like it's never used as a target for receiving tokens. I would say not (?)

test/bridge/integration/TokenBridge.t.sol
```diff
@@ -0,0 +1,145 @@
...
+        tokenBridge.send{
+            value: 0.1 ether
+        }(address(shareToken), shareBalance, receiver.toBytes32(), DESTINATION_CHAIN_ID, user);
+
+        assertEq(shareToken.balanceOf(address(tokenBridge)), 0);
+        assertGt(user.balance, 0.99 ether); // Got refunded
+    }
+
+    /// forge-config: default.isolate = true
+    function testSendWithRelayerSuccess() public {
```
lemunozm, line 100: IIUC, the relayer does not imply any new interaction with the system regarding the normal `testSendSuccess()`. I think the good usage of the relayer is already covered by UTs, could it be?

lemunozm (commented):
Some minor organizational comments, agree with William regarding the allowance/approval part. IIUC, we should change that.

But code looks really good!

PR #774, opened 2026-02-09
Ref: LaunchDeployer using EnvConfig

lemunozm:
### Product requirements

* Refactorize `LaunchDeployer` using `EnvConfig`
* Simplify inputs in `FullDeployer`
* extract layer zero `SetConfigParam` from connections and add it to layer zero config (now we have bytecode to do this in the Batcher and I think it scales better)
  * Happy if you can check this @onnovisser 
* Rename `adminSafe` to `protocolSafe`

Already verified with claude that the change in the deployer is equivalent, so the refactor doesn't add, change, or remove any logic.

### DIFF
--- a/script/FullDeployer.s.sol
+++ b/script/FullDeployer.s.sol
@@ -154,10 +154,7 @@ contract FullActionBatcher is CoreActionBatcher {
 
         report.batchRequestManager.rely(address(report.root));
 
-        if (address(report.layerZeroAdapter) != address(0)) report.layerZeroAdapter.rely(address(report.root));
-        if (address(report.wormholeAdapter) != address(0)) report.wormholeAdapter.rely(address(report.root));
-        if (address(report.axelarAdapter) != address(0)) report.axelarAdapter.rely(address(report.root));
-        if (address(report.chainlinkAdapter) != address(0)) report.chainlinkAdapter.rely(address(report.root));
+        _relyAdapters(report, address(report.root));
 
         // Rely spoke
         report.asyncRequestManager.rely(address(report.core.spoke));
@@ -181,25 +178,13 @@ contract FullActionBatcher is CoreActionBatcher {
         report.root.rely(address(report.protocolGuardian));
         report.tokenRecoverer.rely(address(report.protocolGuardian));
         // Permanent ward for ongoing adapter maintenance
-        if (address(report.layerZeroAdapter) != address(0)) {
-            report.layerZeroAdapter.rely(address(report.protocolGuardian));
-        }
-        if (address(report.wormholeAdapter) != address(0)) {
-            report.wormholeAdapter.rely(address(report.protocolGuardian));
-        }
-        if (address(report.axelarAdapter) != address(0)) report.axelarAdapter.rely(address(report.protocolGuardian));
-        if (address(report.chainlinkAdapter) != address(0)) {
-            report.chainlinkAdapter.rely(address(report.protocolGuardian));
-        }
+        _relyAdapters(report, address(report.protocolGuardian));
 
         // Rely opsGuardian
         report.core.multiAdapter.rely(address(report.opsGuardian));
         report.core.hub.rely(address(report.opsGuardian));
         // Temporal ward for initial adapter wiring
-        if (address(report.layerZeroAdapter) != address(0)) report.layerZeroAdapter.rely(address(report.opsGuardian));
-        if (address(report.wormholeAdapter) != address(0)) report.wormholeAdapter.rely(address(report.opsGuardian));
-        if (address(report.axelarAdapter) != address(0)) report.axelarAdapter.rely(address(report.opsGuardian));
-        if (address(report.chainlinkAdapter) != address(0)) report.chainlinkAdapter.rely(address(report.opsGuardian));
+        _relyAdapters(report, address(report.opsGuardian));
 
         // Rely tokenRecoverer
         report.root.rely(address(report.tokenRecoverer));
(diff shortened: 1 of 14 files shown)


script/utils/EnvConfig.s.sol
```diff
@@ -296,7 +298,48 @@ library Env {
 
 library EnvConfigLib {
     function etherscanApiKey(EnvConfig memory) internal view returns (string memory) {
-        return _envString("ETHERSCAN_API_KEY");
+        return prettyEnvString("ETHERSCAN_API_KEY");
+    }
+
+    function buildZeroConfigParams(EnvConfig memory config) internal view returns (SetConfigParam[] memory params) {
```
onnovisser:
```suggestion
    function buildLayerZeroConfigParams(EnvConfig memory config) internal view returns (SetConfigParam[] memory params) {
```

script/LaunchDeployer.s.sol
```diff
@@ -105,172 +52,64 @@ contract LaunchDeployer is FullDeployer {
...
+            opsSafe: ISafe(config.network.opsAdmin),
+            core: CoreInput({
+                centrifugeId: config.network.centrifugeId, version: version, txLimits: config.network.buildBatchLimits()
+            }),
+            adapters: AdaptersInput({
+                layerZero: LayerZeroInput({
+                    shouldDeploy: config.adapters.layerZero.deploy,
+                    endpoint: config.adapters.layerZero.endpoint,
+                    delegate: config.network.protocolAdmin,
+                    configParams: config.buildZeroConfigParams()
```
onnovisser:
```suggestion
                    configParams: config.buildLayerZeroConfigParams()
```

lemunozm: Good catch!

script/FullDeployer.s.sol
```diff
@@ -278,9 +275,9 @@ contract FullActionBatcher is CoreActionBatcher {
                     .wire(connections.centrifugeId, abi.encode(connections.layerZeroId, report.layerZeroAdapter));
                 adapters[n++] = report.layerZeroAdapter;
 
-                if (connections.layerZeroConfigParams.length > 0) {
+                if (input.adapters.layerZero.configParams.length > 0) {
                     _setLayerZeroUlnConfig(
-                        report.layerZeroAdapter, connections.layerZeroId, connections.layerZeroConfigParams
+                        report.layerZeroAdapter, connections.layerZeroId, input.adapters.layerZero.configParams[i]
```
wischli, line 265: Nit: We silently assume this array to be aligned with `connections`. A length mismatch would cause an revert with no debug msg. IMO, a require at the to of the loop would make this more robust.

wischli (dismissed):
Incredible improvements! Just a minor nit that can be ignored.

script/FullDeployer.s.sol
lemunozm: Yes, agree with it. Actually, two things should match:
- The length
- The order

I think we can cover this as part of the `EvmConfig` parser. I mean, if both can give the same results, the env file can not be parsed. WDYT? That way we don't add more lines to the deployed Batched

lemunozm: Finally checked only the length in the FullDeployer. I think checking the order is a bit overkill by now because both will iterate connectTo in the same way. Also not easy to check.

script/FullDeployer.s.sol
```diff
@@ -633,6 +625,11 @@ contract FullDeployer is CoreDeployer {
             require(input.adapters.layerZero.endpoint != address(0), "LayerZero endpoint address cannot be zero");
             require(input.adapters.layerZero.endpoint.code.length > 0, "LayerZero endpoint must be a deployed contract");
             require(input.adapters.layerZero.delegate != address(0), "LayerZero delegate address cannot be zero");
+            require(
+                input.adapters.layerZero.configParams.length == 0
+                    || input.adapters.layerZero.configParams.length != input.adapters.connections.length,
```
wischli: Flipped condition
```suggestion
                    || input.adapters.layerZero.configParams.length == input.adapters.connections.length,
```

wischli (commented):
~~One minor CR~~

script/FullDeployer.s.sol
wischli: Ah, I was too late.

wischli (dismissed):
LGTM!

script/utils/EnvConfig.s.sol
```diff
@@ -296,7 +298,52 @@ library Env {
...
+        for (uint256 i; i < connectsTo.length; i++) {
+            EnvConfig memory remoteConfig = Env.load(connectsTo[i]);
+
+            require(
+                config.adapters.layerZero.dvns.length == remoteConfig.adapters.layerZero.dvns.length,
+                "DVNs count mismatch between local and remote config"
+            );
+            for (uint256 j; j < config.adapters.layerZero.dvns.length; j++) {
+                require(
+                    config.adapters.layerZero.dvns[j] == remoteConfig.adapters.layerZero.dvns[j],
```
onnovisser: The DVN addresses don't necessarily match, see here for example: https://docs.layerzero.network/v2/deployments/dvn-addresses?dvns=LayerZero+Labs&chains=ethereum%2Cbase%2Carbitrum%2Cavalanche%2Cplumephoenix%2Cbsc
It's the DVNs themselves that have to match, but there's no way to tell from the config

lemunozm: Ahhh ok. Wow that's difficult to ensure. I'll remove that check then

onnovisser: Yes it's annoying 🤷

onnovisser (approved):

PR #776, opened 2026-02-10
Refactor: deployers and ActionBatcher

lemunozm:
## Product requirements

* `FullActionBatcher` had bytecode issues that forbid to scale following the current architecture
* Modify `FullDeployer` is complicated and error-prone (i.e: forgetting to call `register("contract", address)` and different calls to get things deployed)
* Use `FullDeployer` is also cumbersome when used in tests and in `LaunchDeployer`.
* Fix create3. The compiled create3 bytecode from our repo was wrong (not sure why 🤔). To workaround that, we were deploying in integration tests always with create1. Now integration tests are correctly deployed with create3. 
  * Note that for mainnet deployment, it was never used because it was already deployed there

### `FullActionBatcher` changes
- Divided into 3 action batchers (to split conceptual contract levels):
  - `CoreActionBatcher`: 13,410 bytes
  - `NonCoreActionBatcher`: 6,322 bytes (**open to other names**)
  - `AdapterActionBatcher`: 7,133 bytes
  NOTE: have them split force us to avoid i.e: rely from nonCore to core, which is the correct logical behavior.
- Used the constructor pattern instead, it's everything inside the constructor. So we no longer need `revoke()` and `lock()` methods or store a `deployer` on it. The constructor allows bigger bytecode on it, > 40KB.
- Action batchers moved to `src/deployment/ActionBatcher.sol` 

### `Core/FullDeployer` changes
* Unify `CoreDeployer` and `FullDeployer` into `FullDeployer`, now with 3 methods: `deployCore`, `deployNonCore` and `deployAdapters` mimic the batcher conceptual separation.
* Use the upstream create3 bytecode, so we no longer need to workaround tests with create1. Closest test behavior to production.
* Remove the need to call `register("contractName", contractAddress)` for each time.

### `LaunchDeployer`
* Simplified `LaunchDeployer` and test deployment: 
  * Now the batcher is a fully abstract concept (i.e: test don't need to manage it)
  * Just one call `deployFull(input)` to fully make a deployment: no revokes, no locks, no batchers in between


script/FullDeployer.s.sol
```diff
@@ -400,17 +192,184 @@ contract FullDeployer is CoreDeployer {
...
 
+        version = version_;
+        deployer = deployer_;
+    }
+
+    /// @dev Generates a deterministic salt based on contract name and optional VERSION
+    function generateSalt(string memory contractName) internal returns (bytes32 salt) {
+        salt = makeSalt(contractName, version, deployer);
+        register(contractName, computeCreate3Address(salt, deployer));
+    }
```
lemunozm: This automatically register all contracts

script/FullDeployer.s.sol
```diff
@@ -400,17 +192,184 @@ contract FullDeployer is CoreDeployer {
...
+            input.adapters.layerZero.delegate,
+            vm.toString(address(axelarAdapter))
+        );
+    }
+
+    function removeFullDeployerAccess() public {
+        coreBatcher.revokeCore(coreReport());
+        nonCoreBatcher.revokeNonCore(nonCoreReport());
+        adapterBatcher.revokeAdapters(adaptersReport());
+    }
```
lemunozm: This is the new deployment pipeline, already encapsulated in `FullDeployer` (so `LaunchDeployer` or integration test don't need to deal with it)

test/spell/migration/validation/InvestmentFlowExecutor.sol
```diff
@@ -60,7 +60,7 @@ struct InvestmentFlowResult {
 }
 
 struct InvestmentFlowContext {
-    FullReport report;
+    NonCoreReport report;
```
lemunozm, line 63: This is temporal. Ideally, this should use `EnvConfig.Contracts`, this is a WIP for the validation refactor (the same for the fork test)

script/FullDeployer.s.sol
```diff
@@ -400,17 +185,196 @@ contract FullDeployer is CoreDeployer {
...
+                        input.protocolSafe,
+                        input.adapters.connections,
+                        input.adapters.layerZero.configParams,
+                        input.adapters.layerZero.delegate,
+                        vm.toString(address(axelarAdapter))
+                    )
+                )
+            )
+        );
+    }
```
lemunozm, line 222: This is the new deployment pipeline (everything encapsulated in `FullDeployer`).

1. The addresses of batchers are previewed (NOte no one can frontrun us because we use protected create3 addresses by the sender)
2. We deploy the contracts
3. We rely/deny/file contracts on the go during the action batch constructor

script/OnlyAdapters.s.sol
```diff
@@ -1,28 +1,21 @@
...
-import "forge-std/Script.sol";
-
 import {AxelarAdapter} from "../src/adapters/AxelarAdapter.sol";
 import {WormholeAdapter} from "../src/adapters/WormholeAdapter.sol";
 import {LayerZeroAdapter} from "../src/adapters/LayerZeroAdapter.sol";
 
 /// @title OnlyAdapters
 /// @notice Deploys only messaging adapters, reusing existing core addresses from env/<network>.json
-contract OnlyAdapters is Script, JsonRegistry, CoreDeployer {
+contract OnlyAdapters is FullDeployer {
```
wischli: **[Medium] `OnlyAdapters` registry will write wrong addresses**

`version` is set (line 30) but `deployer` is never set (remains `address(0)`). The new `generateSalt` auto-registers via `computeCreate3Address(salt, deployer)` with `deployer=address(0)`.

But CreateX uses a **different guard salt computation** when the sender bytes in the salt are all zeros (chain-id based) vs when they match `msg.sender` (deployer-based). So the address registered by `generateSalt` won't match the actual deployed address.

The old code used `register("wormholeAdapter", address(wormholeAdapter))` — the actual deployed address — which was correct.

**Fix:** Call `_init(version, msg.sender)` (or set `deployer = tx.origin`) before any `generateSalt` calls in `run()`. Or override registration to use the actual address returned by `create3`.

src/deployment/ActionBatchers.sol
```diff
@@ -0,0 +1,522 @@
...
+        report.subsidyManager.deny(address(this));
+
+        report.freezeOnlyHook.deny(address(this));
+        report.fullRestrictionsHook.deny(address(this));
+        report.freelyTransferableHook.deny(address(this));
+        report.redemptionRestrictionsHook.deny(address(this));
+
+        report.batchRequestManager.deny(address(this));
+
+        report.core.root.deny(address(this));
```
wischli: **[Medium] `QueueManager` missing `rely(root)` + `deny(address(this))` — wards permanently bricked**

`QueueManager` inherits `Auth` and is deployed with `nonCoreBatcherAddr` as its initial ward. But this constructor never touches `QueueManager` at all — no `rely(root)`, no `deny(address(this))`.

After deployment:
- Root is NOT a ward on QueueManager
- The dead nonCoreBatcher address is the only ward (permanently unusable)
- `Root.relyContract()` also requires Root to already be a ward on the target, so there is **no recovery path**

QueueManager's operational methods still work (they use `contractUpdater` checks), but `rely()`/`deny()` are permanently bricked.

```solidity
// Add before the revoke block:
report.queueManager.rely(root);

// Add to the revoke block:
report.queueManager.deny(address(this));
```

script/FullDeployer.s.sol
```diff
@@ -434,7 +398,9 @@ contract FullDeployer is CoreDeployer {
                 abi.encodePacked(type(OpsGuardian).creationCode, abi.encode(ISafe(address(batcher)), hub, multiAdapter))
```
wischli: **[Info] Guardian safe bootstrap pattern**

Both guardians are deployed with `ISafe(address(batcher))` as their safe, then `CoreActionBatcher` calls `file("safe", ...)` / `file("opsSafe", ...)` to set the real safes.

This works because it's all atomic in the constructor (no observable intermediate state), but worth noting for auditors that the initial state has the batcher as the safe.

script/FullDeployer.s.sol
```diff
@@ -400,17 +185,196 @@ contract FullDeployer is CoreDeployer {
...
+    function _init(bytes32 version_, address deployer_) internal {
+        // NOTE: This implementation must be idempotent
+        setUpCreateXFactory();
 
+        version = version_;
+        deployer = deployer_;
+    }
+
+    /// @dev Generates a deterministic salt based on contract name and optional VERSION
+    function generateSalt(string memory contractName) internal returns (bytes32 salt) {
```
wischli: NIT: IMO this should be renamed then as it doesn't just generate a salt but also register it. Should update docs as well.

script/OnlyAdapters.s.sol
wischli: I suppose I am missing something but shouldn't we set the `deployer` here via `super._init` somewhere? For `FullDeployer`, we are doing that in `deployFull` but it doesn't seem like that method is applicable to OnlyAdapters`?!

wischli (commented):
Really clean separation and super happy to finally use create3 also in ITs. One question and one nit.

script/FullDeployer.s.sol
lemunozm: Good point!

script/OnlyAdapters.s.sol
lemunozm: Good catch! The version was set but the deployer not. Will fix!

script/FullDeployer.s.sol
lemunozm: Renamed to `createSalt` as indicator of something that generates + register stuff: [039c9a6](https://github.com/centrifuge/protocol/pull/776/commits/039c9a68c180c45d8dee74c843cb82f0a7f6dcb7)

script/OnlyAdapters.s.sol
lemunozm: Fixed here: [75cb97f](https://github.com/centrifuge/protocol/pull/776/commits/75cb97fff3aca268fc6e052ce633110de1e1b6de)

But later I realized that we don't need to inherit from the FullDeployer, we can just isolate the logic about salt/create3 addresses into a `BaseDeployer` (also useful for future scripts that needs to deploy subsets of contracts):  [d9b7801](https://github.com/centrifuge/protocol/pull/776/commits/d9b78016d6f75147b469829f7233342ab21a5501)

script/OnlyAdapters.s.sol
```diff
@@ -65,14 +60,16 @@ contract OnlyAdapters is Script, JsonRegistry, CoreDeployer {
         vm.startBroadcast();
         captureStartBlock();
 
+        _init(vm.envOr("VERSION", string("")).toBytes32(), msg.sender);
```
wischli, line 63: Nit: We set `deployerEOA = tx.origin` but use `msg.sender` here for salt computation. AFAIU, in forge scripts `msg.sender == tx.origin` right? Can we make this more consistent or add a comment?

wischli (approved):

script/OnlyAdapters.s.sol
lemunozm: Yeah, both are the same, but we should unify. I'll add this as task for when I refactorize this script to use EnvConfig instead 👍🏻

PR #778, opened 2026-02-16
Refactor: Spoke => Spoke, SpokeHandler and SpokeRegistry

lemunozm:
### Description

- Makes Spoke to follow the same pattern as Hub:
  - Spoke => High level with no state methods for user interaction
  - SpokeHandler => Methods comming from Hub though messages
  - SpokeRegistry => State easily mutated using simple functions

## Spoke diagram

image

A future vaults implementation will use mostly both: 
  - `Spoke`: to perform requests
    - NOTE: `SpokeRegistry` can be retriever from `Spoke`
  - `VaultRegistry` to look for the registered vault

Ideally processing a message should not call directly to `VaultRegistry`, instead should call to `SpokeHandler`, which would call `VaultRegistry`. I mean both `SpokeRegistry` and Vault registry should be at the same level handled by SpokeHandler. I think I can have this without migrating `VaultRegistry` just by coping `VaultRegisty.updateVault()` method into `SpokeHandler`. WDYT cc @hieronx?

## Messaging diagram

image


src/vaults/AsyncRequestManager.sol
```diff
@@ -51,6 +52,7 @@ contract AsyncRequestManager is Auth, IAsyncRequestManager, ITrustedContractUpda
     using RequestCallbackMessageLib for *;
 
     ISpoke public spoke;
+    ISpokeRegistry public spokeRegistry;
```
lemunozm: We can also do spoke.spokeRegistry() to remove this extra dependency, but should save more gas if we have both

src/core/spoke/BalanceSheet.sol
```diff
@@ -64,7 +64,7 @@ contract BalanceSheet is Auth, BatchedMulticall, Recoverable, IBalanceSheet, IBa
 
     /// @inheritdoc IBalanceSheet
     function file(bytes32 what, address data) external auth {
-        if (what == "spoke") spoke = ISpoke(data);
```
lemunozm, line 67: Note for myself: needs to be called "spoke" to have the same bytecode as v3.1.0

src/core/spoke/VaultRegistry.sol
```diff
@@ -34,7 +34,7 @@ contract VaultRegistry is Auth, Recoverable, IVaultRegistry, IVaultRegistryGatew
 
     /// @inheritdoc IVaultRegistry
     function file(bytes32 what, address data) external auth {
-        if (what == "spoke") spoke = ISpoke(data);
+        if (what == "spokeRegistry") spokeRegistry = ISpokeRegistry(data);
```
lemunozm, line 37: Same here

src/managers/spoke/MerkleProofManager.sol
```diff
@@ -127,7 +127,7 @@ contract MerkleProofManagerFactory is IMerkleProofManagerFactory {
 
     /// @inheritdoc IMerkleProofManagerFactory
     function newManager(PoolId poolId) external returns (IMerkleProofManager) {
-        require(balanceSheet.spoke().isPoolActive(poolId), InvalidPoolId());
+        require(balanceSheet.spokeRegistry().isPoolActive(poolId), InvalidPoolId());
```
lemunozm: Should be called `spoke()` but maybe not need it as this contract will be also deprecated

src/core/spoke/interfaces/IBalanceSheet.sol
```diff
@@ -382,9 +382,9 @@ interface IBalanceSheet is IBatchedMulticall {
     // View methods
     //----------------------------------------------------------------------------------------------
 
-    /// @notice Returns the spoke contract
-    /// @return The spoke contract instance
-    function spoke() external view returns (ISpoke);
+    /// @notice Returns the spoke registry contract
+    /// @return The spoke registry contract instance
+    function spokeRegistry() external view returns (ISpokeRegistry);
```
lemunozm: Should be called `spoke()` to avoid any changes here

src/core/spoke/VaultRegistry.sol
lemunozm: Not applies if we remove VaultRegistry

lemunozm:
Closed in favor to an internal mirror: https://github.com/centrifuge/protocol-internal/pull/174

lemunozm:
Open here: https://github.com/centrifuge/protocol-internal/pull/174

PR #780, opened 2026-02-17
Ref: Generic validators

lemunozm:
### Product requirements

* Make validators spell-agnostic: I should can have some validators for spell-x and others for spell-y. I should be able to run them independiently.
* Use `EnvConfig` to read most of the properties.

### Design notes

* `ValidationOrchestator` is now `ValdidationExecutor`, and it only executes a list of validators. The list of validators is created outside, depending on the spell context.
* Removed PRE/POST info from validator. A validator just validates, what should validate depend of where you register the validator. 3 options: pre, cache, post.
* `GraphQLStore` is now `CacheStore`.
  * Now it can cache anything, not only GraphQL stuff.
  * No longer knows how to query: if you need to query + cache, you can do: `cache.set(query.queryGraphQl())`
  * The cache always writes to a file.
* Other minor simplifications.

### Organization
```
test/integration/spell
├── utils
│  └── validation
│     ├── BaseValidator.sol
│     ├── InvestmentFlowExecutor.sol
│     ├── README.md
│     ├── TestContracts.sol
│     └── ValidationExecutor.sol
├── example-spell
│  ├── validators // Validators are per spell
│  │  └── Validate_Example.sol
│  └── ExampleTest.t.sol
├── // Other spell tests...
```


test/spell/validation/BaseValidator.sol
```diff
@@ -0,0 +1,99 @@
...
+import {Test} from "forge-std/Test.sol";
+import {stdJson} from "forge-std/StdJson.sol";
+
+struct Contracts {
+    /// @notice Always from the indexer, will represent the previous deployed version
+    LiveContracts live;
+
+    /// @notice Represents the modified version. Pre validators don't have latest version.
+    TestContracts latest;
+}
```
lemunozm: To clarify this:

- `LifeContracts` is the struct coming from `config.contracts`. It contains the format of the evm files, because life contracts must always match with these files.

- `TestContracts` uses a different struct, because it comes from the deployer Reports. It mimics the struct of the deployer because any change done in the new version should have been added to the deployer.

TLDR: Both need to be different because they represent different protocol states/versions.

script/utils/CacheStore.sol
```diff
@@ -0,0 +1,75 @@
+// SPDX-License-Identifier: AGPL-3.0-only
+pragma solidity 0.8.28;
+
+import {Vm} from "forge-std/Vm.sol";
+
+Vm constant vm = Vm(address(uint160(uint256(keccak256("hevm cheat code")))));
```
wischli, line 6: Observation: `Vm constant vm` is declared identically in CacheStore, GraphQLQuery, and JsonUtils. We could consider a shared `VmAccess.sol` if this pattern keeps spreading

test/integration/spell/utils/validation/BaseValidator.sol
```diff
@@ -0,0 +1,99 @@
...
+/// @notice Abstract base class for pre/post migration validators
+/// @dev Each validator must implement validate(ValidationContext)
+/// @dev JSON Parsing: Use stdJson helpers (readUint, readString) per field instead of
+///      vm.parseJson + abi.decode, which fails silently with mixed-type structs.
+abstract contract BaseValidator is Test {
+    using stdJson for string;
+    using JsonUtils for *;
+
+    string public name;
+    ValidationError[] _errors; // Array of all errors found (empty if passed)
```
wischli, line 42: AFAICT, errors are never reset. What if a validator is used across pre- and post-phase, could we run into issues?

test/integration/spell/utils/validation/ValidationExecutor.sol
```diff
@@ -0,0 +1,133 @@
...
+            isMainnet: config.network.isMainnet()
+        });
+    }
+
+    function runPreValidation(BaseValidator[] memory validators, bool shouldRevert) external {
+        ctx.contracts.latest = empty;
+        _execute(validators, "PRE-SPELL", shouldRevert);
+    }
+
+    function runCacheValidation(BaseValidator[] memory validators) external {
```
wischli, line 44: Maybe we should add a comment this nukes the cache?

wischli (dismissed):
This very well represents what I had in my mind on how to use validators after the migration. Great idea on removing the `abstract GraphQLQuery` to remove potential problems with diamond inheritance!

A few nits but none of them are blocking at all!

script/utils/CacheStore.sol
lemunozm: Yeah, could be an option, although maybe it's simpler to just copy the line instead of the import? 🤔

test/integration/spell/utils/validation/BaseValidator.sol
lemunozm: Good question! By the structure of a validator, it only makes sense now to be used in one stage, because it can only do one thing inside the `validate()` implementation. Maybe we can add a boolean to protect against its usage in several phases

test/integration/spell/utils/validation/ValidationExecutor.sol
lemunozm: Makes sense

script/utils/CacheStore.sol
wischli: Yeah, just something I noted for myself in case we duplicate this 10x at some point 😅

wischli (approved):

PR #782, opened 2026-02-18
Network unrelated secrets management

lemunozm:
### Product requirements

* We need to be able to generate secrets in .env independent from a specific network. Use case is able to run scripts & fork tests that works for all supported chains at the same time.
* We want to avoid a stocastic test failure in LayerZeroForkTest due reading a public api key. To solve this we need to read from our private alchemy key, but that means load secrets in CI.

### Design notes

* RPC_URL is no longer an envvar. Instead rpc urls are composed on the fly using env.network.baseRpcUrl + secret.
* `script/deploy/deploy.py <network> dump:config` is no longer an option.
* Added `script/deploy/load_secrets.py` that loads all project-related secrets into the `.env` file. Run once in the project lifetime. Similar work to the previous command. Splited to the `deploy.py` script because it does not really depends on the deployment phase.
* Refactor LayerZeroForkTest to use EnvConfig and the alchemy endpoint. Now it can just use live contracts so we can also remove the deployment phase inside the test.
* Add new CI job to run fork and spell tests with secrets

**NOTE: already checked that the test log can not leak any secret in their logs**


gpmayorga (dismissed):

lemunozm:
Verified with the new names:

```
script/deploy/load_secrets.py                                                              ┃
==> Dumping secrets to .env                                                                                                                     ┃
  → Fetching ETHERSCAN_API_KEY                                                                                                                  ┃
    • Loading protocol-etherscan-api from Google Secrets using gcloud CLI                                                                       ┃
    ✓ ETHERSCAN_API_KEY loaded                                                                                                                  ┃
  → Fetching PRIVATE_KEY                                                                                                                        ┃
    • Loading protocol-testnet-private-key from Google Secrets using gcloud CLI                                                                 ┃
    ✓ PRIVATE_KEY loaded                                                                                                                        ┃
  → Fetching ALCHEMY_API_KEY                                                                                                                    ┃
    • Loading protocol-alchemy-api from Google Secrets using gcloud CLI                                                                         ┃
    ✓ ALCHEMY_API_KEY loaded                                                                                                                    ┃
  → Fetching PLUME_API_KEY                                                                                                                      ┃
    • Loading protocol-plume-api from Google Secrets using gcloud CLI                                                                           ┃
    ✓ PLUME_API_KEY loaded                                                                                                                      ┃
  → Fetching PHAROS_API_KEY                                                                                                                     ┃
    • Loading protocol-pharos-api from Google Secrets using gcloud CLI                                                                          ┃
    ✓ PHAROS_API_KEY loaded                                                                                                                     ┃
    ✓ Secrets written to .env
```

gpmayorga (approved):

PR #783, opened 2026-02-19
chore: archive spell 006 (v3.0.1 to v3.1.0)

wischli:
### Product requirements

* Archive the v3.0.1 to v3.1 migration spell (`0xe97ac43A22B8Df15D53503cf8001F12c6B349327`) in `env/spell/` so that integrators and the team have a historical record of the full protocol migration, consistent with spells 001 through 005
* Enable future debugging of on chain state by preserving the exact spell source and a summary of what was migrated

### Design notes

* Uses a subdirectory (`env/spell/006_migration_v3.1/`) instead of the flat per network file pattern because this spell deployed identical bytecode via CREATE3 across all six chains with no per network constants
  * duplicating the same 683 line file six times adds no information
* Includes a `README.md` documenting the execution flow (`castGlobal` / `castPool` / `lock`), the dynamic GraphQL parameter sourcing, all migrated state categories, and the defensive try catch patterns used during migration
* Import paths in the archived `MigrationSpell.sol` are adjusted to resolve from `env/spell/006_migration_v3.1/` relative to the project root
* Validation test suite and migration scripts are available on the `spell-006_v3.1.0` tag under `test/spell/migration/`

### DIFF
--- a/env/spell/006_migration_v3.1/MigrationSpell.sol
+++ b/env/spell/006_migration_v3.1/MigrationSpell.sol
@@ -1,0 +1,686 @@
+// SPDX-License-Identifier: BUSL-1.1
+// Deployed Address: 0xe97ac43A22B8Df15D53503cf8001F12c6B349327 (all chains via CREATE3)
+// V3.0.1 → V3.1 Full Protocol Migration
+// See README.md in this directory for execution details
+pragma solidity 0.8.28;
+
+import {D18} from "../../src/misc/types/D18.sol";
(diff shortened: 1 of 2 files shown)


lemunozm (commented):
~~Q. For spells that receive inputs at execution time, should we also copy their `.s.sol`? To know how the values were populated?~~

Never mind, with the tag is enough!! Love our archive ❤️

lemunozm (approved):

PR #784, opened 2026-02-19
Add CI job for verifier factory contracts

lemunozm:
### Product requirements

* We want an easy way to verify contracts for externals in different chains. The current problem is we need to run the verifier script in an specific commit inside the repo

### Design notes

* Runnable github action that fetches the repo in a specific tag, copy the src/ into the current state and then run the `script/VerifyFactoryContracts.s.sol` script
* Because the script is idempotent and fast for already verified contracts, it runs automatically in all configured networks from `env/*.json` files
* Added extra ward to CI jobs to protect secrets
* Some fixes in the script to make it work fine in CI


.github/workflows/ci.yml
```diff
@@ -204,6 +210,14 @@ jobs:
...
         run: |
           script/deploy/setup.sh
+      - name: Load secrets
+        run: python3 script/deploy/load_secrets.py
+      - name: Mask secrets
+        run: |
+          while IFS='=' read -r key value; do
+            [[ -z "$key" || "$key" == \#* || -z "$value" ]] && continue
+            echo "::add-mask::$value"
+          done < .env
```
lemunozm, line 232: This is a especial addition to the CI jobs that uses secrets that avoid leak those secret in some logs by error

script/utils/EnvConfig.s.sol
```diff
@@ -219,15 +219,19 @@ library Env {
             return config;
         }
 
-        // Core
+        // Admin
         config.root = _parseContractAddress(json, "root");
+        config.tokenRecoverer = _parseContractAddress(json, "tokenRecoverer");
+        config.protocolGuardian = _parseContractAddress(json, "protocolGuardian");
+        config.opsGuardian = _parseContractAddress(json, "opsGuardian");
```
lemunozm, line 223: own NIT: Just a minor organization of the contracts

.github/workflows/verify-factory-contracts.yml
```diff
@@ -0,0 +1,102 @@
...
+    outputs:
+      networks: ${{ steps.list.outputs.networks }}
+    steps:
+      - uses: actions/checkout@v3
+        with:
+          fetch-depth: 0
+          submodules: recursive
+
+      - name: Checkout src/ from deployed version
+        run: git checkout ${{ inputs.source_version }} -- src/
```
wischli: AFAICT, we also need to include `lib/` and `foundry.toml`. Did you test this manually?

.github/workflows/ci.yml
wischli: Doesn't Github upload the tarball of an action as an artifact that's publicly available, including the secrets in the `.env`? Also, we don't need the `PRIVATE_KEY` for verification.

wischli (commented):
How can I test this action before we merge it? Definitely need @gpmayorga eyes on this before merging.

It would be great to have an error reporting, either to slack or by creating a Github issue in this repo in case this action fails. Did you look into that?

.github/workflows/verify-factory-contracts.yml
lemunozm: Yes, I already tested this.

Note the repo that is downloaded is from `main`, but later we add the `src` from the tag to be able to do `forge verify-contract` with that abi. As long as `src` compiles in its own, we're fine (at least by now)

.github/workflows/ci.yml
lemunozm: Good question 🤔, then that would be something to fix...

lemunozm:
> How can I test this action before we merge it? Definitely need @gpmayorga eyes on this before merging.

We can not test this until it's merged (GitHub limitations). But what I did is to modify the trigger to push trigger, and I could saw all jobs worked correctly, verifying it.

> It would be great to have an error reporting, either to slack or by creating a Github issue in this repo in case this action fails. Did you look into that?

This CI is triggered by a specific action. So the person who trigger this will see all their jobs passing/failing, and they will evaluate consecutively. I don't think we need further error reporting (at least by now)

.github/workflows/ci.yml
lemunozm: Very good catch @wischli! Now here ([61bc4c7](https://github.com/centrifuge/protocol/pull/784/commits/61bc4c70a051317ca41f91e2ca8b3352aee15592)), I've moved the loading secret part to each individual job, so the tarball never has an `.env` file.

gpmayorga: Unless you specify a file upload on the job the files are normally lost.
The private key is accessible to any job in this repository that calls the Google cloud login step  whether we put it in an environment file or not, it doesn't really matter

.github/workflows/verify-factory-contracts.yml
```diff
@@ -0,0 +1,104 @@
...
+          retention-days: 1
+
+  verify:
+    needs: setup
+    runs-on: ubuntu-latest
+    timeout-minutes: 30
+    permissions:
+      contents: "read"
+      id-token: "write"
+    strategy:
```
gpmayorga, line 35: Wait. Why are we uploading a tarball with yh entire workspace?
This is very dangerous for security

gpmayorga (changes requested):
I read the script and I'm not sure I understand why we need to package the entire repo, upload it and then download it within the same job?
A much easier approach without having to deal with upload/download would be:
1. Setup -> establish the matrix (nothing else)
2. Verify -> `git checkout $COMMIT -- /src` then do what you need to do for each enviornment.

.github/workflows/verify-factory-contracts.yml
lemunozm: We have a job per network, to easily see which network has issues. To alleviate the process, it performs the common part in on job before doing the proper job for all networks.

But looking now, after some refactors this has become a bit outdated because, anyway, we need to do most of the work separately. I'll change this!

lemunozm: Fixed @gpmayorga

gpmayorga (approved):

PR #789, opened 2026-02-23
Deployment: Improve version management

lemunozm:
### Product requirements

* Please read this thread: https://kflabs.slack.com/archives/C07PG2EUR9C/p1771436616145599
  * One diff: Instead of PREFIX I've called it SUFFIX. This is because for testnet we're already using suffix after the version to distinguish deployments. i.e: v3.1-rev2
* Focus: prepare a scalable version system for future migrations, spells and additions

### Design notes

* Removed `VERSION` and instead:
  * Added SUFFIX to distinguish full domain deployments.
  * Added version tags for each contract along with the names.
* Added `version` field in env/*.json files
* Added `suffix` field at deployment level in env/*.json files.
* Fix `txHash` and `blockNumber` for `root` contract in Base and BnB chains.
* Update Tag Env CI job to use the latest version found in the contracts in the env file.

**NOTE: the addresses in mainnet remain the same after these changes**. Anyways, we should verify this automatically, by generating a test that double-checks the ABI in those versions are the same as the repo.


script/BaseDeployer.s.sol
```diff
@@ -1,36 +1,64 @@
...
-        register(contractName, computeCreate3Address(salt, deployer));
+    function _isLegacyVersion(string memory version) private pure returns (bool) {
+        return keccak256(abi.encodePacked(version)) == keccak256(abi.encodePacked("3"));
+    }
+
+    function _makeSalt(string memory contractName, string memory version) private view returns (bytes32) {
+        // Legacy contracts had a different salt computation (i.e: root in some chains)
+        if (_isLegacyVersion(version)) {
+            return keccak256(abi.encodePacked(contractName, keccak256(abi.encodePacked(version))));
+        }
```
lemunozm, line 32: It's good to support the v3.0.0 deployment to have full traceability between the env json content and the addresses. The idea is that given the contract name, the version, and the prefix, I can compute the address of the contract at any time (without checking old protocol versions)

I think this is useful for: 
- debugging addresses in the future
- verifiers that need to check out the repo in a specific version to fetch the bytecode.
- tests that ensure different bytecode changes increase the version number (something for future PRs)

script/BaseDeployer.s.sol
```diff
@@ -1,36 +1,66 @@
...
+    function _makeSalt(string memory contractName, string memory version) private view returns (bytes32) {
+        require(deployer != address(0), "BaseDeployer::_init() must be called!");
+
+        // Legacy contracts had a different salt computation (i.e: root in some chains)
+        if (_isLegacyVersion(version)) {
+            return keccak256(abi.encodePacked(contractName, keccak256(abi.encodePacked(version))));
+        }
+
+        bytes32 versionHash =
+            bytes(suffix).length > 0 ? bytes32(bytes(string.concat(version, "-", suffix))) : bytes32(bytes(version));
```
wischli: Wondering if we should add a safeguard for ensuring the input is strictly at most 32 bytes. I know it's very unrealistic but doesn't hurt to be safe here?

```suggestion
            require(bytes(string.concat(version, "-", suffix)).length <= 32, "salt too long");
            bytes(suffix).length > 0 ? bytes32(bytes(string.concat(version, "-", suffix))) : bytes32(bytes(version));
```

script/FullDeployer.s.sol
```diff
@@ -177,16 +177,16 @@ contract FullDeployer is BaseDeployer, Constants {
...
     function deployFull(DeployerInput memory input, address deployer_) public {
-        _init(input.version, deployer_);
+        _init(input.suffix, deployer_);
 
-        address coreBatcherAddr = previewCreate3Address("coreBatcher");
-        address nonCoreBatcherAddr = previewCreate3Address("nonCoreBatcher");
-        address adapterBatcherAddr = previewCreate3Address("adapterBatcher");
+        address coreBatcherAddr = previewCreate3Address("coreBatcher", "v3.1");
+        address nonCoreBatcherAddr = previewCreate3Address("nonCoreBatcher", "v3.1");
+        address adapterBatcherAddr = previewCreate3Address("adapterBatcher", "v3.1");
```
wischli: Wondering if we should make the version constant to streamline deploying future versions and reducing error likelihood by missing a "v3.x" bump.

script/deploy/deploy.py
```diff
@@ -110,10 +112,12 @@ def validate_arguments(args, root_dir: pathlib.Path):
                 print_info("  - anvil (local)")
         raise SystemExit(1)
 
-    # Check if VERSION environment variable is set for deployment steps
-    if args.step.startswith("deploy:") and not os.environ.get("VERSION") and not args.dry_run:
-        print_warning("VERSION environment variable not set. Create3 address collisions may occur.")
-        print_info("Consider running: VERSION=XYZ python3 deploy.py ...")
+    # Inform about PREFIX usage for deployment steps
```
wischli: Nit:
```suggestion
    # Inform about SUFFIX usage for deployment steps
```

wischli (dismissed):
Thanks for fixing and cleaning up our deployment version control!

script/BaseDeployer.s.sol
lemunozm: Yeah, let's do this. Agree

script/FullDeployer.s.sol
lemunozm: Yeah, that's a good question I already asked myself. Maybe it makes sense, yeah. Each new version will add a new constant.

script/deploy/deploy.py
lemunozm: Done

script/BaseDeployer.s.sol
lemunozm: Done: [93066e0](https://github.com/centrifuge/protocol/pull/789/commits/93066e00ccbb5c12046f813ab2f8f156783321bb)

script/FullDeployer.s.sol
lemunozm: Done: [6377166](https://github.com/centrifuge/protocol/pull/789/commits/63771662dba789c99ad6f03afa5e71096af8b2c0)

wischli (approved):
LSGTM!

PR #793, opened 2026-02-25
New adapter wiring configuration process

lemunozm:
### Product requirements

* Thread: https://kflabs.slack.com/archives/C04GNE3JP6Z/p1766098313344409
* Issues from the current approach we want to solve:
  * Threshold can not be set for specific network pairs
  * We can not set different adapters per network pairs
  * We don't have an especific option to say if we want to wire an adapter. Right now, if it's deployed it's also wired (unless we remove the ids)
  * Very error-prone

### Design notes

* Extract the connection info from `env/*.json` files to `env/connections/*.json` files.
* Follow a rule pattern:
  * Rules are applied top-down, last rule is the effective rule.
  * We can go from generic to specific applying specific behavior to some pairs from the general configuration (see yaml file)
  * Only if when a connection is found for some chains, the associated adapters are wired for that pair
  * **To understand this file, please check the `env/connections/README.md`**
* Extra change: update `WireAdapters.s.sol` to use `EnvConfig` and add chainlink support
* Fix all missing adapter information (that was removed to avoid wiring)


env/connections/README.md
```diff
@@ -0,0 +1,115 @@
...
+Each rule defines a pair of sides, the adapters used between them, and the quorum threshold.
+
+**`chains`** is a 2-element array (a pair). Each side can be:
+
+| Format | Meaning | Example |
+|--------|---------|---------|
+| `"ALIAS"` | Reference to an alias | `"ALL"` |
+| `["net1", "net2"]` | Literal list of networks | `["pharos", "monad"]` |
+
+The two sides are permuted (cartesian product), creating a connection for every pair across them. For example, `[["A", "B"], ["C", "D"]]` produces connections: (A,C), (A,D), (B,C), (B,D). Matching is also symmetric: both the A-to-B and B-to-A directions are covered.
```
wischli, line 48: Loving the cartesian product here!

test/integration/EnvFiles.t.sol
```diff
@@ -87,3 +94,49 @@ contract EnvTestnetFilesTest is Test {
...
+                    assertTrue(chain1.adapters.wormhole.deploy, _err(pair, "wormhole"));
+                    assertTrue(chain2.adapters.wormhole.deploy, _err(pair, "wormhole"));
+                }
+                if (connections[j].axelar) {
+                    assertTrue(chain1.adapters.axelar.deploy, _err(pair, "axelar"));
+                    assertTrue(chain2.adapters.axelar.deploy, _err(pair, "axelar"));
+                }
+                if (connections[j].chainlink) {
+                    assertTrue(chain1.adapters.layerZero.deploy, _err(pair, "layerZero"));
+                    assertTrue(chain2.adapters.layerZero.deploy, _err(pair, "layerZero"));
```
wischli: CP error 
```suggestion
                    assertTrue(chain1.adapters.chainlink.deploy, _err(pair, "layerZero"));
                    assertTrue(chain2.adapters.chainlink.deploy, _err(pair, "layerZero"));
```

script/testnet/WireAdapters.s.sol
```diff
@@ -20,168 +21,74 @@ import {JsonRegistry} from "../utils/JsonRegistry.s.sol";
...
-                    remoteAdapters[count] = IAdapter(sourceWormholeAddr);
-                    count++;
-                    bytes memory wormholeData = abi.encode(
-                        uint16(vm.parseJsonUint(remoteConfig, "$.adapters.wormhole.wormholeId")), remoteWormholeAddr
-                    );
-                    opsGuardian.wire(sourceWormholeAddr, remoteCentrifugeId, wormholeData);
-                    console.log("Wired WormholeAdapter from source", sourceNetwork, "to destination", remoteNetwork);
-                }
+        for (uint256 i = 0; i < connections.length; i++) {
+            Connection memory connection = connections[i];
```
wischli, line 40: Am I missing something or are we missing to clear connections for each iteration here?


```suggestion\
            delete adapters;
            Connection memory connection = connections[i];
```

script/utils/EnvConfig.s.sol
```diff
@@ -340,17 +344,39 @@ library EnvConfigLib {
...
+        internal
+        view
+        returns (AdapterConnections[] memory adapterConnections_)
+    {
+        Connection[] memory connections = config.network.connections();
+        adapterConnections_ = new AdapterConnections[](connections.length);
+
+        for (uint256 i; i < connections.length; i++) {
+            EnvConfig memory remoteConfig = Env.load(connections[i].network);
+            Connection memory connection = config.network.connections()[i];
```
wischli: Nit: Redundant call
```suggestion
            Connection memory connection = connections[i];
```

script/deploy/lib/load_config.py
```diff
@@ -181,3 +194,56 @@ def _setup_rpc(self) -> str:
             return rpc_url
         except (subprocess.CalledProcessError, subprocess.TimeoutExpired):
             raise RuntimeError(f"RPC connection failed. URL: {rpc_url}.")
+
+
+def load_connected_networks(network_name: str, environment: str, root_dir: pathlib.Path) -> list:
```
wischli, line 199: I might be missing something here but is the asymmetry between this and the Solidity side, which only collect from aliases, intended? E.g., if a network appears only in a literal rule side, Python would include it as a potential peer but Solidity wouldn't

wischli (commented):
Oh what an improvement, great job @lemunozm 🎉

env/connections/README.md
lemunozm: Yeah, after playing with different ways, I think this the one that allows simple configurations but also powerful enough to describe all cases.

test/integration/EnvFiles.t.sol
lemunozm: Nice catch! copy paste errors by me

script/testnet/WireAdapters.s.sol
lemunozm: Another nice catch!

script/deploy/lib/load_config.py
lemunozm: Good catch again! Really nice one. No, it's not intended. The solidity script should also get the networks that appear in the rules directly as literals. I changed that part some times until getting the final solution and I broke that part :S

lemunozm: It's solved right now, but I think could be interesting to use https://github.com/OpenZeppelin/openzeppelin-contracts/blob/master/contracts/utils/structs/EnumerableSet.sol instead to handle the sets. It would simplify a lot the `EnvConfigurationConfig` file

lemunozm: We can not use that contract because it uses a storage and EnvConnectionConfig is a library. But I've got extracting the "set" behavior into a `StringSet` file, now it's much more readable

script/testnet/WireAdapters.s.sol
```diff
@@ -20,168 +21,76 @@ import {JsonRegistry} from "../utils/JsonRegistry.s.sol";
...
-                opsGuardian.initAdapters(remoteCentrifugeId, adaptersToRegister, threshold, recoveryIndex);
-                console.log("Registered", count, "source adapters on", vm.envString("NETWORK"));
-                console.log("Registered adapters for destination", remoteNetwork);
+            if (adapters.length == 0) {
+                console.log("Skipping registration for", connection.network, ": no compatible adapters");
+                continue;
             }
+
+            opsGuardian.initAdapters(
+                remote.network.centrifugeId, adapters, connection.threshold, uint8(adapters.length)
```
wischli, line 89: Just want to make sure its intended we switch from `uint8(count - 1)` previously to `adapters.length`.

lemunozm: Yeah, I changed this intentionally. I think previously it was wrong because that would mean one of the adapters was a recovery one, but that's not true.

wischli (approved):

PR #795, opened 2026-02-25
Spell: Set missing hook wards in spell

onnovisser:
### Product requirements

ContractUpdate was missing as ward on hooks. Set them in the V2CleanupSpell and in the deployer.

### Design notes

* Needed more changes in the spell than I'd like, due to added chains having a different root address and/or not having v2 root there.

### TODOs

- [ ]


src/spell/V2CleaningsSpell.sol
```diff
@@ -7,7 +7,12 @@ import {IERC20} from "../misc/interfaces/IERC20.sol";
 import {Root} from "../admin/Root.sol";
 
 Root constant ROOT_V2 = Root(0x0C1fDfd6a1331a875EA013F3897fc8a76ada5DfC);
-Root constant ROOT_V3 = Root(0x7Ed48C31f2fdC40d37407cBaBf0870B2b688368f);
+
+address constant CONTRACT_UPDATER = 0x3B150B19245D2C366bc8f18c775b725DFB298F71;
+address constant FREEZE_ONLY_HOOK = 0xd5B243F05b2906F1f6C80c6096945faADa0731C1;
+address constant FULL_RESTRICTIONS_HOOK = 0x8E680873b4C77e6088b4Ba0aBD59d100c3D224a4;
+address constant FREELY_TRANSFERABLE_HOOK = 0x2a9B9C14851Baf7AD19f26607C9171CA1E7a1A61;
+address constant REDEMPTION_RESTRICTIONS_HOOK = 0xE5423eD8602Fa0F263e17b6212d88Efe42317f06;
```
lemunozm, line 15: verified ✅

test/integration/spell/V2Cleanings.t.sol
```diff
@@ -105,14 +143,42 @@ contract V2CleaningsSpellTest is Test {
...
+        _testCase(
+            "hyper-evm", string.concat("https://hyperliquid-mainnet.g.alchemy.com/v2/", vm.envString("ALCHEMY_API_KEY"))
+        );
+    }
+
+    function testV2CleaningsMonadMainnet() external {
+        _testCase("monad", string.concat("https://monad-mainnet.g.alchemy.com/v2/", vm.envString("ALCHEMY_API_KEY")));
+    }
+
+    function testV2CleaningsPlumeMainnet() external {
```
lemunozm, line 180: We also need to add it in pharos. It has its own `PHAROS_API_KEY`. I think we can normalize this with `envconfig.rpcUrl()` in the tests, which already has the key on it.

Locally you'll need to fetch the keys using `script/deploy/load_secrets.py` first (only once for project)

lemunozm (commented):
The change is surgically clean 👌🏻. All edge cases I could see are solved.

lemunozm:
ward CI job fails because we need to modify `Deployer.t.sol` to reflect these changes there.

src/spell/V2CleaningsSpell.sol
```diff
@@ -113,4 +127,11 @@ contract V2CleaningsSpell {
             ROOT_V2.denyContract(address(ESCROW_V2), address(this));
         }
     }
+
+    function _relyContractUpdaterOnHooks(Root rootV3) internal {
```
wischli, line 133: Nit, we could check if `code.length > 0` to be fully safe but I agree it's redundant here.

test/integration/spell/V2Cleanings.t.sol
```diff
@@ -4,13 +4,21 @@ pragma solidity 0.8.28;
...
 
+import {Root} from "../../../src/admin/Root.sol";
+
+import {EnvConfig, Env} from "../../../script/utils/EnvConfig.s.sol";
+
 import "forge-std/Test.sol";
 
 import {V2CleaningsSpell} from "../../../src/spell/V2CleaningsSpell.sol";
 import {
     V2CleaningsSpell,
```
wischli: Nit: Duplicate important. I'll have Claude investigate why our script is missing this edge-case.

test/integration/spell/V2Cleanings.t.sol
```diff
@@ -33,41 +41,60 @@ contract V2CleaningsSpellTest is Test {
...
         // ----- PRE SPELL -----
 
-        uint256 preTreasuryValue = _usdc().balanceOf(TREASURY);
-        uint256 preEscrowValue = _usdc().balanceOf(ESCROW_V2);
+        IERC20 usdc = _usdc();
+        uint256 preTreasuryValue;
+        uint256 preEscrowValue;
+        if (address(usdc) != address(0)) {
+            preTreasuryValue = _usdc().balanceOf(TREASURY);
+            preEscrowValue = _usdc().balanceOf(ESCROW_V2);
```
wischli: Nit: Either use `usdc` var or drop the declaration
```suggestion
            preTreasuryValue = usdc.balanceOf(TREASURY);
            preEscrowValue = usdc.balanceOf(ESCROW_V2);
```

src/spell/V2CleaningsSpell.sol
wischli: verified

wischli (commented):
Don't we need to do this for all the perPool-hooks as well?

onnovisser:
> Don't we need to do this for all the perPool-hooks as well?

@wischli Added it to the pool hooks deploy script as well, though I think we can skip it for existing hooks, unless they need the hookManager functionality
/cc @mustermeiszer

test/integration/spell/V2Cleanings.t.sol
wischli: Import script fixed in https://github.com/centrifuge/protocol/pull/798

wischli (dismissed):
LGTM!

wischli (approved):
Re-approving!

PR #798, opened 2026-02-26
fix: duplicate import lines edge-case

wischli:
### Product requirements

* Fix the `fix_imports.py` script to handle duplicate import lines from the same path, where multiple imports pull overlapping symbols from a single file
* Prevent broken Solidity files caused by the script failing to properly remove and reconstruct multi-line imports which was caught in https://github.com/centrifuge/protocol/pull/795#discussion_r2858414985

### Design notes

* Added `find_duplicate_imports()` to detect when multiple import statements reference the same path with overlapping symbol sets
* Added `merge_duplicate_imports()` to consolidate duplicate imports into a single statement per path, preserving symbol order and deduplicating
* Switched import removal from string replacement to regex (`re.sub` with `re.MULTILINE | re.DOTALL`), which correctly handles whitespace normalization differences in multi-line imports

### DIFF
--- a/src/vaults/SyncManager.sol
+++ b/src/vaults/SyncManager.sol
@@ -2,9 +2,12 @@
 pragma solidity 0.8.28;
 
 import {IBaseVault} from "./interfaces/IBaseVault.sol";
-import {IDepositManager} from "./interfaces/IVaultManagers.sol";
-import {ISyncDepositManager} from "./interfaces/IVaultManagers.sol";
-import {ISyncManager, ISyncDepositValuation} from "./interfaces/IVaultManagers.sol";
+import {
+    IDepositManager,
+    ISyncDepositManager,
+    ISyncManager,
+    ISyncDepositValuation
+} from "./interfaces/IVaultManagers.sol";
 
 import {Auth} from "../misc/Auth.sol";
 import {D18} from "../misc/types/D18.sol";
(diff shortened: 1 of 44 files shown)


lemunozm (dismissed):
The CI failure shows correctly the fix

src/vaults/SyncManager.sol
```diff
@@ -2,9 +2,12 @@
...
 import {IBaseVault} from "./interfaces/IBaseVault.sol";
-import {IDepositManager} from "./interfaces/IVaultManagers.sol";
-import {ISyncDepositManager} from "./interfaces/IVaultManagers.sol";
-import {ISyncManager, ISyncDepositValuation} from "./interfaces/IVaultManagers.sol";
+import {
+    IDepositManager,
+    ISyncDepositManager,
+    ISyncManager,
+    ISyncDepositValuation
+} from "./interfaces/IVaultManagers.sol";
```
wischli, line 10: @hieronx Any blocker from your side for this PR since it reformats some of our non-core contracts? As you can see, the script failed to merge multiple entries for the same import path previously.

onnovisser (approved):

src/vaults/SyncManager.sol
hieronx: Yeah seems we need to.

hieronx: It's not inbetween audit and commit hash we deploy from, so not as bad

PR #799, opened 2026-02-26
docs: add missing storage interface declarations

wischli:
### Product requirements

* Interface files throughout `src/` were missing declarations for public state variable getters and explicit view/pure functions that exist on the implementing contracts
* Add all missing view/pure function declarations to their corresponding interfaces so that callers can interact with the protocol entirely through interfaces.

### Design notes

* ~143 function declarations were added across 37 interface files
* Return types for struct-valued mappings use the flattened auto-generated getter signature (individual fields), not the struct type, to match what Solidity actually exposes.
* Adding functions to `IBatchRequestManager` changed its ERC-165 interface ID from `0xaf00afbd` to `0x5cdb8e3c`
* New imports were added to interface files where the return types reference other interfaces (e.g., `IRoot`, `IHub`, `IGateway`).

**Intentionally excluded (reviewers: speak up if you disagree)**

* **Implementation constants** (~14): `BATCH_LOCATORS_SLOT()`, `GLOBAL_POOL()` (x3), `MAINNET_CENTRIFUGE_ID()`, `BASE_COST()`, `RECOVERY_TOKEN_EXTRA_COST()`, `DEFAULT_SUPPORTED_TX_LIMIT()`, `RECEIVE_COST()` (x4), `FREEZE_BIT()`. These are internal tuning knobs and sentinel values that integrators should not depend on. They can change between versions without notice and exposing them in interfaces suggests they are part of the contract's public API.
* **Transient storage getters** (2): `debited()` and `credited()` on `IAccounting`. These are `uint128 public transient` variables that always return 0 when read externally (transient storage is cleared after each transaction). Declaring them in the interface would mislead integrators into thinking they carry meaningful state.
* **Factory-specific getters**: `IVaultFactory` is shared across `AsyncVaultFactory` and `SyncDepositVaultFactory`. Factory-specific public variables like `asyncRequestManager()` or `syncDepositManager()` were not added to avoid breaking the shared interface.
* **RecoveryAdapter**: Has no dedicated interface and one is not being created as part of this change.


lemunozm:
> Adding functions to IBatchRequestManager changed its ERC-165 interface ID from 0xaf00afbd to 0x5cdb8e3c

Thanks for pointing this. I understand the same happens for other interfaces that already support the interfaceId, but should not be an error for integrators. Our methods that check the interfaceId are already compiled.

> Transient storage getters (2): debited() and credited() on IAccounting. These are uint128 public transient variables that always return 0 when read externally (transient storage is cleared after each transaction). Declaring them in the interface would mislead integrators into thinking they carry meaningful state.

I think that could not be fully true, from a hook, you could read those values != 0 inside of a transaction.

> RecoveryAdapter: Has no dedicated interface and one is not being created as part of this change.

Also good point. Not sure if adding an inheritance in the contract to a possible new interface modifies the bytecode, maybe not, and this can be added. Although this contract will never be used directly by externals

lemunozm:
To avoid the Tag Env CI issue, I think we just need to set in the pharos env the commit `0931a54f`(from the tag deploy-mainnet-v3.1-0931a54 that was already created in `deploy-v3.1.0` branch). Basically I think the correct commit added in env/pharos.json as deploy commit is not ok.

lemunozm (dismissed):
LGTM!

wischli:
> > Transient storage getters (2): debited() and credited() on IAccounting. These are uint128 public transient variables that always return 0 when read externally (transient storage is cleared after each transaction). Declaring them in the interface would mislead integrators into thinking they carry meaningful state.
> 
> I think that could not be fully true, from a hook, you could read those values != 0 inside of a transaction.

That's a valid point as hooks or any contract called within the `unlock()/lock()` window can read non-zero values. Added both to `IAccounting` in 209e9bebc with `@dev` NatSpec clarifying they're only non-zero during an active session. Also updated the PR description to remove them from the exclusion list. 

> > RecoveryAdapter: Has no dedicated interface and one is not being created as part of this change.
> 
> Also good point. Not sure if adding an inheritance in the contract to a possible new interface modifies the bytecode, maybe not, and this can be added. Although this contract will never be used directly by externals

Agreed, out of scope. Interface inheritance wouldn't change runtime bytecode (only ABI metadata), so it can be added later if needed.

lemunozm (approved):
LGTM!

onnovisser (approved):
LGTM too! Also tending towards having the IAccounting transient storage be part of the interface.

PR #801, opened 2026-03-06
Fix Cloudflare update condition in registry workflow

gpmayorga:
## Summary
- switch the Cloudflare update gate to `mainnet_changed` / `testnet_changed` outputs from the IPFS upload step
- ensure the Cloudflare update step runs only when at least one registry changed


gpmayorga:
Closing: after reviewing CI logs, the original condition was working correctly. The / CID outputs are only written when a registry actually changed, making them equivalent to the  flags. No fix needed.

PR #802, opened 2026-03-09
feat: Script to wire existing networks to newly deployed network

onnovisser:
### Product requirements

*  Wire existing chains to a newly deployed chain via a reusable script.
                                                                                                                                                                                                                                               
### Design notes

* Selects adapters from connections json
* Setting the LZ DVN config can only be done by the Protocol Safe.

### TODOs

- [ ]


script/WireToNewNetwork.s.sol
```diff
@@ -0,0 +1,194 @@
...
+///      The script reads adapter configuration from env/<network>.json and
+///      env/connections/mainnet.json to determine which adapters to wire.
+///
+///      Example usage:
+///        NETWORK=ethereum forge script script/WireToNewNetwork.s.sol --rpc-url $ETH_RPC_URL --broadcast
+///        NETWORK=base forge script script/WireToNewNetwork.s.sol --rpc-url $BASE_RPC_URL --broadcast
+contract WireToNewNetwork is Script {
+    using Safe for *;
+
+    string constant TARGET = "monad";
```
wischli: While for now we need Monad wiring, this should read from ENV so the script is agnostic for the network
```suggestion
    string constant TARGET = vm.envString("TARGET");
```

script/WireToNewNetwork.s.sol
```diff
@@ -0,0 +1,194 @@
...
+
+    function run() external {
+        vm.startBroadcast();
+        string memory networkName = vm.envString("NETWORK");
+        wire(networkName, LEDGER_DERIVATION_PATH);
+        configureLzDvns(networkName, LEDGER_DERIVATION_PATH);
+        vm.stopBroadcast();
+    }
+
+    function wire(string memory networkName, string memory derivationPath_) public {
```
wischli: Maybe we want to read `adapter.destinations(centrifugeId)` to skip already wired adapters to not run into `AdapterAlreadyWired` wasting gas if the script is not dry-run?

script/WireToNewNetwork.s.sol
```diff
@@ -0,0 +1,194 @@
...
+        Connection memory targetConn = _findTargetConnection(source);
+        derivationPath = derivationPath_;
+        opsGuardian = source.contracts.opsGuardian;
+
+        if (bytes(derivationPath).length > 0) {
+            safe.initialize(source.network.opsAdmin);
+            nonce = safe.getNonce();
+        }
+
+        IAdapter[] memory adapters = new IAdapter[](2);
```
wischli: We should build this dynamically based on the configuration

test/integration/fork/WireToNewNetworkForkTest.t.sol
```diff
@@ -0,0 +1,135 @@
...
+                    assertEq(remoteAdapter, target.contracts.chainlinkAdapter, "Wrong Chainlink destination adapter");
+                    found = true;
+                    break;
+                }
+            }
+        }
+        assertTrue(found, "SendPayload not emitted by MultiAdapter for Chainlink");
+    }
+
+    function testWireEthereum() external {
```
wischli, line 140: I suppose eventually we want to run the test for every network.

test/integration/fork/WireToNewNetworkForkTest.t.sol
```diff
@@ -0,0 +1,156 @@
...
+        string memory targetName = "monad";
+        EnvConfig memory target = Env.load(targetName);
+
+        vm.createSelectFork(source.network.rpcUrl());
+
+        vm.startPrank(source.network.opsAdmin);
+        wire(networkName, targetName, "");
+        vm.stopPrank();
+
+        vm.startPrank(target.network.protocolAdmin);
```
wischli: Shouldn't this read from source instead of target?
```suggestion
        vm.startPrank(source.network.protocolAdmin);
```

script/WireToNewNetwork.s.sol
wischli: As discussed, we can do that in a later iteration which makes this script generic for any network.

script/WireToNewNetwork.s.sol
```diff
@@ -0,0 +1,207 @@
...
+            address lzAdapter = source.contracts.layerZeroAdapter;
+            bytes memory data = abi.encode(target.adapters.layerZero.layerZeroEid, target.contracts.layerZeroAdapter);
+            _call(abi.encodeCall(IOpsGuardian.wire, (lzAdapter, target.network.centrifugeId, data)));
+            adapters[0] = IAdapter(lzAdapter);
+        }
+
+        {
+            address chainlinkAdapter = source.contracts.chainlinkAdapter;
+            bytes memory data = abi.encode(target.adapters.chainlink.chainSelector, target.contracts.chainlinkAdapter);
+            _call(abi.encodeCall(IOpsGuardian.wire, (chainlinkAdapter, target.network.centrifugeId, data)));
```
wischli, line 112: Since this script focuses on Monad for now, we need to add a Monad specific entry to `env/connections/mainnet.json` at the bottom which overwrites the current `ALL <> ALL` inclusion of Monad to be be aligned. Otherwise this script's `initAdapters` call will set adapters that contradict the connection config.

wischli (commented):
Almost done IMO!

script/WireToNewNetwork.s.sol
```diff
@@ -0,0 +1,223 @@
...
+        if (targetConn.wormhole) {
+            address wormholeAdapter = source.contracts.wormholeAdapter;
+            bytes memory data = abi.encode(target.adapters.wormhole.wormholeId, target.contracts.wormholeAdapter);
+            _call(abi.encodeCall(IOpsGuardian.wire, (wormholeAdapter, target.network.centrifugeId, data)));
+            adapters[idx++] = IAdapter(wormholeAdapter);
+        }
+
+        if (targetConn.axelar) {
+            address axelarAdapter = source.contracts.axelarAdapter;
+            bytes memory data = abi.encode(target.adapters.axelar.axelarId, target.contracts.axelarAdapter);
```
wischli: We need to cast the axelar adapter address to string, otherwise it will silently fail to relay via Axelar which has happened multiple times already. It's a real pain.


```suggestion
            bytes memory data = abi.encode(target.adapters.axelar.axelarId, vm.toString(target.contracts.axelarAdapter));
```

script/WireToNewNetwork.s.sol
```diff
@@ -0,0 +1,223 @@
...
+            data[2] = abi.encodeCall(ILayerZeroEndpointV2Like.setConfig, (lzAdapter, sendLib, params));
+            targets[3] = endpoint;
+            data[3] = abi.encodeCall(ILayerZeroEndpointV2Like.setConfig, (lzAdapter, recvLib, params));
+            (address to, bytes memory batchData) = protocolSafe.getProposeTransactionsTargetAndData(targets, data);
+            bytes memory signature =
+                protocolSafe.sign(to, batchData, Enum.Operation.DelegateCall, msg.sender, derivationPath);
+            protocolSafe.proposeTransactionsWithSignature(targets, data, msg.sender, signature);
+        } else {
+            lzEndpoint.setSendLibrary(lzAdapter, targetEid, sendLib);
+            lzEndpoint.setReceiveLibrary(lzAdapter, targetEid, recvLib, 0);
```
wischli, line 173: Aren't these two calls redundant because we set to the values we read at L146f.?

script/WireToNewNetwork.s.sol
onnovisser: wow, good catch.

script/WireToNewNetwork.s.sol
onnovisser: In a way yes, by default LZ will use the default library. But to be extra careful I set it explicitly, because the default library could potentially change at some point. Even if the new version would be backwards compatible, it would then bypass the DVN configs we've set.

wischli: I see, makes sense. Thanks for the explanation!

test/integration/fork/WireToNewNetworkForkTest.t.sol
```diff
@@ -0,0 +1,165 @@
...
+/// @notice Fork test that runs WireToNewNetwork.wire() and verifies the underlying bridge
+///         contracts emit events with the correct destination EID/chain ID and receiving adapter address.
+/// @dev    Extends WireToNewNetwork so the same wiring logic is tested end-to-end.
+///         Run per-source-chain; each test forks that chain and simulates the wiring.
+contract WireToNewNetworkForkTest is WireToNewNetwork, Test {
+    using CastLib for address;
+
+    receive() external payable {}
+
+    function _testCase(string memory networkName) internal {
```
wischli, line 30: Nit: After looking over this once more, I feel like we should also validate the post wiring adapter configuration. AFAICT, if `initAdapters` were misconfigured (e.g., wrong threshold, wrong quorum), the test could still pass. So we shuld add something like

```
IMultiAdapter multiAdapter = IMultiAdapter(source.contracts.multiAdapter);
assertGt(multiAdapter.quorum(target.network.centrifugeId, PoolId.wrap(0)), 0, "Quorum not set");
assertEq(multiAdapter.threshold(target.network.centrifugeId, PoolId.wrap(0)), targetConn.threshold, "Wrong threshold");
```

script/WireToNewNetwork.s.sol
```diff
@@ -0,0 +1,223 @@
...
+    function _configureLzDvns(EnvConfig memory source, EnvConfig memory target) internal {
+        address lzAdapter = source.contracts.layerZeroAdapter;
+        ILayerZeroEndpointV2Like lzEndpoint = ILayerZeroEndpointV2Like(address(LayerZeroAdapter(lzAdapter).endpoint()));
+        uint32 targetEid = target.adapters.layerZero.layerZeroEid;
+
+        SetConfigParam[] memory params = new SetConfigParam[](1);
+        params[0] = _buildLzConfigParam(source, targetEid);
+
+        address sendLib = lzEndpoint.defaultSendLibrary(targetEid);
+        address recvLib = lzEndpoint.defaultReceiveLibrary(targetEid);
```
wischli, line 151: Hypothetically, these could return `address(0)` if LayerZero hasn't configured default libraries for the `targetEid` on the source chain. I think we should add a validation layer here to be safe


```suggestion
        address sendLib = lzEndpoint.defaultSendLibrary(targetEid);
        address recvLib = lzEndpoint.defaultReceiveLibrary(targetEid);
        require(sendLib != address(0) && recvLib != address(0), "LZ default libraries not configured for target EID");
```

script/WireToNewNetwork.s.sol
```diff
@@ -0,0 +1,223 @@
...
+        if (targetConn.layerZero) adapterCount++;
+        if (targetConn.wormhole) adapterCount++;
+        if (targetConn.axelar) adapterCount++;
+        if (targetConn.chainlink) adapterCount++;
+
+        IAdapter[] memory adapters = new IAdapter[](adapterCount);
+        uint256 idx;
+
+        if (targetConn.layerZero) {
+            address lzAdapter = source.contracts.layerZeroAdapter;
```
wischli, line 85: Nit for later generic use: If `targetConn.layerZero` is true but `source.contracts.layerZeroAdapter == address(0)` due to the adapter not being deployed yet on source chain the script calls `OpsGuardian.wire(address(0), ...)` which is harder to debug as its a low-level revert. Suppose we should ensure none of the addresses point to the default 0 one.

test/integration/fork/WireToNewNetworkForkTest.t.sol
```diff
@@ -0,0 +1,165 @@
...
+        _testCase("optimism");
+    }
+
+    function testWirePlume() external {
+        _testCase("plume");
+    }
+
+    // function testWirePharos() external {
+    //     _testCase("pharos");
+    // }
```
wischli, line 175: Nit: We should add a comment why this is disabled.

wischli (commented):
Did another round of reviews. Really only nits that you can ignore for now and can be tackled in a follow-up when we make this generic. Just wanted to flag them now so we have them on our radar.

wischli (dismissed):

test/integration/fork/WireToNewNetworkForkTest.t.sol
onnovisser: I'll try enabling it, maybe the CI has a Pharos api key

onnovisser: Ah right, that one was already wired to Monad

wischli (approved):

lemunozm (commented):
Good script here for future iterations as well!

PR #805, opened 2026-03-18
ci: Add live validators

wischli:
### Merge strategy

This PR targets `main` for review purposes. Once approved, only the `weekly-validation.yml` CI workflow should be merged into `main`. The validator and investment flow test code should be pushed to the separate `live-validators` branch, which is pinned to the latest deployed version and runs on a weekly schedule. This keeps `main` free of fork test code that is tightly coupled to the currently deployed contracts. Which was an issue we discussed in the past. 

### Product requirements

* Replace monolithic `ForkTestLiveValidation.sol` (1805 lines) with modular validator architecture: One validator per concern, indexer-driven, zero hardcoded addresses
* Add investment flow fork tests exercising full deposit/redeem cycles on all live vaults across all networks (async, cross-chain async, sync deposit)
* Detect `VaultRegistry` routing corruption from the `unlinkVault` bug ([ref](https://kflabs.slack.com/archives/C07PG2EUR9C/p1773825775019939))

### Design notes

* Validators are independent `BaseValidator` subclasses, each querying the GraphQL indexer for its data. Adding a validator = one file + one array entry in `LiveValidationForkTest`.
* `InvestmentFlowExecutor` handles local async, cross-chain async, and cross-chain sync deposit flows with shared redeem logic.
* `weekly-validation.yml` on `main` checks out the `live-validators` branch on a Monday schedule via GCP-authed fork tests across 9 networks.

### Notes

* There are expected validation failures based on live misconfiguration ([ref1](https://kflabs.slack.com/archives/C07PG2EUR9C/p1773825775019939), [ref2](https://kflabs.slack.com/archives/C07PG2EUR9C/p1773827735981709))


test/integration/fork/validators/Validate_FileConfigurations.sol
```diff
@@ -0,0 +1,225 @@
...
+        if (
+            c.gateway.code.length == 0 || c.multiAdapter.code.length == 0 || c.messageDispatcher.code.length == 0
+                || c.messageProcessor.code.length == 0 || c.spoke.code.length == 0 || c.balanceSheet.code.length == 0
+                || c.hub.code.length == 0 || c.asyncRequestManager.code.length == 0 || c.syncManager.code.length == 0
+        ) {
+            _errors.push(
+                _buildError("code", "core contracts", "> 0", "0", "One or more core contracts have no deployed code")
+            );
+            return;
+        }
```
lemunozm: In which case this case will fail?

test/integration/fork/validators/Validate_ContractWards.sol
```diff
@@ -0,0 +1,164 @@
...
+
+        _checkWard(c.spoke, c.messageDispatcher, "spoke <- messageDispatcher");
+        _checkWard(c.balanceSheet, c.messageDispatcher, "balanceSheet <- messageDispatcher");
+        _checkWard(c.contractUpdater, c.messageDispatcher, "contractUpdater <- messageDispatcher");
+        if (c.vaultRegistry != address(0)) {
+            _checkWard(c.vaultRegistry, c.messageDispatcher, "vaultRegistry <- messageDispatcher");
+        }
+        if (c.hubHandler != address(0)) {
+            _checkWard(c.hubHandler, c.messageDispatcher, "hubHandler <- messageDispatcher");
+        }
```
lemunozm: In which case, `vaultRegister` or `hubHandler` will be 0? I think never can happen, and if happens, it should be an error. Same for other checks in this file

test/integration/fork/validators/Validate_AdapterConfigurations.sol
```diff
@@ -0,0 +1,611 @@
...
+        ) return;
+
+        EnvConfig memory localConfig = Env.load(ctx.networkName);
+        Connection[] memory connections = localConfig.network.connections();
+
+        if (connections.length == 0) return;
+
+        for (uint256 i; i < connections.length; i++) {
+            _validateConnection(c, connections[i]);
+        }
```
lemunozm: Love it!

test/integration/fork/InvestmentFlowForkTest.t.sol
```diff
@@ -0,0 +1,169 @@
...
+        _testCase("optimism");
+    }
+
+    function testInvestmentFlows_HyperEvm() external {
+        _testCase("hyper-evm");
+    }
+
+    function testInvestmentFlows_Monad() external {
+        _testCase("monad");
+    }
```
lemunozm: I think we can avoid especial methods if we do:

```solidity
    function testInvestmentFlows_HyperEvm() external {
        _testCase(vm.envString("NETWORK"));
    }
```

Later CI just needs to set a specific NETWORK envvar. In this way, each CI call can also be organized per network. WDYT?

test/integration/fork/LiveValidationForkTest.t.sol
```diff
@@ -0,0 +1,77 @@
...
+        validators[6] = new Validate_Vaults();
+        validators[7] = new Validate_HookPoolEscrow();
+
+        ValidationExecutor executor = new ValidationExecutor(networkName, "live-validation");
+        executor.runPreValidation(validators, true);
+    }
+
+    function testLiveValidation_Ethereum() external {
+        _testCase("ethereum");
+    }
```
lemunozm: Same here

lemunozm (commented):
Some NITs, but I need to tell you this refactor is awesome!! Much more legible and organized. Really awesome 🙌🏻

test/integration/fork/validators/Validate_FileConfigurations.sol
wischli: True that this will only fail for a bad fork URL, env config pointing to wrong network, or forking at a block before deployment. In normal operation it should never fire. I prefer adding a one-liner comment over removing this catch-all guard.

test/integration/fork/validators/Validate_ContractWards.sol
wischli: You are right. Will remove the `if != address(0) guards since `_checkWard` already handles zero addresses gracefully. I would keep the guards for adapter checks though since these cannot be assumed to be deployed on all networks.

test/integration/fork/InvestmentFlowForkTest.t.sol
wischli: For now, before we create the env-repo, I will replace this with your recommendation, good catch!

test/integration/fork/validators/Validate_ContractWards.sol
lemunozm: Fine for those!

lemunozm: Maybe for the adapters we could read the EnvConfig, and perform the exact check in each chain. Just an idea we don't need to implement now

lemunozm (dismissed):
All changes look good! Just seems the NETWORK env var is not accesible 🤔

wischli:
> All changes look good! Just seems the NETWORK env var is not accesible 🤔

Good catch! Should be fixed in [9ef021e](https://github.com/centrifuge/protocol/pull/805/commits/9ef021e5e54513c36acb111cf5f48a29ac3d72f4)

lemunozm (dismissed):
LGTM!

>  Once approved, only the weekly-validation.yml CI workflow should be merged into main

Just one point. How do you want to merge this PR based on the above?

wischli:
> LGTM!
> 
> > Once approved, only the weekly-validation.yml CI workflow should be merged into main
> 
> Just one point. How do you want to merge this PR based on the above?

Happy to discuss this in the sync beforehand. My idea was to reduce this PR just to the `weekly-validation.yml` which gets merged into main and push the remaining code to some branch. Right now, that branch would have to be named `live-validations` as declared in the YML but no strong opinion on the branch name in case you have a better recommendation that also fits with the env work.

lemunozm:
Ok, fine to me! I would expect other actions there apart from validations (for example, the VerifyFactoryContracts script). So maybe we could just name it `live`, WDYT?

I'll move the factory verifier there once all of this gets merged

lemunozm (approved):

lemunozm:
Is this ready @wischli ?

PR #806, opened 2026-03-25
Swappable GasService through OpsAdmin

lemunozm:
### Product requirements

* We want a way to easily & fast modify gasService values in all chains to react to chain changes.
  * Thread: https://kflabs.slack.com/archives/C07PG2EUR9C/p1768227747451099
  * Task: https://www.notion.so/Feat-Update-GasService-via-Guardian-2e62eac24e17808eb954f8fc2353bfca?source=copy_link

#### Solution A (failed).
I initially allow direct modification to GasService values using OpsAdmin making GasService mutable. But this comes with some issues:
* Extra SLOAD(s).
* We need to do a lot of calls one per (chain,message) tuple.
  * i.e: if some message change its value I need to update it in each chain for all chains O(n*n).
  * I could use "default" values but I would need another SLOAD.
* Some messages with internal diferentiations are not easy to represent for dynamic updates
  * i.e: UpdateVault comes in some flavors: DeployAndLink, Link, Unlink, each of them with different values but same MessageType to index.

#### Solution B
- Hardcode every change in the GasService bytecode in an immutable way (as it is).
- Call `script/deploy-gas-service.sh` to push a safe OpsAdmin propostion to all changes to acept this new GasService.

### DIFF
--- a/script/DeployGasService.s.sol
+++ b/script/DeployGasService.s.sol
@@ -14,10 +14,10 @@ import {Safe, Enum} from "safe-utils/Safe.sol";
 
 /// @title DeployGasService
 /// @notice Deploys a new GasService and proposes an OpsGuardian.setGasService call via the ops Safe.
-/// @dev Set NETWORK env var to the network name (e.g., "ethereum", "base", "arbitrum").
+/// @dev Set NETWORK and VERSION env vars (e.g., "ethereum", "v3.1.1").
 ///
 ///      Example usage:
-///        NETWORK=ethereum forge script script/DeployGasService.s.sol --rpc-url $ETH_RPC_URL --broadcast
+///        NETWORK=ethereum VERSION=v3.1.1 forge script script/DeployGasService.s.sol --rpc-url $ETH_RPC_URL --broadcast
 contract DeployGasService is Script, JsonRegistry {
     using Safe for *;
 
@@ -28,14 +28,10 @@ contract DeployGasService is Script, JsonRegistry {
         startDeploymentOutput();
         vm.startBroadcast();
 
-        string memory network = vm.envString("NETWORK");
-        EnvConfig memory config = Env.load(network);
+        EnvConfig memory config = Env.load(vm.envString("NETWORK"));
 
         GasService gasService = new GasService(config.network.buildBatchLimits());
-
-        string memory json = vm.readFile(string.concat("env/", network, ".json"));
-        string memory currentVersion = vm.parseJsonString(json, ".contracts.gasService.version");
-        register("gasService", address(gasService), _nextVersion(currentVersion));
+        register("gasService", address(gasService), vm.envString("VERSION"));
 
         address opsGuardian = config.contracts.opsGuardian;
         bytes memory data = abi.encodeCall(IOpsGuardian.setGasService, (IGasService(address(gasService))));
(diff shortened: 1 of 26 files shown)


script/DeployGasService.s.sol
```diff
@@ -0,0 +1,44 @@
...
+
+    string constant LEDGER_DERIVATION_PATH = "m/44'/60'/0'/0/0";
+    Safe.Client safe;
+
+    function run() external {
+        vm.startBroadcast();
+
+        EnvConfig memory config = Env.load(vm.envString("NETWORK"));
+
+        GasService gasService = new GasService(config.network.buildBatchLimits());
```
onnovisser, line 33: Shouldn't the new GasService also be included in the env JSON?

lemunozm: Yeah! Good one, I'll add as part of the script

hieronx:
@lemunozm one note: I think we should actually move `GasService` out of core. Why: every time we update benchmarks / handle some new chain, we will be updating this file in the git repo, and redeploying it. The philosophy of the core is that it's an immutable set of contracts that at some point will never need to be updated; the gas service does not fit that description.

lemunozm:
I agree 👍🏻. Also, from the last iteration regarding this, now we have 2 interfaces `IMessageProperties` (the one that core reads for the things it needs) and `IGasService`. So now the core doesn't have any knowledge of the GasService contract.

Is it ok if we move this to `admin` module? I some way, it contains administration stuff regarding the protocol gas for the supported chains.

hieronx:
> Is it ok if we move this to admin module? I some way, it contains administration stuff regarding the protocol gas for the supported chains.

Seems fine yes!

script/DeployGasService.s.sol
lemunozm: Added!

src/deployment/ActionBatchers.sol
```diff
@@ -230,6 +230,7 @@ contract CoreActionBatcher is Constants {
 
         // Rely opsGuardian
         report.multiAdapter.rely(address(report.opsGuardian));
+        report.gateway.rely(address(report.opsGuardian));
```
wischli, line 233: I am not sure if we should give the OpsGuardian a Gateway ward as it allows for updating the message processor or replacing adapters or even calling `handle` directly to bypass multi-adapter quorum for inbound messages. So I feel like this is in conflict with our trust model. Though, IIUC, this would only be possible in a future version of the OpsGuardian which would support allow it to forward arbitrary calls which it currently cannot?!

src/admin/OpsGuardian.sol
```diff
@@ -41,10 +45,16 @@ contract OpsGuardian is IOpsGuardian {
         if (what == "opsSafe") opsSafe = ISafe(data);
         else if (what == "hub") hub = ICreatePool(data);
         else if (what == "multiAdapter") multiAdapter = IMultiAdapter(data);
+        else if (what == "gateway") gateway = IGateway(data);
         else revert FileUnrecognizedParam();
         emit File(what, data);
     }
 
+    function setGasService(IGasService gasService) external onlySafe {
```
wischli, line 54: Nits: 
1. I think we should an event here such in case it matters for the SDK/Apps.
2. Missing UTs for this setter
3. Missing `@inheritdoc` header

src/admin/interfaces/IOpsGuardian.sol
```diff
@@ -39,6 +40,10 @@ interface IOpsGuardian {
     /// @param data New value for the parameter
     function file(bytes32 what, address data) external;
 
+    /// @notice Updates the gas service
+    /// @param gasService the new gas service to use
+    function setGasService(IGasService gasService) external;
```
wischli, line 46: Nit: We should add a `gateway` view ABI to this interface

src/admin/interfaces/IOpsGuardian.sol
wischli: Nit: Needs to be updated for the new gateway addition

wischli (commented):
Just a few nits and one raised concern that I wanted to discuss but doesn't require any codechange probably.

src/deployment/ActionBatchers.sol
hieronx: This is a very good catch! Definitely a security issue imo.

lemunozm: Note that the ward is from the `OpsGuardian`, not `OpsSafe`. The `OpsGuardian` can only do what is coded on it. No one can call `handle()` or bypass communication actions.

src/admin/OpsGuardian.sol
lemunozm: The events are already in the `file()` methods called by this method, not sure if we want to add another event for this 🤔, in the same way we dont add an event for createPool or init adapters.

Good point with the UTs and documentation and the rest of the NITs!

lemunozm: Changes: [b411c9c](https://github.com/centrifuge/protocol/pull/806/commits/b411c9cc33694ee8f4ae83a87d28426f0b0e79fc)

src/deployment/ActionBatchers.sol
hieronx: Yeah you are right, I was misreading it, this is okay. We just need to be cautious with what we add to OpsGuardian, but that is anyways important.

lemunozm:
Reopen in https://github.com/centrifuge/protocol-internal/pull/190

PR #807, opened 2026-03-31
ci(registry): PR validation, preview comments, and tightened triggers

gpmayorga:
## Summary

- **Validate env file schemas on PRs** — a fast-fail gate (`validate-env-schema.js`) catches broken JSON, missing `network.chainId`, invalid addresses, and structural renames before registry generation even starts.
- **Validate generated registry against indexer requirements** — a post-generation gate (`validate-registry.js`) checks the 6 hard requirements the Ponder indexer needs: `version`, `previousRegistry.ipfsHash`, `startBlock`, `address`, `blockNumber`, and ABI completeness. Errors block the PR; warnings are shown but don't block.
- **PR preview comment** — every PR touching `env/**` gets a comment with the registry summary (version, chains, contracts, ABIs) plus any validation errors/warnings, so reviewers can see exactly what would be published.
- **Tightened triggers** — the workflow now only runs when `env/**`, `script/registry/**`, or the workflow file itself changes. The `release` trigger is removed (env file merges to main are the canonical signal). This keeps the GHA log clean.

### Pipeline flow

```mermaid
flowchart TD
    A[env file changed] --> B[Validate env schemas]
    B -->|fail| X1[PR blocked]
    B -->|pass| C[Detect changed environments]
    C --> D[Generate registry]
    D --> E[Validate registry output]
    E -->|errors| X2[PR blocked]
    E -->|pass/warnings| F[Post PR comment]
    F --> G{Merged to main?}
    G -->|no| H[Done]
    G -->|yes| I[Pin to IPFS]
    I --> J[Update Cloudflare DNS]
```

### Indexer hard requirements (enforced by `validate-registry.js`)

| Field | Rule |
|-------|------|
| `version` | Non-empty string |
| `previousRegistry.ipfsHash` | Non-null if a live registry exists |
| `chains.<chainId>.deployment.startBlock` | Must be a number |
| `chains.<chainId>.contracts.<name>.address` | Non-empty string |
| `chains.<chainId>.contracts.<name>.blockNumber` | Must be a number |
| `abis.<ContractName>` | Must exist for every contract in chains |

PR #808, opened 2026-04-07
feat(registry): delta deprecations and per-tag ABI cache

gpmayorga:
## Summary

This PR improves the contract registry generator (`abi-registry.js`) in two ways:

### Deprecated contracts in deltas
When a contract exists in the **previous** registry (e.g. v3.0) but is **absent** from local `env/*.json`, the delta now includes that contract with `address: null` (and null `blockNumber` / `txHash`). Downstream indexers can use this to stop indexing at the new deployment boundary. Deprecated entries do not require ABIs in the delta.

Compare against a specific previous pin with `SOURCE_IPFS=<cid>` when regenerating (e.g. v3.0 CID under v3.1's `previousRegistry`).

### Per-contract ABI builds (version tags)
ABIs are no longer taken from a single `forge build` at repo tip. Each env contract's `version` field is resolved to a **git tag** (as-is, then `v` prefix, then `.0` semver patch, e.g. `v3.1` → `v3.1.0`). Forge output is built per tag into `cache/abi-registry/<tag>/out/` (gitignored) and reused on later runs.

`processContracts` now preserves `version` on registry contract objects so tag resolution works for all chains.

### ABI name aliases
Env names that differ from Forge artifact names are mapped (e.g. `TokenFactory` → `ShareToken`, `NavManager` → `NAVManager`, hook `*Hook` → hook contract names).

### Top-level `registry.version`
When present, the value is normalized to **`vX.Y.Z`**. It is derived as the max of `deploymentInfo` protocol labels and per-contract `version` strings (so mixed `3` / `v3.1` resolves correctly). The field is **omitted** when unknown; `validate-registry.js` treats a missing top-level version as optional (provenance uses `deploymentInfo.gitCommit`).

### Refactor: shared ABI cache module
Per-tag cache logic lives in **`script/registry/utils/abi-cache.js`** so other tooling can reuse it. **`script/registry/build-abi-cache.js`** warms `cache/abi-registry/<tag>/` without running the full registry generator.

### Docs
`script/registry/README.md` updated for the new flow and deprecated `address` semantics. `.claude/rules/registry.md` documents the cache module and CLI.

---

## Fix: Smart auto-publish with validation gate (follow-up to #807)

> **Context:** Merging PR #807 to `main` unintentionally triggered the registry workflow and pinned a new delta to IPFS + updated Cloudflare, even though no contract deployments had changed. Only CI script and doc changes were in that PR. This section addresses the root cause and adds defence-in-depth.

### Root cause
The `push` trigger on `main` fired whenever any `env/**` or `script/registry/**` file changed — including purely cosmetic/CI changes. The old guard (`mainnet == 'true' || testnet == 'true'`) only checked whether env files were touched, not whether those changes contained actual contract address/block updates.

### Changes

**`validate-registry.js`** — adds `publishable` to the validation summary:
```
publishable: errors.length === 0 && chainIds.length > 0
```
A registry is only publishable when it passes validation _and_ contains at least one chain with contract changes.

**`pin-to-ipfs.js`** — two new guards:
- **Zero-chains guard:** skips pinning if the generated delta has no chains (i.e., nothing the indexer depends on changed). `FORCE_PIN` bypasses this for manual corrections.
- **`FORCE_PIN="both"` fix:** `"both"` now correctly bypasses the guard for all environments (was previously only matching `"mainnet"` or `"testnet"` literally).

**`registry.yml`** — three behaviour changes:
1. **`workflow_dispatch` with structured input** (`mainnet` / `testnet` / `both`): allows ops to manually re-pin without modifying code. The detect-changed-environments step respects this input instead of git diff.
2. **"Validate and gate publish" step** in the `pin-to-ipfs` job: re-runs `validate-registry.js` on downloaded artifacts and _removes_ any registry file with `publishable=false` before the pin step runs. `workflow_dispatch` bypasses this gate (manual override for corrections).
3. **Cloudflare update gated to `push` events only**: `workflow_dispatch` can pin to IPFS (for corrections), but _never_ auto-updates the Cloudflare DNSLink. This prevents any developer with write access from overwriting the live registry pointer without an authorized manual step — important since GitHub Environment approval gates are not available on the current plan.

### New workflow matrix

| Trigger | Contract changes? | IPFS pin | Cloudflare update |
|---|---|---|---|
| `pull_request` | any | ❌ | ❌ |
| `push` to main | no | ❌ | ❌ |
| `push` to main | yes | ✅ | ✅ |
| `workflow_dispatch` | — | ✅ (force) | ❌ (manual only) |


gpmayorga:
#808 needs to be merged first

filo87 (approved):

gpmayorga:
@filo87 here's the 3.1 registry recreated with this deprecation logic

https://ipfs.centrifuge.io/ipfs/bafybeihrpdt7n4ikxkxvewk7lrglwwxxz7eupb2jgekouhytbkae27pfra 

```
guardian	{ address: null, blockNumber: null, txHash: null }
hubHelpers	{ address: null, blockNumber: null, txHash: null }
routerEscrow	{ address: null, blockNumber: null, txHash: null }
globalEscrow	{ address: null, blockNumber: null, txHash: null }
```

gpmayorga:
Documentation: **ABI cache on-disk layout** is now spelled out in `script/registry/README.md` (section *ABI cache layout (repo root)*), with the module header in `script/registry/utils/abi-cache.js` and a pointer in `.claude/rules/registry.md` for AI/human discoverability.

script/registry/utils/abi-cache.js
```diff
@@ -0,0 +1,212 @@
...
+
+/** Env-style name → Forge artifact JSON basename (without .json). */
+export const ABI_NAME_ALIASES = {
+    Token: "ShareToken",
+    NavManager: "NAVManager",
+    FreezeOnlyHook: "FreezeOnly",
+    FullRestrictionsHook: "FullRestrictions",
+    FreelyTransferableHook: "FreelyTransferable",
+    RedemptionRestrictionsHook: "RedemptionRestrictions",
+};
```
lemunozm: Wondering if instead we can just update the names in the env files or in the places where it's called `*Hook`, for example. And avoid to maintain this (this also exists in #807, maybe it could be factorized if we can not get rid of it)

.claude/rules/registry.md
```diff
@@ -0,0 +1,51 @@
...
+## ABI generation
+
+- **`utils/abi-cache.js`** owns the per-tag cache (`ensureAbiCache`, `collectContractTags`, …); `abi-registry.js` and `build-abi-cache.js` call into it.
+- ABIs come from **per-contract `version` in env** → resolved git tag → `cache/abi-registry/<tag>/out/` (worktree + `forge build --skip test`). Not from a single deployment commit’s `out/`.
+- **`DEPLOYMENT_COMMIT`** (env) is metadata only (`registry.deploymentInfo.gitCommit`), not ABI selection.
+- After `packAbis`, **`stripContractVersionsForRegistryOutput`** removes per-contract `version` from serialized JSON (smaller artifacts); **`version` stays in repo `env/*.json`**.
+
+## Delta mode & deprecations
+
+- **Delta only** (not `--full`): compare each chain to **previous registry** (default live URL or `SOURCE_IPFS`).
```
lemunozm, line 36: Q. If I want to run these locally, how do I know the `SOURCE_IPFS` value?

lemunozm (commented):
Looks really nice! Willing to see how it will work in production

lemunozm (dismissed):

script/registry/utils/abi-cache.js
gpmayorga: Renaming chains.*.contracts keys (and the corresponding abis entries) to match Solidity artifact names would be a breaking change for the published registry JSON: the indexer, GraphQL schema, and likely the SDK all assume today’s stable key names. Fixing that properly is a coordinated migration (registry + indexer + clients), not something we should fold into this PR.

In summary:

The table has 6 rows, each requiring a different fix:

**4 *Hook rows** (FreezeOnlyHook → FreezeOnly, FullRestrictionsHook → FullRestrictions, FreelyTransferableHook → FreelyTransferable, RedemptionRestrictionsHook → RedemptionRestrictions)

Straightforward env rename. In env/*.json rename the keys freezeOnlyHook → freezeOnly, fullRestrictionsHook → fullRestrictions, etc. Then emit a new registry delta with those new keys, tombstone the old ones, and update any downstream that uses those key strings (api-v3 pnpm update-registry, SDK if it references them). 4 rows gone.

**NavManager → NAVManager**

The env key navManager → capitalized → NavManager, but the Forge artifact file is NAVManager.json. The only clean fix here is renaming the Solidity contract from NAVManager to NavManager (+ redeployment), because you can't get NAVManager from env navManager with a simple first-letter capitalize. Alternatively we live with this row.

**Token → ShareToken**

This comes from stripping Factory off TokenFactory → Token, then looking up Token.json, which doesn't exist (the artifact is ShareToken.json). Fix: rename TokenFactory → ShareTokenFactory in Solidity (so derived base becomes ShareToken), which again means a redeployment. Or keep this row.


**The 4 `*Hook` rows can be removed in a registry-only migration (env rename + new delta + api-v3 update). The other 2 require a Solidity rename + redeployment.**

gpmayorga: I think it is best to leave the mapping for now and think about the renaming effort separately

.claude/rules/registry.md
gpmayorga: Added how to fetch the SOURCE_IPFS to the readme file

script/registry/utils/abi-cache.js
lemunozm: ITs' fair! Thanks for the explanation

gpmayorga:
@filo87 this is ready to go. Does your approval means you've tested the null fields to deprecate contracts already?

lemunozm (approved):
LGMT!

PR #809, opened 2026-04-08
ci: add auto labelling

wischli:
### Product requirements

Same as https://github.com/centrifuge/protocol-internal/pull/172

* Automate PR labeling to reduce manual overhead and enforce consistent use of the new label taxonomy across `centrifuge/protocol` and `centrifuge/protocol-internal`
* Apply scope labels (`scope:contracts`, `scope:tests`, `scope:scripts`, `scope:tooling`, `scope:infra`) based on which file paths are changed
* Flag `audit:needed` whenever contract source code is modified
* Infer type labels (`type:test`, `type:docs`, `type:chore`) when all changed files fall into a single category

### Design notes

* Uses `actions/github-script@v7` with the Octokit API directly; no third party labeling actions
* The workflow only adds labels, never removes them, so manually applied labels are preserved
* `scope:tooling` and `scope:scripts` are mutually exclusive: tooling applies only when every changed file is under `script/` and none contain `broadcast` or `deploy` in the path
* `src/interfaces/` changes are excluded from `scope:contracts` since interface files alone do not require audit
* `env/` directory changes are classified as infra alongside `.github/`, `lib/`, and root config files
* A comment is posted on the PR the first time the workflow runs (tracked via an HTML marker to avoid duplicates on subsequent pushes), summarizing the auto-applied labels so the author can verify
* Repo agnostic: uses `context.repo` so the same workflow file works on both protocol repos without modification
* Labels NOT set by CI (require human judgment): `type:feature`, `type:bug`, `type:refactor`, `type:idea`, `type:migration`, all `version:*` labels, `audit:in-review`/`audit:fix`/`audit:passed`, `breaking`, `hotfix`, `blocked`, `sync:internal`, `sync:public`

### DIFF
--- a/.github/workflows/auto-label.yml
+++ b/.github/workflows/auto-label.yml
@@ -4,7 +4,7 @@
 # Labeling rules (all matching labels are applied):
 #
 # Scope labels:
-#   scope:contracts  — any file under src/ (excluding src/interfaces/)
+#   scope:contracts  — any file under src/ (excluding */interfaces/ dirs)
 #   scope:tests      — any file under test/
 #   scope:scripts    — any file under script/ (unless scope:tooling applies)
 #   scope:tooling    — ALL files under script/ AND none contain "broadcast" or "deploy"
@@ -71,15 +71,15 @@ jobs:
               const lower = f.toLowerCase();
               let matched = false;
 
-              // Contract source (excluding interfaces)
-              if (f.startsWith('src/') && !f.startsWith('src/interfaces/')) {
+              // Contract source (excluding interface directories scattered
+              // throughout src/, e.g. src/core/hub/interfaces/, src/misc/interfaces/)
+              if (f.startsWith('src/') && !f.includes('/interfaces/')) {
                 hasContract = true;
                 matched = true;
               }
 
-              // Interface-only files under src/interfaces/ are not contracts,
-              // but they are not "other" either — they live in src/ scope
-              if (f.startsWith('src/interfaces/')) {
+              // Interface-only files are not contracts but still belong to src/
+              if (f.startsWith('src/') && f.includes('/interfaces/')) {
                 matched = true;
               }
 
--- a/.github/workflows/auto-label.yml
+++ b/.github/workflows/auto-label.yml
@@ -1,0 +1,202 @@
+# Auto-label PRs based on changed file paths.
+# Works identically on centrifuge/protocol and centrifuge/protocol-internal.
+#
+# Labeling rules (all matching labels are applied):
+#
+# Scope labels:
+#   scope:contracts  — any file under src/ (excluding src/interfaces/)
+#   scope:tests      — any file under test/
+#   scope:scripts    — any file under script/ (unless scope:tooling applies)
+#   scope:tooling    — ALL files under script/ AND none contain "broadcast" or "deploy"
+#   scope:infra      — .github/, lib/, env/, foundry.toml, slither.config.json,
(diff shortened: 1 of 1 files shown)


.github/workflows/auto-label.yml
```diff
@@ -0,0 +1,202 @@
...
+
+              // Contract source (excluding interfaces)
+              if (f.startsWith('src/') && !f.startsWith('src/interfaces/')) {
+                hasContract = true;
+                matched = true;
+              }
+
+              // Interface-only files under src/interfaces/ are not contracts,
+              // but they are not "other" either — they live in src/ scope
+              if (f.startsWith('src/interfaces/')) {
```
onnovisser: Interfaces aren't directly under `src`, but in many subfolders, so maybe `f.includes("/interfaces/")`

wischli: Good catch, fixed in [872e29f](https://github.com/centrifuge/protocol/pull/809/commits/872e29f892eafea3e34fa807738f857521d15bab)

.github/workflows/auto-label.yml
```diff
@@ -0,0 +1,202 @@
+# Auto-label PRs based on changed file paths.
+# Works identically on centrifuge/protocol and centrifuge/protocol-internal.
+#
+# Labeling rules (all matching labels are applied):
+#
+# Scope labels:
+#   scope:contracts  — any file under src/ (excluding */interfaces/ dirs)
+#   scope:tests      — any file under test/
+#   scope:scripts    — any file under script/ (unless scope:tooling applies)
+#   scope:tooling    — ALL files under script/ AND none contain "broadcast" or "deploy"
```
lemunozm: broadcast is in `.gitignore`, am I missing something?

.github/workflows/auto-label.yml
```diff
@@ -0,0 +1,202 @@
+# Auto-label PRs based on changed file paths.
+# Works identically on centrifuge/protocol and centrifuge/protocol-internal.
+#
+# Labeling rules (all matching labels are applied):
+#
+# Scope labels:
```
lemunozm, line 6: Maybe could be interested to also add: `score:env` for those commits where we update the env files after deploying. Also `score:invariant` could be interesting to distinguish from normal tests

.github/workflows/auto-label.yml
```diff
@@ -0,0 +1,202 @@
...
+#   scope:contracts  — any file under src/ (excluding */interfaces/ dirs)
```
lemunozm, line 7: Right now we have `src/deployment` where we store contracts that mostly belong to deploying instead of code itself. But it's still src because we deploy them. Maybe we should move them to `scope:scripts`, or create a new `scope:deploy` that could be interesting. But for the later we should organize first `script/` folder

lemunozm (commented):
Some minor proposal & details, but LGTM!

.github/workflows/auto-label.yml
wischli: Great idea! Added both labels

.github/workflows/auto-label.yml
wischli: I like the idea of `scope:deploy` but at a later time when we re-orged.

.github/workflows/auto-label.yml
wischli: No, you are right. Fixed.

lemunozm (approved):
LGTM!

PR #812, opened 2026-04-15
Tests: Remove hub & spoke BaseTest

lemunozm:
### Product requirements

* Thread: https://kflabs.slack.com/archives/C07PG2EUR9C/p1774530508166949

### Design notes

* Remove hub/BaseTest
* Remove spoke/BaseTest
* Remove hub and spoke Mocks and the mythic CentrifugeChain mock (finally!)
* Remove duplicated coverage in some tests.
* Everything under `Integration.t.sol`
  * Easier to reason, and also easier for auditors that want to create PoC. Previously a lot of PoC used BaseTest which leads to a false-positive PoC. Now using Integration the environment is more real with less mocks.
* Integration tests now moved to `test/integration/<module>/`


test/core/hub/integration/Cases.t.sol
```diff
@@ -1,293 +0,0 @@
...
-                GAIN_ACCOUNT,
-                LOSS_ACCOUNT
-            );
-        }
-        hub.updateVault{value: GAS}(poolId, scId, USDC_C2, bytes32("factory"), VaultUpdateKind.DeployAndLink, 0, REFUND);
-
-        cv.popMessage(); // For setAdapters
-
-        MessageLib.NotifyPool memory m0 = MessageLib.deserializeNotifyPool(cv.popMessage());
-        assertEq(m0.poolId, poolId.raw());
```
wischli, line 77: Nit: The deleted `testPoolCreation` verified that `hub.{notifyPool, notifyShareClass, setRequestManager, updateBalanceSheetManager, updateVault}` each produced a correctly-serialized outbound message*via `cv.popMessage() + deserialize…` whereas the new version of this test doesn't cary an of these paths.

I understand these serializations are unit-tested in `Message{Lib, Dispatcher}.t.sol` but I feel like the compositional check, i.e. that the hub wires the right field values into the dispatcher, is gone. Maybe we could add them back to `test/integration/hub/BatchingAndPayment.t.sol` (or a new `HubDispatch.t.sol`) asserting message bytes on `MockAdapter.values_bytes("send")` for at least one `notifyShareClass{value: GAS}` call.

test/integration/vaults/VaultBaseTest.sol
```diff
@@ -0,0 +1,475 @@
...
+    // --- Helpers ---
+
+    function deployVault(
+        VaultKind vaultKind,
+        uint8 shareTokenDecimals,
+        address hook,
+        bytes16 scId,
+        address asset,
+        uint256 assetTokenId,
+        uint16 /* TODO: destinationChain */
```
wischli: Are we planning on utilizing this? Otherwise, we could as well drop it.

test/integration/vaults/VaultBaseTest.sol
```diff
@@ -0,0 +1,475 @@
...
+import {IBaseVault} from "../../../src/vaults/interfaces/IBaseVault.sol";
+import {SyncDepositVault} from "../../../src/vaults/SyncDepositVault.sol";
+import {RequestCallbackMessageLib} from "../../../src/vaults/libraries/RequestCallbackMessageLib.sol";
+
+import "forge-std/Test.sol";
+
+import {CentrifugeIntegrationTest} from "../Integration.t.sol";
+
+/// @dev Direct centrifuge chain simulator — calls spoke/vaultRegistry/syncManager directly as a ward
+///      instead of routing through adapters. Replaces MockCentrifugeChain in VaultBaseTest.
```
wischli: Nit: I think we can drop references to removed code here and in lines 128 and 179.

test/integration/vaults/VaultBaseTest.sol
```diff
@@ -0,0 +1,475 @@
+// SPDX-License-Identifier: BUSL-1.1
+pragma solidity 0.8.28;
+
+// Re-exported for test files importing from VaultBaseTest
+
```
wischli: Is this intentional or can the comment be removed?

wischli (dismissed):
Just a few nits, thanks for this heavy and long-desired refactor!

test/core/hub/integration/Cases.t.sol
lemunozm: Yeah, agree with this. Not sure how far we want to go checking each step of a end to end composition. I understand that if the end-to-end works it's because the values are correctly added/read from the message (without requiring it to check it here). Would you agree with this last sentence? If not, I'll add it.

test/integration/vaults/VaultBaseTest.sol
lemunozm: It's intentional but not really need it, removed!

test/integration/vaults/VaultBaseTest.sol
lemunozm: This already was in the previous tests IIRC. Removed now!

test/core/hub/integration/Cases.t.sol
wischli: Yeah, let's work with the current coverage.

test/integration/vaults/VaultBaseTest.sol
wischli: Do you still want to fix it? Not a blocker, just asking before I approve.

lemunozm: IIUC I did, unless I understand you wrong

lemunozm: Ah I didn't with the reference xD

lemunozm: Now it should IIUC

wischli (approved):

PR #815, opened 2026-04-27
Deploy (testnet): Onchain PM contracts, updated env files

lemunozm:
### Product requirements

* Add updated env files (compiled with suffix=rev2 to mimic current deployment suffix in testnet)


env/arbitrum-sepolia.json
```diff
@@ -312,14 +312,38 @@
...
   "deploymentInfo": {
     "deploy:protocol": {
-      "gitCommit": "24032569",
-      "timestamp": "2026-01-28T15:42:39Z",
-      "suffix": "rev2",
-      "startBlock": 237622348
+      "gitCommit": "41e19975",
+      "timestamp": "2026-04-27T13:42:19Z",
+      "suffix": "",
+      "startBlock": 10742915
```
lemunozm: @gpmayorga IIUC, we no longer need deploymentInfo section right? I think we discussed this but I didn't remember the outcome.

Right now these changes comes from running `update_network_config.py`.

NOTE: suffix is `""` but it is `rev2` as expected. The problem here is I deployed without using deploy.py (using directly foundry script instead), so the `SUFFIX` envvar is not updated when `update_network_config.py` is running

lemunozm: Also I think this data should not be updated for each new contract deployed, because that `startBlock` should be the first in time IIUC

gpmayorga: > @gpmayorga IIUC, we no longer need deploymentInfo section right? I think we discussed this but I didn't remember the outcome.
> 
> Right now these changes comes from running `update_network_config.py`.
> 
> NOTE: suffix is `""` but it is `rev2` as expected. The problem here is I deployed without using deploy.py (using directly foundry script instead), so the `SUFFIX` envvar is not updated when `update_network_config.py` is running

We did discuss trimming env surface area, but nothing in the current pipeline has removed the need for this block. validate-env-schema.js still requires a deploymentInfo.*.startBlock whenever any contract has a numeric blockNumber (see the error path that mentions indexers / chain-level listeners). The registry generator then copies that into the published registry via getDeploymentStartBlock in abi-registry.js. Downstream, api-v3 also uses chain.deployment.startBlock as the fallback when a contract has no creation block (e.g. onchainPMFactory with blockNumber: null) 
So we cannot drop the deploymentInfo section without changing schema, generation, and consumers.

gpmayorga: > Also I think this data should not be updated for each new contract deployed, because that `startBlock` should be the first in time IIUC

Agreed. startBlock should stay the earliest anchor for that chain’s deployment in the env, not something we bump on every small add-on deploy.

I think the value on Arbitrum Sepolia was simply wrong. The pipeline that calculates deploymentInfo.startBlock has evolved since the 3.1 testnet work (e.g. how deploymentStartBlock gets into env/latest/*-latest.json and when Python falls back to broadcast + heuristics), so an old number can linger and only show up when we touch the file again.

Relevant logic:
https://github.com/centrifuge/protocol/blob/5ec5501ce94043e83aee998778e9714018289627/script/deploy/lib/verifier.py#L398-L403

https://github.com/centrifuge/protocol/blob/5ec5501ce94043e83aee998778e9714018289627/script/deploy/lib/verifier.py#L453-L498

gpmayorga (dismissed):
Even with the missing block numbers there's a default for the indexer to do its job, even though less efficiently for those two contracts but that's a very small impact.

Tx hash: we added that feature but it is not currently being consumed by the indexer or any other tool, we did thought however that it would be an interesting thing to have as that allows for discovery of the deployment of that contract.

gpmayorga:
Screenshot 2026-04-27 at 18 37 10


@lemunozm  DO check if this is the intended result, AKA this is what was deployed to testnet

gpmayorga (approved):

PR #816, opened 2026-04-29
registry: add patch-mode (replace last registry) via null-version layer

gpmayorga:
## Summary

Lets follow-up PRs to the same protocol version (or post-deploy fix-ups of a missed contract) effectively **replace** the last published registry — without orphaning anything on IPFS — by emitting a **null-version patch layer**. The indexer (api-v3 `fetch-registry.mjs`) already supports two layered behaviours that this PR plugs into:

- **`mergeNullVersionPatchesIntoPredecessors`** — a registry with `version: null` is merged into its chronological predecessor.
- **`collapseConsecutiveRegistriesWithSameFileSlug`** — consecutive registries that resolve to the same file slug are merged with newer-wins semantics.

Output mode is now one of:

- `delta` — append onto live tip (existing behaviour; new resolved version)
- `patch` — same `previousRegistry` pointer, `registry.version = null`
- `full` — base snapshot, no `previousRegistry`

Selected via the existing **`REGISTRY_MODE`** env var (`auto` (default), `delta`, `patch`, `full`); auto picks `patch` when local-resolved version equals the live tip's version, and `delta` otherwise. `--full` keeps working.

The mode is surfaced in:

- `abi-registry.js` stdout summary
- `validate-registry.js` sidecar (`summary.mode`) and stdout
- `registry.yml` PR-preview comment row, e.g. `Mode: patch (replaces last v3.2 registry on indexer merge)`

`validate-registry.js` is updated to allow `registry.version = null` exactly when `previousRegistry.ipfsHash` is set (it's a real patch on top of an ancestor); `delta` and `full` modes still require a non-empty string version.

`README.md` documents the modes, the auto rule, and the use cases (multi-PR same version, hotfix, recovery).

### Why a null-version patch instead of true "skip live in linked list"?

Both produce the same end state in the indexer, but the patch mechanism:
- is the canonical mechanism the indexer was already designed for,
- preserves history (no orphans on IPFS),
- requires no delta-base swap (current diff-vs-live logic stays correct),
- supports stacking multiple patches naturally.

The trade-off is a slightly longer linked list, which the indexer collapses on read.

## Test plan

Verified locally (script/registry/README.md "Generating registries locally" commands):

- [x] `mainnet` auto → `patch` (live `v3.1.0` == local `v3.1.0`); `validate-registry` passes.
- [x] `mainnet REGISTRY_MODE=delta` override → `delta`; validator passes.
- [x] `testnet` auto → `delta` (live `version` is `null`, local `v3.1.0`); validator reports zero-chain warning since env on `main` already matches the broken live's contracts. Once a real env change lands (e.g. PR #815) the testnet path will produce a non-empty delta.
- [x] `validate-registry` errors on `version: null` when there is no `previousRegistry.ipfsHash` (the "patch with no ancestor" guard).
- [x] PR preview comment renders `Mode: patch (replaces last \`<v>\` registry on indexer merge)` / `Mode: delta (append onto \`<v>\`)` / `Mode: full snapshot`.

CI on this PR will additionally exercise both auto paths (mainnet → patch, testnet → delta on `main` env) end-to-end through the GitHub Actions workflow.

## Follow-up notes

- Once this lands, PR #815's testnet publish (which adds OnchainPM) will still go out as `delta` (its `v3.2.0` does not match the broken `null` live). After that, any further v3.2 contributions to testnet will auto-promote to `patch` and seamlessly fold into the v3.2 tip on the indexer side.
- Mainnet's existing live (`v3.1.0`) means **any future PR that bumps env to `v3.1.x` on mainnet without bumping major/minor will auto-emit a patch** — this is the multi-PR-same-version flow Luis described.


lemunozm (approved):
LGTM

lemunozm:
Merged with latest main

lemunozm (approved):

PR #822, opened 2026-05-20
Remove files that should be tracked internally

lemunozm:
### Product requirements

* This PR leaves the public `protocol` repo to work with the new repo organization

Wait to merge this once internal/main is sync ✅


wischli (commented):
Let's discuss in the next sync if its acceptable that any external party which clones this repository after applying the `internalize` changes, will probably be quite confused. I think we should add a public readme explaining the emptyness of the public repository. 

Apart from that, this PR looks good with https://github.com/centrifuge/protocol-internal/pull/189 as precondition.

lemunozm:
Makes total sense! We should adapt the README

lemunozm:
Added: [aff9fab](https://github.com/centrifuge/protocol/pull/822/commits/aff9fab159046717743b881c6fb0715ef7c2f565)

wischli (dismissed):

lemunozm:
I think with how the internal CI job works, this comment I added in the README will be removed.

I think the simplest solution is to also add this comment to internal/main, would you agree? I'm gonna normalize the comment is such a way, it is understandable independent from which repo is read

wischli (approved):

